PR #5, opened 2023-10-25
Moves 7409 to final.

steven2308:

PR #6, opened 2023-10-25
jekyll: ignore erc-template.md

lightclient:
Fixes the issue first noted in #5.


lightclient:
Okay there are still issues with HTMLProofer because we haven't updated the relative links here to use `./erc-x.md` format. I'm going to merge this as it is still relevant and will not require passing HTMLProofer to merge.

PR #21, opened 2023-10-26
Add ERC: Trusted Hint Registry

strumswell:
Proposing an EIP for a Trusted Hint Registry. Number 7506 has already been assigned to us. In response to: https://github.com/ethereum/EIPs/pull/7547#issuecomment-1780217238


strumswell:
Still waiting for review...

ERCS/erc-7506.md
```diff
@@ -0,0 +1,463 @@
+---
+eip: 7506
+title: Trusted Hint Registry
+description: Registry of trusted hints for decentralized ecosystems.
```
SamWilsn, line 4: Your description doesn't really add any new information that wasn't in the title. I'd recommend maybe expanding what a trusted hint is so that readers can get the gist of this proposal at a glance.

ERCS/erc-7506.md
```diff
@@ -0,0 +1,463 @@
...
+requires: 712
+---
+
+## Abstract
+
+This ERC proposes the "Trusted Hint Registry," a standardized on-chain metadata management interface. It addresses the
+critical need for trust establishment in decentralized ecosystems, enabling entities to efficiently manage hints that
+bolster claim integrity and verifiability. The addition of various management features around delegation, ownership, and
+meta transactions aim to create an adaptable registry for various use cases and requirements within decentralized
+ecosystems.
```
SamWilsn, line 20: This reads more like an introduction. The abstract should be a terse technical summary of the EIP, and not just an introduction. The reader should get a high-level, but technical, understanding of how the EIP works just from this section.

ERCS/erc-7506.md
```diff
@@ -0,0 +1,463 @@
...
+and off-chain—can be reliably issued, verified, and interpreted, thus serving as an essential building block for the
+credible operation of any decentralized ecosystem. Therefore, the Trusted Hint Registry is not just an addition to the
+ecosystem but a necessary evolution in the complex topology of decentralized trust.
+
+## Specification
+
+The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL NOT”, “SHOULD”, “SHOULD NOT”, “RECOMMENDED”, “MAY”, and
+“OPTIONAL” in this document are to be interpreted as described in RFC 2119.
+
+This EIP specifies a contract called`TrustedHintRegistry`that is deployed once and may then be commonly used by a wider
```
SamWilsn, line 60:
```suggestion
This EIP specifies a contract called `TrustedHintRegistry` that is deployed once and may then be commonly used by a wider
```

ERCS/erc-7506.md
```diff
@@ -0,0 +1,463 @@
...
+audience. Governance is deliberately excluded from this ERC.
```
SamWilsn, line 61: If there is only a single registry contract deployed, does there need to be a standard? EIPs/ERCs generally shine when multiple independent implementers need to interoperate without directly communicating. ERC-20 is a prime example: wallets shouldn't have to write bespoke code for each token, and each token wants to work with as many wallets as possible. Having a token standard makes sense.

In this proposal, you define a single registry contract. Once it's deployed, anyone can use its public interface, no standard required.

That said, perhaps I'm misinterpreting this statement? If there's only one instance ever deployed, I'm not sure why any of the functions defined later on would ever not be implemented. Several (like `getMetadata`) seem to be optional.

---

We have made exceptions to this rule for some ERCs (notably [ERC-7409](https://eips.ethereum.org/EIPS/eip-7409).) If you'd like to go down that route, I'd suggest specifying the same deployment methodology as ERC-7409.

ERCS/erc-7506.md
```diff
@@ -0,0 +1,463 @@
...
+##### setMetadataDelegatedSigned
+
+A method with the following signature **MAY** be implemented to set metadata for a hint as a pre-approved delegate of
+the hint list with a raw signature. The raw signature **MUST** be generated following the Meta Transactions section.
+
+```solidity
+function setMetadataDelegatedSigned(address _namespace, bytes32 _list, bytes32 _key, bytes32 _value, bytes calldata _metadata, address _signer, bytes calldata _signature) public;
+```
+
+****
```
SamWilsn, line 299:
```suggestion
```

ERCS/erc-7506.md
```diff
@@ -0,0 +1,463 @@
...
+- **`transaction sender`**: An Ethereum address that takes signed data from a **transaction signer** and commits it
+  wrapped in a transaction to the smart contract.
+
+A **transaction signer** **MAY** be able to deliver a signed payload off-band to a **transaction sender** that initiates
+the Ethereum interaction with the smart contract. The signed payload **MUST** be limited to being used only
+once (see Signed Hash and Nonce).
+
+#### Signed Hash
+
+The signature of the **transaction signer** **MUST** conform to EIP-712. This helps users understand
```
SamWilsn, line 379:
```suggestion
The signature of the **transaction signer** **MUST** conform to [EIP-712](./eip-712.md). This helps users understand
```

ERCS/erc-7506.md
strumswell: Great question as this section definitely needs some work to be more clear. Let me try to explain it in more detail.

TL;DR: We should remove the sentence as the standard should not care about it. Multiple deployments will probably still be needed to reflect ecosystems needs. Standardizing the contract makes it possible to have a common interface, e.g., for those general checks needed to do while verifying claims across ecosystems.

One area where this ERC might be helpful is, e.g., digital identities (see, e.g.,off-chain standards like W3C Decentralized Identifiers and Verifiable Credentials or your on-chain addresses). In general, those digital identities can represent some on/off-chain entity and can issue signed claims to other entities, receive them, or present them to proof something to some other entity.

In reality, there are often ecosystem-specific claims that are being issued by specific organizations vetted trusted in their respected ecosystems. Think of KYC for DeFi, where me issuing a KYC claim attesting your KYC status is worth nothing, but Coinbase doing it might be trusted in a DeFi ecosystem. To keep track of those trusted entities/ identities, the ecosystem can come together and publish them as trusted hints in this proposed trusted hint registry. The sentence `[...] deployed once and may then be commonly used by a wider
audience. [...]` is supposed to reflect that many entities can be part of that specific ecosystem and use the one contract to publish trusted hints in the same namespace/list combo or separated in different namespace/list combos. So one group might be allowed to define trusted entities for KYC claims, another for AML claims, and so on.

The idea is, that every ecosystem might have their own deployment of this registry to publish trusted hints, that consuming parties, e.g. during verification of claims, can consume. Some methods are marked optional because they might be overkill for some ecosystems but present a standardized way on list/ hint management features in case they are needed. The core methods about setting hints are required, while nice-to-have management features (like adding delegates, meta txn, ...) are kept optional to not overcomplicate things in case an ecosystem might not need it. Ecosystems will probably also have the need for use-case specific methods that would warrant their own deployment.

I know the example above with trusted entities is very specific, but the idea of trusted hints can be expanded further, as they can represent revocation information about issued claims, timestamping for re-verifiability for off-chain claims, and more. 

At the company I work with, we heavily use W3C Decentralized Identifiers and Verifiable Credentials for corporate identities and signed claims that represent, e.g., an enterprises authorized trading partner status needed to legally exchange drugs in your supply chain according to the US [DSCSA](https://www.fda.gov/drugs/drug-supply-chain-integrity/drug-supply-chain-security-act-dscsa). In this W3C off-chain claim world, the identifier part is already solved with the did:ethr method based on [ERC-1056](https://eips.ethereum.org/EIPS/eip-1056). Though, a lot of the other core problems around trusted issuers and revocation are not solved yet with a resilient, decentralized architecture. This ERC is supposed to close this gap, while being, imo, relevant to the core Ethereum community for on-chain claims. We were already approached by other consortiums and their ecosystems to talk about this hint registry because it solves a lot of problems every ecosystem faces. This standard makes it possible to define a common language on how to set and retrieve this information across ecosystems and deployments on Ethereum.

We should probably remove that sentence because there is no good reason that the standards defines this, as this questions would have to be answered on another layer. I will incorporate fixes to your other feedback in a commit later.

SamWilsn: Perfect, thanks a lot for taking the time to write that out! I think you've convinced me that this is a good fit for an EIP.

---

As an aside, you can probably take some of this comment and add it to the EIP itself (likely motivation.)

ERCS/erc-7506.md
strumswell: I tried to make it clearer on what it actually is in my newest commit. Let me know what you think.

ERCS/erc-7506.md
strumswell: I tried to give more of a short summary of what the spec offers in my newest revision. Let me know if find this more clearer.

SamWilsn (approved):

PR #29, opened 2023-10-26
Update ERC-6982: Move to Last Call

sullof:
New PR on this new repo to move ERC-6982 to Last Call


sullof:
I tested if fixing the other ERCs passes the checks. I will revert the changes, later.

sullof:
I reverted it.

PR #58, opened 2023-10-28
Update ERC-6551: Registry and execution interface changes

jaydenwindle:

 - The build passes.


jaydenwindle:
@SamWilsn it looks like the changes in this PR were blocked because ERC-6551 depends on EIP-155, which is not included in the new ERCs repo. The walidator still fails after removing EIP-155 from the list of dependencies and changing the EIP-155 link to point to the EIP website, so I've removed all references to EIP-155 in order to allow these changes to pass.

This is not ideal, as I would prefer to explicitly reference EIP-155 to avoid ambiguity. I'm sure there are many other ERCs that depend on core EIPs like this. How would you recommend handling this edge case?

PR #76, opened 2023-10-31
Update ERC-1363: Update specification natspec comments

vittominacori:
This PR is part of some changes I would like to make on ERC1363 to follow the modern EIP1 recommendation.

This PR:

* Add reference to RFC 2119 and RFC 8174 in specification
* Update natspec comments on interfaces

Note that this PR will not pass the CI, because of many changes required by validator and that are not addressed here but will be addressed in other PR.

I.g. external links, broken link, new line after h1, etc.

Other PR will address other sections of ERC1363


SamWilsn:
These changes are a bit more involved than we allow for final EIPs. I think the best option would be to open a new EIP with the updated content.

PR #42, opened 2023-10-27
Fixed links in ERCS markdown files

cfries:
The HTML Proofer currently fails because all .md files in ERC reference eip-nnnn.md file but these files have been renamed to erc-nnnn.md. This PR just renames the links in the ERCS md files.
Please check if you find this PR reasonable:

- Replaced (regexp) `\(eip-(\d\d+)\.md\)` by `(erc-$1.md)`
- Replaced (regexp) `\(\./eip-(\d\d+)\.md\)` by `(./erc-$1.md)`
- Replaced (regexp) `\.\./assets/eip-(\d\d+)\/`by `../assets/erc-$1/`
- Replaced (regexp) `\: \./eip-(\d\d+)\.md`by `: ./erc-$1.md`
- Replaced (regexp) `\((\./|\.\./EIPS/)eip-(\d\d+)(\.md|)((#[\w\-]*|))\)` by `(erc-$2$3$4)`
- Replaced (regexp) `\./eip-(\d\d+)\.md` by `./erc-$1.md`

Note: The HTML Proofer will still fail because there are links to EIPs that are no longer in this repository (cross-linking is required). The above will also rename the EIPs to ERCs that are EIPs. Those likely require cross linking as there are no files in the ERCS folder. For those one could consider using abolute links:

- Replaced (regexp) `\(\./(eip|erc)-155(\.md|)(#.*|)\)` by `(https\://eips.ethereum.org/EIPS/eip-155$3)`
- Replaced (regexp) `\(\./(eip|erc)-712(\.md|)(#.*|)\)` by `(https\://eips.ethereum.org/EIPS/eip-712$3)`
- ... and so on for the numbers 86, 101, 170, 658, 1014, 1102, 1153, 1193, 1559, 2255, 2938, 3074, 3085, 3220, 3855, 5069, 5757

(I am using this PR to find our if there is a set of reg exp that will fix all or most of the broken links - I hope you don't mind)

The above regexps reduce the number of errors reported by the HTMLProofer from

```
HTML-Proofer found 1807 failures!
```

to

```
HTML-Proofer found 27 failures!
```

(There is maybe a one regexp for all - the list above is a result from step by step carefully fixing issues)

PR #79, opened 2023-10-31
Update ERC-4337: Fix typos and broken links

randomishwalk:


ERCS/erc-4337.md
```diff
@@ -694,7 +694,7 @@ Still, it might require putting a "semi-valid" signature (e.g. a signature in th
 
 **Parameters**: same as `eth_sendUserOperation`
   gas limits (and prices) parameters are optional, but are used if specified.
-  `maxFeePerGas` and `maxPriorityFeePerGas` default to zero, so no payment is required by neither account nor paymaster.
+  `maxFeePerGas` and `maxPriorityFeePerGas` default to zero, so no payment is required by either account nor paymaster.
```
drortirosh, line 697: either/or ...

ERCS/erc-4337.md
```diff
@@ -993,15 +993,15 @@ An array of reputation entries with the fields:
 
 ## Backwards Compatibility
 
-This EIP does not change the consensus layer, so there are no backwards compatibility issues for Ethereum as a whole. Unfortunately it is not easily compatible with pre-[ERC-4337](./eip-4337.md) accounts, because those accounts do not have a `validateUserOp` function. If the account has a function for authorizing a trusted op submitter, then this could be fixed by creating an [ERC-4337](./eip-4337.md) compatible account that re-implements the verification logic as a wrapper and setting it to be the original account's trusted op submitter.
+This EIP does not change the consensus layer, so there are no backwards compatibility issues for Ethereum as a whole. Unfortunately it is not easily compatible with pre-[ERC-4337](./erc-4337.md) accounts, because those accounts do not have a `validateUserOp` function. If the account has a function for authorizing a trusted op submitter, then this could be fixed by creating an [ERC-4337](./erc-4337.md) compatible account that re-implements the verification logic as a wrapper and setting it to be the original account's trusted op submitter.
```
drortirosh, line 996: afaik, it is enough to reference the actual doc only once, so better to remove the duplications, instead of updating them all.

drortirosh (changes requested):

ERCS/erc-4337.md
Joachim-Lebrun: even if it sounds counter-intuitive, you should keep the `./eip-XXX` relative links here, at least for now, as can be seen in issue #86 
PR #42 is supposed to fix all md files in one shot later on, but for now the HTML Proofer run will fail if you put the `./erc-XXX` relative links, and prevent you from getting the PR approved

randomishwalk: Noted thanks @Joachim-Lebrun! That is counter-intuitive indeed :)

ElDobleRR (approved):

SamWilsn:
`eipw` is correct here. You'll need to return to relative links, even though the files are not in this repository.

PR #98, opened 2023-11-10
Add ERC: MIME type for Web3 URL in Auto Mode

nand2:
Takes previous https://github.com/ethereum/EIPs/pull/7087 from before the ERC migration, and edited : 
Now add several new features to the [ERC-6860](./erc-6860.md) auto mode : 

- Determine the MIME type of a URL via additional query parameters.
- Ability to name fields when ``?returns=`` is used.

The main aim is to facilitate the fetching of data of general smartcontracts that were not designed for the ``web3://`` protocol.


nand2:
<activity>
We are still working on it :)

nand2:
I guess we are at a stage we are ready for a review.
Thanks!

ERCS/erc-7087.md
```diff
@@ -0,0 +1,163 @@
+---
+eip: 7087
+title: New features for Web3 URL in Auto Mode
```
SamWilsn, line 3: "New Features" doesn't differentiate this proposal enough. Imagine five years down the road, having a dozen "New Features" proposals. It'd be hard to tell them apart.

ERCS/erc-7087.md
```diff
@@ -0,0 +1,163 @@
...
+description: Add features to Web3 URL in auto mode
```
SamWilsn, line 4: Try not to restate your title in the description. The description should _expand_ on the title.

ERCS/erc-7087.md
```diff
@@ -0,0 +1,163 @@
...
+status: Draft
+type: Standards Track
+category: ERC
+created: 2023-05-28
+requires: 6860
+---
+
+## Abstract
+
+This standard extends the [ERC-6860](./eip-6860.md) `web3://` standard in auto mode with new features :
```
SamWilsn, line 16:
```suggestion
This standard extends the [ERC-6860](./eip-6860.md) `web3://` standard in auto mode with new features:
```

ERCS/erc-7087.md
```diff
@@ -0,0 +1,163 @@
...
+
+- Determine the MIME type of a URL via additional query parameters.
+- Ability to name fields when ``?returns=`` is used.
```
SamWilsn, line 19: Are these features independent? If so, they should probably live in their own EIPs.

SamWilsn, line 19: Your abstract is a bit too terse. You can expand a bit on how this proposal is implemented. The reader should get a high level technical understanding of the proposal just by reading the abstract.

ERCS/erc-7087.md
```diff
@@ -0,0 +1,163 @@
...
+
+## Motivation
+
+The web3 URL standard defines a way to determine the MIME type of a link in auto mode via the suffix of the last argument. However, most of existing contracts on Ethereum do not expect such suffix and thus the protocol cannot determine and return the right MIME type (e.g., contracts returning SVG images with token id as parameter). The standard proposes additional query parameters for a Web3 URL to determine the MIME type.
+
+Additionally, when ``?returns=`` is used with returning complex structured data, an usual step after fetching the URL return is to map the returned arrays into proper named structs. To skip this tedious task, this standard propose a way to name the fields directly in ``?returns=``.
```
SamWilsn, line 25: When you say "determine", do you mean "indicate"? As in the URL indicates to the HTTP server what mime type to return?

ERCS/erc-7087.md
```diff
@@ -0,0 +1,163 @@
...
+retTypeReserved = ":" / "," / "(" / ")"
+
+```
+
+When field names are specified, the JSON returned is no longer an array but an object. As an basic example, ``?returns=(field1:uint)`` will return ``{"field1": 1234}``. 
+
+If ``?returns=`` specifies multiple types, all must have a field name. At a tuple definition level or at the base level, if one type does not have a field name, then the formatting revert to a array at this level. As a basic example, ``?returns=(field1:(subField1:uint,string))`` will return ``{"field1": [1234, "abcd"]}``. 
+
+At a tuple definition level or at the base level, if a field name is duplicate, an error will be throw.
+
```
SamWilsn, line 91: You should define what "an error will be thrown" means. Is that an HTTP error code?

ERCS/erc-7087.md
```diff
@@ -0,0 +1,163 @@
...
+                / "!" / "$" / "'" / "*" / "+" / ";" / "@"
+retTypeReserved = ":" / "," / "(" / ")"
+
+```
+
+When field names are specified, the JSON returned is no longer an array but an object. As an basic example, ``?returns=(field1:uint)`` will return ``{"field1": 1234}``. 
+
+If ``?returns=`` specifies multiple types, all must have a field name. At a tuple definition level or at the base level, if one type does not have a field name, then the formatting revert to a array at this level. As a basic example, ``?returns=(field1:(subField1:uint,string))`` will return ``{"field1": [1234, "abcd"]}``. 
+
+At a tuple definition level or at the base level, if a field name is duplicate, an error will be throw.
```
SamWilsn, line 90:
```suggestion
At a tuple definition level or at the base level, if a field name is duplicate, an error will be thrown.
```

nand2:
Thanks for your review!
I will address all of this shortly.

ERCS/erc-7087.md
nand2: Good point!

ERCS/erc-7087.md
nand2: Ok! I have removed the second part, will put it on an separate pull request later.

ERCS/erc-7087.md
nand2: Yes indeed, "indicate" when the returned data MIME type is implicit (not returned by the smartcontract) -- so in this case an additional query parameter in the URL indicate what is the MIME type to be returned as ``content-type`` when returning the result.

But additonally, in the case of the data returned being a data URL, "determine" by having an additional query parameter indicate that the data is a data URL, so that the MIME type can be extracted from the returned data.

I have rewritten the abstract and motivation sections, hopefully making all this clearer.

ERCS/erc-7087.md
nand2: To be adressed on the splitted ERC -- but basically no, it would be an error like errors you get when the domain name cannot be resolved, or the certificate is wrong : it doesn't even reach the stage of getting an answer with HTTP code, it stops before.

nand2:
Thanks again for the review!
I have adressed your points and I have rewritten the abstract and motivation part, to be more verbose and hopefully more clear and explanatory.
Let me know if these parts are clear enough now.

ERCS/erc-7087.md
```diff
@@ -0,0 +1,107 @@
...
+This standard extends the [ERC-6860](./eip-6860.md) `web3://` standard: in smart contracts not designed for `web3://` (thus using auto mode), the MIME type of the returned data is either implicit (not advertized by the smart contract) or included within the returned data ([RFC 2397](https://www.rfc-editor.org/rfc/rfc2397) data URLs). This standard defines additional query parameters so that a MIME type can be returned when fetching a `web3://` URL in these scenarios.
+
+
+## Motivation
+
+When returning data to the web browser, a `Content-Type` header indicating the MIME type of the data is strongly recommended, or the data may be incorrectly interpreted and displayed by the web browser.
+
+The `web3://` protocol has 2 modes: manual and auto. 
+
+- The manual mode is used on smart contracts explicitely requesting this mode (via an interface), so they are expected to signal the MIME type of the returned data, with the mechanism described in [ERC-6860](./eip-6860.md).
```
SamWilsn, line 25:
```suggestion
- The manual mode is used on smart contracts explicitly requesting this mode (via an interface), so they are expected to signal the MIME type of the returned data, with the mechanism described in [ERC-6860](./eip-6860.md). 
```

ERCS/erc-7087.md
```diff
@@ -0,0 +1,107 @@
...
+
+
+## Motivation
+
+When returning data to the web browser, a `Content-Type` header indicating the MIME type of the data is strongly recommended, or the data may be incorrectly interpreted and displayed by the web browser.
+
+The `web3://` protocol has 2 modes: manual and auto. 
+
+- The manual mode is used on smart contracts explicitely requesting this mode (via an interface), so they are expected to signal the MIME type of the returned data, with the mechanism described in [ERC-6860](./eip-6860.md). 
+- On the other hand, the auto mode is used on both smart contracts specifically requesting the mode, and for all the others not signaling anything. While we can expect smart contracts explicitely requesting auto mode to signal the MIME type of the returned data, we cannot expect it for the others contracts.
```
SamWilsn, line 26:
```suggestion
- On the other hand, the auto mode is used on both smart contracts specifically requesting the mode, and for all the others not signalling anything. While we can expect smart contracts explicitly requesting auto mode to signal the MIME type of the returned data, we cannot expect it for the others contracts.
```

ERCS/erc-7087.md
SamWilsn: "error is thrown" still exists in the updated proposal, but I don't see this definition in it. I'd recommend explaining it in the body of the document.

nand2: Sorry, I thought there were all on the other section.
I have replaced "an error is thrown" with "the URL is not fetched and an error message is displayed to the user".

SamWilsn (approved):

Issue #110, opened 2023-11-20
Unrelated errors reported from HTML Proofer


minkyn:
### Pull Request

https://github.com/ethereum/ERCs/pull/91

### What happened?

The validation of HTML Proofer output a bunch of errors that have nothing to do with the files changed in my PR.

https://github.com/ethereum/ERCs/actions/runs/6929548434/job/18847469212

### Relevant log output

```shell
Run bundle exec htmlproofer --allow-missing-href --disable-external --assume-extension '.html' --log-level=:info --cache='{"timeframe":{"external":"6w"}}' --checks 'Links,Images,Scripts,OpenGraph' --no-check-sri --ignore-empty-alt --no-enforce_https ./_site
Running 4 checks (Images, Links, OpenGraph, Scripts) in ["./_site"] on *.html files ...


Checking 5425 internal links
Checking internal link hashes in 703 files
Ran on 743 files!


For the Images check, the following failures were found:

* At ./_site/EIPS/eip-6358.html:226:

  internal image ../assets/erc-6358/img/o-dlt.png does not exist

For the Links > Internal check, the following failures were found:

* At ./_site/EIPS/eip-7066.html:219:

  internally linking to ./erc-721.md, which does not exist

* At ./_site/EIPS/eip-7066.html:229:

  internally linking to ./erc-721.md, which does not exist

* At ./_site/EIPS/eip-7066.html:237:

  internally linking to ./erc-721.md, which does not exist

* At ./_site/EIPS/eip-7066.html:264:

  internally linking to ./erc-721.md, which does not exist

* At ./_site/EIPS/eip-7066.html:284:

  internally linking to ./erc-721.md, which does not exist

* At ./_site/EIPS/eip-7066.html:372:

  internally linking to ./erc-721.md, which does not exist

* At ./_site/EIPS/eip-7066.html:374:

  internally linking to ./erc-721.md, which does not exist

* At ./_site/EIPS/eip-7066.html:384:

  internally linking to ../assets/erc-7066/test/test.js, which does not exist

* At ./_site/EIPS/eip-7066.html:394:

  internally linking to ../assets/erc-7066/IERC7066.sol, which does not exist

* At ./_site/EIPS/eip-7066.html:396:

  internally linking to ../assets/erc-7066/ERC7066.sol, which does not exist

* At ./_site/EIPS/eip-7066.html:406:

  internally linking to ./erc-721.md, which does not exist

* At ./_site/EIPS/eip-7092.html:228:

[...]

PR #117, opened 2023-11-23
Disable some eipw lints

SamWilsn:


etan-status:
May also need to do this in the EIPs repo:

- https://github.com/ethereum/ERCs/issues/110
- https://github.com/ethereum/EIPs/pull/7936

PR #124, opened 2023-11-27
Update ERC-6358: Fix links

SamWilsn:


ERCS/erc-7092.md
```diff
@@ -17,13 +17,13 @@ This proposal introduces fixed-income financial bonds with key characteristics d
 
 ## Motivation
 
-Fixed-income instruments are a widely utilized asset class for corporations and other entities raising funds. However, transitioning to tokenized bonds is challenging due to existing standards like [ERC-3475](./erc-3475.md), which introduces unfamiliar concepts and leads to unnecessary gas consumption. Additionally, the lack of named variables like coupon, maturity date, and principal, makes it difficult to implement ERC-3475 since developers need to remember which metadata is assigned to each parameter.
```
Edoumou, line 20: Hello, the use of `eip-` instead of `erc` prevent tests to pass. So why do we need to change the links back ? Did you fix that issue ? Could you guarantee that approving this will not break the ERCs ?

PR #146, opened 2023-12-06
Update ERC-6909: Move to Review

jtriley-eth:


SamWilsn (commented):
* Your abstract is a bit too terse. It should include a high-level technical description of how it solves the problem. In this case, maybe you could include a few sentences describing what it removes/adds compared to ERC-1155?
* The interface ID shouldn't be in the rationale section. It belongs in the Specification section.
* Same goes for the metadata extension. Since you're introducing new functionality, and not just explaining the reasoning behind your specification, it _has_ to go in the specification section.

jtriley-eth:
diff isn't showing bc i merged a different branch, but recommended fixes are done

SamWilsn (commented):
You have a "SHOULD NOT" and a "MAY" requirement in your rationale section. All requirements should be in the specification section.

PR #155, opened 2023-12-11
Update Template: Authentication SBT using Credential

c1ick:
discussion to : https://ethereum-magicians.org/t/erc-7574-authentication-sbt-using-credential/17264


erc-template.md
```diff
@@ -1,13 +1,13 @@
 ---
```
abcoathup, line 1:
```suggestion
---
eip: 7574
```
Numbering changed to sequential from 7500 and is no longer the PR number.

Please create a new file for the ERC rather than updating the template, i.e. erc-7574.md

ERCS/erc-7574.md
```diff
@@ -0,0 +1,191 @@
...
+-->
+
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
+
+
+### MetaData Interface
+
+
+NFT metadata structured by the official [ERC-721](./eip-721.md) metadata standard or the Enjin Metadata suggestions are as follows referring to OpenSea docs.
```
SamWilsn, line 74: > the Enjin Metadata suggestions are as follows referring to OpenSea docs

We don't allow external links (with some exceptions defined in EIP-1.) This section should contain sufficient information for an implementer to build the software, which in this case probably means some JSON schema definitions instead of calling out to the Enjin/OpenSea docs.

ERCS/erc-7574.md
```diff
@@ -0,0 +1,191 @@
...
+   "name":" KOREA KYC",
+   "attributes" : { ... },
+   "Iusser" : { ...},
+   "credentialNumber" : { ... },
+ }
+```
+
+We have defined a structure to represent VerifiedPresentation. It includes the user's wallet address, issuer, user name, and Credential Number
+
+```jsx
```
SamWilsn, line 104:
````suggestion
```solidity
````

Perhaps?

File: ERCS/erc-7574.md [hunk 4]
```diff
@@ -0,0 +1,191 @@
...
+Following that, here is the contract that verifies a user's Credential proof and issues SBTs:
+
+
+### Contract Interface
+
+
+In this scenario, a **`verify`** function is created using Zokrates. Upon successful verification, the generated **`verify`** function calls the SBT issuance contract, enabling the entire process.
+
+```
+/ SPDX-License-Identifier: CC0-1.0
```
SamWilsn, line 122:
```suggestion
// SPDX-License-Identifier: CC0-1.0
```

ERCS/erc-7574.md
```diff
@@ -0,0 +1,191 @@
...
+}
+```
+
+Following that, here is the contract that verifies a user's Credential proof and issues SBTs:
+
+
+### Contract Interface
+
+
+In this scenario, a **`verify`** function is created using Zokrates. Upon successful verification, the generated **`verify`** function calls the SBT issuance contract, enabling the entire process.
```
SamWilsn, line 119: > using Zokrates

Hm, so this definitely doesn't provide enough information to implementers. If the format of the `Proof` structure is "whatever Zokrates generates", then everyone implementing this standard has to use Zokrates, and there isn't much point of having an ERC at all.

ERCs really shine when multiple independent parties have to coordinate without otherwise communicating. ERC-20 is a great example. Every token wants to work with every wallet, and vice versa; perfect for an ERC.

ERCS/erc-7574.md
```diff
@@ -0,0 +1,191 @@
...
+
+<!--
+  The rationale fleshes out the specification by describing what motivated the design and why particular design decisions were made. It should describe alternate designs that were considered and related work, e.g. how the feature is supported in other languages.
+
+  The current placeholder is acceptable for a draft.
+
+  TODO: Remove this comment before submitting
+-->
+
+This is a system that verifies the basic DID signature by integrating the Ethereum extension program Zokrates. Following the verification of the DID signature, it proceeds with issuing SBT, allowing the validation of a user's identity information.
```
SamWilsn, line 158: Your Rationale section is a bit light. You want to explain the reasoning behind technical choices you made within the ERC itself.

My pet analogy for this is:

> Motivation: "why do we need a shed?"
> Rationale: "why is the shed red?"

PR #174, opened 2023-12-26
Update ERC-4337: Fix typos

kopy-kat:
Fixing typos

### DIFF
--- a/ERCS/erc-4337.md
+++ b/ERCS/erc-4337.md
@@ -379,9 +379,9 @@ While simulating `userOp` validation, the client should make sure that:
    1. must not use value (except from account to the entrypoint)
    2. must not revert with out-of-gas
    3. destination address must have code (EXTCODESIZE>0) or be a standard Ethereum precompile defined at addresses from `0x01` to `0x09`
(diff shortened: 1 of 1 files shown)


kopy-kat:
Another thing in the specs that is not entirely clear is whether the following rule gives the account factory and paymaster as an example of an entity that can be staked and thus access their self storage or whether this rule ONLY applies to the factory and paymaster but not any other external contract

https://github.com/ethereum/ERCs/blob/8568c155452785f4de2e30ae665abe5d431281ca/ERCS/erc-4337.md?plain=1#L375C11-L375C11

drortirosh (approved):

drortirosh:
note that we're in the process of moving the rules from erc4337 itself into
a separate document (erc-7562)

drortirosh:
actually, it is better to re-submit against the version in the eth-infinitism/account-abstraction/erc, as that's what we'll push back into the main ethereum/erc repo.

kopy-kat:
looks like these are already fixed in `/erc` on `develop` branch

PR #184, opened 2024-01-04
Update ERC-5169: Add contributing author

JamesSmartCell:
This PR is to add an author who contributed to the original concept.


xinbenlv (approved):
Existing lead author wants to add another author, I think it's ok.

SamWilsn (changes requested):
This is final, so we should probably do a Call for Input.

SamWilsn:
As discussed on https://github.com/ethcatherders/EIPIP/issues/308, the prevailing opinion among editors is to not modify a final proposal.

PR #192, opened 2024-01-08
Website: Fixed rendering of headers of ERC pages, Navbar and footer

dhanushlnaik:
Website : Fixed rendering of headers of ERC pages, Navbar and footer


lightclient (commented):
The ERCs are loaded in a CI workflow in the EIPs repo to deploy the site eips.ethereum.org. This change won't affect that website because the jekyll code would live in ethereum/eips.

PR #201, opened 2024-01-12
Update ERC-7201: Fix solidity formula

RenanSouza2:
This PR fixes one formula in the ERC7201, I am not one of the authors

### DIFF
--- a/ERCS/erc-7201.md
+++ b/ERCS/erc-7201.md
@@ -36,7 +36,7 @@ A namespace in a contract should be implemented as a struct type. These structs 
 
 ### Formula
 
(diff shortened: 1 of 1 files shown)


Amxx:
I don't think this change is needed. If `id` is a string, you can call `keccak256(id)` directly.

I was able to confirm (in remix) that
```
    bytes32 public constant MAIN_STORAGE_LOCATION =
        keccak256(abi.encode(uint256(keccak256("example.main")) - 1)) & ~bytes32(uint256(0xff));
```
does compile (tested with 0.8.7)

RenanSouza2:
> I don't think this change is needed. If `id` is a string, you can call `keccak256(id)` directly.
> 
> I was able to confirm (in remix) that
> 
> ```
> bytes32 public constant MAIN_STORAGE_LOCATION =
>     keccak256(abi.encode(uint256(keccak256("example.main")) - 1)) & ~bytes32(uint256(0xff));
> ```
> 
> does compile (tested with 0.8.7)

it works with hardcoded string like `"example.main"` but not with a variable type string memory

Amxx (approved):

ernestognw (approved):
> it works with hardcoded string like `"example.main"` but not with a variable type string memory

Yea I can confirm.

SamWilsn (changes requested):
https://github.com/ethcatherders/EIPIP/issues/327

Amxx (approved):

SamWilsn (approved):

PR #232, opened 2024-02-02
Update ERC-4337: fix debug_bundler_dumpreputation status type

arddluma:
Changed `debug_bundler_dumpreputation` response to `number` instead of `string` as per [spec-tests](https://github.com/eth-infinitism/bundler-spec-tests/blob/master/tests/single/reputation/test_reputation.py#L20-L23)


jetonk:
🙌

drortirosh:
note that the [schema](https://github.com/eth-infinitism/bundler-spec/blob/update-actions-versions/src/schemas/bundle.yaml#L38)
defines the reputation as an enum (string), not as a number

PR #236, opened 2024-02-04
Add ERC: Puppet Proxy Contract

CodeSandwich:
Here's the puppet proxy contract EIP. 

I'm aware that there's a strict numbering convention for the EIPs, but if it's bypassable, it would be great if this EIP got number 1133. It would place the puppet proxy mentally close to ERC-1167, the clone proxy, which is a similar concept.


abcoathup:
> I'm aware that there's a strict numbering convention for the EIPs, but if it's bypassable, it would be great if this EIP got number 1133. It would place the puppet proxy mentally close to ERC-1167, the clone proxy, which is a similar concept.

EIP/RIP/ERC numbers are assigned incrementally.  
From: https://eips.ethereum.org/EIPS/eip-1
Once the EIP is ready for the repository, the EIP editor will:
Assign an EIP number (generally incremental; editors can reassign if number sniping is suspected)

ERCS/eip-draft_puppet_proxy_contract.md
```diff
@@ -0,0 +1,342 @@
+---
```
abcoathup, line 1:
```suggestion
---
eip: 7613
```

abcoathup: Assigning next sequential EIP/ERC/RIP number.
Numbering changed to sequential from 7500 and is no longer the PR number.

ERCS/eip-draft_puppet_proxy_contract.md
```diff
@@ -0,0 +1,342 @@
...
+title: Puppet proxy contract
+description: The proxy that if called by its deployer, delegates to the address passed to it in the calldata.
+author: Igor Żuk (@CodeSandwich)
+discussions-to: https://ethereum-magicians.org/t/eip-puppet-proxy-contract/18482
```
abcoathup, line 5:
```suggestion
discussions-to: https://ethereum-magicians.org/t/erc-7613-puppet-proxy-contract/18482
```

CodeSandwich: Done

mattstam (commented):
Cool idea! Implementation looks good. Only thing I'd like to see is more of is an explanation of why someone would use this, or why a protocol would deploy these for their users.

CodeSandwich:
@mattstam There's the "motivation" section, do the usage examples need more thorough explanation? The alternative approaches are described in the "rationale" section, they are separated to follow the template ERC structure.

mattstam:
I mean expand on this:
> each user of a protocol can get their own address where assets can be sent and stored.

Who is the deployer of these accounts, the protocol or the users? What types of protocols would want this behavior?

ERCS/erc-7613.md
```diff
@@ -0,0 +1,343 @@
...
+created: 2024-02-04
+---
+
+## Abstract
+
+A puppet is a contract that when called, acts like an empty account, it doesn't do anything and it has no API. The only exception is that if it's called by the address that deployed it, it delegates the call to the address passed to it in calldata. This gives the deployer the ability to execute any logic they want in the context of the puppet.
+
+## Motivation
+
+A puppet can be used as an alternative account of its deployer. It has a different address, so it has a separate set of asset balances. This enables sophisticated accounting, e.g. each user of a protocol can get their own address where assets can be sent and stored.
```
CodeSandwich, line 19:
```suggestion
A puppet can be used as an alternative account of its deployer. It has a different address, so it has a separate set of asset balances. This enables sophisticated accounting, e.g. each user of a protocol can get their own address where assets can be sent and stored. The user may call the protocol contract, which in turn will deploy a new puppet and consider it assigned to the user. If the puppet is deployed under a predictable address, e.g. by using the user's address as the CREATE2 salt, the puppet may not even need to be deployed before funds are sent to its address. From now on the protocol will consider all the assets sent to the puppet as owned by the user. If the protocol needs to move the funds out from the puppet address, it can call the puppet ordering it to delegate to a function transferring the assets to arbitrary addresses, or making arbitrary calls triggering approved transfers to other contracts.

Puppets can be used as an alternative to approved transfers when loading funds into the protocol. Any contract and any wallet can transfer the funds to the puppet address assigned to the user without making any approvals or calling the protocol contracts. Funds can be loaded across multiple transactions and potentially from multiple sources. To funnel funds from another protocol, there's no need for integration in the 3rd party contracts as long as they are capable of transferring funds to an arbitrary address. Wallets limited to plain ERC-20 transfers and stripped of any web3 functionality can be used to load funds into the protocol. The users of the fully featured wallets don't need to sign opaque calldata blobs that may be harmful or approve the protocol to take their tokens, they only need to make a transfer, which is a simple process with a familiar UX. When the funds are already stored in the puppet assigned to the user, somebody needs to call the protocol so it's notified that the funds were loaded. Depending on the protocol and its API this call may or may not be permissionless potentially making the UX even more convenient with gasless transactions or 3rd parties covering the gas cost. Some protocols don't need the users to specify what needs to be done with the loaded funds or they allow the users to configure that in advance. Most of the protocols using approved transfers to load funds may benefit from using the puppets.
```

mattstam: Looks good 🙏

CodeSandwich: Pushed.

ERCS/erc-7613.md
```diff
@@ -0,0 +1,345 @@
+---
+eip: 7613
+title: Puppet proxy contract
```
SamWilsn, line 3:
```suggestion
title: Puppet Proxy Contract
```

ERCS/erc-7613.md
```diff
@@ -0,0 +1,345 @@
...
+description: The proxy that if called by its deployer, delegates to the address passed to it in the calldata.
```
SamWilsn, line 4:
```suggestion
description: A proxy that, if called by its deployer, delegates to an implementation specified in calldata.
```

ERCS/erc-7613.md
```diff
@@ -0,0 +1,345 @@
...
+discussions-to: https://ethereum-magicians.org/t/eip-7613-puppet-proxy-contract/18482
+status: Draft
+type: Standards Track
+category: ERC
+created: 2024-02-04
+---
+
+## Abstract
+
+A puppet is a contract that when called, acts like an empty account, it doesn't do anything and it has no API. The only exception is that if it's called by the address that deployed it, it delegates the call to the address passed to it in calldata. This gives the deployer the ability to execute any logic they want in the context of the puppet.
```
SamWilsn, line 15:
```suggestion
A puppet is a contract that, when called, acts like an empty account. It doesn't do anything and it has no API, except when it is called by the address that deployed it. In that case, it delegates the call to the address passed to it in calldata. This gives the deployer the ability to execute any logic they want in the context of the puppet.
```

ERCS/erc-7613.md
```diff
@@ -0,0 +1,345 @@
...
... Funds can be loaded across multiple transactions and potentially from multiple sources. To funnel funds from another protocol, there's no need for integration in the 3rd party contracts as long as they are capable of transferring funds to an arbitrary address. Wallets limited to plain ERC-20 transfers and stripped of any web3 functionality can be used to load funds into the protocol. The users of the fully featured wallets don't need to sign opaque calldata blobs that may be harmful or approve the protocol to take their tokens, they only need to make a transfer, which is a simple process with a familiar UX. When the funds are already stored in the puppet assigned to the user, somebody needs to call the protocol so it's notified that the funds were loaded. Depending on the protocol and its API this call may or may not be permissionless potentially making the UX even more convenient with gasless transactions or 3rd parties covering the gas cost. Some protocols don't need the users to specify what needs to be done with the loaded funds or they allow the users to configure that in advance. Most of the protocols using approved transfers to load funds may benefit from using the puppets.
```
SamWilsn, line 21:
```suggestion
Puppets can be used as an alternative to approved transfers when loading funds into the protocol. Any contract and any wallet can transfer the funds to the puppet address assigned to the user without making any approvals or calling the protocol contracts. Funds can be loaded across multiple transactions and potentially from multiple sources. To funnel funds from another protocol, there's no need for integration in the 3rd party contracts as long as they are capable of transferring funds to an arbitrary address. Wallets limited to plain [ERC-20](./eip-20.md) transfers and stripped of any web3 functionality can be used to load funds into the protocol. The users of the fully featured wallets don't need to sign opaque calldata blobs that may be harmful or approve the protocol to take their tokens, they only need to make a transfer, which is a simple process with a familiar UX. When the funds are already stored in the puppet assigned to the user, somebody needs to call the protocol so it's notified that the funds were loaded. Depending on the protocol and its API this call may or may not be permissionless potentially making the UX even more convenient with gasless transactions or 3rd parties covering the gas cost. Some protocols don't need the users to specify what needs to be done with the loaded funds or they allow the users to configure that in advance. Most of the protocols using approved transfers to load funds may benefit from using the puppets.
```

ERCS/erc-7613.md
```diff
@@ -0,0 +1,345 @@
...
+
+The puppet's logic doesn't need to be ever upgraded, to change its behavior the deployer needs to change the address it passes to the puppet to delegate to or the calldata it passes for delegation. The entire fleet of puppets deployed by a single contract can be upgraded by upgrading the contract that deployed them, without using beacons. A nice trick is that the deployer can make the puppet delegate to the address holding the deployer's own logic, so the puppet's logic is encapsulated in the deployer's.
```
SamWilsn, line 23:
```suggestion
The puppet's logic doesn't need to be ever upgraded. To change its behavior the deployer needs to change the address it passes to the puppet to delegate to or the calldata it passes for delegation. The entire fleet of puppets deployed by a single contract can be upgraded by upgrading the contract that deployed them, without using beacons. A nice trick is that the deployer can make the puppet delegate to the address holding the deployer's own logic, so the puppet's logic is encapsulated in the deployer's.
```

ERCS/erc-7613.md
```diff
@@ -0,0 +1,345 @@
...
+// | RETURNDATASIZE | 3D            | [return size] 0 [success]
+// | SWAP2          | 91            | [success] 0 [return size]
+// | PUSH1          | 60 40         | [success branch] [success] 0 [return size]
+// | JUMPI          | 57            | 0 [return size]
+// | REVERT         | FD            |
+// | JUMPDEST       | 5B            | 0 [return size]
+// | RETURN         | F3            |
+```
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
```
SamWilsn, line 126: Please move this paragraph to the top of your specification section (if you intend to use these keywords), or remove it entirely (if you don't.)

ERCS/erc-7613.md
```diff
@@ -0,0 +1,345 @@
...
+}
+```
+
+## Security Considerations
+
+The bytecode is made to resemble clone proxy's wherever it makes sense to simplify auditing.
+
+ABI-encoding the delegation address protects the deployer from being tricked by a 3rd party into calling the puppet and making it delegate to an arbitrary address. Such scenario would only be possible if the deployer called on the puppet a function with the selector `0x00000000`, which as of now doesn't come from any reasonably named function.
+
+Needs discussion.
```
SamWilsn, line 341:
```suggestion
<!-- Needs discussion. -->
```

Use HTML-style comments and the linter will make sure you deal with them.

SamWilsn (changes requested):
This is looking pretty decent, but the biggest blocker is convincing me that this needs to be an ERC.

ERCs shine when multiple independent implementations need to interoperate without communicating beforehand. ERC-20 is a great example: every wallet wants to support every token, and every token wants to support every wallet, but there are simply too many combinations of token+wallet to build bespoke integrations for each pair.

This proposal seems like it would be a better fit for a library (with associated documentation), or for a smart contract wallet implementation.

To get this merged, please explain (in the document itself) who the parties are that need to coordinate on this topic, and why you need a particular implementation instead of just an interface.

ERCS/erc-7613.md
CodeSandwich: Moved.

CodeSandwich:
Hi @SamWilsn, thanks for the review and the forum discussion!

> ERCs shine when multiple independent implementations need to interoperate without communicating beforehand. ERC-20 is a great example

> This proposal seems like it would be a better fit for a library (with associated documentation), or for a smart contract wallet implementation.

Examples of ERCs very similar to this one are 1167, 3448 and 7511. This ERC and all 3 other proxies are standardized by specifying the exact bytecodes of their init codes, with their interfaces and usage emerging from these bytecodes. None of these ERCs define how these raw bytecodes are handled in practice, it's up to the designers of tools and libraries. For example searching on GitHub for the ERC-1167 init code yields over 17K results, and each of them is a different tool, library or a piece of ad-hoc code for handling the proxies. Such proliferation is possible because the standard enforces only the raw bytecode giving a lot of elasticity, anybody can easily write their own library using the preferred language and framework. If they only agree to use the same bytecode, they can derive each others' CREATE2 addresses, they know how to deploy the proxies and how to use them.

> please explain (in the document itself) who the parties are that need to coordinate on this topic, and why you need a particular implementation instead of just an interface.

This is touched on in the "motivation" section:

`Because the bytecode is not compiled, it can be reliably deployed under a predictable CREATE2 address regardless of the compiler version.`

There's a scenario described that is only feasible because the bytecode is standardized and known by the actor sending the funds. It would be much harder to work with if the puppet bytecode wasn't fixed because depending on the compilation process it would generate a different CREATE2 address derived from a different input:

`If the puppet is deployed under a predictable address, e.g. by using the user's address as the CREATE2 salt, the puppet may not even need to be deployed before funds are sent to its address.`

There's another scenario which is practical only because the users can rely on the fixed bytecode resulting in a predictable CREATE2 address:

`It can be also used as a full replacement of the CREATE3 factory by using a puppet deployed using CREATE2 to deploy arbitrary code using plain CREATE.`

Is this described clearly and deeply enough or does it need more explanation?

SamWilsn:
> This is touched on in the "motivation" section: [...]

It's clear why you've chosen bytecode over solidity. The document does a good job of explaining that. What's missing is why you need deterministic addresses, and why a third party (who isn't the puppeteer) would care what the address ends up being.

Essentially you need to make the case for why you can't just specify an interface and you _need_ to specify an implementation.

I think you could get away with rewording this chunk and putting it in the document:

> Such proliferation is possible because the standard enforces only the raw bytecode giving a lot of elasticity, anybody can easily write their own library using the preferred language and framework. If they only agree to use the same bytecode, they can derive each others' CREATE2 addresses, they know how to deploy the proxies and how to use them.

CodeSandwich:
I pushed explicit interoperability motivation for standardizing the exact creation code. @SamWilsn is this good and does it provide the necessary information?

CodeSandwich:
Is there anything blocking this PR? What are the next steps?

SamWilsn (approved):

PR #250, opened 2024-02-10
Add ERC: Pre-Authorized Service Payment Protocol

bizliaoyuan:
This EIP proposes a decentralized mechanism for users to pre-authorize specific service providers to deduct funds from their accounts without requiring explicit approval for each transaction. It aims to improve transaction efficiency and user experience in decentralized applications (DApps) and services.

### DIFF
--- a/ERCS/erc-7620.md
+++ b/ERCS/erc-7620.md
@@ -66,7 +66,7 @@ By introducing this protocol, the aim is to optimize the utilization of authoriz
 
 ## Reference Implementation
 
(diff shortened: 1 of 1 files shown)


ERCS/erc-7620.md
```diff
@@ -0,0 +1,77 @@
+---
+eip: 7620
```
abcoathup, line 2: EIP numbers are assigned by EIP/RIP/ERC editors/associates.  7620 has already been assigned to an EIP.

PR #283, opened 2024-02-28
Update ERC-165: Clarify misleading usage of `pure` attribute

xaler5:
The example provided after the phrase still uses `view` attribute. Either the example or the description before that should change.

If the example is changed, it would be ambiguous whether the ERC-165 imposes restriction on the usage of `view` or `pure` for the `supportsInterface` function.

### DIFF
--- a/ERCS/erc-165.md
+++ b/ERCS/erc-165.md
@@ -198,7 +198,7 @@ contract Lisa is ERC165MappingImplementation, Simpson {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


Jean68172 (approved):

SamWilsn:
As discussed on EIPIP (ethcatherders/EIPIP#336), we don't feel like the clarification justifies changing a Final ERC.

PR #301, opened 2024-03-06
Update ERC-5573: ERC-5573 typos

chris13524:
Fixes some details:
- Missing trailing slashes in some URIs; test cases didn't match
- Duplicate `:` in ReCaps URI spec
- Javascript -> JavaScript
- Misspelling of receive

### DIFF
--- a/ERCS/erc-5573.md
+++ b/ERCS/erc-5573.md
@@ -85,7 +85,7 @@ A ReCap Capability is identified by their ReCap URI that resolves to a ReCap Det
 
 ##### ReCap URI Scheme
 
-A ReCap URI starts with `urn:recap:` followed by `:` and the unpadded base64url-encoded payload of the ReCap Details Object. Note, the term base64url is defined in RFC4648 - Base 64 Encoding with URL and Filename Safe Alphabet. If present, a Recap URI MUST occupy the final entry of the SIWE resource list.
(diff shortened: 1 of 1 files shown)


bumblefudge:
NB @skgbafa

bumblefudge:
also, should
```
- `att`: (CONDITIONAL) If present, `att` MUST be a JSON object where each key is a URI and each value is an object containing Ability Strings as keys and a corresponding value which is an array of qualifications to the action (i.e. a restriction or requirement). The keys of the object MUST be ordered lexicographically.
```
read
```
- `att`: (CONDITIONAL) If present, `att` MUST be a JSON object where each of 1 or more key(s) is a URI and each value is an object containing Ability Strings as keys and a corresponding value which is an array of qualifications to the action (i.e. a restriction or requirement). The keys of the object MUST be ordered lexicographically.
```
or
```
- `att`: (CONDITIONAL) If present, `att` MUST be a JSON object where each of 0 or more key(s) is a URI and each value is an object containing Ability Strings as keys and a corresponding value which is an array of qualifications to the action (i.e. a restriction or requirement). The keys of the object MUST be ordered lexicographically.
```
?
I think as written `att` could be present as an empty object but I can see implementers assuming one or the other interpretation and failing to interoperate.

llbartekll (approved):

awoie (approved):

PR #302, opened 2024-03-07
Update ERC-7579: Update erc-7579 interface param names

plusminushalf:
Fixes typo, rename `moduleType` to `moduleTypeId`

This change can be verified from the reference implementation - https://github.com/erc7579/erc7579-implementation/blob/main/src/MSAAdvanced.sol#L249


SamWilsn:
I believe these changes have been made to the document. Please re-open if that isn't the case.

PR #303, opened 2024-03-08
Update ERC-5573: require single recap in resources

chris13524:
"a Recap URI MUST occupy the final entry of the SIWE resource list" is somewhat ambiguous as to if there can be more (implied ignored) `urn:recap` URIs elsewhere in the resource list. Making this more explicit that there must be a single `urn:recap` URI.

Also adjust spacing and indent to be consistent with elsewhere in the spec & fix some caps typos.

### DIFF
--- a/ERCS/erc-5573.md
+++ b/ERCS/erc-5573.md
@@ -85,7 +85,7 @@ A ReCap Capability is identified by their ReCap URI that resolves to a ReCap Det
 
 ##### ReCap URI Scheme
 
-A ReCap URI starts with `urn:recap:` followed by the unpadded base64url-encoded payload of the ReCap Details Object. Note, the term base64url is defined in RFC4648 - Base 64 Encoding with URL and Filename Safe Alphabet. There MUST be at most one ReCap URI in the SIWE resource list and if present, it MUST occupy the final entry.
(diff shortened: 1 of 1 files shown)


ERCS/erc-5573.md
```diff
@@ -85,7 +85,7 @@ A ReCap Capability is identified by their ReCap URI that resolves to a ReCap Det
 
 ##### ReCap URI Scheme
 
-A ReCap URI starts with `urn:recap:` followed by the unpadded base64url-encoded payload of the ReCap Details Object. Note, the term base64url is defined in RFC4648 - Base 64 Encoding with URL and Filename Safe Alphabet. If present, a Recap URI MUST occupy the final entry of the SIWE resource list.
+A ReCap URI starts with `urn:recap:` followed by the unpadded base64url-encoded payload of the ReCap Details Object. Note, the term base64url is defined in RFC4648 - Base 64 Encoding with URL and Filename Safe Alphabet. There MUST be at most one ReCap URI in the SIWE resource list and if present, it MUST occupy the final entry.
```
chris13524, line 88:
```suggestion
A ReCap URI starts with `urn:recap:` followed by the unpadded base64url-encoded payload of the ReCap Details Object. Note, the term base64url is defined in RFC4648 - Base 64 Encoding with URL and Filename Safe Alphabet. There MUST be at most one ReCap URI in the SIWE resource list and, if present, it MUST occupy the final entry.
```

ganchoradkov (approved):
Always nice to have explicit rules how something should work otherwise it's up to one's interpretation 💯

awoie (commented):
What are your thoughts @chunningham?

chris13524:
I still think this is worth addressing/merging

PR #305, opened 2024-03-09
Add ERC: Fractionally Represented Non-Fungible Token

0xacme:

 - The build passes.

### DIFF
--- a/ERCS/erc-7651.md
+++ b/ERCS/erc-7651.md
@@ -13,7 +13,7 @@ requires: 20, 165, 721
 
 ## Abstract
 
-This proposal introduces a standard for fractionally represented non-fungible tokens, allowing NFTs to be managed and owned fractionally within a single contract. This approach enables [ERC-721](./erc-721.md) NFTs to coexist with [ERC-20](./erc-20.md) tokens seamlessly, enhancing liquidity and access without dividing the NFT itself, and without the need for an explicit conversion step. The standard includes mechanisms for both fractional and whole token transfers, approvals, and event emissions.
+This proposal introduces a standard for fractionally represented non-fungible tokens, allowing NFTs to be managed and owned fractionally within a single contract. This approach enables [ERC-721](./eip-721.md) NFTs to coexist with [ERC-20](./eip-20.md) tokens seamlessly, enhancing liquidity and access without dividing the NFT itself, and without the need for an explicit conversion step. The standard includes mechanisms for both fractional and whole token transfers, approvals, and event emissions.
 
 ## Motivation
 
@@ -27,13 +27,13 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 
 ### Fractionally Represented Non-Fungible Token Interface
 
-All [ERC-7651](./erc-7651.md) compliant contracts MUST implement the ERC-7651 and [ERC-165](./erc-165.md) interfaces.
+All [ERC-7651](./eip-7651.md) compliant contracts MUST implement the ERC-7651 and [ERC-165](./eip-165.md) interfaces.
 
 Compliant contracts MUST emit fractional Approval or Transfer events on approval or transfer of tokens in fractional representation.
 
 Compliant contracts MUST additionally emit non-fungible ApprovalForAll, Approval or Transfer on approval for all, approval, and transfer in non-fungible representation.
 
-Note that this interface draws from existing [ERC-721](./erc-721.md) and [ERC-20](./erc-20.md) standards, but is not fully backwards compatible with either.
+Note that this interface draws from existing [ERC-721](./eip-721.md) and [ERC-20](./eip-20.md) standards, but is not fully backwards compatible with either.
 
 ```solidity
 interface IERC7651 is IERC165 {
@@ -199,7 +199,7 @@ interface IERC165 {
 
 ### Fractionally Represented Non-Fungible Token Metadata Interface
 
-This is a RECOMMENDED interface, identical in definition to the [ERC-721](./erc-721.md) Metadata Interface. Rather than using this interface directly, a distinct metadata interface should be used here to avoid confusion surrounding ERC-721 inheritance.
+This is a RECOMMENDED interface, identical in definition to the [ERC-721](./eip-721.md) Metadata Interface. Rather than using this interface directly, a distinct metadata interface should be used here to avoid confusion surrounding ERC-721 inheritance.
 
 ```solidity
 /// @title ERC-7651 Fractional Non-Fungible Token Standard, optional metadata extension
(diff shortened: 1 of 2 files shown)


ERCS/erc-draft_fractional_nft.md
```diff
@@ -0,0 +1,348 @@
+---
+title: Fractionally Represented Non-Fungible Token Standard
+description: A standard for fractionally represented non-fungible tokens.
+author: Acme (@0xacme), Calder (@caldereth)
+discussions-to: <URL>
```
xinbenlv, line 5: Please open a thread in Ethereum Magician foruem

ERCS/erc-draft_fractional_nft.md
```diff
@@ -0,0 +1,348 @@
+---
```
abcoathup, line 1:
```suggestion
---
eip: 7651
```

ERCS/erc-draft_fractional_nft.md
```diff
@@ -0,0 +1,348 @@
...
+title: Fractionally Represented Non-Fungible Token Standard
```
abcoathup, line 2:
```suggestion
title: Fractionally Represented Non-Fungible Token
```
ERCs don't need to have standard in the title, its implied.

ERCS/erc-draft_fractional_nft.md
0xacme: Added in latest commit. Thanks!

ERCS/erc-7651.md
```diff
@@ -0,0 +1,349 @@
...
+status: Draft
+type: Standards Track
+category: ERC
+created: 2024-03-05
+requires: 20, 165, 721
+---
+
+## Abstract
+
+This proposal introduces a standard for fractionally represented non-fungible tokens, allowing NFTs to be managed and owned fractionally within a single contract. This approach enables [ERC-721](./eip-721.md) NFTs to coexist with [ERC-20](./eip-20.md) tokens seamlessly, enhancing liquidity and access without dividing the NFT itself, and without the need for an explicit conversion step. The standard includes mechanisms for both fractional and whole token transfers, approvals, and event emissions.
```
xinbenlv, line 16:
```suggestion
This proposal introduces a standard for fractionally represented non-fungible tokens, allowing NFTs to be managed and owned fractionally within a single contract. This approach enables NFTs to coexist with fungible token seamlessly, enhancing liquidity and access without dividing the NFT itself, and without the need for an explicit conversion step. The standard includes mechanisms for both fractional and whole token transfers, approvals, and event emissions. This project takes inspiration from [ERC-721](./eip-721.md) and [ERC-20](./eip-20.md), but is not fully compatible with either of them.
```

xinbenlv: Let's ensure it's clear that this ERC doesn't comply with ERC-20 and ERC-721 so readers make more informed decision when adopting

ERCS/erc-7651.md
```diff
@@ -0,0 +1,349 @@
...
+
+### Fractionally Represented Non-Fungible Token Interface
+
+All [ERC-7651](./eip-7651.md) compliant contracts MUST implement the ERC-7651 and [ERC-165](./eip-165.md) interfaces.
+
+Compliant contracts MUST emit fractional Approval or Transfer events on approval or transfer of tokens in fractional representation.
+
+Compliant contracts MUST additionally emit non-fungible ApprovalForAll, Approval or Transfer on approval for all, approval, and transfer in non-fungible representation.
+
+Note that this interface draws from existing [ERC-721](./eip-721.md) and [ERC-20](./eip-20.md) standards, but is not fully backwards compatible with either.
```
xinbenlv, line 36:
```suggestion
Note that this interface draws similar named functions and events from existing [ERC-721](./eip-721.md) and [ERC-20](./eip-20.md) standards, but is not fully backwards compatible with either.
```

ERCS/erc-7651.md
```diff
@@ -0,0 +1,349 @@
...
+    /// @return `true` if the contract implements `interfaceID` and
+    ///         `interfaceID` is not 0xffffffff, `false` otherwise
+    function supportsInterface(bytes4 interfaceID_) external view returns (bool);
+}
+```
+
+### Fractionally Represented Non-Fungible Token Metadata Interface
+
+This is a RECOMMENDED interface, identical in definition to the [ERC-721](./eip-721.md) Metadata Interface. Rather than using this interface directly, a distinct metadata interface should be used here to avoid confusion surrounding ERC-721 inheritance.
+
```
xinbenlv, line 203: I think we shall mention the ERC165 interfaceId will therefore be the same

ERCS/erc-7651.md
```diff
@@ -0,0 +1,349 @@
...
+The `balanceOf` function as defined in both [ERC-20](./eip-20.md) and [ERC-721](./eip-721.md) standards varies, in practice, to represent either fractional or whole token ownership respectively. Given fractional non-fungible tokens should adhere to an underlying fractional representation, it follows that this function should return a balance in that representation. This does, however, imply that fractional NFT contracts cannot fully adhere to the `balanceOf` specification provided by ERC-721.
+
+### Success Return Values
+
+The `transfer` and `approve` functions both return a boolean value indicating success or failure. This is non-standard for the [ERC-721](./eip-721.md) specification, though is standard for [ERC-20](./eip-20.md). Fractional non-fungible tokens adhere to a returned boolean value to meet minimum expectations for the ERC-20 standard, acknowledging that this deviates from a state of ideal backwards compatibility.
+
+## Security Considerations
+
+Needs discussion.
```
xinbenlv, line 345: I think given some interface and events are identical, we shall consider a security issue when an ERC-721 or ERC-20 client (dApp/wallet/DEX) will unintentionally misinterpret a contract of this ERC as ERC-721 or ERC-20, such as Uniswap, what will happen. Given the potential financial implication, please include a least a placeholder for such discussion before checking in the Draft

ERCS/erc-7651.md
sullof: A solution used in other cases is to enforce the interfaceId using a magic number, but I think that it would introduce more issues. I suggest that the author adds an extra function that does nothing but is there only to produce a different interfaceId, like
```
function isIERC7651Metadata() external pure returns(bool);
```
that can only return `true`.

ERCS/erc-7651.md
sullof: I agree. The problem is particularly relevant when discussing the fungible component because ERC-20 does not extend ERC165, so exchanges may assume that an ERC7651 is actually an ERC20. I see fewer issues with ERC721 because exchanges like OpenSea check if the token supports the IERC721 interface.

ERCS/erc-7651.md
0xacme: Sounds good, updated in the latest commit!

ERCS/erc-7651.md
0xacme: I decided to just add clarification for now. Given this is a draft, I figure it would be better to discuss the possibility of introducing some means of distinguishing the interfaceId's on the discussion thread.

ERCS/erc-7651.md
0xacme: Totally agree, we've given a lot of consideration to this, though I know there are still a few edgecases. I can kick off discussion here shortly by raising a few. In the meantime, added a more direct placeholder here that calls out the need for discussion surrounding interface misinterpretation.

ERCS/erc-7651.md
0xacme: Updated in latest commit.

xinbenlv (approved):

PR #328, opened 2024-03-17
Update ERC-7521: Add Prohibited Address Description

Karotoka:
Adding the description of the functions included in the PR [here](https://github.com/essential-contributions/ERC-7521/pull/140#issue-2190864203).

Attempted to mirror the existing structure but feel free to edit or reject if inappropriate.

### DIFF
--- a/ERCS/erc-7521.md
+++ b/ERCS/erc-7521.md
@@ -94,6 +94,17 @@ function generalizedIntentDelegateCall
     external returns (bool);
 ```
 
+The core interface allows the setting of Prohibited Addresses which will be excluded from intent.sender and intent.intentData:
+
+```solidity
+function setProhibitedAddresses
+    (address[] calldata addresses, UserIntent calldata intent, bytes32 intentHash)
+    external returns (uint256)
+
+function isProhibitedAddress
+    (UserIntent calldata intent, bytes32 intentHash)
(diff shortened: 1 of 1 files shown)


pixelcircuits:
This would be better suited as an intent standard for ERC-7521 rather than being baked into the entrypoint itself. This way it's up to users to opt in and include a segment on their intent with these extra constraints.

Also, the full list of possibly touched contracts isn't fully known by just looking at intent data. For example, imagine a wrapper contract that just forwards calls to another contract, with the intent of obscuring the target contracts address.

Karotoka:
This makes sense to me as the functionality is designed to be optional.

Can you direct me towards the correct repo that this should live in and I will submit it there? I'm a poor wordcell struggling through GitHub and had a hunch this may have been the wrong spot.

As for the possibly touched contracts - yes, I understand that structure and packing of `intentData` may cause problems extracting prohibited addresses. I don't think this can be resolved in the ERC7521 standard itself. The goal is to set a framework that users/builders can iterate on depending on their specs.

pixelcircuits:
I think it makes the most sense to write an example intent standard contract and add it [here](https://github.com/essential-contributions/ERC-7521/tree/master/src/standards). At least one that whitelists a solver by checking `msg.origin`.

PR #342, opened 2024-03-23
Update ERC-2098: variable v is not right in example.

chen4903:
I have read and tried using the example of [EIP-2098](https://eips.ethereum.org/EIPS/eip-2098), but I have found some problems. I poc this error in solidity: 
```solidity
// SPDX-License-Identifier: GPL-3.0

pragma solidity =0.8.23;

contract TestEIP2098{
    function to_compact(uint256 r, uint256 s, uint256 v) public pure returns(uint256, uint256){
        // example 1
        // input:
        // r        = 0x68a020a209d3d56c46f38cc50a33f704f4a9a10a59377f8dd762ac66910e9b90
        // s        = 0x7e865ad05c4035ab5792787d4a0297a43617ae897930a6fe4d822b8faea52064
        // v        = 27
        // output:
        // r                  = 0x68a020a209d3d56c46f38cc50a33f704f4a9a10a59377f8dd762ac66910e9b90
        // yParityAndS here   = 0xfe865ad05c4035ab5792787d4a0297a43617ae897930a6fe4d822b8faea52064 (true)
        // yParityAndS in EIP = 0x7e865ad05c4035ab5792787d4a0297a43617ae897930a6fe4d822b8faea52064 (false)

        // example 2
        // input:
        // r        = 0x9328da16089fcba9bececa81663203989f2df5fe1faa6291a45381c81bd17f76
        // s        = 0x139c6d6b623b42da56557e5e734a43dc83345ddfadec52cbe24d0cc64f550793
        // v        = 28
        // output:
        // r                  = 0x9328da16089fcba9bececa81663203989f2df5fe1faa6291a45381c81bd17f76
        // yParityAndS here   = 0x139c6d6b623b42da56557e5e734a43dc83345ddfadec52cbe24d0cc64f550793 (true)
        // yParityAndS in EIP = 0x939c6d6b623b42da56557e5e734a43dc83345ddfadec52cbe24d0cc64f550793 (false)

        // But if you set v to 28 in example 1, you can get the same result in EIP. 
        // The same as example2, if you set v to 27, you can get the same result in EIP. 

        return (r, (v << 255) | s);
    } 
}
```
So I think there is some problem with variable `v` in the example of EIP-2098. Modify the value of `v`: 27=>28 in example 1, 28=>27 in example2.


ricmoo:
Thanks for the ping. I’ll look into it shortly. I
At first look, it looks like they are correct.

SamWilsn:
How about at second look? :innocent:

lightclient:
@chen4903 can you please provide a program that we can verify the proposed signature is correct.

PR #348, opened 2024-03-26
Add ERC: AI Agent NFTs

marleymarl:

 - The build passes.


ERCS/erc-7661.md
```diff
@@ -0,0 +1,103 @@
+---
+eip: 7661
```
abcoathup, line 2:
```suggestion
eip: 7662
```
Assigning next sequential EIP/ERC/RIP number.  Numbers are assigned by editors & associates, please don't self assign.

ERCS/erc-7661.md
```diff
@@ -0,0 +1,103 @@
...
+title: AI Agent NFTs
+description: A specification for NFTs that represent AI Agents.
+author: Greg Marlin (@0marleymarl)
+discussions-to: https://ethereum-magicians.org/t/add-erc-ai-agent-nfts/19371
```
abcoathup, line 6:
```suggestion
discussions-to: https://ethereum-magicians.org/t/erc-7662-ai-agent-nfts/19371
```

ERCS/erc-7661.md
marleymarl: Apologies - the green checkmark seeker in me took over. I've updated the filename correctly.

ERCS/erc-7662.md
```diff
@@ -0,0 +1,103 @@
...
+## Specification
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
+
+
+All ERC-XXXX compliant contracts MUST implement the standard [ERC-721](./eip-721.md) functionality for minting and transferring NFTs, and MUST additionally implement this standard's Agent struct
+
+```solidity
+   
+struct Agent {
```
xinbenlv, line 32: Can you elaborate why this needs to be stored onchain as opposed to be stored as ERC721 Metadata?

marleymarl: The primary reason for not storing them directly in the metadata is for the case if someone didn’t encrypt the prompts. If they are stored onchain someone could still discover them but it would be a lot harder for a casual user, for example looking at the NFT on a marketplace. Many marketplaces list a link to the TokenURI that can easily be looked up, and thus the casual user could easily just grab the prompts and not buy the NFT, whereas if it's stored onchain they can't. If stored onchain unencrypted, a determined user could e.g. look up the info on a block explorer or look it up from a self-built app but considering most of these I expect to be priced fairly low then it’s easier to just pay for it, for most cases. For high-value agents, then those sellers should obviously opt for the encryption.

Since I’m recommending encryption of the prompts anyway, and will be using that in the reference implementation, it could make sense to adapt the Metadata Standard for this since encrypted prompt URIs could be exposed in the metadata, however I’ve tried to avoid enforcing the need for encryption in the standard if that makes sense. 

Another consideration for putting it onchain is to be a foundation for further future extensibility in terms of building out new data features to that struct (that can be accessed with the same call i.e. Agents[id]) and opening up the possibility of other smart contracts interacting with the data in that struct.

ERCS/erc-7662.md
```diff
@@ -0,0 +1,103 @@
+---
+eip: 7662
+title: AI Agent NFTs
+description: A specification for NFTs that represent AI Agents.
+author: Greg Marlin (@0marleymarl)
```
abcoathup, line 5:
```suggestion
author: Greg Marlin (@marleymarl)
```
This should be your GitHub username

xinbenlv: Good catch @abcoathup  LOL

marleymarl: Was going to say, nice catch, thanks - fixed.

ERCS/erc-7662.md
```diff
@@ -0,0 +1,273 @@
+---
+eip: 7662
+title: AI Agent NFTs
+description: A specification for NFTs that represent AI Agents.
```
SamWilsn, line 4: Your description doesn't include any new information that wasn't in your title, and it has a bit of fluff ("A specification for".) You should use the description to elaborate on the ideas introduced in the title as succinctly as possible.

For example, what is an AI Agent?

ERCS/erc-7662.md
```diff
@@ -0,0 +1,273 @@
...
+status: Draft
+type: Standards Track
+category: ERC
+created: 2024-03-26
+requires: 721
+---
+
+## Abstract
+
+This proposal introduces a standard for AI agent NFTs. In order for AI Agents to be created and traded as NFTs it doesn't make sense to put the prompts in the token metadata, therefore it requires a standard custom struct. It also needs doesn't make sense to store the prompts directly onchain as they can be quite large, therefore this standard proposes they be stored as decentralized storage URLs. This standard also proposes two options on how this data should be made private to the owner of the NFT, with the favored implementation option being encrypting the data using custom contract parameters for decryption that decrypt only to the owner of the NFT.
```
SamWilsn, line 16:
```suggestion
This proposal introduces a standard for AI agent NFTs. When AI Agents are created and traded as NFTs, it doesn't make sense to put the prompts in the token metadata, therefore it requires a standard custom struct. It also doesn't make sense to store the prompts directly onchain as they can be quite large, therefore this standard proposes they be stored as decentralized storage URLs. This standard also proposes two options on how this data should be made private to the owner of the NFT, with the favored implementation option being encrypting the data using custom contract parameters for decryption that decrypt only to the owner of the NFT. 
```

SamWilsn, line 16: Your abstract does a good job of covering a technical overview, but it doesn't really explain the goal of the proposal very well. After reading the abstract, the reader should have _both_ a technical overview and an idea of the problem the proposal is trying to solve.

ERCS/erc-7662.md
```diff
@@ -0,0 +1,273 @@
...
+
+The creation and trading of AI Agent NFTs are a natural fit and offer the potential for an entirely new onchain market. This requires some custom data to be embedded in the NFT through a custom struct and this needs to be standardized so that any marketplace or AI Agent management product, among others, know how to create and parse AI Agent NFTs.  
+
+
+## Specification
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
+
+
+All ERC-XXXX compliant contracts MUST implement the standard [ERC-721](./eip-721.md) functionality for minting and transferring NFTs, and MUST additionally implement this standard's Agent struct
```
SamWilsn, line 28: It is unclear how a contract should "implement" a struct. Normally you implement an interface.

ERCS/erc-7662.md
```diff
@@ -0,0 +1,273 @@
...
+```
+
+and MUST implement the mapping between NFT Token ID and its Agent information.
+
+```solidity
+   
+mapping(uint => Agent) public Agents;
+```
+
+It is RECOMMENDED that this mapping is public and that the URIs for User Prompt and System Prompt are made private through encryption with decryption set to the holder of the NFT via custom contract parameters set during encryption, such as those provided by Lit Protocol.
```
SamWilsn, line 49: "Lit Protocol" is an external link, and we have a fairly strict policy on those. Please include a specific encryption scheme or explicitly leave it undefined. The scheme can be included in this proposal, another proposal, or in a permitted external origin.

Please see [EIP-1](https://eips.ethereum.org/EIPS/eip-1#linking-to-external-resources) for allowed origins.

ERCS/erc-7662.md
```diff
@@ -0,0 +1,273 @@
...
+   
+function mintAgent(address _recipient, string memory _name, string memory _description, string memory _model, string memory _userPromptURI, string memory _systemPromptURI, string memory _imageURI, string memory _category, string memory _tokenURI) public {
+        tokenIds++;
+        bool _promptsEncrypted = false;
+        Agents[tokenIds] = Agent(_name, _description, _model, _userPromptURI, _systemPromptURI, _imageURI, _category, _promptsEncrypted);
+
+        _mint(_recipient, tokenIds);
+        collectionIds[_recipient].push(tokenIds);
+        _setTokenURI(tokenIds, _tokenURI);
+        emit AgentCreated(_name, _description, _model, _category, _recipient, tokenIds);
```
SamWilsn, line 65: Please do not specify the implementation of functions in an ERC. Describe the externally visible behaviour of the function, but the actual code should be left to implementers.

ERCS/erc-7662.md
```diff
@@ -0,0 +1,273 @@
...
+
+### Reference Implementation
+
+This is being currently implemented in a CEO AI product for creating, managing and using AI Agents Onchain through a DApp interface. In this implementation, Lit Protocol is being used to encrypt the prompts using custom EVMContractParameters that only decrypt for the holder of the NFT and using Arweave to store the URLs of this encrypted data. To facilitate that and make DApp handling easier, some parameters were added to Agent and the addEncryptedPrompts function is added that enables adding the encrypted prompt URIs after first minting the NFT (as the tokenId of the NFT is needed for setting the encryption/decryption conditions)
+
+```solidity
+
+// SPDX-License-Identifier: MIT
```
SamWilsn, line 114: Code in the proposal itself must be licensed under `CC0-1.0`. You may put MIT code in the assets directory if you wish.

ERCS/erc-7662.md
```diff
@@ -0,0 +1,273 @@
...
+    
+}
+
+
+```
+
+
+
+## Security Considerations
+
```
SamWilsn, line 269:
```suggestion

<!-- TODO -->
```

ERCS/erc-7662.md
marleymarl: Good point - will add, thanks.

ERCS/erc-7662.md
marleymarl: Better - agreed, thanks.

marleymarl: Got it, will add that, thanks again.

ERCS/erc-7662.md
marleymarl: Will make that update, thanks.

ERCS/erc-7662.md
marleymarl: Understood. Will consider and update.

SamWilsn:
Looking good so far; still a few comments to address.

SamWilsn (approved):
Do try and address the remaining comments while in draft.

PR #358, opened 2024-04-02
Add ERC: Temporary Approval Extension for ERC-20

byshape:
Among all cases of `ERC-20` token transactions, a popular one is when smart contracts approve token spending to other contracts. Often tokens are approved for only one transaction.

Following the `ERC-20` standard, if a smart contract wants to approve the spending of tokens to another smart contract for only one transaction, this causes the allowance saved in storage to be updated and retrieved.

Token allowances utilising `EIP-1153` transient storage are a cheaper alternative to the regular storage allowances.


ERCS/eip-draft-tae-erc20.md
```diff
@@ -0,0 +1,55 @@
+---
+title: Transient Approval Extension for ERC-20
+description: An interface for ERC-20 approvals via EIP-1153 transient storage
+author: Xenia Shape (@byshape), Mikhail Melnik (@ZumZoom)
+discussions-to: # TODO: update
```
abcoathup, line 5: Please create a topic in https://ethereum-magicians.org/ to discuss this draft ERC.
Include a link in the topic to this PR.

byshape: Thank you for your comment, all done.

ZumZoom:
Commit [85f7d71](https://github.com/ethereum/ERCs/pull/358/commits/85f7d71c1f3bb212bacd0568f80f0076f59ba593) removes EIP-1153 reference from the description section but it seems that EIP Walidator needs to be fixed instead:

https://github.com/ethereum/ERCs/actions/runs/8536634399/job/23385573486#step:3:14

`Error: error[preamble-refs-description]: unable to read file ``erc-1153.md``: Io: JsValue(Error: ENOENT: no such file or directory, open 'ERCS/erc-1153.md'`

It seems that it disallows referencing any EIP in the description section.

ERCS/erc-7674.md
```diff
@@ -0,0 +1,56 @@
...
+
+Compliant contracts MUST implement 1 new function in addition to `ERC-20`:
+```solidity
+function transientApprove(address spender, uint256 value) public returns (bool success)
+```
+Call to `transientApprove(spender, value)` allows `spender` to withdraw within the same transaction from `msg.sender` multiple times, up to the `value` amount.
+
+Compliant contracts MUST use the transient storage `EIP-1153` to keep the temporary allowance. For each `owner` and `spender`, the slot MUST be uniquely selected to avoid slot collision. Each slot index SHOULD be derived from the base slot index for transient allowances, `owner` and `spender` addresses. Slot MAY be derived as `keccak256(spender . keccak256(owner . p))` where `.` is concatenation and `p` is `keccak256` from the string uniquely defining transient allowances in the namespace of the implementing contract.
+
+Compliant contracts MUST add a transient allowance check to the `transferFrom` function. The permanent allowance can only be spent after the temporary allowance has been exhausted.
```
Amxx, line 34:
```suggestion
Compliant contracts MUST add a transient allowance check to the `transferFrom` function. The permanent allowance SHOULD only be spent after the temporary allowance has been exhausted.
```

ERCS/erc-7674.md
```diff
@@ -0,0 +1,56 @@
...
+
+Compliant contracts MUST add a transient allowance to the permanent one when returning the allowed amount to spend in the `allowance` function.
```
Amxx, line 36:
```suggestion
Compliant contracts MUST add a transient allowance to the permanent one when returning the allowed amount to spend in the `allowance` function. In case the sum of the transient and permanent allowance overflow, `type(uint256).max` SHOULD be returned.
```

Amxx:
IMO "transient" is a technical name that refers to the underlying mechanism used here, but should not be exposed in the ABI or in the "docuementation" or this feature. I propose using "temporary allowance" instead of "transient allowance", as I believe this naming would be more user-friendly for people that have no knowledge of EIP-1153

Amxx:
If we have consensus, we should state that:
- No event is required when setting a temporary/transient allowance
- This ERC does not break ERC20's transferFrom, because setting a temporary/transient allowance falls in the scope of "explicitelly authorizing the transfer"

byshape:
Mikhail @ZumZoom and I have collected some opinions on naming transient/temporary, with 99% leaning towards temporary as the friendlier option. So I replaced uses wherever it was appropriate.

ERCS/erc-7674.md
```diff
@@ -0,0 +1,58 @@
...
+
+The key words "MUST", "SHOULD", "MAY" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
+
+Compliant contracts MUST implement 1 new function in addition to `ERC-20`:
+```solidity
+function temporaryApprove(address spender, uint256 value) public returns (bool success)
+```
+Call to `temporaryApprove(spender, value)` allows `spender` to withdraw within the same transaction from `msg.sender` multiple times, up to the `value` amount.
+
+Compliant contracts MUST use the transient storage `EIP-1153` to keep the temporary allowance. For each `owner` and `spender`, the slot MUST be uniquely selected to avoid slot collision. Each slot index SHOULD be derived from the base slot index for temporary allowances, `owner` and `spender` addresses. Slot MAY be derived as `keccak256(spender . keccak256(owner . p))` where `.` is concatenation and `p` is `keccak256` from the string uniquely defining temporary allowances in the namespace of the implementing contract.
```
ernestognw, line 32: > Compliant contracts MUST use the transient storage `EIP-1153` to keep the temporary allowance.

Is this enforceable, though? I agree with @Amxx that "transient" shouldn't be exposed in the ABI or documentation, which makes me think that requiring the use of EIP-1153 is not even needed for the interface and the semantics of `temporaryApprove`.

ERCS/erc-7674.md
```diff
@@ -0,0 +1,58 @@
...
+
+All functionality of the `ERC-20` standard is backward compatible except for the `allowance` function.
+
+## Reference Implementation
+
+TBD
+
+## Security Considerations
+
+The method of deriving slot identifiers to store temporary allowances must avoid collision with other transient storage slots.
```
ernestognw, line 54: I acknowledge I'm biased, but why not to suggest using https://eips.ethereum.org/EIPS/eip-7201 right away to derive the slot? It has the characteristics we're looking for imo.

ernestognw:
I left a review, thanks for pointing out @Amxx!

The ERC looks good, but since it's defining an interface for a temporary approval, I don't see why to restrict this to transient storage. While it makes sense, I think the benefits of this kind of approval can be achieved regardless of transient storage support in a context where using storage may not be a big deal (e.g. an L2)

I would suggest generalizing the interface and its semantics while keeping the suggestions for using transient storage if EIP-1153 is available, but I understand the intention of making a transient-approval ERC 😅

ERCS/erc-7674.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Specification
+
+The key words "MUST", "SHOULD", "MAY" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
+
+Compliant contracts MUST implement 1 new function in addition to `ERC-20`:
+```solidity
+function temporaryApprove(address spender, uint256 value) public returns (bool success)
+```
+Call to `temporaryApprove(spender, value)` allows `spender` to withdraw within the same transaction from `msg.sender` multiple times, up to the `value` amount.
```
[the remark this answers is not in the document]
ernestognw: I think this should be updated regardless.

ernestognw (commented):
Left some comments in case you're open to generalize it. Feel free to disregard.

byshape:
@ernestognw Thank you for your comments!

> While it makes sense, I think the benefits of this kind of approval can be achieved regardless of transient storage support in a context where using storage may not be a big deal (e.g. an L2)

Do you already have any ideas on how this could be implemented?

ernestognw:
A quick thought was another mapping with approvals per block number (or timestamp), in this way it's not needed to clean up the value at the end since it will just get invalidated immediately.

Here's a quick [PoC](A quick thought was another mapping with approvals per block number (or timestamp), in this way it's not needed to clean up the value at the end since it will just get invalidated immediately.

https://gist.github.com/ernestognw/9bc89668722090c5976f42531032a81a)

byshape:
I prefer to keep the temporary storage for this ERC within a transaction rather than a block.
`EIP-1153` already operates within these limits so that compatible chains will be able to use this functionality. The rest may try to implement this logic in any other non-conflicting way.

ERCS/erc-7674.md
byshape: Suggestion has been applied, MUST was replaced with MAY.

ERCS/erc-7674.md
byshape: Some may want to go in a different direction. So I think the wording should be flexible here.

ERCS/erc-7674.md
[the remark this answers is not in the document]
byshape: If we're talking about the same transaction, I think it's better to mention only `value` since `allowance(owner, spender)` can be spent (or not) later too.

ERCS/erc-7674.md
```diff
@@ -0,0 +1,58 @@
+---
+eip: 7674
+title: Temporary Approval Extension for ERC-20
+description: An interface for ERC-20 approvals via transient storage
```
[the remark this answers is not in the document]
Amxx: Transient is not single-block though, its single-transaction. I like the idea, but we would need the wording to accuratly describe what transient does.

Maybe
```suggestion
description: An interface for ephemeral ERC-20 approvals
```

ERCS/erc-7674.md
```diff
@@ -0,0 +1,58 @@
...
+status: Draft
+type: Standards Track
+category: ERC
+created: 2024-04-02
+requires: 20, 1153
+---
+
+## Abstract
+
+This specification defines the minimum interface required to temporarily approve `ERC-20` tokens for spending within a single transaction.
```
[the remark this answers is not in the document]
Amxx: same comment: "single block" is incorrect.

ERCS/erc-7674.md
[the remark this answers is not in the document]
ernestognw: Yeah, 100% agree wth your suggestion 👍🏻

ERCS/erc-7674.md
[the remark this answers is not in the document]
ernestognw: Yes. Updated accordingly

ERCS/erc-7674.md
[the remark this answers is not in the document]
Amxx:
```suggestion
Call to `temporaryApprove(spender, value)` allows `spender` to withdraw within the same transaction from `msg.sender` multiple times, up to the `value` amount. This temporary allowance is to be considered in addition to the normal (persistent) ERC20 allowance, the total value than spender is able to spend during the transaction is thus capped by the sum of the temporary and the normal (pensistent) allowances.
```

ERCS/erc-7674.md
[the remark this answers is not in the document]
ZumZoom: I actually prefer this wording instead of adding extra text in parens:

```suggestion
This specification defines the minimum interface required to temporarily approve `ERC-20` tokens for spending within the same transaction.
```

SamWilsn, line 16:
```suggestion
This specification defines the minimum interface required to temporarily approve [ERC-20](./eip-20.md) tokens for spending within a single transaction.
```

ERCS/erc-7674.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Motivation
+
+`EIP-1153` allows to use a cheaper way to temporarily store allowances.
```
SamWilsn, line 20:
```suggestion
[EIP-1153](./eip-1153.md) introduces a cheaper way to temporarily store allowances.
```

ERCS/erc-7674.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Specification
+
+The key words "MUST", "SHOULD", "MAY" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
+
+Compliant contracts MUST implement 1 new function in addition to `ERC-20`:
```
SamWilsn, line 26: Please remove all EIP names in backticks and replace them with links. Backticks are only for inline code.

ERCS/erc-7674.md
```diff
@@ -0,0 +1,58 @@
...
+
+The storage for the temporary allowances MUST be different to that of the regular allowance. Compliant contracts MAY use the transient storage `EIP-1153` to keep the temporary allowance. For each `owner` and `spender`, the slot MUST be uniquely selected to avoid slot collision. Each slot index SHOULD be derived from the base slot index for temporary allowances, `owner` and `spender` addresses. Slot MAY be derived as `keccak256(spender . keccak256(owner . p))` where `.` is concatenation and `p` is `keccak256` from the string uniquely defining temporary allowances in the namespace of the implementing contract.
+
+Compliant contracts MUST add a temporary allowance check to the `transferFrom` function. The permanent allowance SHOULD only be spent after the temporary allowance has been exhausted.
+
+Compliant contracts MUST add a temporary allowance to the permanent one when returning the allowed amount to spend in the `allowance` function. In case the sum of the temporary and permanent allowance overflow, `type(uint256).max` MUST be returned.
+
+No event is required when setting a temporary allowance, but compliant contracts MAY implement it.
```
SamWilsn, line 38:
```suggestion
No event is required when setting a temporary allowance, but compliant contracts MAY emit it.
```

ERCS/erc-7674.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Rationale
+
+The main goal of this standard is to make it cheaper to approve `ERC-20` tokens for a single transaction with minimal interface extension to allow easier integration of a compliant contract into existing infrastructure. This affects the backward compatibility of the `allowance` function. However, the required changes to the `transferFrom` function implementation satisfy the requirement to explicitly authorize the spender to transfer tokens.
```
SamWilsn, line 42: Comments about the proposal as a whole belong in the Motivation section. The Rationale section should be used to explain choices made within your proposal.

ERCS/erc-7674.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Backwards Compatibility
+
+All functionality of the `ERC-20` standard is backward compatible except for the `allowance` function.
+
+## Reference Implementation
+
+TBD
```
SamWilsn, line 50: If you use an HTML-style comment, the linter will make sure you deal with it before going to review.

```suggestion
TBD <!-- TODO -->
```

Amxx:
@byshape created https://github.com/byshape/ERCs/pull/1 to update this PR. We should stry to get it merge sonner rather than latter.

ERCS/erc-7674.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Specification
+
+The key words "MUST", "SHOULD", "MAY" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
+
+Compliant contracts MUST implement 1 new function in addition to [ERC-20](./eip-20.md):
+```solidity
+function temporaryApprove(address spender, uint256 value) public returns (bool success)
+```
+Call to `temporaryApprove(spender, value)` allows `spender` to withdraw within the same transaction from `msg.sender` multiple times, up to the `value` amount.  This temporary allowance is to be considered in addition to the normal (persistent) [ERC-20](./eip-20.md) allowance, the total value that spender is able to spend during the transaction is thus capped by the sum of the temporary and the normal (persistent) allowances.
```
frangio, line 30: >This temporary allowance is to be considered in addition to the normal (persistent) ERC-20 allowance

This forces an SLOAD of the current allowance, resulting in higher costs than one would want from this feature. I think it should be a temporary allowance override instead.

Amxx: I disagree. 

If the temporary allowance takes priority (this should be mentionned somewhere if it isn't already), and if the temporary allowance is sufficient to cover the need of some function that consumes allowance (`transferFrom`), then the allowance can be consumed from the temporary part without having to go consume anything in the persistent part.

You only need to lean (and update) the persistent part if the temporary part is not sufficient.

An analogy would be:
- you want to cook rice,
- you have 2kg of rice in your kitchen (fast access - eq to temporary)
- you have 50kg of rice stored in you basement (slow access - eq to persistent)

it is true that:
- you have 52kg available
- to measure how much you have available, you have to check both the kitchen and the basement
- if you need to cook diner, and that diner requires 500g of rice, the kitchen is sufficient, and you don't need to go to the basement. Rice in the basement is available if you need, but you don't need to access it unless there isn't enough in the kitchen for what you are planning to cook.

Amxx: What do you think of

```suggestion
Call to `temporaryApprove(spender, value)` allows `spender` to withdraw within the same transaction from `msg.sender` multiple times, up to the `value` amount.  This temporary allowance is to be considered in addition to the normal (persistent) [ERC-20](./eip-20.md) allowance, the total value that spender is able to spend during the transaction is thus capped by the sum of the temporary and the normal (persistent) allowances. 

While it SHOULD be possible for a `transferFrom` operation to consume both time of allowance, the consumption of the temporary allowance SHOULD take priority over the consumption of the persistent allowance. Therefore, if the temporary allowance is sufficient for executing a `transferFrom` operation, the persistent allowance SHOULD not be loaded/updated from the storage. Consumption of persistent allowance, which implies storage accesses, SHOULD be performed only if the temporary allowance is not sufficient for the operation being executed.
```

ERCS/erc-7674.md
```diff
@@ -0,0 +1,58 @@
...
+
+The storage for the temporary allowances MUST be different to that of the regular allowance. Compliant contracts MAY use the transient storage [EIP-1153](./eip-1153.md) to keep the temporary allowance. For each `owner` and `spender`, the slot MUST be uniquely selected to avoid slot collision. Each slot index SHOULD be derived from the base slot index for temporary allowances, `owner` and `spender` addresses. Slot MAY be derived as `keccak256(spender . keccak256(owner . p))` where `.` is concatenation and `p` is `keccak256` from the string uniquely defining temporary allowances in the namespace of the implementing contract.
+
+Compliant contracts MUST add a temporary allowance check to the `transferFrom` function. The permanent allowance SHOULD only be spent after the temporary allowance has been exhausted.
+
+Compliant contracts MUST add a temporary allowance to the permanent one when returning the allowed amount to spend in the `allowance` function. In case the sum of the temporary and permanent allowance overflow, `type(uint256).max` MUST be returned.
+
+No event is required when setting a temporary allowance, but compliant contracts MAY emit a specific event.
```
SamWilsn, line 38: You should specify the event, or clarify what "specific" means here. The easiest would be to define the event, but leave it optional.

ERCS/erc-7674.md
```diff
@@ -0,0 +1,58 @@
...
+```solidity
+function temporaryApprove(address spender, uint256 value) public returns (bool success)
+```
+Call to `temporaryApprove(spender, value)` allows `spender` to withdraw within the same transaction from `msg.sender` multiple times, up to the `value` amount.  This temporary allowance is to be considered in addition to the normal (persistent) [ERC-20](./eip-20.md) allowance, the total value that spender is able to spend during the transaction is thus capped by the sum of the temporary and the normal (persistent) allowances.
+
+The storage for the temporary allowances MUST be different to that of the regular allowance. Compliant contracts MAY use the transient storage [EIP-1153](./eip-1153.md) to keep the temporary allowance. For each `owner` and `spender`, the slot MUST be uniquely selected to avoid slot collision. Each slot index SHOULD be derived from the base slot index for temporary allowances, `owner` and `spender` addresses. Slot MAY be derived as `keccak256(spender . keccak256(owner . p))` where `.` is concatenation and `p` is `keccak256` from the string uniquely defining temporary allowances in the namespace of the implementing contract.
```
SamWilsn, line 32: This is all still implementation detail. You should only specify the externally visible behaviour in the Specification section. You can put recommendations in the Reference Implementation section. How about:

```suggestion
Each temporary allowance MUST persist until the end of the transaction that created it (unless overwritten by another call to `temporaryApprove`.) Each temporary allowance MUST be cleared at the end of the transaction that created it. See [Using Transient Storage](#using-transient-storage) for an example.
```

Then moving this content to the Reference Implementation section under a "Using Transient Storage" heading.

ERCS/erc-7674.md
frangio: Yes, my confusion was because of the effect on the `allowance()` function. If the allowances have to be added, this forces `allowance` to read from both. However, in the context of `transferFrom` it should not be necessary to do this, as you've described.

I am not sure if the effect on `allowance` is something that we should care about.

ERCS/erc-7674.md
Amxx: > (unless overwritten by another call to `temporaryApprove` or consumed by a call to `transferFrom`).

ERCS/erc-7674.md
```diff
@@ -0,0 +1,58 @@
...
+status: Draft
+type: Standards Track
+category: ERC
+created: 2024-04-02
+requires: 20, 1153
+---
+
+## Abstract
+
+This specification defines the minimum interface required to temporarily approve [ERC-20](./eip-20.md) tokens for spending within the same transaction.
```
ZumZoom, line 16:
```suggestion
This specification defines the minimum interface required to temporarily approve [ERC-20](./erc-20.md) tokens for spending within the same transaction.
```

ERCS/erc-7674.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Motivation
+
+User are often require to set token approval that will only be used once. It is common to leave unexpected approvals after these interactions. [EIP-1153](./eip-1153.md) introduces a cheap way to handle temporarily allowances.
+
+## Specification
+
+The key words "MUST", "SHOULD", "MAY" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
+
+Compliant contracts MUST implement 1 new function in addition to [ERC-20](./eip-20.md):
```
ZumZoom, line 26:
```suggestion
Compliant contracts MUST implement 1 new function in addition to [ERC-20](./erc-20.md):
```

ERCS/erc-7674.md
ZumZoom, line 30:
```suggestion
Call to `temporaryApprove(spender, value)` allows `spender` to withdraw within the same transaction from `msg.sender` multiple times, up to the `value` amount.  This temporary allowance is to be considered in addition to the normal (persistent) [ERC-20](./erc-20.md) allowance, the total value that spender is able to spend during the transaction is thus capped by the sum of the temporary and the normal (persistent) allowances.
```

ERCS/erc-7674.md
```diff
@@ -0,0 +1,58 @@
...
+
+No event is required when setting a temporary allowance, but compliant contracts MAY emit a specific event.
+
+## Rationale
+
+It was decided to make minimal interface extension to allow easier integration of a compliant contract into the existing infrastructure. This affects the backward compatibility of the `allowance` function. However, the required changes to the `transferFrom` function implementation satisfy the requirement to explicitly authorize the spender to transfer tokens.
+
+## Backwards Compatibility
+
+All functionality of the [ERC-20](./eip-20.md) standard is backward compatible except for the `allowance` function.
```
ZumZoom, line 46:
```suggestion
All functionality of the [ERC-20](./erc-20.md) standard is backward compatible except for the `allowance` function.
```

ERCS/erc-7674.md
ZumZoom, line 32:
```suggestion
The storage for the temporary allowances MUST be different to that of the regular allowance. Compliant contracts MAY use the transient storage [EIP-1153](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1153.md) to keep the temporary allowance. For each `owner` and `spender`, the slot MUST be uniquely selected to avoid slot collision. Each slot index SHOULD be derived from the base slot index for temporary allowances, `owner` and `spender` addresses. Slot MAY be derived as `keccak256(spender . keccak256(owner . p))` where `.` is concatenation and `p` is `keccak256` from the string uniquely defining temporary allowances in the namespace of the implementing contract.
```

SamWilsn:
I took the liberty of fixing your EIP/ERC links. You do need to use `[ERC-20](./erc-20.md)` when linking to ERCs because of a quirk in our rendering system.

SamWilsn (approved):

PR #360, opened 2024-04-04
Add ERC: Paymaster Web Service Capability

lukasrosario:
Defines an EIP-5792 capability for paymaster web services along with a proposal for standardized paymaster web service RPC methods.

https://ethereum-magicians.org/t/add-erc-paymaster-web-service-capability/19530


ERCS/erc-draft_pm_capability.md
```diff
@@ -0,0 +1,285 @@
...
+
+#### `pm_getPaymasterStubData`
+
+Returns stub values to be used in paymaster-related fields of an unsigned user operation for gas estimation. Accepts an unsigned user operation, entrypoint address, chain id, and a context object. Paymaster service providers can define fields that app developers should use in the context object. 
+
+##### `pm_getPaymasterStubData` RPC Specification
+
+```typescript
+// [userOp, entryPoint, chainId, context]
+type GetPaymasterStubDataParams = [
```
wilsoncusack, line 37: Not sure this needs to take any args? Aside maybe sender, but even then I am skeptical of entry point detection. I think maybe it should be passed. I expect most paymaster services have pretty static values here

ERCS/erc-draft_pm_capability.md
```diff
@@ -0,0 +1,285 @@
...
+    "preVerificationGas": "0x...",
+    "maxFeePerGas": "0x...",
+    "maxPriorityFeePerGas": "0x...",
+    "paymasterAndData": "0x...",
+    "signature": "0x..."
+  },
+  "0x5FF137D4b0FDCD49DcA30c7CF57E578a026d2789",
+  "0x2105",
+  {
+    "policyId": "962b252c-a726-4a37-8d86-333ce0a07299"
```
wilsoncusack, line 80: what is policyId?

ERCS/erc-draft_pm_capability.md
wilsoncusack: Oh I see paymaster address is passed

wilsoncusack: Actually I take it back, it's probably useful to give the paymaster all of this incase they want to use in their dummy data.

ERCS/erc-draft_pm_capability.md
```diff
@@ -0,0 +1,285 @@
...
+{
+  "paymasterAndData": "0x..."
+}
+```
+
+If using entrypoint v0.7:
+
+```json
+{
+  "paymaster": "0x...",
```
wilsoncusack, line 178: I'm confused why these are separate here and not for v0.6, given in both cases in the actual user op they are concatenated? I think maybe we should just return in whatever format is needed to estimate gas.

ERCS/erc-draft_pm_capability.md
lukasrosario: this is illustrative. paymaster services should define fields they might look for here.

ERCS/erc-draft_pm_capability.md
lukasrosario: they are not concatenated in the user op. for v0.6 it is a `paymasterAndData` field and for v0.7 they are explicitly separate `paymaster` and `paymasterData` fields.

ERCS/erc-draft_pm_capability.md
```diff
@@ -0,0 +1,285 @@
+---
```
abcoathup, line 1:
```suggestion
---
eip: 7677
```

ERCS/erc-draft_pm_capability.md
```diff
@@ -0,0 +1,285 @@
...
+title: Paymaster Web Service Capability
+description: A standard for apps to communicate with smart wallets about paymaster web services
+author: Lukas Rosario (@lukasrosario), Dror Tirosh (@drortirosh), Wilson Cusack (@wilsoncusack)
+discussions-to: https://ethereum-magicians.org/t/add-erc-paymaster-web-service-capability/19530
```
abcoathup, line 5:
```suggestion
discussions-to: https://ethereum-magicians.org/t/erc-7677-paymaster-web-service-capability/19530
```

ERCS/erc-7677.md
```diff
@@ -0,0 +1,296 @@
...
+
+Below is a diagram illustrating the full `wallet_sendCalls` flow, including how a wallet might implement the interaction.
+
+![flow](../assets/eip-7677/0.png)
+
+## Rationale
+
+### Gas Estimation
+
+The current loose standard for paymaster services is to implement `pm_sponsorUserOperation`. This method returns values for paymaster-realted user operation fields and updated gas values. The problem with this method is that paymaster service providers have different ways of estimating gas, which results in different estimated gas values. Sometimes these estimates can be incorrect. As a result we believe it’s better to leave gas estimation up to the wallet, as it has more context on its onchain implementation, and then ask paymaster services to sponsor given the estimates defined by the wallet.
```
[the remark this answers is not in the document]
lukasrosario: right makes sense. i'll update this

ERCS/erc-7677.md
```diff
@@ -0,0 +1,296 @@
...
+  "paymaster": "0x...",
+  "paymasterData": "0x..."
+}
+```
+
+#### `pm_getPaymasterData`
+
+Returns values to be used in paymaster-related fields of a signed user operation. These are not stub values and will be used during user operation submission to a bundler. Similar to `pm_getPaymasterStubData`, accepts an unsigned user operation, entrypoint address, chain id, and a context object.
+
+This method also returns a `sponsor` object with `name` and `icon` fields. The `name` field is the name of the party sponsoring the transaction, and the `icon` field is a URI pointing to an image. Wallet developers MAY choose to display sponsor information to users.
```
bumblefudge, line 110: It might be worth sticking some normative requirements in here to keep the URI field from being a JS injection vector, of [the sort that some reviewer (apologies I can't remember which) inserted into ERC-6963 at the last minute](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-6963.md?plain=1#L76).  Alternately, you could mention this URI security in security considerations, but I recommend doing it normatively here in the very section that every [wallet] implementer will read.

ERCS/erc-7677.md
```diff
@@ -0,0 +1,296 @@
...
+    nonce: `0x${string}`;
+    initCode: `0x${string}`;
+    callData: `0x${string}`;
+    callGasLimit: `0x${string}`;
+    verificationGasLimit: `0x${string}`;
+    preVerificationGas: `0x${string}`;
+    maxFeePerGas: `0x${string}`;
+    maxPriorityFeePerGas: `0x${string}`;
+  },
+  `0x${string}`,
```
bumblefudge, line 51: Why does the RPC specification jump straight to a TypeScript data shape without any explanation of semantics? I think it would be good to explain, for example, that the first of these hex strings is the address and the second the EIP155 chainId converted to hex (which would make sense given the examples), at the very least by normative reference over to [the exact section of 5792](https://eips.ethereum.org/EIPS/eip-5792#wallet_getcapabilities-rpc-specification), although now that I notice, it's also just a data shape without any semantics there as well.

Concretely, though, I would mention that in [one of the 5792 examples](https://eips.ethereum.org/EIPS/eip-5792#wallet_getcapabilities-example-return-value), there are TWO chainIds given with different capability objects in each, which would not match this TS definition, which models a flat structure with only one chainId (and only one capability!).

ERCS/erc-7677.md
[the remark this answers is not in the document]
bumblefudge: wordsmithed a bit (and caught a typo!)
```suggestion
The emerging developer convention for exposing paymaster services is to implement the `pm_sponsorUserOperation` interface. This method returns values for paymaster-related user operation fields and updated gas values. The problem with uniformly using this method in a way that  abstracts out gas estimation way is that paymaster service providers have different ways of estimating gas, making such abstracted estimates unreliable. As a result, we believe it’s better to leave gas estimation up to the wallet, as it has more context on its own onchain implementation, and then ask paymaster services to sponsor after fetching more accurate estimates directly from the wallet.
```

ERCS/erc-7677.md
lukasrosario: above this it says `Accepts an unsigned user operation, entrypoint address, chain id, and a context object.` There is also a comment above the type `// [userOp, entryPoint, chainId, context]`. Added comments to each line to try to make it more clear though.

The specification here is related to a single `wallet_sendCalls` call, which is specific to a single chain id, where this type defines how a wallet would communicate with a paymaster service. The return value from `wallet_getCapabilities` is just expressing the available capabilities of the wallet per chain. I don't think I understand the relationship you're seeing between the 5792 example and what is written here.

ERCS/erc-7677.md
lukasrosario: added the same note here.

ERCS/erc-7677.md
bumblefudge: That works too! I think I just have a[n admittedly unpopular] preference for very explicit and verbose declarations of what each property means and what shape is allowed for its values, plus what to do if the wrong shape is passed.

My comment about the 5792 example was confused, apologies.  Here, `chainId` is just a top-level string (of which there can only be one), in the 5792 example I was pointing to a top-level chainId used as a key, whose value was an object full of capabilities. No action needed.

lukasrosario: gotcha. all good!

ERCS/erc-7677.md
```diff
@@ -0,0 +1,296 @@
...
+    "nonce": "0x...",
+    "initCode": "0x",
+    "callData": "0x...",
+    "callGasLimit": "0x...",
+    "verificationGasLimit": "0x...",
+    "preVerificationGas": "0x...",
+    "maxFeePerGas": "0x...",
+    "maxPriorityFeePerGas": "0x...",
+    "paymasterAndData": "0x...",
+    "signature": "0x..."
```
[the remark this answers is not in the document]
wilsoncusack: Should probably leave off request, yeah.

ERCS/erc-7677.md
```diff
@@ -0,0 +1,296 @@
...
+  {
+    sender: `0x${string}`;
+    nonce: `0x${string}`;
+    initCode: `0x${string}`;
+    callData: `0x${string}`;
+    callGasLimit: `0x${string}`;
+    verificationGasLimit: `0x${string}`;
+    preVerificationGas: `0x${string}`;
+    maxFeePerGas: `0x${string}`;
+    maxPriorityFeePerGas: `0x${string}`;
```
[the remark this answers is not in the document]
wilsoncusack: The user will sign after the paymasterAndData field is properly populated. We should make more clear in the text, if it's not.

ERCS/erc-7677.md
```diff
@@ -0,0 +1,296 @@
...
+{
+  "sponsor": {
+    "name": "My App",
+    "icon": "https://..."
+  },
+  "paymasterAndData": "0x..."
+}
+```
+
+If using entrypoint v0.7:
```
wilsoncusack, line 181: I think I disagree with having different specs for v0.7 vs v0.6. In both cases paymasterAndData is a concatenation of different values. It easier to just supply the paymasterAndData field so wallet doesn't have to handle.

ERCS/erc-7677.md
```diff
@@ -0,0 +1,296 @@
...
+    maxPriorityFeePerGas: `0x${string}`;
+  }, // userOp
+  `0x${string}`, // Entrypoint
+  `0x${string}`, // Chain ID
+  Record<string, any> // Context
+];
+
+type GetPaymasterDataResult = {
+  sponsor?: { name: string; icon: string } // Sponsor info
+} & Record<string, `0x${string}`>; // Paymaster values
```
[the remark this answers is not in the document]
wilsoncusack: I actually disagree with having these differ by entry point @lukasrosario https://github.com/ethereum/ERCs/pull/360/files/f9e15bfd482be08ac8a3147e8756a8c6a07c0e1d#r1559567260

ERCS/erc-7677.md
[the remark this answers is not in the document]
lukasrosario: removed

ERCS/erc-7677.md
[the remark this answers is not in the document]
lukasrosario: discussed with Wilson offline, these fields should be separate as currently written in the draft. ie `paymasterAndData` for v0.6 and `paymaster` / `paymasterData` for v0.7.

@pedrouid the intention is to make this compatible with potential future changes, where the fields might include something other than `paymasterAndData`, `paymaster`, `paymasterData`. what do you think about

```
type GetPaymasterDataResult = {
  sponsor?: { name: string; icon: string } // Sponsor info
  paymaster?: string // Paymaster address (for v0.7 or higher)
  paymasterData?: string;
  paymasterAndData?: string; // Paymaster Data
} & Record<string, `0x${string}`>;
```

makes the v0.6 and v0.7 fields explicit, but still allows for future changes.

ERCS/erc-7677.md
lukasrosario: discussed with Wilson offline, these fields should be separate as currently written in the draft. ie paymasterAndData for v0.6 and paymaster / paymasterData for v0.7.

the reasoning is that while these fields are packed onchain (i.e. once user op reaches the entrypoint, it will still be `paymasterAndData`), wallets submit ops to bundlers unpacked, so keeping them separate in the paymaster response would be easiest for wallets to handle.

ERCS/erc-7677.md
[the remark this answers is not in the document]
lukasrosario: added a note on why theres no signature here. lmk if you think it's still unclear though

ERCS/erc-7677.md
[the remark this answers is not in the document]
dancoombs: For `pm_getPaymasterStubData` on v0.7, I'd recommend having this return type (if you plan on having a structured schema) contain both `paymasterVerificationGasLimit` and `paymasterPostOpGasLimit`. The paymaster service can hardcode gas limits that it knows works with its contract. When provided through the `eth_estimateUserOperationGas` endpoint bundlers should skip estimating these values (but will validate once that they work), saving computation.

Many (most?) paymaster functions have pretty static gas consumption, so it doesn't make sense to estimate gas for them each time. Instead just use a hardcoded limit that is known to work for all inputs and onchain states.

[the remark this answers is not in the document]
hazim-j: @dancoombs could you elaborate on the difficulties with estimating `paymasterVerificationGasLimit ` and `paymasterPostOpGasLimit`?

My initial impression is that it might be difficult to estimate although not impossible if using custom tracers?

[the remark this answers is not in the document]
dancoombs: There are not difficulties estimating `paymasterVerificationGasLimit`.

My suggestion above is that, regardless of estimation difficulties, paymaster may want to signal to their users the gas values that should be used. This will allow the user/bundler to skip estimating paymaster gas limits.

Many accounts/paymasters have very static verification functions that don't need to be estimated for each user op, and a static limit value can be used that the paymaster or account developer knows will work each time, saving the bundler from performing the same computation each time. Potentially latency or even RPC cost savings would trickle down to the user/dapp. This suggestion allows the paymaster to communicate those limits.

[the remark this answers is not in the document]
lukasrosario: I'm pretty hesitant to now bring bundlers into this mix. 

@dancoombs is it likely / have you observed that paymaster gas values are insufficiently estimated by bundlers?

That was the motivation for going with the 2-step flow here (proposed by Dror). I.e. we routinely observed insufficient gas estimates from paymasters so we decided wallets / bundlers should estimate gas themselves. I could get behind your proposal if the issue is that bundlers similarly provide insufficient paymaster gas estimates. I am less convinced if this is just for cost savings.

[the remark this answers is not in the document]
dancoombs: How is this bringing bundlers into the mix? Its suggesting that paymasters supply their own gas limits if they know what those gas limits need to be.

There are scenarios where bundlers/wallets don't have enough context to provide sufficient gas estimates, especially for postOp, and the paymaster needs to supply to get stable execution.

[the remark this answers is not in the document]
lukasrosario: We would need to specify that bundlers should skip paymaster gas estimation if paymaster gas values are supplied during an `eth_estimateUserOperationGas` call.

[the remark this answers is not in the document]
lukasrosario: Like I said though, if you're saying that it would be easy for bundlers to misestimate paymaster gas values-resulting in failed user op submission-then I can get on board with this. It sounds like you are saying bundlers misestimating the paymaster gas values could be common? Is that right?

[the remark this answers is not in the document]
dancoombs: > We would need to specify that bundlers should skip paymaster gas estimation if paymaster gas values are supplied during an `eth_estimateUserOperationGas` call.

This is already the [case](https://github.com/eth-infinitism/account-abstraction/blob/develop/erc/ERCS/erc-4337.md#-eth_estimateuseroperationgas).

```
gas limits (and prices) parameters are optional, but are used if specified.
```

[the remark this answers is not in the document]
lukasrosario: Ah right. Ok, I can get on board with this. @hazim-j what do you think?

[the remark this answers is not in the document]
hazim-j: I think I could get behind this. But it might be worth defining a concrete example for when a bundler's `paymasterPostOpGasLimit` can become unreliable and adding it to the rationale section. My understanding of this on a high level is if a state change happens between simulation and execution that causes two different code paths in `postOp` then I can see how that would cause an unreliable estimation.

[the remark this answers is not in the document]
lukasrosario: i've added the paymaster gas values to the `pm_getPaymasterStubData` response. going to resolve this for now to move things along and we can continue discussion on it if necessary

ERCS/erc-7677.md
```diff
@@ -0,0 +1,350 @@
...
+
+type GetPaymasterDataResult = {
+  sponsor?: { name: string; icon: string } // Sponsor info
+  paymaster?: string // Paymaster address (entrypoint v0.7)
+  paymasterData?: string; // Paymaster data (entrypoint v0.7)
+  paymasterAndData?: string; // Paymaster and data (entrypoint v0.6)
+}
+```
+
+###### `pm_getPaymasterData` Example Parameters
```
livingrockrises, line 163: Paymaster services are supposed to detect which account implementation is it from the sender's address?
otherwise how would one decode data based on policy-id?

livingrockrises (commented):
review i

ERCS/erc-7677.md
lukasrosario: detecting the account implementation is not a strict requirement of paymaster services, and it is ultimately up to paymaster services to decide how to validate policy ids or whatever other abstractions they might have.

you could reliably do this without needing to know the implementation though by simulating the incoming user op, no need to decode the calldata (although you could if you know the implementation).

arcticfloyd1984: Shouldn't this be a requirement as it is mentioned that 
`Paymaster services MUST detect which entrypoint version the account is using and return the correct fields.`
how is a paymaster supposed to detect this without knowing the account implementation?
Afaik all smart accounts store the supported entry point on the smart account in a simple variable and reading that might be the best way to know this information.
To make that possible I think the paymaster services should have some knowledge about the implementation

ERCS/erc-7677.md
```diff
@@ -0,0 +1,350 @@
...
+##### `pm_getPaymasterStubData` RPC Specification
+
+Note that the user operation parameter does not include a signature, as the user signs after all other fields are populated.
+
+```typescript
+// [userOp, entryPoint, chainId, context]
+type GetPaymasterStubDataParams = [
+  // Below is specific to Entrypoint v0.6 but this API can be used with other entrypoint versions too
+  {
+    sender: `0x${string}`;
```
drortirosh, line 45: these are the fields of UserOperation v0.6.
On v0.7, there are few differences (factory, factoryData, paymaster fields, packed gas fields).

ERCS/erc-7677.md
```diff
@@ -0,0 +1,350 @@
...
+    nonce: `0x${string}`;
+    initCode: `0x${string}`;
+    callData: `0x${string}`;
+    callGasLimit: `0x${string}`;
+    verificationGasLimit: `0x${string}`;
+    preVerificationGas: `0x${string}`;
+    maxFeePerGas: `0x${string}`;
+    maxPriorityFeePerGas: `0x${string}`;
+  }, // userOp
+  `0x${string}`, // EntryPoint
```
drortirosh, line 55: it is better to have named parameters instead of indexed.
e.g. 
```js
{ userOp: {}, chainId: "0x..", entryPoint: "0x...", version: "0.7", context: {} }
```
And also be better to add "version", instead of rely on entrypoint address

ERCS/erc-7677.md
```diff
@@ -0,0 +1,350 @@
...
+```
+
+If using entrypoint v0.7, with paymaster gas estimates:
+
+```json
+{
+  "paymaster": "0x...",
+  "paymasterData": "0x...",
+  "paymasterVerificationGasLimit": "0x...",
+  "paymasterPostOpGasLimit": "0x..."
```
drortirosh, line 121: in v0.7, postOpGasLimit **cannot** be estimated by app/wallet, and must be returned by the serve
verificationGasLimit MAY be estimated, though if the service knows the value, it can place it anyway.

ERCS/erc-7677.md
drortirosh: paymaster knows the version from the entrypoint address (in any case, a paymaster support explicit entrypoint instance, not "any".
it would still be nice if at the API level a "version" field is added..

ERCS/erc-7677.md
```diff
@@ -0,0 +1,350 @@
...
+  "paymaster": "0x...",
+  "paymasterData": "0x..."
+}
+```
+
+If using entrypoint v0.7, with paymaster gas estimates:
+
+```json
+{
+  "paymaster": "0x...",
```
drortirosh, line 118: There are cases where the stub/final separation is not needed, e.g: when the paymaster is a TokenPaymaster, and the paymasterData doesn't contain "signature".
In such cases, the 2nd RPC call to the paymaster MAY be skipped, by returning `isFinal: true` by this RPC call.

ERCS/erc-7677.md
```diff
@@ -0,0 +1,350 @@
...
+  "paymasterVerificationGasLimit": "0x...",
+  "paymasterPostOpGasLimit": "0x..."
+}
+```
+
+#### `pm_getPaymasterData`
+
+Returns values to be used in paymaster-related fields of a signed user operation. These are not stub values and will be used during user operation submission to a bundler. Similar to `pm_getPaymasterStubData`, accepts an unsigned user operation, entrypoint address, chain id, and a context object.
+
+This method also returns a `sponsor` object with `name` and `icon` fields. The `name` field is the name of the party sponsoring the transaction, and the `icon` field is a URI pointing to an image. Wallet developers MAY choose to display sponsor information to users. The `icon` string MUST be a data URI as defined in [RFC-2397]. The image SHOULD be a square with 96x96px minimum resolution. The image format is RECOMMENDED to be either lossless or vector based such as PNG, WebP or SVG to make the image easy to render on the wallet. Since SVG images can execute Javascript, wallets MUST render SVG images using the `<img>` tag to ensure no untrusted Javascript execution can occur.
```
drortirosh, line 129: is the sponsor structure a "MUST" ?

ERCS/erc-7677.md
lukasrosario: curious about your reasoning here. i like the clarity of named arguments but most ethereum-related JSON-RPC methods i can think of use positional arguments. is there a reason you think we should stray from that loose convention other than clarity?

lukasrosario: same question for the `version` arg, is that just for clarity?

ERCS/erc-7677.md
```diff
@@ -0,0 +1,350 @@
...
+    initCode: `0x${string}`;
+    callData: `0x${string}`;
+    callGasLimit: `0x${string}`;
+    verificationGasLimit: `0x${string}`;
+    preVerificationGas: `0x${string}`;
+    maxFeePerGas: `0x${string}`;
+    maxPriorityFeePerGas: `0x${string}`;
+  }, // userOp
+  `0x${string}`, // EntryPoint
+  `0x${string}`, // Chain ID
```
arpan-jain, line 56: Is there a specific reason, to prefix 0x in front of chainId, instead of plain string or a number.

lukasrosario: chain ids are uint256 so all possible values could not be represented by a js number, and hex is pretty standardized in such cases vs the number as a decimal string

assets/erc-7677/0.png
SamWilsn, line 1: If you can replace these with SVGs, we'd appreciate it

SamWilsn (approved):

PR #361, opened 2024-04-05
Add ERC: UserOperation Builder

arein:

 - The build passes.


ERCS/erc-XXXX.md
```diff
@@ -0,0 +1,244 @@
+---
+eip: XXXX
+title: Smart Account Interfaces
+description: A simple way for apps, wallets, and smart account tooling to interact with Smart Accounts
+author: Derek Chiang (@derekchiang), Fil Makarov (@filmakarov), Garvit Khatri (@plusminushalf), Kristof Gazso (@kristofgazso), Derek Rein (@arein), Juan Caballero (@bumblefudge), Tomas Rocchi (@tomiir), Jake Moxey (@jxom)
```
jxom: I don't think I had much of an impact here – so happy to omit my name to weight the credit on others! 😄

Jircs1 (commented):
👍

Jircs1 (approved):
👍

ERCS/erc-7678.md
```diff
@@ -0,0 +1,223 @@
...
+        address create2Factory, 
+        bytes memory factoryData,
+        address userOpBuilder, 
+        bytes memory userOpBuilderCalldata
+    ) { 
+        if (address(smartAccount).code.length == 0) {
+            (bool success, bytes memory err) = create2Factory.call(factoryData);
+            if (!success) revert CounterfactualDeployFailed(err);
+        }
+
```
Ivshti, line 156: it's really important to check again for  `if (address(smartAccount).code.length == 0)` here because the call may have succeeded but not deployed the correct contract (or not deployed anything at all), which will cause a silent fail later on, as a call to an EOA will just succeed.

Ivshti (changes requested):

ERCS/erc-7678.md
```diff
@@ -0,0 +1,223 @@
+---
+eip: 7678
```
abcoathup, line 2:
```suggestion
eip: 7679
```
Numbers are assigned by ERC editors/associates. Please don't self assign. Number sniping is not allowed.
Manually assigning next sequential number.


ERCS/erc-7678.md
```diff
@@ -0,0 +1,223 @@
...
+title: UserOperation Builder
+description: Construct UserOperations without being coupled with account-specific logic.
+author: Derek Chiang (@derekchiang), Garvit Khatri (@plusminushalf), Fil Makarov (@filmakarov), Kristof Gazso (@kristofgazso), Derek Rein (@arein), Tomas Rocchi (@tomiir), bumblefudge (@bumblefudge)
+discussions-to: https://ethereum-magicians.org/t/erc-xxx-smart-account-interfaces/19547
```
abcoathup, line 6:
```suggestion
discussions-to: https://ethereum-magicians.org/t/erc-7679-smart-account-interfaces/19547
```

ERCS/erc-7678.md
arein: CC @filmakarov

ERCS/erc-7679.md
```diff
@@ -0,0 +1,223 @@
...
+    constructor(
+        address smartAccount,
+        address create2Factory, 
+        bytes memory factoryData,
+        address userOpBuilder, 
+        bytes memory userOpBuilderCalldata
+    ) { 
+        if (address(smartAccount).code.length == 0) {
+            (bool success, bytes memory err) = create2Factory.call(factoryData);
+            if (!success) revert CounterfactualDeployFailed(err);
```
filmakarov, line 154:
```suggestion
            (bool success, bytes memory ret) = create2Factory.call(factoryData);
            if (!success || address(smartAccount).code.length == 0) revert CounterfactualDeployFailed(ret);
```

ERCS/erc-7678.md
filmakarov: suggested a change below.
Thank you @Ivshti

pedrouid:
@abcoathup we weren’t sniping a number as 7678 doesn’t really mean anything

but we made an educated guess looking at EIPs, RIPs and ERCs

Why did the number change to 7679? I can’t find any 7678 in other repos

ERCS/erc-7678.md
bumblefudge: apologies, my mistake! i was just trying [wrongly, I think,] to resolve a Walidator error about the filename

ERCS/erc-7679.md
```diff
@@ -0,0 +1,223 @@
...
+}
+```
+
+### Using the UserOp builder
+
+To build a UserOp using the UserOp builder, the building party should proceed as follows:
+
+1. Obtain the address of `UserOpBuilder` and a `context` from the account owner.  The `context` is an opaque bytes array from the perspective of the building party.  The smart account implementation may need the `context` in order to properly figure out the UserOp fields.  See the “Rationale” section for more info.
+2. Execute a multicall (batched `eth_call`s) of `getNonce`, `getCallData`, `getDummySignature` with the `context` and executions.  The building party will now have obtained the nonce, calldata, and dummy signature (see “Rationale” for what a dummy signature is).
+3. Fill out a UserOp with the data obtained previously.  This UserOp must be valid except for the `signature` field.  Then call (via `eth_call`) `getSignature` with the UserOp and `context` to obtain a completely valid UserOp.
```
moldy530, line 123: How is this envisioned to work exactly? signature are typically generated off-chain by some signer (EOA or 3rd party KMS)

ERCS/erc-7678.md
abcoathup: No worries.  To avoid any perception of number sniping I manually bump self assigners numbers by 1.

ERCS/erc-7679.md
plusminushalf: `getSignature` is not meant for generating a signature on chain but rather formatting the signature in a way that will be accepted by the account. So signature by the Dapp's session key will still be generated off-chain and can be put in `userOperation.signature` along with the permissionsContext that was passed by the wallet. This on-chain function will just return a fully formatted signature which when entryPoint will pass when calling `validateUserOp`, the on-chain wallet will be able to parse and verify.

ERCS/erc-7678.md
filmakarov: @arein I think this can be marked as resolved?

ERCS/erc-7679.md
howydev: makes sense!

It seems like smart account contracts will integrate this interface, but if we're querying the account itself, it seems redundant to include `address smartAccount` in all the calls. could you explain whats the use case for that?

howydev: i didn't get the use case for this function from the naming or comments. i suggest renaming the function to formatSignature, or updating comments to clarify that it takes in an unformatted signature and outputs a formatted one

filmakarov: > it seems redundant to include address smartAccount in all the calls. could you explain whats the use case for that?

This interface won't be implemented by the Smart Account itself, but deployed as a standalone contract per Smart Account version. That's why SA address is included as an arg

howydev: Gotcha, thanks for clarifying!

howydev: Hmm can you share more about having it in the smart account vs external contract? Accounts are proxies so there shouldn't be any additional init costs. Also, I imagine that looking for this additional contract adds a bunch of logistical overhead whereas if we do the account, wallets/clients/SDKs should already have the address of the account handy

ERCS/erc-7679.md
```diff
@@ -0,0 +1,223 @@
...
+
+The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL NOT”, “SHOULD”, “SHOULD NOT”, “RECOMMENDED”, “MAY”, and “OPTIONAL” in this document are to be interpreted as described in RFC 2119.
+
+### UserOp builder
+
+To conform to this standard, a smart account implementation must provide a “UserOp builder” contract that implements the IUserOperationBuilder interface, as defined below:
+
+
+```solidity
+struct Execution {
```
jxom, line 43: Curious why it is termed as "execution" instead of "call"? Feel like we should make terminology symmetrical to [EIP-5792](https://eips.ethereum.org/EIPS/eip-5792).

ERCS/erc-7679.md
```diff
@@ -0,0 +1,223 @@
...
+     * @dev Returns a correctly encoded signature, given a UserOp that
+     * has been correctly filled out except for the signature field.
+     * @param smartAccount is the address of the UserOp sender.
+     * @param userOperation is the UserOp.  Every field of the UserOp should
+     * be valid except for the signature field.  The "PackedUserOperation"
+     * struct is as defined in ERC-4337.
+     * @param context is the data required for the UserOp builder to
+     * properly compute the requested field for the UserOp.
+     */
+    function getSignature(
```
drortirosh, line 109: It is highly dangerous to trust an external entity (rpc node) to create a signature. 
You can use view function to generate the hash to sign (similar to getUserOpHash), and maybe "embedsig" if you need to populate the "signature" field with other data in addition to the signature itself. 
But never trust an external entity to hold private information and generate a signature.

drortirosh, line 109: As mentioned below, this method should be replaced with 2 methods:
getDayaToSign (returning a hash and maybe sig type)
formatSignature (pack the signature created externally as a "signature" field)

ERCS/erc-7679.md
```diff
@@ -0,0 +1,223 @@
...
+- In addition to the `UserOpBuilder` address and the `context`, the building party also obtains the `factory` and `factoryData` as defined in ERC-4337.
+- When calling one of the view functions on the UserOp builder, the building party may use `eth_call` to deploy the `CounterfactualCall` contract with `factory` and `factoryData` (see below).  The `CounterfactualCall` contract would deploy the account before calling the view functions on the UserOp builder.
+- When filling out the UserOp, the building party includes `factory` and `factoryData`.
+
+### Counterfactual call
+
+The counterfactual call contract is inspired by [ERC-6492](./eip-6492.md), which devised a mechanism to execute `isValidSignature` (see [ERC-1271](./eip-1271.md)) against a pre-deployed (counterfactual) contract.
+
+```solidity
+contract CounterfactualCall {
```
drortirosh, line 141: You don't need a special handler. Just do
address(factory).call(factoryData)
To create the account, and return its address

derekchiang: The purpose of this handler is to call one of the 7679 functions against it, like `getNonceCallData`.  Am I missing something?

ERCS/erc-7679.md
derekchiang: Yeah the intention is for the `userOperation.signature` to include the "raw signature" so that this function can format it.  Agreed that `formatSignature` could be the better name.

Can you explain `getDataToSign`?  What's it supposed to take?  A completely-filled UserOp (except for `signature`)?

arein:
@abcoathup any change you could take a look? This has been through quite a few rounds and is ready to be merged - thank you

ERCS/erc-7679.md
```diff
@@ -0,0 +1,212 @@
...
+category: ERC
+created: 2024-04-05
+requires: 4337
+---
+
+## Abstract
+
+Different [ERC-4337](./eip-4337.md) smart account implementations encode their signature, nonce, and calldata differently.  This makes it difficult for DApps, wallets, and smart account toolings to integrate with smart accounts without integrating with account-specific SDKs, which introduces vendor lock-in and hurts smart account adoption.
+
+We propose a standard way for smart account implementations to put their account-specific encoding logic on-chain.
```
SamWilsn, line 18: Could you include a bit more technical detail on how you put the encoding logic on chain? Just a sentence or two to give the reader a high level overview.

ERCS/erc-7679.md
```diff
@@ -0,0 +1,212 @@
...
+- In addition to the `UserOpBuilder` address and the `context`, the building party also obtains the `factory` and `factoryData` as defined in ERC-4337.
+- When calling one of the view functions on the UserOp builder, the building party may use `eth_call` to deploy the `CounterfactualCall` contract with `factory` and `factoryData` (see below).  The `CounterfactualCall` contract would deploy the account before calling the view functions on the UserOp builder.
+- When filling out the UserOp, the building party includes `factory` and `factoryData`.
+
+### Counterfactual call
+
+The counterfactual call contract is inspired by [ERC-6492](./eip-6492.md), which devised a mechanism to execute `isValidSignature` (see [ERC-1271](./eip-1271.md)) against a pre-deployed (counterfactual) contract.
+
+```solidity
+contract CounterfactualCall {
```
SamWilsn, line 128: Does this contract need to be exactly this code, or will any similar code also work? If it's the latter, move the implementation to the Reference Implementation section and only describe the visible behaviour of the counterfactual call contract here.

SamWilsn (approved):

PR #369, opened 2024-04-08
Add ERC: Dual Nature Multi Token Standard

sennett-lau:
This proposal delineates the integration of the fungible ERC-20 token contract with the semi-fungible ERC-1155 multi-token standard, enabling cohesive operations between both standards within a single contract framework. It defines a mechanism for combining two token contracts and synchronizing operations between them.

### DIFF
--- a/ERCS/erc-dnmt.md
+++ b/ERCS/erc-dnmt.md
@@ -1,0 +1,129 @@
+---
+eip: dnmt
+title: Dual Nature Multi Token Standard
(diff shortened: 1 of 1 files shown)


ERCS/erc-dnmt.md
```diff
@@ -0,0 +1,129 @@
+---
+eip: dnmt
```
SamWilsn, line 2:
```suggestion
eip: 7681
```

SamWilsn (approved):

PR #375, opened 2024-04-11
Add ERC: Cross Chain Intents

marktoda:
This commit adds an ERC for a Cross Chain Intents Standard which is meant to unify off-chain messages and on-chain settlement smart contracts to enable sharing of infrastructure, filler networks, and orders across cross-chain bridging and trading systems.


ERCS/erc-7622.md
```diff
@@ -0,0 +1,170 @@
+---
+eip: 7622
```
SamWilsn, line 2:
```suggestion
eip: 7683
```

xinbenlv (dismissed):
Saw @SamWilsn issued a number. 

Content format is good enough as a draft. I will leave some technical questions on FEM thread

SamWilsn (approved):
Re-approving after minor edit.

PR #379, opened 2024-04-15
Update ERC-5559: Move to Draft

sshmatrix:
The following proposal is a superseding version of [EIP-5559: Cross Chain Write Deferral Protocol](https://ethereum-magicians.org/t/eip-5559-cross-chain-write-deferral-protocol/10576), aka CCIP-Write, replacing Ethereum L1 storage with L2 chains and cryptographically secure databases with an aim to cut gas costs and further privacy while retaining the secure aspects of on-chain storage. Methods in this document specifically target security and cost-effectiveness of write deferrals in context of databases. The cross-chain data written with these methods can be retrieved by generic EIP-3668-compliant contracts completing the cross-chain data life cycle.

### DIFF
--- a/ERCS/erc-5559.md
+++ b/ERCS/erc-5559.md
@@ -2,7 +2,7 @@
 eip: 5559
 title: "Cross-chain Write Deferral Protocol"
 description: The cross-chain write deferral protocol provides a mechanism to replace L1 storage with L2 and databases through off-chain handlers
-author: Avneet Singh (@sshmatrix), 0xc0de4c0ffee (@0xc0de4c0ffee), Nick Johnson (@arachnid), (@0xpaulio)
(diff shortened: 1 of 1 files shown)


abcoathup:
@sshmatrix your intention appears to be to create a new ERC that supersedes ERC-5559.  If this is the case we should issue this a new ERC number.  Let me know and I will issue one.

Arachnid:
@abcoathup We're working with @sshmatrix to update 5559, which is still a draft.

Arachnid:
Are you able to take a look at the text from the previous version and see if there are sections that are worth reusing? Particularly motivation and justification sections.

sshmatrix:
@Arachnid I've borrowed and merged appropriate text from previous version in the 'Abstract' and 'Rationale' sections. Please have a look and let me know if it reads well.

PR #380, opened 2024-04-15
Add ERC: Smart Blobs

charmful0x:
https://ethereum-magicians.org/t/erc-7689-smart-blobs/19672


ERCS/eip-draft_smart_blob.md
```diff
@@ -0,0 +1,270 @@
...
+- `inputs` : stringified JSON of the contract's function call object
+
+```ts
+function encodeBvmData(data: string): number[] {
+  const encodedData = data.split("").map((char) => char.charCodeAt(0));
+  return encodeBvmData;
+}
+```
+
+**Additional information about Transactions:**
```
wjmelements, line 75: Use markdown headers instead of bold

ERCS/eip-draft_smart_blob.md
```diff
@@ -0,0 +1,269 @@
+---
```
abcoathup, line 1:
```suggestion
---
eip: 7689
```

ERCS/eip-draft_smart_blob.md
```diff
@@ -0,0 +1,269 @@
...
+title: Smart Blobs
+description: State machine based on lazy evaluation on top of blobspace
+author: Rani El Husseini (@charmful0x), Benjamin Brandall (@xylophonez), Nik Rykov (@angrymouse)
+discussions-to: https://ethereum-magicians.org/t/add-erc-smart-blobs/19672
```
abcoathup, line 5:
```suggestion
discussions-to: https://ethereum-magicians.org/t/erc-7689-smart-blobs/19672
```

ERCS/erc-7689.md
```diff
@@ -0,0 +1,268 @@
...
+```
+***N.B: Smart Blobs sequencers are centralized and do not offer censorship resistance guarantees. The end user must rely on the good faith of the sequencer.***
+
+## Rationale
+This proposal hopes to accomplish the following:
+* Standardizing the usage of EIP-4844 blobs within the scope of creating execution machines on top of the blobspace
+* Utilize SmartWeave's lazy evaluation paradigm within the EVM space.
+* Extend the capabilities of lazy evaluation by using window-timegated DA
+* Utilize blobspace instead of calldata to leverage off-chain execution more cheaply
+* Write Smart Blobs in any interpreted language, or in a compiled language provided it uses WASM (JavaScript, Python, Rust, Lua, etc.)
```
SamWilsn, line 242: The Rationale section is used to explain choices made in the proposal, while this content seems to be justifying the EIP as a whole (and so belongs in the Motivation section.) My favourite analogy is:

> Rationale: We painted the shed red because...

SamWilsn (changes requested):
This proposal reads more like documentation for a specific project than a specification for a protocol. For example, `curl` commands aren't exactly common in EIPs.

You need to rewrite this as a document describing the expected behaviour of the system and the integration points with other systems/users.

ERCS/erc-7689.md
```diff
@@ -0,0 +1,268 @@
...
+
+
+## Backwards Compatibility
+There is no existing standard for smart blobs as described in this EIP, indicating that the proposal introduces new standard without conflicting with or altering any current implementations.
+
+
+## Reference Implementation
+
+A reference implementation of this proposal is currently being developed by the weaveVM organization. The ongoing development can be accessed on GitHub at the following URL: `https://github.com/weavevm/blobvm-core`.
```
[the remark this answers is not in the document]
charmful0x: do i remove the Reference Implementation as a whole or just the external URL?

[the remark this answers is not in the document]
SamWilsn: You can either remove the whole section, or include a reference implementation in the assets directory for this proposal (preferably under CC0-1.0, but any permissive non-copyleft license is allowed.)

We don't allow external links (see [here](https://ethereum.github.io/eipw/markdown-rel-links/).) In particular, linking to a reference implementation could be seen as unfairly favouring one implementation over others.

xylophonez:
@SamWilsn we've made revisions based on your feedback. Would be great if you could re-review. Cheers!

ERCS/erc-7689.md
```diff
@@ -0,0 +1,256 @@
...
+  "state":[]
+}
+```
+#### Contract call
+```json
+{
+  "type": 2,
+  "inputs": [],
+}
+```
```
SamWilsn, line 58: Not required for merging, but personally I'd like to see [JSON Schema](https://json-schema.org/specification) whenever an EIP describes a JSON blob.

ERCS/erc-7689.md
```diff
@@ -0,0 +1,256 @@
...
+
+```plaintext
+tx_gas = l1_gas_fees + (262604 * winston_byte_price * 1e-12 * ar_usd_price / eth_usd_price) * bvm_multiplier
+```
+
+#### Equation Terms Breakdown:
+
+- `l1_gas_fees`: The gas paid to post the transaction to the EVM network.
+- `262604`: The total byte size of an `EIP-4844` transaction when archiving on Arweave. This includes data, KZG commitments, and proof.
+- `winston_byte_price`: The cost price per byte on Arweave. This is dynamic and can be checked at `https://arweave.net/price/262604`.
```
SamWilsn, line 211:
```suggestion
- `winston_byte_price`: The cost price per byte on Arweave. This is dynamic.
```

We do not permit external links.

SamWilsn (commented):
I'm afraid there still isn't enough depth here for an EIP. I could entirely be wrong. I'm not an expert on Arweave, or even on L2s.

Could I, after reading just this document, implement a compatible sequencer?

What does "indexes the state changes in the cloud" mean? Is the encoding of blob data when sent to Arweave well specified?

Like... I just get the vibe that there's a ton more here that needs to be written before someone could actually implement this proposal.

angrymouse:
Agreed, this probably should be abstracted from arweave per se, but maybe just generalized "permanent storage network with probabilistic DA guarantee". `winston_byte_price` probably should be `storage_network_byte_price` or so.
I think arweave belongs much more to "implementation detail". In fact, the more mediums you store/fetch historical data from - the more secure overall Smart Blobs scheme becomes.

SamWilsn:
Just checking in to see if you're still interested in pursuing this proposal.

PR #386, opened 2024-04-17
Update ERC-2333: Move to Last Call

CarlBeek:
Bring ERC 2333, 2334, 2335 up to date.

These ERCs are already the standard for BLS keys as used within the Beacon chain. This PR brings the ERCs up to date and sets them to `Last Call` so they can become `Final` ASAP to line up with their actual status.


CarlBeek:
Taking this straight to `final`, there doesn't seem much sense in going via Last Call as this is already the standard across BLS key derivation

joshim2021 (approved):

PR #388, opened 2024-04-18
Update ERC-5559: Move to Draft

sshmatrix:
The following proposal is a superseding version of [EIP-5559: Cross Chain Write Deferral Protocol](https://ethereum-magicians.org/t/eip-5559-cross-chain-write-deferral-protocol/10576), aka CCIP-Write, replacing Ethereum L1 storage with L2 chains and cryptographically secure databases with an aim to cut gas costs and further privacy while retaining the secure aspects of on-chain storage. Methods in this document specifically target security and cost-effectiveness of write deferrals in context of databases. The cross-chain data written with these methods can be retrieved by generic EIP-3668-compliant contracts completing the cross-chain data life cycle.

### DIFF
--- a/ERCS/erc-5559.md
+++ b/ERCS/erc-5559.md
@@ -2,7 +2,7 @@
 eip: 5559
 title: "Cross-chain Write Deferral Protocol"
 description: The cross-chain write deferral protocol provides a mechanism to replace L1 storage with L2 and databases through off-chain handlers
-author: Avneet Singh (@sshmatrix), 0xc0de4c0ffee (@0xc0de4c0ffee), Nick Johnson (@arachnid), (@0xpaulio)
+author: Avneet Singh (@sshmatrix), 0xc0de4c0ffee (@0xc0de4c0ffee), Nick Johnson (@arachnid), Paul Gauvreau (@0xpaulio)
 discussions-to: https://ethereum-magicians.org/t/erc-5559-cross-chain-write-deferral-protocol/19664
 status: Draft
 type: Standards Track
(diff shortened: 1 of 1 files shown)


ERCS/erc-5559.md
```diff
@@ -1,471 +1,449 @@
 ---
 eip: 5559
-title: "Cross Chain Write Deferral Protocol"
-description: The cross chain write deferral protocol provides a mechanism to defer the storage & resolution of mutations to off-chain handlers
-author: Paul Gauvreau (@0xpaulio), Nick Johnson (@arachnid)
-discussions-to: https://ethereum-magicians.org/t/eip-cross-chain-write-deferral-protocol/10576
-status: Stagnant
+title: "Cross-chain Write Deferral Protocol"
+description: The cross-chain write deferral protocol provides a mechanism to replace L1 storage with L2 and databases through off-chain handlers
```
makoto, line 4: why do you replace `defer` to `replace` when the title uses the word `Deferral`?

makoto: Also do you need to change the title from `Cross Chain` to `Cross-chain`? It looks inconsistent

ERCS/erc-5559.md
```diff
@@ -1,471 +1,449 @@
...
+author: Avneet Singh (@sshmatrix), 0xc0de4c0ffee (@0xc0de4c0ffee), Nick Johnson (@arachnid), Makoto Inoue (@makoto), Paul Gauvreau (@0xpaulio)
+discussions-to: https://ethereum-magicians.org/t/eip-5559-cross-chain-write-deferral-protocol/19664
+status: Draft
 type: Standards Track
 category: ERC
-created: 2022-06-23
-requires: 712
+created: 2024-04-15
```
makoto, line 10: The original creation date should stay as is

ERCS/erc-5559.md
```diff
@@ -1,471 +1,449 @@
...
-
-This standard can be implemented in conjunction with [EIP-3668](./eip-3668) to provide a mechanism in which protocols can reside on and be interfaced through an L1 contract on mainnet, while being able to resolve and mutate data stored in external systems.
+The following standard provides a mechanism by which smart contracts can defer certain tasks to external providers. In particular, protocols can reduce the gas fees associated with storing data on mainnet by deferring the handling of write operations to another system or network. These storage handlers act as an extension to the core L1 contract. Methods in this document specifically target security and cost-effectiveness of write deferrals to two handler types: L2 and databases. The cross-chain data written with these methods can be retrieved by generic [EIP-3668](./eip-3668)-compliant contracts, thus completing the cross-chain data life cycle. This document, alongside [EIP-3668](./eip-3668), is a meaningful step toward a secure infrastructure for cross-chain write deferrals and data retrievals.
```
makoto, line 15: The abstract section looks contextually identical. Do you really need the change?

ERCS/erc-5559.md
```diff
@@ -1,471 +1,449 @@
...
+    revert StorageHandledByDatabase( 
+        gatewayUrl
+    );
+};
 ```
 
-In the case in which no reversion occurs, data is stored in the L1 contract when the transaction is executed.
+Following the revert, the client must take these steps:
 
-### Data Stored in an L2
```
makoto, line 72: I would suggest that you keep `Data Stored in an L1` and `Data Stored in an L2` because there are no changes

ERCS/erc-5559.md
sshmatrix: > do you need to change the title from `Cross Chain` to `Cross-chain`

Correct terminology is `off-chain`, `on-chain` and `cross-chain`, not `off chain`, `on chain` or `cross chain`

ERCS/erc-5559.md
sshmatrix: Why? Is that the standard practise?

ERCS/erc-5559.md
sshmatrix: Original document uses the word 'mutation' a lot. We don't use that word since it is confusing and jargon. That led to this re-write.

ERCS/erc-5559.md
sshmatrix: The language in this document is completely different. By using your suggestions, the entire document will read as if some sections were written by a completely different individual than the rest of the document. It destroys coherence.

ERCS/erc-5559.md
```diff
@@ -1,471 +1,449 @@
...
+1. Request the user for a secret signature `sigKeygen` to generate a deterministic `dataSigner` keypair,
```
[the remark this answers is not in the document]
sshmatrix: > The only benefit I am able to comprehend is that your solution can provide optional password

I don't know what you mean by this quite honestly. Can you please rephrase? The benefits have nothing to do with `password` which is not a feature.

True benefits that you missed out start from line 200:

```
Since the derived signer is wallet-specific, it can 

- sign batch data for multiple keys for a given node, and 
- sign batches of data for multiple nodes owned by a wallet

simultaneously in the background without ever prompting the user.
```

ERCS/erc-5559.md
sshmatrix: > Data Stored in an L1

I will add this back in the next commit

ERCS/erc-5559.md
makoto: @Arachnid can probably comment on it but I would be very strange that creation date changes as if all the prior specification didn't exist.

ERCS/erc-5559.md
makoto: https://chain.link/cross-chain is using as `Cross-chain` so you are right

ERCS/erc-5559.md
sshmatrix: > > Data Stored in an L1
> 
> I will add this back in the next commit

Fixed in https://github.com/ethereum/ERCs/pull/388/commits/eb07bce4d395e1fcefefa3d3b77e06ecb3f22733

ERCS/erc-5559.md
sshmatrix: @SamWilsn & @abcoathup can probably clear this up ↑

sshmatrix:
After discussions with other authors of the proposal, we have decided that this draft makes more sense as a standalone ERC, not a PR to EIP-5559, in order to preserve existing users. The new pull request is now here: https://github.com/ethereum/ERCs/pull/404

PR #403, opened 2024-04-29
Update ERC-1363: Update text to modern recommendations

vittominacori:
In order to be more clear and understandable I've updated some text and sections in order to be compliant with the new recommendations.

The content of the ERC was not edited but only rewritten and reorganized in a more clear way (also because of the first version was of 2018).

I've requested these changes in a moment when the EIP repo was migrating and these were rejected. I hope they can be reviewed now that the process is almost completed.


SamWilsn:
For changes to Final proposals, please ignore `eipw` (aka "EIP Walidator".) Only make the minimal changes required to clarify the text or fix errors. Anything else will likely get rejected.

vittominacori:
@SamWilsn yes I know but this is simply moving content to the right section as defined by EIP1, removing external links, and adjust some text.

SamWilsn:
As decided in https://github.com/ethcatherders/EIPIP/issues/346, we feel this pull request is changing too much for a Final proposal.

PR #424, opened 2024-05-10
Update ERC-7432: Removed non-unique roles

ernanirst:
After speaking with other developers about ERC-7432, we reach the conclusion that the interface is becoming overly complex by attempting to support multiple use cases.

We are simplifying this EIP by removing the non-unique roles use case.


JanKortes212 (changes requested):
hi guys help me please! whu can inform me how to colect ethereum airdrop from ghithub? thank you in advance !

PR #433, opened 2024-05-21
Add ERC: Smart Contract Delegation

rojotek:
Adds a new EIP that specifies a delegation interface for smart contract accounts.

This proposal defines interfaces for smart contracts to delegate capabilities to other smart contracts,
smart contract accounts, or Externally Owned Accounts (EOAs). The delegating contract MUST
implement ERC-1271 along with the `IDelegator` interface. This framework empowers a
delegating contract with the ability to delegate any actions it has the authority to perform,
thereby enabling more flexible and scalable contract interactions. This standard outlines the
minimal interface necessary to facilitate such delegation.

Additionally, this proposal is compatible with ERC-4337, although its implementation does not
necessitate ERC-4337.

PR #436, opened 2024-05-24
Add ERC: Grant Permissions from Wallets

pedrouid:
Adds JSON-RPC method for granting permissions from a wallet


ERCS/eip-draft-request-permissions-from-wallets.md
```diff
@@ -0,0 +1,436 @@
+---
```
abcoathup, line 1:
```suggestion
---
eip: 7715
```

ERCS/eip-draft-request-permissions-from-wallets.md
```diff
@@ -0,0 +1,436 @@
...
+title:  Request Permissions from Wallets
```
abcoathup, line 2:
```suggestion
title: Request Permissions from Wallets
```

ERCS/eip-draft-request-permissions-from-wallets.md
```diff
@@ -0,0 +1,436 @@
...
+description: Adds JSON-RPC method for requesting permissions from a wallet
+author: Luka Isailovic (@lukaisailovic), Derek Rein (@arein), Dan Finlay (danfinlay), Derek Chiang (@derekchiang), Fil Makarov (@filmakarov), Pedro Gomes (@pedrouid)
+discussions-to: https://ethereum-magicians.org/t/erc-request-permissions-from-wallets/20100
```
abcoathup, line 5:
```suggestion
discussions-to: https://ethereum-magicians.org/t/erc-7715-request-permissions-from-wallets/20100
```

ERCS/erc-7715.md
```diff
@@ -0,0 +1,437 @@
...
+
+### Request Specification
+
+```tsx
+type IssuePermissionsRequestParams = {
+  signer?: {
+    type: {
+      name: string,
+      uuid?: string,
+    },
```
lukaisailovic, line 57:
```suggestion
    type: string
```

ERCS/erc-7715.md
```diff
@@ -0,0 +1,437 @@
...
+    data: any
+  }
+ 
+  permissions: {
+    type: {
+	    name: string,
+      uuid?: string,
+    }
```
lukaisailovic, line 65:
```suggestion
    type: string
```

ERCS/erc-7715.md
```diff
@@ -0,0 +1,437 @@
...
+
+`signer` is an optional field that identifies the key or account associated with the session.  When not specified, it’s implied that the wallet will manage the session.  See the “Wallet-managed Sessions” section for details.
+
+`permissions` is a list of permissions.  See “Permission Types” section for details.
+
+Each permission can be either `required` or not.  If a permission is `required`, the wallet MUST grant the permission if it approves the request.  Otherwise, the wallet may reject the permission even if it approves the request (presumably because it grants other permissions).
+
+`expiry` is a UNIX timestamp that specifies the time by which this session MUST expire.
+
+`type` is an object that MUST include a `name`. If a `uuid` key is included, the wallet MUST match against its own supported permission, to reduce the risk of method name collisions.
```
lukaisailovic, line 82:
```suggestion
```

ERCS/erc-7715.md
```diff
@@ -0,0 +1,437 @@
...
+			id:'did:key:zQ3shokFTS3brHcDQrn82RUDfCZESWL1ZdCEJwekUDPQiYBme',
+		}
+	}
+	
+	permissions: [
+		{
+			type: {
+			  name: 'erc20_spending_limit',
+			  uuid: '175e6c32-3009-483f-b320-0826f2fefa57',
+			},
```
lukaisailovic, line 100:
```suggestion
			type: 'erc20_spending_limit'
```

ERCS/erc-7715.md
```diff
@@ -0,0 +1,437 @@
...
+				erc20Address: '0x...',
+				limit: '0x...',
+			}
+			required: true
+		},
+		{
+			type: {
+			  name: 'gas_limit',
+			  uuid: 'e9a0d387-8648-4928-9a5c-1a2fbbeda931',
+			},
```
lukaisailovic, line 111:
```suggestion
			type: 'gas_limit',
```

ERCS/erc-7715.md
```diff
@@ -0,0 +1,437 @@
...
+### Response Specification
+
+```tsx
+
+type IssuePermissionsResponse = {
+  grantedPermissions: {
+    type: {
+      name: string,
+      uuid?: string,
+    },
```
lukaisailovic, line 132:
```suggestion
    type: string
```

ERCS/erc-7715.md
```diff
@@ -0,0 +1,437 @@
...
+`wallet_issuePermissions` response example:
+
+```tsx
+{
+  grantedPermissions: [
+    {
+      type: {
+			  name: 'erc20_spending_limit',
+			  uuid: '175e6c32-3009-483f-b320-0826f2fefa57',
+			},
```
lukaisailovic, line 167:
```suggestion
      type: 'erc20_spending_limit',
```

ERCS/erc-7715.md
```diff
@@ -0,0 +1,437 @@
...
+      data: {
+        erc20Address:'0x...',
+        limit: '0x...',
+      },
+    },
+    {
+		  type: {
+			  name: 'gas_limit',
+			  uuid: 'e9a0d387-8648-4928-9a5c-1a2fbbeda931',
+			},
```
lukaisailovic, line 177:
```suggestion
		  type: 'gas_limit',
```

ERCS/erc-7715.md
```diff
@@ -0,0 +1,437 @@
...
+}
+```
+
+## Signer Types
+
+This ERC does not specify an exhaustive list of signer types, since we expect more types of signers to be developed as wallets get more advanced.  A signer type is valid as long as both the DApp and the wallet are willing to support it.
+
+However, if more than two signers use the same type name, a DApp could request with one type of signer while the wallet grants another type.  Therefore, it’s important that signer types are globally unique.
+
+Therefore, standardized signer types spec SHOULD include a UUID key, to help distinguish an intended specification/implementation, where `uuid` is a UUID. The `type` object can be extended later to support additional fields, like a version, or expected type enforcement, to further ensure the host wallet is acting in the way the requesting application expects.
```
lukaisailovic, line 197:
```suggestion
The object can be extended later to support additional fields, like a version, or expected type enforcement, to further ensure the host wallet is acting in the way the requesting application expects.
```

ERCS/erc-7715.md
```diff
@@ -0,0 +1,437 @@
...
+    // TODO
+  }
+}
+
+// The number of calls the session can make during each interval
+type RateLimitPermission = {
+  type: {
+    name: 'contract-call',
+  }
+  data: {
```
lukaisailovic, line 278:
```suggestion
  type: 'native-token-limit',
  data: {
    amount: "0x...",  // wei in hex
  }
}

// ERC20 spending limit
type ERC20LimitPermission = {
  type: 'erc20-limit',
  data: {
    erc20Address: "0x..."  // ERC20 token address
    amount: "0x...",  // wei in hex
  }
}

// Gas spending limit
type GasLimitPermission = {
  type: 'gas-limit',
  data: {
    amount: "0x...",  // wei in hex
  }
}

// Allowlist of contract calls
type ContractCallPermission = {
  type: 'contract-call',
  data: {
    // TODO
  }
}

// The number of calls the session can make during each interval
type RateLimitPermission = {
  type: 'rate-limit',
  data: {
```

ERCS/erc-7715.md
```diff
@@ -0,0 +1,437 @@
...
+The wallet SHOULD include `signerTypes` (`string[]`) and `permissionTypes` (`string[]`) in the response, to specify the signer types and permission types it supports.
+Example:
+
+```json
+{
+  "0x123": {
+    "permissions": {
+      "supported": true,
+      "signerTypes": ["keys", "account"],
+      "permissionTypes": [{ name: "erc20-limit" }, { name: "contract-call" }]
```
lukaisailovic, line 331:
```suggestion
      "permissionTypes": ["erc20-limit", "contract-call"]
```

ERCS/erc-7715.md
```diff
@@ -0,0 +1,466 @@
...
+  }[]
+      
+  expiry: number
+  
+  signerData: {
+    userOpBuilder?: `0x${string}`
+    submitToAddress?: `0x${string}`
+  },
+  
+	factory?: `0x${string}`,
```
xinbenlv, line 127: Can we change tab / space indentation to be consistent within the file?

ERCS/erc-7715.md
```diff
@@ -0,0 +1,466 @@
+---
+eip: 7715
+title: Request Permissions from Wallets
+description: Adds JSON-RPC method for requesting permissions from a wallet
```
xinbenlv, line 4: Editorial: Can you clarify "who" requesting permissions from wallet? It's both important from a description to provide the intended requester as design target but also to imply how does wallet identify who IS being approved, which is a technical question. 

Techincal: I will ask the question as a peer reviewer in the discussion thread

ERCS/erc-7715.md
```diff
@@ -0,0 +1,466 @@
...
+  },
+  factory?: `0x${string}`,
+  factoryData?: string,
+	
+  permissionsContext: string
+}
+```
+
+- `grantedPermissions` is a list of permissions that were granted.  The wallet MUST grant `required` permissions if the request was approved.
+- `expiry` is the expiry time that was granted.  The wallet MAY grant a different expiry time than the one that the DApp requested.
```
himanshuchawla009, line 132: can we describe about `signerData.userOpBuilder` and `signerData.submitToAddress` fields.

ERCS/erc-7715.md
```diff
@@ -0,0 +1,466 @@
...
+---
+
+## Abstract
+
+We define a new JSON-RPC method `wallet_issuePermissions` for DApp to request permissions from a wallet in order to execute transactions on users’s behalf.  This enables two use cases:
+
+- Executing transactions for users without a wallet connection.
+- Executing transactions for users with a wallet connection that is scoped with permissions.
+
+##Motivation
```
glitch-txs, line 21:
```suggestion
## Motivation
```

Typo?

ERCS/erc-7715.md
```diff
@@ -0,0 +1,466 @@
...
+
+In the context of AA, there are multiple vendor-specific implementations of session keys, which are temporary keys imbued with specific permissions.  However, since the implementations are vendor-specific, it’s impossible for DApps to “request” session keys from wallets in a unified way, regardless of the specific wallet implementations.
+
+## Specification
+
+The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL NOT”, “SHOULD”, “SHOULD NOT”, “RECOMMENDED”, “MAY”, and “OPTIONAL” in this document are to be interpreted as described in RFC 2119.
+
+### `wallet_issuePermissions`
+
+We introduce a `wallet_issuePermissions` method for the DApp to request permissions from the wallet.  We’d prefer `wallet_requestPermissions` but it’s already taken with [ERC-2255](./eip-2255.md).
```
glitch-txs, line 47:
```suggestion
We introduce a `wallet_issuePermissions` method for the DApp to request permissions from the wallet.
```

ERCS/erc-7715.md
```diff
@@ -0,0 +1,466 @@
...
+    type: string
+    data: any
+    required: boolean
+  }[]
+  
+  expiry: number,
+}
+```
+
+`signer` is an optional field that identifies the key or account associated with the session.  When not specified, it’s implied that the wallet will manage the session.  See the “Wallet-managed Sessions” section for details.
```
glitch-txs, line 68:
```suggestion
`signer` is an optional field that identifies the key or account associated with the session.  When not specified, it’s implied that the wallet will manage the session. See the “Wallet-managed Sessions” section for details.
```

ERCS/erc-7715.md
```diff
@@ -0,0 +1,466 @@
...
+  data: {
+    count: number,  // the number of times during each interval
+    interval: number,  // in seconds
+  }
+}
+```
+
+### Wallet-managed Sessions
+
+If `signer` is not specified in the request, the wallet manages the session.  If the wallet approves the request, it MUST accept [ERC-5792](./eip-5792.md)’s `wallet_sendCalls` with the `permissions` capability, which MAY include the session with a `permissionsContext`.  For example:
```
glitch-txs, line 251:
```suggestion
If `signer` is not specified in the request, the wallet manages the session. If the wallet approves the request, it MUST accept [ERC-5792](./eip-5792.md)’s `wallet_sendCalls` with the `permissions` capability, which MAY include the session with a `permissionsContext`.  For example:
```

ERCS/erc-7715.md
```diff
@@ -0,0 +1,466 @@
...
+  data: '0x'
+};
+
+// Alice sends the transaction by calling redeemDelegation on Bob's account
+const tx = await bob.sendTransaction({
+  to: submitToAddress,
+  data: bob.interface.encodeFunctionData('redeemDelegation', [
+    permissionsContext,
+    action
+  ])
```
jxom, line 418: should we make the pseudocode more concise + less opinionated?
```suggestion
  data: abi.encode(bytes permissionsContext, string action)  
```

ERCS/erc-7715.md
```diff
@@ -0,0 +1,466 @@
...
+// Alice sends the transaction by calling redeemDelegation on Bob's account
+const tx = await bob.sendTransaction({
+  to: submitToAddress,
+  data: bob.interface.encodeFunctionData('redeemDelegation', [
+    permissionsContext,
+    action
+  ])  
+});
+
+await tx.wait();
```
jxom, line 421: seems unnessessary

```suggestion
```

ERCS/erc-7715.md
```diff
@@ -0,0 +1,466 @@
...
+});
+
+// Extract the permissionsContext and submitToAddress
+const permissionsContext = permissionsResponse.permissionsContext;
+const submitToAddress = permissionsResponse.signerData.submitToAddress;
+
+// Alice forms the action she wants Bob's account to take
+const action = {
+  to: alice.address,
+  value: ethers.utils.parseEther('0.5'),
```
jxom, line 408:
```suggestion
  value: 0.5 ether,
```

ERCS/erc-7715.md
```diff
@@ -0,0 +1,466 @@
...
+    signer: {
+      type: 'account',
+      data: {
+        id: bob.address
+      }  
+    },
+    permissions: [{
+      type: 'native-token-limit',
+      data: {
+        amount: ethers.utils.parseEther('1.0')
```
jxom, line 393:
```suggestion
        amount: 1.0 ether
```

ERCS/erc-7715.md
```diff
@@ -0,0 +1,466 @@
...
+
+```jsx
+const getSignature = async ({ address, userOp, permissionsContext }: GetSignatureArgs) => {
+    return readContract(config, {
+      abi: BUILDER_CONTRACT_ABI,
+      address: BUILDER_CONTRACT_ADDRESS,
+      functionName: 'getSignature',
+      args: [address, userOp, permissionsContext]
+    })
+  }
```
jxom, line 334: Should we make this code pseudocode + less opinionated?

```suggestion
address addr = "0x..."
PackedUserOperation userOp = { ... }
bytes permissionsContext = { ... }
bytes signature = builderContract.getSignature(addr, userOp, permissionsContext)
```

ERCS/erc-7715.md
```diff
@@ -0,0 +1,466 @@
...
+  factoryData?: string,
+	
+  permissionsContext: string
+}
+```
+
+- `grantedPermissions` is a list of permissions that were granted.  The wallet MUST grant `required` permissions if the request was approved.
+- `expiry` is the expiry time that was granted.  The wallet MAY grant a different expiry time than the one that the DApp requested.
+- `factory` and `factoryData` are as defined in [ERC-4337](./eip-4337.md).  They are either both specified, or none.  If the account has not yet been deployed, the wallet MUST return `factory` and `factoryData`, and the DApp MUST deploy the account by calling the `factory` contract with `factoryData` as the calldata.
+- `permissionsContext` is an opaque string that identifies the session. It MAY be the `context` as defined in [ERC-7679](./eip-7679.md) and [ERC-7710](./eip-7710.md).  See “Rationale” for details.
```
jxom, line 134: Would be good if this property was named something more opaque and simple as well, such as `id`.

Current naming makes it sound like it is only meant to be used in a "context" sense (ie. with ERC-7679 + ERC-7710).

ERCS/erc-7715.md
```diff
@@ -0,0 +1,466 @@
...
+  factory?: `0x${string}`,
+  factoryData?: string,
+	
+  permissionsContext: string
+}
+```
+
+- `grantedPermissions` is a list of permissions that were granted.  The wallet MUST grant `required` permissions if the request was approved.
+- `expiry` is the expiry time that was granted.  The wallet MAY grant a different expiry time than the one that the DApp requested.
+- `factory` and `factoryData` are as defined in [ERC-4337](./eip-4337.md).  They are either both specified, or none.  If the account has not yet been deployed, the wallet MUST return `factory` and `factoryData`, and the DApp MUST deploy the account by calling the `factory` contract with `factoryData` as the calldata.
```
jxom, line 133:
```suggestion
- `factory` and `factoryData` are as defined in [ERC-4337](./eip-4337.md).  They are either both specified, or none.  If the account has not yet been deployed, the wallet MUST return `factory` and `factoryData`, and the DApp MUST deploy the account by calling the `factory` contract with `factoryData` as the calldata.
```

Could be good to add background for these in "Rationale".

ERCS/erc-7715.md
```diff
@@ -0,0 +1,466 @@
...
+}
+```
+
+`signer` is an optional field that identifies the key or account associated with the session.  When not specified, it’s implied that the wallet will manage the session.  See the “Wallet-managed Sessions” section for details.
+
+`permissions` is a list of permissions.  See “Permission Types” section for details.
+
+Each permission can be either `required` or not.  If a permission is `required`, the wallet MUST grant the permission if it approves the request.  Otherwise, the wallet may reject the permission even if it approves the request (presumably because it grants other permissions).
+
+`expiry` is a UNIX timestamp that specifies the time by which this session MUST expire.
```
jxom, line 74:
```suggestion
`expiry` is a UNIX timestamp (in seconds) that specifies the time by which this session MUST expire.
```

ERCS/erc-7715.md
```diff
@@ -0,0 +1,466 @@
...
+
+### **Limited Permission Scope**
+
+DApps should only request the permissions they need, with a reasonable expiration time.
+
+Wallets must correctly enforce permissions.  Ultimately, users must trust that their wallet software is implemented correctly, and permissions should be considered a part of the wallet implementation.
+
+### **Phishing Attacks**
+
+Malicious DApps could pose as legitimate applications and trick users into granting broad permissions.  Wallets should always clearly display the permissions to users and warn them against granting dangerous permissions.
```
jxom, line 462:
```suggestion
Malicious DApps could pose as legitimate applications and trick users into granting broad permissions.  Wallets MUST clearly display the permissions to users and warn them against granting dangerous permissions.
```

ERCS/erc-7715.md
```diff
@@ -0,0 +1,466 @@
...
+
+Wallet that don’t support `wallet_issuePermissions` SHOULD return error message if the JSON-RPC method is called.
+
+## **Security Considerations**
+
+### **Limited Permission Scope**
+
+DApps should only request the permissions they need, with a reasonable expiration time.
+
+Wallets must correctly enforce permissions.  Ultimately, users must trust that their wallet software is implemented correctly, and permissions should be considered a part of the wallet implementation.
```
jxom, line 458:
```suggestion
Wallets MUST correctly enforce permissions.  Ultimately, users must trust that their wallet software is implemented correctly, and permissions should be considered a part of the wallet implementation.
```

ERCS/erc-7715.md
```diff
@@ -0,0 +1,466 @@
...
+
+Example:
+
+```json
+{
+  //...
+  "sessionProperties": {
+    "permissions": "true",
+    "signerTypes": "keys,account",
+    "permissionTypes": "erc20-limit,contract-call"
```
jxom, line 311: Do these need to be serialized like this? Why not an array?

ERCS/erc-7715.md
```diff
@@ -0,0 +1,466 @@
...
+
+If the request is malformed or the wallet is unable/unwilling to issue permissions, wallet MUST return an error with a code as defined in [ERC-1193](./eip-1193.md).
+
+`wallet_issuePermissions` response example:
+
+```tsx
+{
+  grantedPermissions: [
+    {
+      type: 'erc20_spending_limit',
```
jxom, line 144:
```suggestion
      type: 'erc20-spending-limit',
```

ERCS/erc-7715.md
```diff
@@ -0,0 +1,466 @@
...
+      data: {
+        erc20Address:'0x...',
+        limit: '0x...',
+      },
+    },
+    {
+      type: 'gas_limit',
```
jxom, line 151:
```suggestion
      type: 'gas-limit',
```

ERCS/erc-7715.md
```diff
@@ -0,0 +1,466 @@
...
+
+`wallet_issuePermissions` response example:
+
+```tsx
+{
+  grantedPermissions: [
+    {
+      type: 'erc20_spending_limit',
+      data: {
+        erc20Address:'0x...',
```
jxom, line 146:
```suggestion
        erc20Address: '0x...',
```

ERCS/erc-7715.md
```diff
@@ -0,0 +1,466 @@
...
+{
+  signer: {
+    type: 'key',
+    data: {
+      id:'did:key:zQ3shokFTS3brHcDQrn82RUDfCZESWL1ZdCEJwekUDPQiYBme',
+      }
+    }
+  permissions: [
+    {
+      type: 'erc20_spending_limit',
```
jxom, line 89:
```suggestion
      type: 'erc20-spending-limit',
```

ERCS/erc-7715.md
```diff
@@ -0,0 +1,466 @@
...
+      data: {
+        erc20Address: '0x...',
+        limit: '0x...',
+      }
+      required: true
+    },
+    {
+      type: 'gas_limit',
```
jxom, line 97:
```suggestion
      type: 'gas-limit',
```

ERCS/erc-7715.md
```diff
@@ -0,0 +1,466 @@
...
+  
+  expiry: number,
+}
+```
+
+`signer` is an optional field that identifies the key or account associated with the session.  When not specified, it’s implied that the wallet will manage the session.  See the “Wallet-managed Sessions” section for details.
+
+`permissions` is a list of permissions.  See “Permission Types” section for details.
+
+Each permission can be either `required` or not.  If a permission is `required`, the wallet MUST grant the permission if it approves the request.  Otherwise, the wallet may reject the permission even if it approves the request (presumably because it grants other permissions).
```
jxom, line 72: Why isn't everything required? Adding this will make it so hard for a wallet to correctly convey to its user what they are really permitting.

Seems odd for an end-user to optionally set a limit, only to have it ignored. What is the use case for that?

ERCS/erc-7715.md
```diff
@@ -0,0 +1,466 @@
...
+}
+
+// A signer representing a multisig signer.
+// Each element of `ids` is a DID.
+type MultiKeySigner = {
+  type: 'keys',
+  data: {
+    ids: string[],
+  }
+}
```
jxom, line 192: Any examples of these? It is a bit unclear how these work.

ERCS/erc-7715.md
```diff
@@ -0,0 +1,466 @@
...
+
+// Allowlist of contract calls
+type ContractCallPermission = {
+  type: 'contract-call',
+  data: {
+    // TODO
+  }
+}
```
jxom, line 237: What is the use-case for this?

ERCS/erc-7715.md
```diff
@@ -0,0 +1,466 @@
...
+
+// The number of calls the session can make during each interval
+type RateLimitPermission = {
+  type: 'rate-limit',
+  data: {
+    count: number,  // the number of times during each interval
+    interval: number,  // in seconds
+  }
+}
```
jxom, line 246: What is the use-case for this?

jxom, line 246: What is an example/use case of this?

ERCS/erc-7715.md
```diff
@@ -0,0 +1,466 @@
...
+      },
+      {
+        "to": "0xd46e8dd67c5d32be8058bb8eb970870f07244567",
+        "value": "0x182183",
+        "data": "0xfbadbaf01"
+      }
+    ],
+    "capabilities": {
+      "permissions": {
+        "context": "<permissionsContext>"
```
jxom, line 273:
```suggestion
        "context": "<permissionsContext>"
```

`id` seems clearer IMO.

jxom:
How do I revoke permissions?

ERCS/erc-7715.md
```diff
@@ -0,0 +1,466 @@
...
+  data: bob.interface.encodeFunctionData('redeemDelegation', [
+    permissionsContext,
+    action
+  ])  
+});
+
+await tx.wait();
+```
+
+## Rationale
```
jxom, line 424: How do we revoke permissions?

ERCS/erc-7715.md
```diff
@@ -0,0 +1,466 @@
...
+status: Draft
+type: Standards Track
+category: ERC
+created: 2024-05-24
+requires: 4337, 5792, 7679, 7710
+---
+
+## Abstract
+
+We define a new JSON-RPC method `wallet_issuePermissions` for DApp to request permissions from a wallet in order to execute transactions on users’s behalf.  This enables two use cases:
```
jxom, line 16: Wonder if `wallet_grantPermissions` would sound better as it may have more symmetry if we ever add a "revoke" term.

ERCS/erc-7715.md
```diff
@@ -0,0 +1,466 @@
...
+
+- Executing transactions for users without a wallet connection.
```
jxom, line 18: How would scheduled/recurring transactions work with this ERC? Is that another permission type, or can it be used with the existing ones listed?

wcalderipe: I assume the user MUST issue permissions with a long expiry time or undefined for permissions that don't expire.

ERCS/erc-7715.md
lukaisailovic: Limiting the call to a single smart contract method. For example, this permissions are valid only for using `purhcaseNft` method on smart contract x with the amount y. Its just an example of how permissions might look like.

ERCS/erc-7715.md
lukaisailovic: Useful for recurring transactions. For example issuing permissions to a key and saying limit 0.5 ETH doesn't stop the key to consume the permissions every minute. Its just an example of how permissions might look like.

ERCS/erc-7715.md
lukaisailovic: When DApp requests permissions, required ones MUST be issued otherwise some flow DApp side will not work. If the permission is not required, user may or may not grant it and the flows dapp side will not break.

ERCS/erc-7715.md
lukaisailovic: CAIP-25 only supports `<string>:<string> `type object

ERCS/erc-7715.md
lukaisailovic: I would argue context is more opaque. `id` associates me with uuid or some kind of number, `permissionsContext` can contain any data.

ERCS/erc-7715.md
danfinlay: This wouldn't be valid javascript.

Also: Feel free to modify or add an example to show viem usage!

jxom: I thought we were representing pseudocode 😜

ERCS/erc-7715.md
jxom: Isn’t it described as an identifier though?

ERCS/erc-7715.md
pedrouid: I like it 👌

ERCS/erc-7715.md
jxom: Makes sense!

ERCS/erc-7715.md
```diff
@@ -0,0 +1,466 @@
...
+- If a DApp leverages [ERC-7710](./eip-7710.md), it could use `permissionsContext` as the `_data` when interacting with the delegation manager.
+- If a DApp leverages in-app sessions, it would use `permissionContext` as an identifier of the session when using `wallet_sendCalls`.
+
+### Non-exhaustive list of signers and permissions
+
+With the advancement in wallet technologies, we expect new types of signers and permissions to be developed.  We considered mandating that each signer and permission must have a UUID in order to avoid collisions, but ultimately decided to stick with the simpler approach for now of simply mandating that these types be defined in ERCs.
+
+## **Backwards Compatibility**
+
+Wallet that don’t support `wallet_issuePermissions` SHOULD return error message if the JSON-RPC method is called.
```
usame-algan, line 450:
```suggestion
Wallets that don’t support `wallet_issuePermissions` SHOULD return an error message if the JSON-RPC method is called.
```

ERCS/erc-7715.md
```diff
@@ -0,0 +1,466 @@
...
+```
+
+### Signer & Permission Types
+
+In this ERC, we specify a list of signers and permissions that we expect to be commonly used.
+
+This ERC does not specify an exhaustive list of signer or permission types, since we expect more signer/permission types to be developed as wallets get more advanced.  A signer/permission type is valid as long as both the DApp and the wallet are willing to support it.
+
+However, if two signers or two permissions share the same type name, a DApp could request with one type of signer/permission while the wallet grants another.  Therefore, it’s important that no two signers or two permissions share the same type.  Therefore, new signer or permission types should be specified in ERCs, either in this ERC as an amendment or in another ERC.
+
```
DmitryBespalov, line 172: Thanks so much for the ERC! After reading the text, I've got a few questions. I might be missing some overall context, but I hope they're relevant.

1) Do those permission requests need an identifier so that the wallet can distinguish between different permissions for different Dapps? Can there be multiple 'issue permission requests' for the same Dapp?

2) Also, does a wallet need to know what specific Dapp is calling these permission requests, so that it can later identify and authenticate the requests with permissions? (It looks like this data would go into `permissionContext`.)

3) For ERC completeness, what about revoking issued permissions? Without this, there seems to be no standard way to stop potentially automated actions.

5) And what about getting and modifying previously issued permissions? Would it be useful for a Dapp to check the current status of those and request changes if needed?

ERCS/erc-7715.md
```diff
@@ -0,0 +1,466 @@
...
+```
+// Alice requests a permission from Bob
+const permissionsResponse = await bob.request({
+  method: 'wallet_issuePermissions',
+  params: [{
+    signer: {
+      type: 'account',
+      data: {
+        id: bob.address
```
schmanu, line 387: Earlier you describe this field as
> `signer` is an optional field that identifies the key or account associated with the session

And here Alice requests a session from Bob. Wouldn't the session's signer then be 

```suggestion
        id: alice.address
```

ERCS/erc-7715.md
lukaisailovic: CAIP-25 is the most efficient way to discover capabilities since it eliminates additional network call.

ERCS/erc-7715.md
lukaisailovic: Description is inaccurate. `permissionsContext` would encapsulate the data to both identify and authenticate the session. It could be a signature over some kind of identifier if the permissions data is stored on chain, but in case its not stored on chain `permissionsContext` could contain signature over entire permissions array. That would force the user to provide the permissions with each userOp, which is cheaper if the storage price is higher than execution price (case of L1).

ERCS/erc-7715.md
```diff
@@ -0,0 +1,466 @@
...
+Each permission can be either `required` or not.  If a permission is `required`, the wallet MUST grant the permission if it approves the request.  Otherwise, the wallet may reject the permission even if it approves the request (presumably because it grants other permissions).
+
+`expiry` is a UNIX timestamp that specifies the time by which this session MUST expire.
+
+
+Request example:
+
+```tsx
+{
+  signer: {
```
lukaisailovic, line 81:
```suggestion
{
  account: '0x...',
  
  signer: {
```

ERCS/erc-7715.md
```diff
@@ -0,0 +1,466 @@
...
+}
+```
+
+`signer` is an optional field that identifies the key or account associated with the session.  When not specified, it’s implied that the wallet will manage the session.  See the “Wallet-managed Sessions” section for details.
+
+`permissions` is a list of permissions.  See “Permission Types” section for details.
+
+Each permission can be either `required` or not.  If a permission is `required`, the wallet MUST grant the permission if it approves the request.  Otherwise, the wallet may reject the permission even if it approves the request (presumably because it grants other permissions).
+
+`expiry` is a UNIX timestamp (in seconds) that specifies the time by which this session MUST expire.
```
kopy-kat, line 76: can wallets enforce that this expiry is below some threshold? seems like the incentive of the dapp would just be to give long expiry since that will be the best ux

ERCS/erc-7715.md
```diff
@@ -0,0 +1,466 @@
...
+  }[]
+      
+  expiry: number
+  
+  signerData: {
+    userOpBuilder?: `0x${string}`
+    managerAddress?: `0x${string}`
+  },
+  factory?: `0x${string}`,
+  factoryData?: string,
```
kopy-kat, line 129:
```suggestion
  factoryData?: `0x${string}`,
```

ERCS/erc-7715.md
```diff
@@ -0,0 +1,466 @@
...
+
+
+Request example:
+
+```tsx
+{
+  signer: {
+    type: 'key',
+    data: {
+      id:'did:key:zQ3shokFTS3brHcDQrn82RUDfCZESWL1ZdCEJwekUDPQiYBme',
```
kopy-kat, line 88: where does the dapp know this credential from?

ERCS/erc-7715.md
```diff
@@ -0,0 +1,466 @@
...
+    },
+    {
+      type: 'gas_limit',
+      data: {
+        limit: '0x...',
+      },
+      required: false
+    }
+  ],
+
```
kopy-kat, line 162: would it make sense here to add signerdata and factory+data to show all the main fields?

ERCS/erc-7715.md
```diff
@@ -0,0 +1,466 @@
...
+    amount: "0x...",  // wei in hex
+  }
+}
+
+// ERC20 spending limit
+type ERC20LimitPermission = {
+  type: 'erc20-limit',
+  data: {
+    erc20Address: "0x..."  // ERC20 token address
+    amount: "0x...",  // wei in hex
```
kopy-kat, line 223:
```suggestion
    amount: "0x...",  // amount in hex
```

ERCS/erc-7715.md
```diff
@@ -0,0 +1,466 @@
...
+  data: {
+    amount: "0x...",  // wei in hex
+  }
+}
+
+// Allowlist of contract calls
+type ContractCallPermission = {
+  type: 'contract-call',
+  data: {
+    // TODO
```
kopy-kat, line 239:
```suggestion
   {
     address: "0x...",  // contract address
     selectors: ["0x...",],  // list of allowed selectors (if list == [] then any is allowed)
  }[]
```

ERCS/erc-7715.md
```diff
@@ -0,0 +1,466 @@
...
+    "calls": [
+      {
+        "to": "0xd46e8dd67c5d32be8058bb8eb970870f07244567",
+        "value": "0x9184e72a",
+        "data": "0xd46e8dd67c5d32be8d46e8dd67c5d32be8058bb8eb970870f072445675058bb8eb970870f072445675"
+      },
+      {
+        "to": "0xd46e8dd67c5d32be8058bb8eb970870f07244567",
+        "value": "0x182183",
+        "data": "0xfbadbaf01"
```
kopy-kat, line 272: this isnt valid calldata fyi

ERCS/erc-7715.md
pedrouid: Any actions here or should we close this thread?

ERCS/erc-7715.md
pedrouid: isn't it better to just add scientific notation here
```suggestion
        amount: 1e18
```

ERCS/erc-7715.md
pedrouid:
```suggestion
  value: 0.5e18
```

ERCS/erc-7715.md
pedrouid: agreed

ERCS/erc-7715.md
pedrouid: technically the permissions are controlled by a smart contract so that would be specific to its implementation

unless you are referring to the dapp actually requesting to revoke?

ERCS/erc-7715.md
jxom: Would be good to update the description of this property!

ERCS/erc-7715.md
pedrouid: did:key are self-resolvable as they are just encoding of public keys

https://w3c-ccg.github.io/did-method-key/

ERCS/erc-7715.md
pedrouid: we could add an RPC error message where the wallet can reject for expiry that is too high

ERCS/erc-7715.md
pedrouid: but technically these are optionals and signerData has both a 7679 and a 7710 scenario

so we would need separate examples for each

ERCS/erc-7715.md
lukaisailovic: This is an example request. The key field is indicating what key will be the signer for requested permissions. It could be dapp managed key or user owned passkey. Either way its in the dapp context

ERCS/erc-7715.md
```diff
@@ -0,0 +1,481 @@
...
+        erc20Address: '0x...',
+        limit: '0x...',
+      },
+    },
+    {
+      type: 'gas-limit',
+      data: {
+        limit: '0x...',
+      },
+      required: false
```
howydev, line 159: should this line be removed?

ERCS/erc-7715.md
howydev: +1 on modifying permissions - curious to hear the flow that you guys have in mind!

lukaisailovic: 1. Wallet will take part in issuing permissions, but after that its up to smart contract to verify them given the signature and `permissionsContext`. Knowing which Dapp is trying to consume the permissions doesn't really matter. There can be multiple issue permissions request for the same DApp. 

2. Permissions are not validated by the wallet. They are issued by the wallet, but after that wallet connection is not required (unless using wallet based signer type). They are verified on-chain.

3. This would require us defining something like `wallet_revokePermissions`. It makes sense, but its not necessarily needed. We wanted to leave revocation up to the account implementatators but it might make sense to include it here as well. 
4. Its pretty hard to make this flow feasible. Some implementations might return some kind of identifier inside the `permissionsContext` while storing the permissions on chain. In that case the modification would be possible. In other cases they would return the entire permissions array (L1 - Storage is expensive).  
Alternative is for modify to be revoke + issue again

ERCS/erc-7715.md
pedrouid, line 159:
```suggestion
```

pedrouid: You are right

ERCS/erc-7715.md
howydev: Thanks for adding more color to the modification flow! 

I think the flow of revoke + issue again doesn't map 1 to 1 with updating permissions, especially with "logistics" based permissions:
1. For interval spend limits, it might (unexpectedly) refresh any tracked spend for an entity/multiple entities during the interval
2. For transactions gated behind time delays, revoke + issue again would empty the queue

howydev: Some way to modify allows the use of `increaseAllowance`/`decreaseAllowance` calls which guards against the ERC20 frontrunning-allowance-consumption "vulnerability" with ERC20's `approve`. Vulnerability is in quotation marks b/c I recall that although this would potentially be an issue, it isn't something that has been exploited a single time in all of eth's history. But worth considering since the flows here would quite heavily dictate the flows that are possible to perform on chain

lukaisailovic: That's a really good point! 
Still, I wouldn't like to limit all implementations by forcing modifying permissions flow that would severely restrict what can be put into the `permissionsContext`. If the specific permission wants to support `increaseAllowance`/`decreaseAllowance` smart contract calls, that would be fine but shouldn't be enforced.

ERCS/erc-7715.md
```diff
@@ -0,0 +1,480 @@
+---
+eip: 7715
+title: Grant Permissions from Wallets
+description: Adds JSON-RPC method for granting permissions from a wallet
+author: Luka Isailovic (@lukaisailovic), Derek Rein (@arein), Dan Finlay (@danfinlay), Derek Chiang (@derekchiang), Fil Makarov (@filmakarov), Pedro Gomes (@pedrouid)
+discussions-to: https://ethereum-magicians.org/t/erc-7715-request-permissions-from-wallets/20100
```
pedrouid, line 6:
```suggestion
discussions-to: https://ethereum-magicians.org/t/erc-7715-grant-permissions-from-wallets/20100
```

ERCS/erc-7715.md
```diff
@@ -0,0 +1,480 @@
...
+      id:'did:key:zQ3shokFTS3brHcDQrn82RUDfCZESWL1ZdCEJwekUDPQiYBme',
+      }
+    }
+  permissions: [
+    {
+      type: 'erc20-spending-limit',
+      data: {
+        erc20Address: '0x...',
+        limit: '0x...',
+      }
```
pedrouid, line 98:
```suggestion
      },
```

ERCS/erc-7715.md
```diff
@@ -0,0 +1,480 @@
...
+        to: "0xd46e8dd67c5d32be8058bb8eb970870f07244567",
+        value: "0x182183",
+        data: "0xfbadbaf01",
+      },
+    ],
+    capabilities: {
+      permissions: {
+        context: "<permissionsContext>",
+      },
+    },
```
ilikesymmetry, line 285: Can we also add a field for `capabilities.permissions.signature`? This way the app can sign things with their session key and easily pass this to the wallet for validation and userOp bundling along with the `context`. Passing an identifier for the session is definitely needed, but the wallet receiving `sendCalls` will also need a signature that provides per-userOp validation.

```suggestion
    capabilities: {
      permissions: {
        context: "<permissionsContext>",
        signature: "<optional signature from session key>"
      },
    },
```

ilikesymmetry: Conversation thread from telegram [here](https://t.me/c/2006770962/1136).

There are currently two paths in 7715:
1. App does everything without the wallet (7679/7710)
2. App leverages wallet connection in order to send transactions

This comment's proposal is a nuance of 2:
2a) wallet provisions session keys for app
2b) app brings its own session key

For modularity of product offerings, I think 2b is very important to support.

Consider a 2x2 grid with one axis for signing done by wallet/app and other axis for userOp submission by wallet/app.

|  | wallet signs | app signs |
| -- | -- | -- |
| wallet submits |  supported | _missing_ |
| app submits     | signTransaction | 7679 |

Wallets are able to prompt signing and submit userOps by definition.

If apps want maximum control they can own signing and use 7679 to submit userOps.

Something is missing to let apps own signing, but lean on the wallet for userOp submission.

To enable this missing box of app signs + wallet submits, we need to pass a signature from app to wallet and therefore I think this new and optional `capabilities.permissions.signature` field is the simplest approach.

ERCS/erc-7715.md
```diff
@@ -0,0 +1,480 @@
...
+  data: {
+    id: `0x${string}`;
+  };
+};
+```
+
+### Permissions Types
+
+```tsx
+// Native token spending limit, e.g. ETH on Ethereum
```
ilikesymmetry, line 216: I think we need to spec these types more. Adding my current interpretation.

```suggestion
// Native token spending limit, e.g. ETH on Ethereum
// This only enables simple transfers and does not support any calldata.
// A session is no longer allowed to spend native tokens when the cumulative amount spent in the session exceeds `amount` or when the session expires, whichever is first.
```

ERCS/erc-7715.md
```diff
@@ -0,0 +1,480 @@
...
+type NativeTokenLimitPermission = {
+  type: "native-token-limit";
+  data: {
+    amount: "0x..."; // wei in hex
```
ilikesymmetry, line 220: Can we make this `value` instead to be more aligned with `msg.value` and ERC20?

```suggestion
    value: "0x..."; // wei in hex
```

ERCS/erc-7715.md
```diff
@@ -0,0 +1,480 @@
...
+};
+
+// ERC20 spending limit
+type ERC20LimitPermission = {
+  type: "erc20-limit";
+  data: {
+    erc20Address: "0x..."; // ERC20 token address
+    amount: "0x..."; // wei in hex
+  };
+};
```
ilikesymmetry, line 231:
```suggestion
// ERC20 spending limit/
// This enables ERC20 transfers and approvals up to a limit value.
// A session is no longer allowed to spend ERC20s when the cumulative amount transferred or increased allowance in the session exceeds `value` or when the session expires, whichever is first.
type ERC20LimitPermission = {
  type: "erc20-limit";
  data: {
    chainId: "eip155:1"; // CAIP-2 chain id
    erc20Address: "0x..."; // ERC20 token address
    value: "0x..."; // wei in hex
  };
};
```

ERCS/erc-7715.md
```diff
@@ -0,0 +1,480 @@
...
+### Permissions Types
+
+```tsx
+// Native token spending limit, e.g. ETH on Ethereum
+type NativeTokenLimitPermission = {
+  type: "native-token-limit";
+  data: {
+    amount: "0x..."; // wei in hex
+  };
+};
```
ilikesymmetry, line 222: I think we should rename `amount` -> `value` to be more obviously aligned with `msg.value` and ERC20.

```suggestion
// Native token spending limit, e.g. ETH on Ethereum
// This only enables simple transfers and does not support any calldata.
// A session is no longer allowed to spend native tokens when the cumulative amount spent in the session exceeds `value` or when the session expires, whichever is first.
type NativeTokenLimitPermission = {
  type: "native-token-limit";
  data: {
    chainId: "eip155:1"; // CAIP-2 chain id
    value: "0x..."; // wei in hex
  };
};
```

ERCS/erc-7715.md
```diff
@@ -0,0 +1,480 @@
...
+
+// Gas spending limit
+type GasLimitPermission = {
+  type: "gas-limit";
+  data: {
+    amount: "0x..."; // wei in hex
+  };
+};
```
ilikesymmetry, line 239:
```suggestion
// Gas spending limit
// This enables spending native token in the account to pay for 4337 userOps up to a specific value.
// A session is no longer allowed to spend gas when the cumulative amount consumed in the session exceeds `value` or when the session expires, whichever is first.
type GasLimitPermission = {
  type: "gas-limit";
  data: {
    chainId: "eip155:1"; // CAIP-2 chain id
    value: "0x..."; // wei in hex
  };
};
```

ERCS/erc-7715.md
```diff
@@ -0,0 +1,480 @@
...
+// An account that can be granted with permissions as in ERC-7710.
+type AccountSigner = {
+  type: "account";
+  data: {
+    id: `0x${string}`;
+  };
+};
+```
+
+### Permissions Types
```
ilikesymmetry, line 213: I do not think we should have **any** canonical permission types in this EIP to keep this proposal tight. That said, I have dropped requested changes to improve the documentation of these examples.

Two general comments:
1. I think we should prefer `value` over `amount` to be more obviously connected to `msg.value` and ERC20 spec.
2. I think all of these a CAIP-2 `chainId` to be more specific about where these permissions can apply? I do not see a notion of chain-specificity in the top-level session definition, but I think it's better to apply it at the permissions layer anyways for modularity of request permissions across multiple networks simultaneously. Are we comfortable scaling multi-chain permissions by just requesting the same object but with a different `chainId` parameter?

ERCS/erc-7715.md
```diff
@@ -0,0 +1,466 @@
...
+
+`wallet_issuePermissions` response example:
+
+```tsx
+{
+  grantedPermissions: [
+    {
+      type: 'erc20_spending_limit',
+      data: {
+        erc20Address:'0x...',
```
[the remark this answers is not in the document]
pedrouid: We just addressed this with the did in the account

ERCS/erc-7715.md
```diff
@@ -0,0 +1,603 @@
...
+          type: 'call-limit',
+          data: {
+            count: 1,
+          },
+        },
+      ],
+      required: true,
+      data: {
+        address: '0x...',
+        allowance: '500000000'
```
ilikesymmetry, line 135: A previous iteration of this proposal used hex encoding for `uint256` values which I agree with for consistency with the Ethereum json-rpc spec: https://ethereum.org/en/developers/docs/apis/json-rpc/#quantities-encoding

ERCS/erc-7715.md
```diff
@@ -0,0 +1,601 @@
...
+  expiry: number;
+
+  signerData: {
+    userOpBuilder?: `0x${string}`;
+    managerAddress?: `0x${string}`;
+  };
+
+  factory?: `0x${string}`;
+
+  factoryData?: `0x${string}`;
```
ilikesymmetry, line 181: Can we bundle these together within a new optional field `accountData`? I don't see why these should be optional individually and it would be consistent with the pattern established with `signerData`.
```suggestion
  accountData?: {
    factory?: `0x${string}`;
    factoryData?: `0x${string}`;
  }
```
```suggestion
  factory?: `0x${string}`;

  factoryData?: `0x${string}`;
```

ERCS/erc-7715.md
```diff
@@ -0,0 +1,601 @@
...
+```tsx
+type GrantPermissionsResponse = {
+  grantedPermissions: Permission[];
+
+  expiry: number;
+
+  signerData: {
+    userOpBuilder?: `0x${string}`;
+    managerAddress?: `0x${string}`;
+  };
```
ilikesymmetry, line 177: nit: It feels inconsistent for 7710 manager to have `Address` suffix while the `userOpBuilder` does not nor the account `factory`. Thoughts on trimming the suffix and "address" is implied from type like the others?

```suggestion
  signerData: {
    userOpBuilder?: `0x${string}`;
    manager?: `0x${string}`;
  };
```

ERCS/erc-7715.md
```diff
@@ -0,0 +1,601 @@
...
+  factoryData?: `0x${string}`;
+
+  permissionsContext: string;
+};
+```
+
+- `grantedPermissions` is a list of permissions that were granted. The wallet MUST grant `required` permissions if the request was approved.
+- `expiry` is the expiry time that was granted. The wallet MAY grant a different expiry time than the one that the DApp requested.
+- `factory` and `factoryData` are as defined in [ERC-4337](./eip-4337.md). They are either both specified, or none. If the account has not yet been deployed, the wallet MUST return `factory` and `factoryData`, and the DApp MUST deploy the account by calling the `factory` contract with `factoryData` as the calldata.
+- `permissionsContext` is an opaque string that identifies the session. It MAY be the `context` as defined in [ERC-7679](./eip-7679.md) and [ERC-7710](./eip-7710.md). See “Rationale” for details.
```
ilikesymmetry, line 190: nit: because this is session-level, it's more intuitive to name this `sessionContext`. When I read `permissionsContext`, my first thought is this context is returned for each permission.

ilikesymmetry, line 190: Missing context on the `userOpBuilder` and `managerAddress` parameters:
```suggestion
- `grantedPermissions` is a list of permissions that were granted. The wallet MUST grant `required` permissions if the request was approved.
- `expiry` is the expiry time that was granted. The wallet MAY grant a different expiry time than the one that the DApp requested.
- `signerData.userOpBuilder` is as defined in [ERC-7679](./eip-7679.md).
- `signerData.managerAddress` is as defined in [ERC-7710](./eip-7710.md).
- `factory` and `factoryData` are as defined in [ERC-4337](./eip-4337.md). They are either both specified, or none. If the account has not yet been deployed, the wallet MUST return `factory` and `factoryData`, and the DApp MUST deploy the account by calling the `factory` contract with `factoryData` as the calldata.
- `permissionsContext` is an opaque string that identifies the session. It MAY be the `context` as defined in [ERC-7679](./eip-7679.md) and [ERC-7710](./eip-7710.md). See “Rationale” for details.
```

ERCS/erc-7715.md
```diff
@@ -0,0 +1,601 @@
...
+        {
+          type: 'gas-limit',
+          data: {
+            limit: '0x...',
+          },
+        },
+        {
+          type: 'call-limit',
+          data: {
+            count: 1,
```
ilikesymmetry, line 211: nit: again on preferring hex encoding all numbers for consistency with Ethereum json-rpc: https://ethereum.org/en/developers/docs/apis/json-rpc/#quantities-encoding

ERCS/erc-7715.md
pedrouid: it should be `delegationManager`

ERCS/erc-7715.md
pedrouid: or maybe we just call both `signerMeta` and `accountMeta`

pedrouid: `signerMeta` should also be optional for `wallet` signer type

ERCS/erc-7715.md
pedrouid: Agreed

ERCS/erc-7715.md
pedrouid: Just added `revokePermissions` to the spec

ERCS/erc-7715.md
```diff
@@ -0,0 +1,621 @@
...
+};
+
+// Alice sends the transaction by calling redeemDelegation on Bob's account
+const tx = await bob.sendTransaction({
+  to: delegationManager,
+  data: bob.interface.encodeFunctionData('redeemDelegation', [
+    permissionsContext,
+    action
+  ])
+});
```
filmakarov, line 575: Is this correct?
1. Why `bob.sendTransaction`? shouldn't it be `alice.sendTransaction`?
2. Why `bob` has interface of `redeemDelegation`? `bob` is Delegator, so it has `ERC7710` interface with `executeDelegatedAction` method. `redeemDelegation` lives in the `ERC7710Manager` interface implemented by `delegationManager`.  
[ERC-7710](https://eip.tools/eip/7710) for reference

ERCS/erc-7715.md
```diff
@@ -0,0 +1,621 @@
...
+    }
+  },
+
+  permissions: [
+    {
+      type: 'native-token-transfer',
+      policies: [
+        {
+          type: 'gas-limit',
+          data: {
```
GregTheGreek, line 127: just did a pretty big dive into this and can't really see concrete definitions with data in the context of policies. It also might make sense to try and define the permissions/policies deeper before getting into examples like these

ERCS/erc-7715.md
```diff
@@ -0,0 +1,621 @@
...
+        {
+          type: 'call-limit',
+          data: {
+            count: 1,
+          },
+        },
+      ],
+      required: true,
+      data: {
+        address: '0x...',
```
erensanlier, line 140: I think there shouldn't be `address` field in `native-token-transfer` permission.

SamWilsn (dismissed):

aminhdev:
.

ERCS/erc-7715.md
```diff
@@ -0,0 +1,551 @@
...
+    data: Record<string, any>;
+  };
+  permission: {
+    type: string; // enum defined by ERCs
+    data: Record<string, any>;
+  };
+  policies: {
+    type: string; // enum defined by ERCs
+    data: Record<string, any>;
+  }[];
```
moldy530, line 67: any reason not to be consistent with `permission` and `policies` either both being an `Array` or both being singletons? 

looking at some examples policies vs permissions below, it's also not super clear what the difference is between them. a `NativeTokenSpendPolicy` seems similar to an `Erc20TokenTransferPermission`. They both define limits on the spend of some funds.

moldy530: When I was first looking at this, I was thinking it would be used like:

```
{
  permission: { type: "owner" },
  policies: [], // no need to for policies because this is a root permission
}
```

or 

```
{
  permission: { type: "session-key", data: { expiry: fifteenMinutesInMs } },
  policies: [
    ... // spend limits, etc
  ]
}
```

ERCS/erc-7715.md
```diff
@@ -0,0 +1,551 @@
...
+
+### Sending transaction with a session
+
+#### ERC-7679 with `Key` type signer
+
+`wallet_grantPermissions` replies with `permissionsContext` and `userOpBuilder` address inside the `signerMeta` field. DApps can use that data with methods provided by [ERC-7679](./eip-7679.md) to build the [ERC-4337](./eip-4337.md) userOp.
+
+[ERC-7679](./eip-7679.md) UserOp Builder contract defines `bytes calldata context` parameter in all of its methods. It’s equivalent to the`permissionsContext` returned by the `wallet_grantPermissions` call.
+
+Example of formatting userOp signature using the [ERC-7679](./eip-7679.md) UserOp Builder
```
VGabriel45, line 400: Clicking ERC-7579 is giving 404.
```suggestion
Example of formatting userOp signature using the [ERC-7679](./erc-7679.md) UserOp Builder
```

ERCS/erc-7715.md
```diff
@@ -0,0 +1,551 @@
...
+    "policyTypes": "rate-limit,gas-limit"
+  }
+}
+```
+
+### Sending transaction with a session
+
+#### ERC-7679 with `Key` type signer
+
+`wallet_grantPermissions` replies with `permissionsContext` and `userOpBuilder` address inside the `signerMeta` field. DApps can use that data with methods provided by [ERC-7679](./eip-7679.md) to build the [ERC-4337](./eip-4337.md) userOp.
```
VGabriel45, line 396: Clicking ERC-4337 and ERC-7679 is giving 404.

```suggestion
`wallet_grantPermissions` replies with `permissionsContext` and `userOpBuilder` address inside the `signerMeta` field. DApps can use that data with methods provided by [ERC-7679](./erc-7679.md) to build the [ERC-4337](./erc-4337.md) userOp.
```

ERCS/erc-7715.md
```diff
@@ -0,0 +1,551 @@
...
+
+[ERC-7679](./eip-7679.md) UserOp Builder contract defines `bytes calldata context` parameter in all of its methods. It’s equivalent to the`permissionsContext` returned by the `wallet_grantPermissions` call.
```
VGabriel45, line 398: Clicking ERC-7579 is giving 404.
```suggestion
[ERC-7679](./erc-7679.md) UserOp Builder contract defines `bytes calldata context` parameter in all of its methods. It’s equivalent to the`permissionsContext` returned by the `wallet_grantPermissions` call.
```

ERCS/erc-7715.md
```diff
@@ -0,0 +1,551 @@
...
+
+With this ERC, DApps can request Wallets to grant permissions and execute transactions on the user's behalf, therefore circumventing the issues above.
+
+### `permissionsContext`
+
+Since this ERC only specifies the interaction between the wallet and the DApp but not how the wallet enforces permissions, we need a flexible way for the wallet to pass along information to the DApp so that it can construct transactions that imbue the permissions.
+
+The `permissionsContext` field is meant to be an opaque string that's maximally flexible and can encode arbitrary information for different permissions schemes. We specifically had three schemes in mind:
+
+- If a DApp leverages [ERC-7679](./eip-7679.md), it could use `permissionsContext` as the `context` parameter when interacting with the UserOp builder.
```
VGabriel45, line 525:
```suggestion
- If a DApp leverages [ERC-7679](./erc-7679.md), it could use `permissionsContext` as the `context` parameter when interacting with the UserOp builder.
```

ERCS/erc-7715.md
```diff
@@ -0,0 +1,551 @@
...
+};
+```
+
+First note that the response contains all of the parameters of the original request and it is not guaranteed that the values received are equivalent to those requested.
+
+`context` is a catch-all to identify a permission for revoking permissions or submitting userOps, and can contain non-identifying data as well. It MAY be the `context` as defined in [ERC-7679](./eip-7679.md) and [ERC-7710](./eip-7710.md). See “Rationale” for details.
+
+`accountMeta` is optional but when present then fields for `factory` and `factoryData` are required as defined in [ERC-4337](./eip-4337.md). They are either both specified, or none. If the account has not yet been deployed, the wallet MUST return `accountMeta`, and the DApp MUST deploy the account by calling the `factory` contract with `factoryData` as the calldata.
+
+`signerMeta` is dependent on the account type. If the signer type is `wallet` then it's not required. If the signer type is `key` or `keys` then `userOpBuilder` is required as defined in [ERC-7679](./eip-7679.md). If the signer type is `account` then `delegationManager` is required as defined in [ERC-7710](./eip-7710.md).
```
VGabriel45, line 141:
```suggestion
`signerMeta` is dependent on the account type. If the signer type is `wallet` then it's not required. If the signer type is `key` or `keys` then `userOpBuilder` is required as defined in [ERC-7679](./erc-7679.md). If the signer type is `account` then `delegationManager` is required as defined in [ERC-7710](./erc-7710.md).
```

VGabriel45 (changes requested):

ERCS/erc-7715.md
```diff
@@ -0,0 +1,621 @@
...
+We define a new JSON-RPC method `wallet_grantPermissions` for DApp to request a Wallet to grant permissions in order to execute transactions on the user’s behalf. This enables two use cases:
+
+- Executing transactions for users without a wallet connection.
+- Executing transactions for users with a wallet connection that is scoped with permissions.
+
+## Motivation
+
+Currently most DApps implement a flow similar to the following:
+
+```mermaid
```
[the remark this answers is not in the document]
SamWilsn: I checked, and no, this will not render.

SamWilsn:
Note to self: this is still blocked on https://github.com/ethereum/ERCs/pull/433.

---

I've locked this PR because too much of the discussion is technical in nature. Please discuss anything technical on the [discussion thread](https://ethereum-magicians.org/t/erc-7715-grant-permissions-from-wallets/20100).

---

Do not update the links to `erc-xxxx.md` as suggested above. That'll break our weird rendering system.

lukaisailovic (approved):

SamWilsn (approved):
This is reasonable for a draft. Do render your mermaid diagrams into SVGs before going into review.

octavio12345300 (commented):
list

Jojad9 (commented):
Approved to this change

PR #501, opened 2024-06-20
Update ERC-7578: Move to Review

V1d0r:
Move to Review


SamWilsn (commented):
Please only use backticks (`\``) for inline code. You're using them in your motivation section for definitions.

---

Those definitions in your motivation section, why are they there? Are they used later in the document, and need specific definitions? If so, please move them to the Specification section. If they're supposed to be motivating examples, maybe try changing the formatting to make that more clear?

---

Where is "set mint" (line 52) defined?

---

On line 54, try to describe the externally visible behaviour of the contract. Like instead of "if the token burns, the properties MUST be deleted", how about "if the token is burned, `getProperties` MUST revert"

---

On line 127, you aren't really stating a requirement on implementers. You should describe the behaviour of the contract _if `setProperties` is not called_, instead of saying how the contract is supposed to be used.

---

On line 139, you mention an `_update` function that must be overridden. `_update` doesn't exist in the ERC-721 standard. I believe it's part of OpenZeppelin's implementation. Instead of referencing a specific implementation here (at the risk of sounding like a broken record) you should describe the externally visible behaviour of the contract in terms of the functions in ERC-721.

gabrielstoica:
Hey @SamWilsn! 

Thanks for your feedback - it was very valuable and we've tried to cover all points as follows:

- Moved all definitions to the specification section;
- Updated the rationale to be less restrictive and define the externally visible behaviour;
- Removed the `setProperties` definition to improve the standard's interoperability;

Let me know what you think!

SamWilsn (approved):

PR #502, opened 2024-06-20
Update ERC-7656: Update naming for clarity

sullof:
After conversations with members of the community, it was clear that the name `IERC7656Linked` was better than `IERC7656Contract` but still not ideal. Then, the decision to change it to `IERC7656Service`, which better aligns with the goal of the ERC, i.e., allowing NFT to own services that extend the capability of the single NFT.


sullof:
@abcoathup I am having bot errors that are unrelated to the PR's proposed changes. 
Do you know which is the best practice in these cases?

abcoathup:
> @abcoathup I am having bot errors that are unrelated to the PR's proposed changes. Do you know which is the best practice in these cases?

@sullof I assume you resolved them.  
For unrelated review-bot issues I suggest reporting in the EIP-editors channel of EthCatHerders Discord.

sullof:
@abcoathup I made another push but obtained the same error. Then, at some moment, the issue got fixed by itself. I have no idea why. Anyway, thanks, I will ask in the Discord server

PR #513, opened 2024-06-30
Add ERC: Vulnerability and Exposure Identifier

dimitrysuen:
To address the absence of a standardized reference method for vulnerabilities and to reduce communication costs between protocol developers and ethical hackers, this ERC proposes a rule for creating unique identifiers for vulnerabilities. The aim is to establish a common language in these scenarios.


ERCS/erc-tbd.md
```diff
@@ -0,0 +1,182 @@
+---
+number: TBD
```
abcoathup, line 2:
```suggestion
eip: 7731
```

ERCS/erc-tbd.md
```diff
@@ -0,0 +1,182 @@
...
+title: Vulnerability and Exposure Identifier Specification and Indexing
+author: john (@0xJohnZW), dimitry (@dimitrysuen), myiwen (@myiwen), msfew (@fewwwww) <msfew@ora.io>
+discussions-to: TBD
```
abcoathup, line 5: Please create a discussions topic in Eth Magicians with a link to this PR.

ERCS/erc-tbd.md
```diff
@@ -0,0 +1,182 @@
+---
+number: TBD
+title: Vulnerability and Exposure Identifier Specification and Indexing
```
abcoathup, line 3:
```suggestion
title: Vulnerability and Exposure Identifier
```
Specification seems superfluous in an ERC title.

ERCS/erc-tbd.md
```diff
@@ -0,0 +1,182 @@
...
+author: john (@0xJohnZW), dimitry (@dimitrysuen), myiwen (@myiwen), msfew (@fewwwww) <msfew@ora.io>
+discussions-to: TBD
+status: Draft
+type: Standards Track
+category: ERC
+created: 2024-05-18
```
abcoathup, line 9:
```suggestion
created: 2024-06-30
```
PR was created June 30 (or 29 depending on timezone).

ERCS/erc-7731.md
```diff
@@ -0,0 +1,182 @@
...
+## Motivation
+
+When discussing security incidents, they are usually referred to as "Protocol Name + Exploit". Also, there are no unique identifiers for ethical hackers to reference the vulnerabilities, leading to high communication costs between the protocol developers and the ethical hackers. We can unify the terminology across various contexts by providing each vulnerability a unique identifier, as a parallel system to the existing ones such as CVE(Common Vulnerability and Exposure).
+
+This EIP can be useful in the following use cases:
+
+* Both audit and bug reports can reference a specific vulnerability with a unique identifier.
+* Security researchers can subscribe to on-chain events to track the latest vulnerabilities easily.
+
+This EIP can benefit the following parties:
```
SamWilsn, line 26:
```suggestion
This proposal can benefit the following parties:
```

ERCS/erc-7731.md
```diff
@@ -0,0 +1,182 @@
...
+* Protocols' security can be enhanced since this EIP encourages open vulnerability reporting.
+* Security researchers can gain more insights on vulnerabilities in more efficient way.
+
+## Specification
+
+The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC-2119.
+
+### Identifier of Vulnerability and Exposure
+
+The identifier of vulnerability and exposure MUST conform to the following format: `Database Prefix + Year + Arbitrary Digits`.
```
SamWilsn, line 38: This is a somewhat ambiguous way of specifying the parts. Are the pluses literal or are they a concatenation operator?

The way you write it later is much better, so how about something like this:

```suggestion
The identifier of vulnerability and exposure SHALL contain three components separated by hyphens (`‐`) in the following order: database prefix, year, and arbitrary digits. An example of this format is `CVE-2024-0097`, where `CVE` is the database prefix, `2024` is the year, and `0097` are the arbitrary digits.
```

ERCS/erc-7731.md
```diff
@@ -0,0 +1,182 @@
...
+
+`Database Prefix` SHOULD be uniquely named by different assigners.
```
SamWilsn, line 40: This is not really an actionable requirement. Does this mean each assigner should choose exactly one unique prefix? What happens if they don't?

ERCS/erc-7731.md
```diff
@@ -0,0 +1,182 @@
...
+
+`Year` MUST be in `YYYY` format.
+
+`Arbitrary Digits` is RECOMMENDED to start from `0001` and increase incrementally.
+
+These three parts MUST be connected by `-`, for example, `CVE-2024-0097`. The length of the identifier MUST NOT exceed 32 bytes.
```
SamWilsn, line 46: > exceed 32 bytes

This leaves the encoding of the identifier unspecified. Is it UTF-8? ASCII?

ERCS/erc-7731.md
```diff
@@ -0,0 +1,182 @@
...
+
+### Identifier of Vulnerability and Exposure
+
+The identifier of vulnerability and exposure MUST conform to the following format: `Database Prefix + Year + Arbitrary Digits`.
+
+`Database Prefix` SHOULD be uniquely named by different assigners.
+
+`Year` MUST be in `YYYY` format.
+
+`Arbitrary Digits` is RECOMMENDED to start from `0001` and increase incrementally.
```
SamWilsn, line 44: What is permitted as a digit? Is Ⅷ (U+2167) a valid digit?

ERCS/erc-7731.md
```diff
@@ -0,0 +1,182 @@
...
+)
+```
+
+`veId` is the identifier of the vulnerability and exposure assigned by `assignerName`. It MUST comply with the **Identifier of Vulnerability and Exposure** format mentioned above.
+
+`assignerName` is the name of the `veId` assigner, and the length of `assignerName` MUST NOT exceed 32 bytes. Smart contracts complying with this EIP SHOULD belong to only one assigner entity. The indexer SHOULD identify the assigner based on the contract address.
+
+`state` is the acceptance status of this `veId` set by the assigner. In this EIP, only two final state values are defined: `Published` and `Rejected`. The assigners can define their own initial or intermediate states based on their business requirements. The `status` value MUST be either `Published` or `Rejected` when the `veId` is in the final state.
+
+`reportCid` MUST be the vulnerability report's IPFS CID(Content Identifier).
```
SamWilsn, line 71:
```suggestion
`reportCid` MUST be the vulnerability report's IPFS Content Identifier (CID).
```

assets/erc-7731/veId-and-refVeId.png
SamWilsn, line 1: We prefer SVG images where possible. If you can't, this PNG is fine.

Mikhai56 (changes requested):
ERCS/erc-7731.md

Mikhai56 (changes requested):
assets/erc-7731/veId-and-refVeId.png

ERCS/erc-7731.md
chaals: Suggest `[0-9]*`

chaals: The veid scheme seems heavily optimised for people who produce atomic notices of vulnerabilities, but I'm looking to provide identifiers for items in a specification that is revised and published as a new version from time to time (so far we have worked on a "12-18 months" revision cycle, but it fundamentally depends on how fast things change in the ecosystem).

How do we increase incrementally? We have a set of vulnerabilities that are taken from something that should be used directly as its own reference that already uses the relevant format (Solidity compiler bugs), plus about twice as many other vulnerabilities that are divided across a couple of different dimensions.

Thinking about how to identify these helpfully, it seems like a bad idea to start from 0001 and increment by 1 for each subsequent item in document order. It would be helpful if ethtrust-2023-0021 refers to the same thing as ethtrust-2025-0021 if they are identical in the two specifications. 

One possible approach would be to add identifiers to each requirement, and then maintain the same identifier unless the requirement changes. But if we did this, I would suggest incrementing with every new requirement, so that if we make a smaller change to the requirement that was say ethtrust-2025-0026, we know that we can identify the updated requirement as ethtrust-2028-0026.

PR #514, opened 2024-07-01
Update ERC-6123: Move to Review

cfries:
As we are actively using this ERC for a while now, we would like to lift it from draft to review.  The only minor change that we that we added recently is the possibility to pass the termination terms that the two parties agree on.

Thanks for reviewing.


cfries:
will close this one and open a new one with an additional function

PR #528, opened 2024-07-09
Update ERC-3770: Move to Draft

ligi:
we should revive this one (especially as it is used and currently pushed by @vbuterin here: https://x.com/VitalikButerin/status/1810633473750683974)

So moving it from stagnant to draft.

Also removing the responsibility on Ethereum-Lists curators (currently only me) to care for similar looking names. This is just not possible. It is just a drawback of this EIP in contrast of CAIP-10.


pedrouid (approved):
👍

usame-algan (approved):

PR #533, opened 2024-07-10
Update ERC-5630: Move to Draft

jdbertron:
The EIP was moved to Stagnant state automatically, but there are many applications waiting for a replacement to the now deprecated support for eth_getEncryptionPublicKey   .  

Authors: J.D. Bertron <jdbertron>


SamWilsn (approved):
Please open a separate PR to move into review.

PR #534, opened 2024-07-10
Update ERC-6123: EIP-6123, new getter function for the mutual termination bool.

Julius278:
Just added a new getter function in the abstract SDC contract.
This could be helpful in case the Token not only needs to know the TradeState but also if the trade was mutually terminated or by an error in the settlement process.


cfries:
The member field  `mutuallyTerminated` was removed by @pekola - not sure if this getter is still needed. Need to discuss with Peter.

SamWilsn:
This pull request has conflicts that must be resolved.

PR #537, opened 2024-07-13
Update ERC-7627: Move to Review

chenly:

 - The build passes.


SamWilsn (commented):
Your Rationale section is answering the question "why do we need this standard?". That kind of content belongs in the Motivation section, while the Rationale section should explain choices made within the proposal. My favourite analogy is:

> Rationale: We chose to paint the shed red because...

---

Everything else looks decent though!

chenly:
As you mentioned, having a wallet store only one public key is not a good design. I modified the implementation to allow users to set multiple public keys and specify expiration times.

PR #538, opened 2024-07-13
Update erc-3770.md

Zartaj0:
The links were not updated after the separation of ERCs and EIPs. Updated in this PR.


SamWilsn:
Links should still use `eip-` because of a quirk of our rendering system.

PR #547, opened 2024-07-20
Update ERC-6860: Add _blockNumber parameter to Web3 URL specification

nishuzumi:
This PR extends the Web3 URL specification to include support for the _blockNumber parameter, allowing users to specify a particular block number when making queries and enabling access to historical blockchain data.

## Summary of changes:

- Modified `pathQuery` to optionally include `evmQuery`
- Added new `evmQuery` rule to handle EVM-specific parameters
- Introduced `blockNumberQuery` rule for _blockNumber parameter

### DIFF
--- a/ERCS/erc-6860.md
+++ b/ERCS/erc-6860.md
@@ -307,8 +307,15 @@ chainid         = %x31-39 *DIGIT
 pathQuery       = mPathQuery [ evmQuery ] ; path+query for manual mode
                 / aPathQuery [ evmQuery ] ; path+query for auto mode
 
-evmQuery        = *( "&" blockNumberQuery )
-blockNumberQuery = "_blockNumber=" 1*DIGIT
+evmQuery        = *[(evmPath evmParams)]
+evmPath         =  "?" ; begins with "/" but not "//" 
+                / "&" 
+evmParams       = blockNumberQuery
+                / gasLimitQuery
+                / fromQuery
+blockNumberQuery= "_blockNumber=" uintArg
+gasLimitQuery   = "_gasLimit=" uintArg
+fromQuery       = "_from=" addressArg
 
 web3UrlRef      = web3URL 
                 / relativeWeb3URL
--- a/ERCS/erc-6860.md
+++ b/ERCS/erc-6860.md
@@ -2,7 +2,7 @@
 eip: 6860
 title: Web3 URL to EVM Call Message Translation
 description: A translation of an HTTP-style Web3 URL to an EVM call message
-author: Qi Zhou (@qizhou), Chao Pi (@pichaoqkc), Sam Wilson (@SamWilsn), Nicolas Deschildre (@nand2) Box (@nishuzumi)
+author: Qi Zhou (@qizhou), Chao Pi (@pichaoqkc), Sam Wilson (@SamWilsn), Nicolas Deschildre (@nand2), Box (@nishuzumi)
 discussions-to: https://ethereum-magicians.org/t/eip-4804-web3-url-to-evm-call-message-translation/8300
 status: Draft
 type: Standards Track
--- a/ERCS/erc-6860.md
+++ b/ERCS/erc-6860.md
@@ -2,7 +2,7 @@
 eip: 6860
 title: Web3 URL to EVM Call Message Translation
 description: A translation of an HTTP-style Web3 URL to an EVM call message
-author: Qi Zhou (@qizhou), Chao Pi (@pichaoqkc), Sam Wilson (@SamWilsn), Nicolas Deschildre (@nand2)
+author: Qi Zhou (@qizhou), Chao Pi (@pichaoqkc), Sam Wilson (@SamWilsn), Nicolas Deschildre (@nand2) Box (@nishuzumi)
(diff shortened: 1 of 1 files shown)


ERCS/erc-6860.md
```diff
@@ -293,8 +304,11 @@ contractName    = address
                 / domainName
 chainid         = %x31-39 *DIGIT
 
-pathQuery       = mPathQuery ; path+query for manual mode
-                / aPathQuery ; path+query for auto mode
+pathQuery       = mPathQuery [ evmQuery ] ; path+query for manual mode
+                / aPathQuery [ evmQuery ] ; path+query for auto mode
+
+evmQuery        = *( "&" blockNumberQuery )
```
qizhou, line 310: I guess this format cannot cover the case of no query in `mPathQuery` and `aPathQuery`.  E.g., 

`web3://w3url.eth/?_blockNumber=123

ERCS/erc-6860.md
```diff
@@ -293,8 +304,11 @@ contractName    = address
...
+blockNumberQuery = "_blockNumber=" 1*DIGIT
```
qizhou, line 311: It will be great to add other `eth_call` parameters such as `gasLimit`.

ERCS/erc-6860.md
nishuzumi: I tried to fix it, does it meet the standard now?

ERCS/erc-6860.md
nishuzumi: I added gasLimit and from, maybe more can be added

nishuzumi:
@qizhou Add some fix code

ERCS/erc-6860.md
```diff
@@ -293,8 +304,18 @@ contractName    = address
                 / domainName
 chainid         = %x31-39 *DIGIT
 
-pathQuery       = mPathQuery ; path+query for manual mode
-                / aPathQuery ; path+query for auto mode
+pathQuery       = mPathQuery [ evmQuery ] ; path+query for manual mode
+                / aPathQuery [ evmQuery ] ; path+query for auto mode
+
+evmQuery        = *[(evmPath evmParams)]
+evmPath         =  "?" ; begins with "/" but not "//"
```
qizhou, line 311: This seems to allow a wired format like

`web3://usdc.eth/balanceOf/vitalik.eth?returns=(uint256)?_blockNumber=1234`?

I think the problem here is that these parameters `_xxxxx=xxx` should be part of `mPathQuery` and `aPathQuery` while appending `evmQuery` may introduce unexpected formats.

nishuzumi:
The rfc format problem has been fixed after the last modification, please kindly review it again
@nand2, @pichaoqkc, @qizhou, @SamWilsn

nand2:
Hi!

Thanks for the PR. Here is my feedback :
- The `_from` attribute is not necessary, it is already supported via the `userinfo` field in `schema "://" [ userinfo "@" ] contractName [ ":" chainid ] pathQuery`.
- In the Auto mode ABNF :
  - Please keep the ordering of lines from so that each new variable in a line is defined in a line below it.
  - The ERC contains both the complete ABNF notation in the appendix, and excerpts within the specification, please do the modifications at both places.
  - Instead of mixing `attribute` and `evmParams` in `aQuery`, it would be better to rework the `attribute` field, and start from there. Rename `attrName` and `attrValue` to something more precise, as the `returns`/`returnTypes` are no longer the only attributes. For example : `attribute = retAttrName "=" retAttrValue / evmAttrName "=" evmAttrValue`
  - For `blockNumber` : I guess we should add the 5 authorized string values ("latest", "earliest", "pending", "safe" or "finalized").
- In the auto mode section in the specification: Please describe briefly what these options do.
- The specification in ABNF notation indicate the structure of a valid `web3://` URL. So, in manual mode, every query (`mQuery = *( pchar / "/" / "?" )`) is already allowed, so there is no need to alter it. On the other hand, the special `_blockNumber` and `_gasLimit` parameters should be described in the manual mode section in the specification (behavior, syntax of value (can reference the syntax definition of auto mode), whether they are forwarded to the smartcontract, etc)

I would also argue that, while prefixing these attributes with `_` makes sense for the manual mode, to prevent conflicts with actual attributes used by the webpage, they are not necessary for the auto mode since there cannot be any conflict.
One could argue that it may be confusing that sometimes it is `_blockNumber` and sometimes it is `blockNumber`, but on the other hand when we craft a web3:// URL, we already know for which mode we are crafting the URL. So:

- auto mode : `web3://usdc.eth/balanceOf/vitalik.eth?returns=(uint)&blockNumber=12000000`
- manual mode : `web3://vitalikblog.eth/general/2024/07/23/circlestarks.html?_blockNumber=12000000`

What do you think?

Other subject : Outside the scope of this ERC, but linked : for resource request mode (erc-6944), I think we would support this there as well, and it would make sense to me that we just do not forward the new attributes `_blockNumber` and `_gasLimit` to the smartcontract.
So for the manual mode : should we also extract these from the calldata sent to the smartcontract? It would make sense to me.

Thanks!

SamWilsn:
I'm not super involved with this standard, so I'll leave this to the other authors.

PR #548, opened 2024-07-20
Update ERC-4337: the -> they

samlaf:

 - The build passes.


Skandiakiwi13:
I understand how to return a Fibonacci sequence using the iterative approach, as well as the dynamic programming approach.

With recursion, I understand the recursion tree for Fibonacci. and how to return Nth Fibonacci number

def fib(num): if num <= 2: return num return fib(num -1) + fib(num-2) print(fib(5)) 

The above code returns 8 which is great, but I want to understand how could I return a list of the sequence by updating the above recursive code.

eg: [1,1,2,3,5,8]

PR #568, opened 2024-08-01
Update ERC-7208: Move to Review

galimba:
We request to push ERC-7208 to Review and Last Call soon, after validating with the community actively using this architecture.

Additionally, we provide an educational implementation that acts as a reference to the interoperability capabilities of this approach.


ERCS/erc-7208.md
```diff
@@ -271,6 +271,8 @@ We present an **educational example** implementation showcasing two types of tok
 **This example has not been audited and should not be used in production environments.**
 
 
+See [Minimalistic Data Index](https://github.com/NexeraProtocol/Minimalistic-Data-Index) implementation.
```
SamWilsn, line 274: The linter is correct here. For the reasons enumerated [here](https://ethereum.github.io/eipw/markdown-rel-links/), please remove this external link.

pash7ka: Hi Sam,
I've removed the external link replacing it with the link to README.md, as you've suggested in the comment above.
But now the HTML prufer fails with
```
For the Links > Internal check, the following failures were found:
* At ./_site/EIPS/eip-7208.html:526:
  internally linking to ../assets/erc-7208/contracts/README.md, which does not exist
HTML-Proofer found 1 failure!
```
which was the reason we've tried with external repo.
Do you have any ideas how to fix it?

ERCS/erc-7208.md
```diff
@@ -271,6 +271,8 @@ We present an **educational example** implementation showcasing two types of tok
 **This example has not been audited and should not be used in production environments.**
 
 
+See [contracts](../assets/erc-7208/contracts/README.md)
```
SamWilsn, line 274: Because of a quirk of our rendering system, this needs to be:

```suggestion
See [contracts](../assets/eip-7208/contracts/README.md)
```

pash7ka: Done! Thank you!

PR #570, opened 2024-08-02
Reverting change tradeId to tradeData for termination.

cfries:
Reverting a previous rename. Just a rename of the arguments to be consistent with documentation and implementation.


cfries:
@SamWilsn  Sorry to bother, but I wonder - why are some PRs merged automatically but this one is not? If I can improve the process to reduce your work, I would be happy to try.

PR #630, opened 2024-09-10
Update ERC-7572: Remove Schema assumptions, Add updated URI information to the event

peersky:
I think this standard proposal is over opinionated yet very actual and needed. 

industry does need standard way of interfacing and defining URIs, yet these do not need to have any schemas pre-defined, that could be encapsulated in separate, follow-up ERC. 

I suggest removing schemas from specification
These are not needed, for example, contract name in EC712 is already defined on blockchain level, hence schema defining `name` field is opinionated and can be implemented as separate standard. 

Also, an emitted event is missing URI information that would be useful for indexers.


peersky:
@SamWilsn @g11tech 

How can I claim championship over this ERC? 

I requested reasonable proposal that allowed to move this ERC quickly forward but could not get it merged due to author inactivity both in GH and in Magicians forum board.

SamWilsn:
@peersky sorry for the, uh, half a year delay in replying. Copy the text, add yourself as an author, and submit as a new draft.

peersky:
> @peersky sorry for the, uh, half a year delay in replying. Copy the text, add yourself as an author, and submit as a new draft.

@SamWilsn you got it back ;) 

You mean open new ERC?

PR #637, opened 2024-09-13
Update ERC-1078: Typo ERC-1078

DoTheBestToGetTheBest:
Made a few small adjustments.

### DIFF
--- a/ERCS/erc-1078.md
+++ b/ERCS/erc-1078.md
@@ -33,7 +33,7 @@ The first step of the process is to request from the user the ENS name that poin
 
 #### 2.a) Create a new identity
 
-If the user doesn’t have an identity, the app should provide the option to create one for them. Each app must have one or more domains they control which they can create immediately subdomains on demand. The app therefore will make these actions on the background:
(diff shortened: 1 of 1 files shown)


ERCS/erc-1078.md
```diff
@@ -33,11 +33,11 @@ The first step of the process is to request from the user the ENS name that poin
 
 #### 2.a) Create a new identity
 
-If the user doesn’t have an identity, the app should provide the option to create one for them. Each app must have one or more domains they control which they can create immediate subdomains on demand. The app therefore will make these actions on the background:
+If the user doesn’t have an identity, the app should provide the option to create one for them. Each app must have one or more domains they control which they can create immediately subdomains on demand. The app therefore will make these actions on the background:
```
SamWilsn, line 36: The adjective form is correct here:

```suggestion
If the user doesn’t have an identity, the app should provide the option to create one for them. Each app must have one or more domains they control which they can create immediate subdomains on demand. The app therefore will make these actions on the background:
```

SamWilsn (approved):

PR #669, opened 2024-10-09
Add ERC: Onchain registration of chain identifiers

yuliyaalexiev:
This ERC proposes deriving chain identifiers by hashing their chain name (along with other relevant information) and using ENS to map chain names to these identifiers, replacing the centralized list currently maintained on GitHub.


ERCS/erc-chain-identifiers.md
```diff
@@ -0,0 +1,105 @@
+---
```
abcoathup, line 1:
```suggestion
---
eip: 7785
```

ERCS/erc-chain-identifiers.md
```diff
@@ -0,0 +1,105 @@
...
+title: On-chain registration of chain identifiers
```
abcoathup, line 3:
```suggestion
title: Onchain registration of chain identifiers
```
Personal preference (I am not an ERC editor)

ERCS/erc-chain-identifiers.md
```diff
@@ -0,0 +1,106 @@
+---
+eip: 7785
+title: On-chain registration of chain identifiers
+description: Derive chain identifiers by hashing their chain name and use ENS to map chain names to these identifiers
+author: Marco Stronati (@paracetamolo), Jeff Lau <jeff-lau@live.com>
+discussions-to: https://ethereum-magicians.org/t/on-chain-registration-of-chain-identifiers/21299
```
abcoathup, line 6: @SamWilsn created a discussions topic for the EIP.  You can either get him to point to this ERC or create a new one.

ERCS/erc-chain-identifiers.md
```diff
@@ -0,0 +1,106 @@
...
+status: Draft
+type: Meta
```
abcoathup, line 8:
```suggestion
type: ERC
```

ERCS/erc-chain-identifiers.md
yuliyaalexiev: I will ask him to update the discussion to the new PR. Thank you.

ERCS/erc-chain-identifiers.md
yuliyaalexiev: I updated this and pushed a new commit - thanks

yuliyaalexiev: @abcoathup - it lookd like type: ERC is not a valid entry - ok to go back to Meta? 
image

abcoathup: Use Standards Track

ERCS/erc-7785.md
```diff
@@ -0,0 +1,107 @@
...
+### Extending chain identifiers
+
+Current chain identifiers are usually chosen arbitrarily to be short. While these identifiers are convenient on a small scale, as their number increases it is more desirable to draw them from a larger space.
+
+We propose to extend the size of identifiers to 32 bytes and to derive them using a cryptographic hash function.
+The input to the function MUST contain the chain name and MAY contain additional information.
+
+An example for a L2:
+```
+chain_id = Keccak-256(CHAIN_NAME, L1_CHAIN_ID, VERSION, BRIDGE)
```
kelemeno, line 49: Some small comments: 
- L1_CHAIN_ID should be settlement chain id to allow e.g. L3s. 
- Bridge should not be the chain address, but the address of the chain registration contracts.

kelemeno: Also we should also probably add a salt value.

ERCS/erc-7785.md
```diff
@@ -0,0 +1,107 @@
+---
+eip: 7785
+title: Onchain registration of chain identifiers
+description: Derive chain identifiers by hashing their chain name and use ENS to map chain names to these identifiers
+author: Marco Stronati (@paracetamolo), Jeff Lau <jeff-lau@live.com>
```
jefflau, line 5:
```suggestion
author: Marco Stronati (@paracetamolo), Jeff Lau <@jefflau>
```

jefflau (changes requested):

ERCS/erc-7785.md
jefflau: Switching to my github username

ERCS/erc-7785.md
Arachnid, line 49: Suggest replacing `,` with `||` to clarify that this is a concatenation operation. This also needs to specify the encoding of `version` in particular.

Arachnid: Also, isn't the bridge address and L1 chain ID alone enough to uniquely identify a chain?

ERCS/erc-7785.md
SamWilsn: That'll be:

```suggestion
author: Marco Stronati (@paracetamolo), Jeff Lau (@jefflau)
```

ERCS/erc-7785.md
sampkaML: > Also, isn't the bridge address and L1 chain ID alone enough to uniquely identify a chain?

This wouldn't necessarily allow say a L3 to be identified.

Arachnid: How about bridge address and parent chain ID, then?

sampkaML: I dont see any reason to reduce the data being sent when all the listed seems to be to be useful information to be able to have

sampkaML: i agree with both of these.

ERCS/erc-7785.md
```diff
@@ -0,0 +1,107 @@
...
+### Extending chain identifiers
+
+Current chain identifiers are usually chosen arbitrarily to be short. While these identifiers are convenient on a small scale, as their number increases it is more desirable to draw them from a larger space.
+
+We propose to extend the size of identifiers to 32 bytes and to derive them using a cryptographic hash function.
+The input to the function MUST contain the chain name and MAY contain additional information.
+
+An example for a L2:
+```
+chain_id = Keccak-256(CHAIN_NAME, L1_CHAIN_ID, VERSION, BRIDGE)
```
sampkaML, line 49: `chain_id = Keccak-256(CHAIN_NAME, SETTLEMENT_CHAIN_ID, VERSION, DEPLOYER_CONTRACT_ADDRESS, SALT)
`

ERCS/erc-7785.md
yuliyaalexiev: PR updated

ERCS/erc-7785.md
```diff
@@ -0,0 +1,107 @@
...
+
+We propose to extend the size of identifiers to 32 bytes and to derive them using a cryptographic hash function.
+The input to the function MUST contain the chain name and MAY contain additional information.
+
+An example for a L2:
+```
+chain_id = Keccak-256(CHAIN_NAME, SETTLEMENT_CHAIN_ID, VERSION, DEPLOYER_CONTRACT_ADDRESS, SALT)
+```
+where:
+- `L1_CHAIN_ID` is the id of the L1 where the L2 settles, it could be Mainnet or a testnet.
```
bumblefudge, line 52:
```suggestion
- `SETTLEMENT_CHAIN_ID` is the id of the L1 where the L2 settles, it could be Mainnet or a testnet.
```
I think ANCHOR_CHAIN_ID is more accurate, because "settlement" could be happening on _either_ technically.

bumblefudge: also, can a given L2 have multiple ANCHOR_CHAIN_IDs and thus multiple L2 `chain_id`s? Might be good to explain how that's prevented if it is, or how it's not a security or UX risk not to prevent it if it isn't..

ERCS/erc-7785.md
```diff
@@ -0,0 +1,107 @@
...
+
+TBD
+
+## Backwards Compatibility
+
+Existing identifiers, that were not derived using the scheme above, can be supported using a reverse mapping from chain identifiers to chain names, so that one can check for uniqueness.
+
+For example the chain name `legacy-rollup.eth` can be resolved to the chain identifier `0x123`.
+Then `0x123` can be resolved in the `chainid.reverse` domain to a `chain_name`.
+If `chain_name == legacy-rollup` then the mapping is valid.
```
bumblefudge, line 87: what if that name is already registered, tho? or what if someone registers `legacy-rollup.eth` to attack/break this because they don't like, or compete with, `legacy-rollup`?

if this goes into action, i'd recommend ENS consider blocklisting/squatting/otherwise-protecting all those chain_names from future registration to prevent abuses and hijinx. plenty of people would love to block/break this fallback for major chains that DON'T move quickly and register L2 names.

ERCS/erc-7785.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Specification
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
+
+### Extending chain identifiers
+
+Current chain identifiers are usually chosen arbitrarily to be short. While these identifiers are convenient on a small scale, as their number increases it is more desirable to draw them from a larger space.
+
+We propose to extend the size of identifiers to 32 bytes and to derive them using a cryptographic hash function.
```
bumblefudge, line 44: Are there any chain_ids registered in eth-lists/chains of this size already? if all keccak chain_id's generated this way have the same length, should eth-lists/chains STOP allowing registrations of that length to prevent [malicious] collisions?

bumblefudge: i like the idea of L2s and L3s using this mechanism and reducing traffic/burden on eth-lists/chains, but eth-lists/chains will still be needed for lots of non-L2 use-cases, so I think the smooth continued operation of eth-lists/chains should be in-scope for this ERC!

ERCS/erc-7785.md
yuliyaalexiev: @bumblefudge, this is addressed in the Bootstrapping and handover section (line 89). All existing rollups will be registered to avoid this scenatio

ERCS/erc-7785.md
yuliyaalexiev: @sampkaML - what are your thoughts on this?

sampkaML: I agree with preventing eth-lists/chains from allowing registrations of this length due to your stated reason of avoiding [malicious] collisions.

However, regarding the continued use of existing eth-lists/chains and their inclusion within the scope of this ERC, I'm not sure I agree. What would be the downside of allowing the current ones to continue but stopping new registrations unless they follow this new format?

bumblefudge: Simple: not everything is an L2 anchored in mainnet (or any other EVM chain). Private networks, permissioned networks, EVM-mode L2s of other L1s, all need chainIds (and can't be accessed via a contract that they anchor to on mainnet)

sampkaML: While I understand the concerns about private networks and other unique use cases, these issues aren't addressed by the current ENS implementation either. Our goal with this ERC isn't to solve every possible problem but to apply the 80/20 rule—improving the system for the vast majority of users. By focusing on the common scenarios where this solution will have the most impact, we can make meaningful progress now and address the remaining specialized cases with tailored solutions in the future.

yuliyaalexiev: @bumblefudge, which chains specifically will be adversely impacted by this change now?

ERCS/erc-7785.md
```diff
@@ -0,0 +1,107 @@
...
+The input to the function MUST contain the chain name and MAY contain additional information.
+
+An example for a L2:
+```
+chain_id = Keccak-256(CHAIN_NAME, SETTLEMENT_CHAIN_ID, VERSION, DEPLOYER_CONTRACT_ADDRESS, SALT)
```
iJaack, line 49: I would suggest allowing `null` or `0` input for the `SETTLEMENT_CHAIN_ID` and `DEPLOYER_CONTRACT_ADDRESS`, so as to allow non-L2 use cases as well.

Also, I believe another input should be the chainID of the DA its posting data to, since it’s becoming very frequent to see chains not posting data to blobs on L1 but to DAs on other EVMs or even on other VMs (see NEAR or Avail).

This new onchain registry could help a lot block explorer providers like us because it tells us immediately how to index the chain, and if there’s some extra indexing to add to scan the chain properly.

ERCS/erc-7785.md
iJaack: Hi, I would suggest to think about this EIP as a solution for any EVM, even outside the L1+L2s ecosystem. Some reasons:
- chain teams already submit their info (name, chainID, etc.) to ethereum-lists/chains because that’s the ultimate source of truth for the list of all EVMs, so asking them to submit a transaction in order to do the same thing they would do already seems like a reasonable compromise
- From our research (I work at routescan.io) there are currently roughly 1200 chains live at the moment, of which roughly 85% is EVM, and even though the pace of growth has increased, we don’t forecast millions of chains within the foreseeable future, so increasing the size  of the identifiers to 32 bytes may not be needed.

ERCS/erc-7785.md
sampkaML: Thank you for your insightful suggestions. I agree that allowing null or zero inputs for SETTLEMENT_CHAIN_ID and DEPLOYER_CONTRACT_ADDRESS would accommodate non-L2 use cases and enhance the versatility of the registry. 

However, my main concern is that I prefer not to dictate which inputs are required or optional within this ERC itself. I believe our primary goal should be to offer a flexible framework that includes all these options, allowing future ERCs—such as those focusing on interoperability—to specify the requirements for these variables, including which can be null or zero. This approach ensures that the ERC remains adaptable to a variety of use cases and can evolve with the ecosystem's needs.

iJaack: I understand, and I believe this needs to become an ERC as quick as possible so that the conversation with the wide community can start. Let's make it as flexible as possible so that future work can address extensions for interoperability.

@sampkaML So you think other info, like DA layer, tech stack and others could be part of a new attestation-based spec / ERC?

mds1, line 49: What is `DEPLOYER_CONTRACT_ADDRESS`? It's not defined below

mds1, line 49: Is `VERSION` arbitrarily specified by the rollup, i.e. it refers to the rollup's version, or does it refer to the version of this ERC to support updates to the scheme defined here?

ERCS/erc-7785.md
```diff
@@ -0,0 +1,107 @@
...
+```
+where:
+- `SETTLEMENT_CHAIN_ID` is the id of the L1 where the L2 settles, it could be Mainnet or a testnet.
+- `VERSION` is to separate the domain of the hash function with an arbitrary string
+- `BRIDGE` is the address of the L2 on the L1
```
mds1, line 54: A few questions here:

1. What does "settle" mean? Is this where calldata / state diffs are posted, or where withdrawal proofs are, etc? 
2. Is this ERC intended to cover rollups only (seems like yes based on `SETTLEMENT_CHAIN_ID`) or any chain (implied by the abstract and motivation)?
3. What if a chain only posts calldata to an alt-DA layer and has no settlement chain ID? This DA layer could be a centralized server that has no chain ID
4. What if a chain has two bridges it accepts deposits from? Or no bridges? See https://youtu.be/Aey1_acXL0k?si=cNdwikh15P6EFNIi

To sidestep these questions, perhaps we simplify this to the approach to unique identification here to just be the chain name, and some type of version identifier. Does adding everything else to the hash add any value? All other metadata can still live onchain in an ENS record of `myRollup.eth`, but doesn't have to be used to derive the chain ID

ERCS/erc-7785.md
```diff
@@ -0,0 +1,107 @@
...
+
+### Chain name resolution
+
+Any ENS name can resolve to a chain identifier as specified in ENSIP-11. The name should resolve to a record containing not only the chain identifier, but also all the optional information necessary to verify the identifier.
```
mds1, line 58: How does resolution work? I have not used ENS for resolving anything other than addresses. Do we need to specify custom resolvers, or certain records to store data in?

ERCS/erc-7785.md
```diff
@@ -0,0 +1,107 @@
...
+If `chain_name == legacy-rollup` then the mapping is valid.
+
+### Bootstrapping and handover
+
+In order to bootstrap the handling of legacy chain identifiers, we imagine the EF populating the `chainid.reverse` domain, a temporary `l2.eth` for names and then handing them over.
+
+- EF populates two subdomains `l2.eth` and `chainid.reverse` using Ethereum lists.
+- A rollup registers a `rollup.eth` and points it to their `chain_id.
+- EF hands over to the rollup `rollup.l2.eth` and `chain_id.chainid.reverse`
+- The rollup updates `chain_id.chainid.reverse` to return `rollup.eth`
```
mds1, line 96: What stops a chain from always using the legacy process and getting whatever chain ID they want? Is this a one time migration somehow? I don't fully understand the EF's role here

ERCS/erc-7785.md
```diff
@@ -0,0 +1,107 @@
...
+
+
+## Security Considerations
```
mds1, line 99: A lot of tooling currently doesn't support uint256 chain IDs, e.g. see MetaMask: https://github.com/MetaMask/metamask-extension/blob/39528b02100a6003f412bbef5e0b560002c945bc/shared/constants/network.ts#L229-L233

Part of this ERC will require an effort to ensure various tooling developers support uint256 chain IDs

mds1 (commented):
Thanks for putting this together! I am very supportive of this idea in general and have some questions regarding the details :)

ERCS/erc-7785.md
iJaack: I believe this is equal to the genesis contract address, it's for L2s/L3s. If it's not intended like this, then further specifications are needed.

ERCS/erc-7785.md
iJaack: Completely agree here on 3. and 4., but it seems this is a first implementation, hence needs to be as general and flexible as possible.

As I was describing [in another comment](https://github.com/ethereum/ERCs/pull/669#discussion_r1838757887), the fields could be reduced to three or four and the rest of the information can be posted as metadata included in an attestation, so that this registry truly becomes the ultimate source of truth and hub for all EVM chains.

ERCS/erc-7785.md
iJaack: I can tell that a lot of chains use random chainID and don’t care about collision until they reach Mainnet (or a bit after that, unfortunately). Chains that don’t adhere to this registry will always be there, but I guess new ERCs could be implemented that design an ‘enforcement plan’ within and outside the Ethereum ecosystem.

ERCS/erc-7785.md
mds1, line 99: What happens when an ENS name expires and has a new owner? The new owner can change data, what problems will this cause for users? We may want to consider a way to freeze or lock the data so only a specific address can update it

SamWilsn:
Gentle reminder to keep technical discussions on the Ethereum Magicians thread.

ERCS/erc-7785.md
```diff
@@ -0,0 +1,107 @@
...
+---
+
+## Abstract
+
+This ERC proposes to derive chain identifiers as a digest of their chain name (and other information) and to use ENS to map chain names to identifiers in place of the centralized list on GitHub.
+A solution to support existing chain identifiers that were not derived following this ERC is also proposed.
+
+## Motivation
+
+The mapping between chain names and identifiers, such as `Mainnet -> 0x1`, is currently maintained in a centralized list on GitHub (see ```ethereum-lists/chains``` repo).
```
SamWilsn, line 20:
```suggestion
The mapping between chain names and identifiers, such as `Mainnet -> 0x1`, is currently maintained in a centralized list.
```

`ethereum-lists/chains` is an external link and must be removed.

ERCS/erc-7785.md
SamWilsn, line 58: > ENSIP-11

This is an external link, and must be removed. We do have a process for allowing certain external links. See [EIP-5757](https://eips.ethereum.org/EIPS/eip-5757).

SamWilsn (changes requested):

ERCS/erc-7785.md
sampkaML: I hear the point that 32-byte identifiers might seem like an overcorrection at the moment, especially if the current chain ecosystem isn’t scaling into the millions anytime soon. But I believe that making room for robust, collision-resistant identifiers now—even if it adds a small hurdle in the short term—ultimately sets us up for long-term success. By implementing larger identifiers early, we ensure that we won’t need multiple incremental expansions down the line as the ecosystem evolves and potentially grows beyond today’s forecasts.

In other words, taking this step now creates a durable standard that can accommodate future growth without repeated adjustments or migration pains. It’s about thinking ahead and choosing a solution that positions us for scalability and reliability in the years to come.

ERCS/erc-7785.md
```diff
@@ -0,0 +1,107 @@
...
+- `VERSION` is to separate the domain of the hash function with an arbitrary string
+- `BRIDGE` is the address of the L2 on the L1
+
+### Chain name resolution
+
+Any ENS name can resolve to a chain identifier as specified in ENSIP-11. The name should resolve to a record containing not only the chain identifier, but also all the optional information necessary to verify the identifier.
+
+For example the chain name `rollup` can be converted to a chain identifier on Mainnet by resolving:
+```
+rollup.eth -> {version : uint, bridge : address, chain_id : chain_id}
```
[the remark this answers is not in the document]
sampkaML: If they are live testnets i think they should be treated same way as a mainnet is.

ERCS/erc-7785.md
sampkaML: I would say this is the version of the ERC.

ERCS/erc-7785.md
sampkaML: This is a great point. Let me put together some ideas for how we can potentially lock the data.

ERCS/erc-7785.md
```diff
@@ -0,0 +1,107 @@
...
+The input to the function MUST contain the chain name and MAY contain additional information.
+
+An example for a L2:
+```
+chain_id = Keccak-256(CHAIN_NAME, SETTLEMENT_CHAIN_ID, VERSION, DEPLOYER_CONTRACT_ADDRESS, SALT)
+```
+where:
+- `SETTLEMENT_CHAIN_ID` is the id of the L1 where the L2 settles, it could be Mainnet or a testnet.
+- `VERSION` is to separate the domain of the hash function with an arbitrary string
+- `DEPLOYER_CONTRACT_ADDRESS` is the address of the L2 on the L1
```
yoavw, line 54: The one contract that is unlikely to ever change for a rollup, even if it changes its underlying technology, is the canonical bridge.  It can't change easily because it holds assets on L1.  Anything else might change in a future upgrade.  Shouldn't we use `CANONICAL_BRIDGE_ADDRESS` rather than `DEPLOYER`?

ERCS/erc-7785.md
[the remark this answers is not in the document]
yoavw: [ERC-7828](https://github.com/ethereum/ERCs/pull/735/files) indicates that they're resolved in the chain specified as the TLD.  E.g. `rollup.sepolia`.

ERCS/erc-7785.md
yoavw: This should be a one-time process during bootstrap.  EF's role would be just to copy current content of `eth-lists/chains` to the new domain.  This operation will be publicly verifiable.  After it's populated, EF should have no way to keep editing the data in the domain.

ERC-7828 can be seen as an 'enforcement plan' because wallets that use chain-specific addresses won't be able to resolve new chains that don't adhere to this process.

For that reason I proposed (above) that `eth-lists/chains` stops accepting new entries when the one-time migration happens.  It'll prevent inadvertent use of incompatible chainids by devs not familiar with the new process.  Chains devs that don't adhere to it could register elsewhere and should be aware that they opt out of ERC-7828 support.

ERCS/erc-7785.md
```diff
@@ -0,0 +1,107 @@
...
+- A rollup registers a `rollup.eth` and points it to their `chain_id.
+- EF hands over to the rollup `rollup.l2.eth` and `chain_id.chainid.reverse`
+- The rollup updates `chain_id.chainid.reverse` to return `rollup.eth`
+
+
+## Security Considerations
+
+Domain spoofing can lead to replay attacks as described above and can be eliminated by deriving new identifiers using a hash function and by checking the reverse mapping for legacy identifiers.
+
+Domain squatting, the practice of ammassing a large number of domains in the hope to selling them later to legitimate users, is a possibility but with an increasing number of L2 registrations we can expect the same problem to appear in the centralized Github list.
```
yoavw, line 103: This domain could also be made more expensive to deter squatting.  L2 projects require significant funding anyway so spending hundreds of dollars on registration won't be a major issue, but will make squatting impractical.

sambacha (commented):
Why does ENS get to be the resolver? 

This effectively gives them de facto namespace control over any L2 that uses this.

dmap seems a much more appropriate solution if your wanting namespace delegation and resolution.

ERCS/erc-7785.md
iJaack: Just to again remind about the chainID collision: there are two chains that currently use the same chainID.

[Nova Network](https://chainid.network/chain/87/) and [Waves](https://docs.waves.tech/en/blockchain/blockchain-network/#connecting-node-to-blockchain-network) use the same chainID, 87.

ERCS/erc-7785.md
```diff
@@ -0,0 +1,107 @@
...
+Desired properties:
+- the ability to register new chain names and identifiers in a censorship-resistant way
+- the ability to resolve chain names and identifiers in a trustless way
+- maintain a unique mapping between names and identifiers
+
+### Chain Identifier Spoofing and Replay Attacks
+
+An important property of the centralized list is that it keeps a one-to-one correspondence between names and indentifiers.
+
+Without this property, an attacker could register a fresh name pointing to an existing identifier. For example `my-testnet` could point to mainnet `0x1`. A user could be tricked into signing a transaction for the innocent looking `my-testnet` while actually signing a transaction for mainnet, a transaction that we attacker can then replay.
```
thedarkjester, line 34:
```suggestion
Without this property, an attacker could register a fresh name pointing to an existing identifier. For example `my-testnet` could point to mainnet `0x1`. A user could be tricked into signing a transaction for the innocent looking `my-testnet` while actually signing a transaction for mainnet, a transaction that the attacker can then replay.
```

ERCS/erc-7785.md
clowestab: It says 

> DEPLOYER_CONTRACT_ADDRESS is the address of the L2 on the L1

As in the Rollup Contract? Its not beyond the realms of possibility that this could change. As could `SETTLEMENT_CHAIN_ID` in principle. Is that the intention?

iJaack: Then here’s the principle again, as I stated in [another comment](https://github.com/ethereum/ERCs/pull/669#discussion_r1838757887): if the `DEPLOYER_CONTRACT_ADDRESS`, does it mean that the Rollup changes, relative to the L1? In the same way, if the `SETTLEMENT_CHAIN_ID` changes, the chain is changing nature, so it’s **another chain**.

That’s because if the user deposits on the Rollup, they do it knowing that there’s a specific configuration (at a high level), for example that the X L2 settles on Ethereum.

If the settlement chain changes to another L1, maybe the users or developers don’t trust it anymore because they trusted Ethereum better. The same could be said about changing the DA layer, and similarly the tech stack (OP, ZK, SN, etc.)

Maybe the alias can be the same, but there’s got to be a way to differentiate the chains when they change their fundamental nature. This is about trust assumption as much as it is about versioning chain releases for the community.

clowestab: > If the settlement chain changes to another L1, maybe the users or developers don’t trust it anymore because they trusted Ethereum better. The same could be said about changing the DA layer, and similarly the tech stack (OP, ZK, SN, etc.)

Absolutely agree on this aspect.

But fundamental/significant changes that effect end users can also occur whilst on the same chain. Many rollups use proxy contracts and upgrade implementations. There is huge inconsistency across chain operators regarding process in this regard. I'd like to see process behind this and agreement on what constitutes a 'chain'. Noting this, the `VERSION` is particularly important in my opinion.

clowestab: Noting [my comment above](https://github.com/ethereum/ERCs/pull/669#discussion_r1938506265), clarity is needed on this.

SamWilsn (approved):

ERCS/erc-7785.md
iJaack: This could be tricky, you're right, but if the change in nature requires a regensis, a different validator set, of different security assumptions, then yes, I think they need to change chainID. It's not the same chain anymore, by definition.

This is not what's happening in the industry, I want to be clear. But there needs to be a standard around it.

Like, why has Fantom 'moved' to Sonic (so new chain name, new chainID, etc.) while Celo hasn't?

What's the difference between the two?

noelportillo (commented):
The first one was a bit

PR #673, opened 2024-10-14
Add ERC: Cross-Chain Messaging Gateway

Amxx:
This proposal describes an interface, and the corresponding workflow, for smart contracts to send arbitrary data through cross-chain messaging protocols. The end goal of this proposal is to have all such messaging protocols accessible via this interface (natively or using "adapters") to improve their composability and interoperability. That would allow a new class of cross-chain native smart contracts to emerge while reducing vendor lock-in. This proposal is modular by design, allowing users to leverage bridge-specific features through attributes while providing simple "universal" access to the simple feature of "just getting a simple message through".


ERCS/erc-yyyy.md
```diff
@@ -0,0 +1,215 @@
+---
+title: Cross-Chain Messaging Gateway
+description: An interface for contracts to send and receive cross-chain messages.
+author: Francisco Giordano (@frangio), Hadrien Croubois (@Amxx), Ernesto Garcia (@ernestognw), CJ Cobb (@cjcobb23)
+discussions-to: <URL>
```
Amxx, line 5:
```suggestion
discussions-to: https://ethereum-magicians.org/t/erc-yyyy-cross-chain-messaging-gateway/21374
```

ERCS/erc-yyyy.md
```diff
@@ -0,0 +1,215 @@
...
+The following standard attributes MAY be supported by a gateway.
+
+- `postProcessingOwner(address)`: The address of the account that shall be in charge of message post-processing.
+
+### Source Gateway
+
+An Source Gateway is a contract that offers a protocol to send a message to a receiver on another chain. It MUST implement `IGatewaySource`.
+
+```solidity
+interface IGatewaySource {
```
Amxx, line 67:
```suggestion
interface IERCXXXXGatewaySource {
```

ERCS/erc-yyyy.md
```diff
@@ -0,0 +1,215 @@
...
+### Destination Gateway
+
+A Destination Gateway is a contract that implements a protocol to validate messages sent on other chains and have them received at their destination.
+
+The gateway can operate in Active or Passive Mode.
+
+In both modes, the destination account of a message, aka the receiver, MUST implement a `receiveMessage` function. The use of this function depends on the mode of the gateway as described in the following sections.
+
+```solidity
+interface IGatewayReceiver {
```
Amxx, line 133:
```suggestion
interface IERCXXXXReceiver {
```

ERCS/erc-yyyy.md
```diff
@@ -0,0 +1,215 @@
...
+The arguments `gateway` and `gatewayMessageKey` are unused in active mode and SHOULD be zero and empty respectively.
+
+#### Passive Mode
+
+The gateway does not directly invoke `receiveMessage`, but provides a means to validate messages. The receiver allows any party to invoke `receiveMessage`, but if the caller is not a known gateway it MUST check that the gateway provided as an argument is a known gateway, and it MUST validate the message against it before accepting it, forwarding the message key.
+
+A gateway acting in passive mode MUST implement `IGatewayDestinationPassive`. If a gateway operates exclusively in active mode, the implementation of this interface is OPTIONAL.
+
+```solidity
+interface IGatewayDestinationPassive {
```
Amxx, line 158:
```suggestion
interface IERCXXXXGatewayDestinationPassive {
```

ERCS/erc-yyyy.md
```diff
@@ -0,0 +1,216 @@
+---
```
abcoathup, line 1:
```suggestion
---
eip: 7786
```

ERCS/erc-yyyy.md
```diff
@@ -0,0 +1,216 @@
...
+title: Cross-Chain Messaging Gateway
+description: An interface for contracts to send and receive cross-chain messages.
+author: Francisco Giordano (@frangio), Hadrien Croubois (@Amxx), Ernesto Garcia (@ernestognw), CJ Cobb (@cjcobb23)
+discussions-to: https://ethereum-magicians.org/t/erc-yyyy-cross-chain-messaging-gateway/21374
```
abcoathup, line 5:
```suggestion
discussions-to: https://ethereum-magicians.org/t/erc-7786-cross-chain-messaging-gateway/21374
```
I updated the title on Eth Magicians to add the assigned ERC number.

Amxx:
@abcoathup, We believe the document listed the "Reference" section is really valuable to the reader. Is there any "ERC compliant" way to link to it ?

Asfand78690 (approved):

Asfand78690 (approved):

Asfand78690 (approved):

ERCS/erc-7786.md
```diff
@@ -0,0 +1,217 @@
...
+    event MessageCreated(bytes32 indexed outboxId, string sender, string receiver, bytes payload, uint256 value, bytes[] attributes);
+    event MessageSent(bytes32 indexed outboxId);
+
+    error UnsupportedAttribute(bytes4 signature);
+
+    function supportsAttribute(bytes4 signature) external view returns (bool);
+
+    function sendMessage(
+        string calldata destination, // CAIP-2 chain identifier
+        string calldata receiver, // CAIP-10 account address
```
cjcobb23, line 78: I think it would be clearer to call this destinationChain and destinationAddress. Without the comments, it's unclear to me which is the chain versus address.

ERCS/erc-7786.md
```diff
@@ -0,0 +1,217 @@
...
+
+Initiates the sending of a message.
+
+Further action MAY be required by the gateway to make the sending of the message effective, such as providing payment for gas. See Post-processing.
+
+MUST revert with `UnsupportedAttribute` if an unsupported attribute key is included. MAY revert if the value of an attribute is not a valid encoding for its expected type.
+
+MAY accept call value (native token) to be sent with the message. MUST revert if call value is included but it is not a feature supported by the gateway. It is unspecified how this value is represented on the destination.
+
+MAY generate and return a unique non-zero *outbox identifier*, otherwise returning zero. This identifier shall be used to track the lifecycle of the message in the outbox in events and for post-processing.
```
cjcobb23, line 101:
```suggestion
MAY generate and return a unique non-zero *outbox identifier*, otherwise returning zero. This identifier can be used to track the lifecycle of the message in the outbox in events and for post-processing.
```

ERCS/erc-7786.md
```diff
@@ -0,0 +1,217 @@
...
+
+MUST emit a `MessageCreated` event, including the optional outbox identifier that is returned by the function.
+
+If an outbox identifier was generated, MUST emit a `MessageSent` event if no post-processing is required.
```
cjcobb23, line 105:
```suggestion
MUST emit a `MessageSent` event if no post-processing is required.
```
Why do we need the first part of the sentence about the outbox id?

ERCS/erc-7786.md
```diff
@@ -0,0 +1,217 @@
...
+
+#### `MessageCreated`
+
+This event signals that a would-be sender has initiated the sending of a message.
+
+If `outboxId` is present, post-processing MAY be required to send the message through the cross-chain channel.
+
+#### `MessageSent`
+
+This event signals that no more post-processing in the source chain is required, and that the message has been sent through the cross-chain channel.
```
cjcobb23, line 115:
```suggestion
This event signals that no more post-processing in the source chain is required, and that the message is able to be sent.
```
We don't really know whether the message was sent or not if the relaying is off chain and decoupled from the protocol, so I think the language should be more about whether the message is ready for routing.

ERCS/erc-7786.md
```diff
@@ -0,0 +1,217 @@
...
+
+#### Post-processing
+
+After a sender has invoked `sendMessage`, further action MAY be required by the gateways to make the message effective. This is called *post-processing*. For example, some payment is typically required to cover the gas of executing the message at the destination.
+
+The exact interface for any such action is out of scope of this ERC. If the `postProcessingOwner` attribute is supported and present, such actions MUST be restricted to the specified account, otherwise they MUST be able to be performed by any party in a way that MUST NOT be able to compromise the eventual receipt of the message.
+
+The gateway MUST emit a `MessageSent` event with the appropriate identifier once no more post-processing is required by the source gateway. Further post-processing MAY be required in the source or destination gateways.
```
cjcobb23, line 123: I think we need to distinguish protocol post-processing vs other post processing. In the Axelar example, gas payment is required and this is a form of post processing, but the gateway has no knowledge of this. So it's unclear based on how this is written if/when the gateway should emit `MessageSent`.

I personally think `MessageSent` could probably use a renaming, and simply means the message is ready for sending. Whether the message is actually sent or not is hard for the gateway to determine and might not be possible, but the gateway does know if anything else needs to be done for the message to be able to be sent.

ERCS/erc-7786.md
```diff
@@ -0,0 +1,217 @@
...
+
+A Destination Gateway is a contract that implements a protocol to validate messages sent on other chains and have them received at their destination.
+
+The gateway can operate in Active or Passive Mode.
+
+In both modes, the destination account of a message, aka the receiver, MUST implement a `receiveMessage` function. The use of this function depends on the mode of the gateway as described in the following sections.
+
+```solidity
+interface IERC7786Receiver {
+    function receiveMessage(
```
cjcobb23, line 135:
```suggestion
    function executeMessage(
```
since the contract is actually executing the payload, and not just receiving it

ERCS/erc-7786.md
```diff
@@ -0,0 +1,217 @@
...
+        address gateway,
+        bytes calldata gatewayMessageKey,
+        string calldata source, // CAIP-2 chain identifier
+        string calldata sender, // CAIP-10 account address
```
cjcobb23, line 139:
```suggestion
        string calldata sourceChain, // CAIP-2 chain identifier
        string calldata sourceAddress, // CAIP-10 account address
```

ERCS/erc-7786.md
```diff
@@ -0,0 +1,217 @@
...
+
+#### Passive Mode
+
+The gateway does not directly invoke `receiveMessage`, but provides a means to validate messages. The receiver allows any party to invoke `receiveMessage`, but if the caller is not a known gateway it MUST check that the gateway provided as an argument is a known gateway, and it MUST validate the message against it before accepting it, forwarding the message key.
+
+A gateway acting in passive mode MUST implement `IERC7786GatewayDestinationPassive`. If a gateway operates exclusively in active mode, the implementation of this interface is OPTIONAL.
+
+```solidity
+interface IERC7786GatewayDestinationPassive {
+    function validateReceivedMessage(
```
cjcobb23, line 160:
```suggestion
    function executeMessage(
```
Can we just call it this? I feel the naming is a bit confusing, since not only does the function validate, it also marks it as executed. Or maybe `validateAndExecute`. The name should be a little more clear about what the function does.

ERCS/erc-7786.md
```diff
@@ -0,0 +1,217 @@
...
+}
+```
+
+##### `validateReceivedMessage`
+
+Checks that there is a valid and as yet unexecuted message whose contents are exactly those passed as arguments and whose receiver is the caller of the function. The message key MAY be an identifier, or another piece of data necessary for validation.
+
+MUST revert if the message is invalid or has already been executed.
+
+TBD: Passing full payload or payload hash (as done by Axelar). Same question for attributes, possibly different answer depending on attribute malleability.
```
cjcobb23, line 176: is passing the payload hash going to be more efficient?

ERCS/erc-7786.md
```diff
@@ -0,0 +1,217 @@
...
+
+#### Dual Active-Passive Mode
+
+A gateway MAY operate in both active and passive modes, or it MAY switch from operating exclusively in active mode to passive mode or vice versa.
+
+A receiver SHOULD support both active and passive modes for any gateway. This is accomplished by first checking whether the caller of `receiveMessage` is a known gateway, and only validating the message if it is not; the first case supports an active mode gateway, while the second case supports a passive mode gateway.
```
cjcobb23, line 182: I feel this section adds complexity. Why is this necessary?

ERCS/erc-7786.md
```diff
@@ -0,0 +1,217 @@
...
+
+### Pending discussion
+
+- How to "reply" to a message? Duplex gateway? Getter for reverse gateway address? Necessary for some applications, e.g., recovery from token bridging failure?
```
cjcobb23, line 186: Can you not just send a message back to the source address?

ERCS/erc-7786.md
```diff
@@ -0,0 +1,217 @@
...
+```
+
+##### `validateReceivedMessage`
+
+Checks that there is a valid and as yet unexecuted message whose contents are exactly those passed as arguments and whose receiver is the caller of the function. The message key MAY be an identifier, or another piece of data necessary for validation.
+
+MUST revert if the message is invalid or has already been executed.
+
+TBD: Passing full payload or payload hash (as done by Axelar). Same question for attributes, possibly different answer depending on attribute malleability.
+
```
cjcobb23, line 177: I think this is a good place for the properties section. We could call it guarantees.

- Only once execution
- Safety: A message is delivered at the destination if and only if it was sent at the source, and the sending transaction is finalized.
- Liveness: A sent message is delivered at the destination eventually, with minimal delay.
- Decentralization: The above properties do not assume trust of any centralized entity. Safety should be ensured through a trustless mechanism such as proofs or a decentralized, open validator set. Liveness should be ensured through permissionless relaying (anyone can relay), and a decentralized, open validator set (if applicable).

Can expand a little bit on this, and maybe in another doc go into greater detail. But I think this is important to specify here, because we don't want totally centralized entities implementing this API.

ERCS/erc-7786.md
Amxx: Thats is documented on line 35 of this document, and it matches the wording in this document https://github.com/0xFableOrg/xchain/tree/7789933bba24e2dd893cf515157af70474e7180b

That being says, I think your changes are not bad. Its a balance between clarity and conciseness

ERCS/erc-7786.md
Amxx: Because if there is no outbox id (if its 0) then the message is implicitelly sent when the `MessageCreated` event is emitted. In that case no `MessageSent` event is emitted.

ERCS/erc-7786.md
```diff
@@ -0,0 +1,217 @@
...
+#### Dual Active-Passive Mode
+
+A gateway MAY operate in both active and passive modes, or it MAY switch from operating exclusively in active mode to passive mode or vice versa.
+
+A receiver SHOULD support both active and passive modes for any gateway. This is accomplished by first checking whether the caller of `receiveMessage` is a known gateway, and only validating the message if it is not; the first case supports an active mode gateway, while the second case supports a passive mode gateway.
+
```
frangio, line 183:
```suggestion

### Properties

The protocol underlying a pair of gateways is expected to guarantee a series of properties. For detailed definition and discussion we refer to XChain Research’s [Cross-chain Interoperability Report](https://github.com/0xFableOrg/xchain/blob/7789933bba24e2dd893cf515157af70474e7180b/README.md).

- The protocol MUST guarantee Safety: A message is delivered at the destination if and only if it was sent at the source, and the sending transaction was finalized.
- The protocol MUST guarantee Liveness: A sent message is delivered at the destination eventually, assuming Liveness and censorship-resistance of the source and destination chains.
- The protocol SHOULD guarantee Timeliness: A sent message is delivered at the destination within a bounded delivery time, which should be documented.
- The above properties SHOULD NOT rely on trust in some centralized actor.

```

ERCS/erc-7786.md
Amxx: What would you think of
- `sourceChain`
- `sender`
- `destinationChain`
- `receiver`
?

ERCS/erc-7786.md
```diff
@@ -181,6 +181,15 @@ A gateway MAY operate in both active and passive modes, or it MAY switch from op
...
 
+### Properties
+
+The protocol underlying a pair of gateways is expected to guarantee a series of properties. For detailed definition and discussion we refer to XChain Research’s [Cross-chain Interoperability Report](https://github.com/0xFableOrg/xchain/blob/7789933bba24e2dd893cf515157af70474e7180b/README.md).
+
+- The protocol MUST guarantee Safety: A message is delivered at the destination if and only if it was sent at the source, and the sending transaction was finalized.
+- The protocol MUST guarantee Liveness: A sent message is delivered at the destination eventually, assuming Liveness and censorship-resistance of the source and destination chains.
+- The protocol SHOULD guarantee Timeliness: A sent message is delivered at the destination within a bounded delivery time, which should be documented.
+- The above properties SHOULD NOT rely on trust in some centralized actor.
+
```
cjcobb23, line 192: Can we add a sentence about only once execution here? I think this is an important property.

ERCS/erc-7786.md
frangio: Added more or less this below.

ERCS/erc-7786.md
Amxx: Since the receiver already has a `executeMessage` this change would create confusion.

ERCS/erc-7786.md
cjcobb23: I think that is clearer, at least the chain fields are explicit, and therefore the other fields meaning can be inferred

ERCS/erc-7786.md
frangio: How about `markExecutedMessage`? If the message is invalid it will revert.

ERCS/erc-7786.md
frangio: It would be slightly more efficient.

ERCS/erc-7786.md
frangio: What do you think about `MessageReady`. What other terms would you propose?

ERCS/erc-7786.md
```diff
@@ -0,0 +1,222 @@
...
+### Properties
+
+The protocol underlying a pair of gateways is expected to guarantee a series of properties. For detailed definition and discussion we refer to XChain Research’s [Cross-chain Interoperability Report](https://github.com/0xFableOrg/xchain/blob/7789933bba24e2dd893cf515157af70474e7180b/README.md).
+
+- The protocol MUST guarantee Safety: A message is delivered at the destination if and only if it was sent at the source, and the sending transaction was finalized.
+- The protocol MUST guarantee Liveness: A sent message is delivered at the destination eventually, assuming Liveness and censorship-resistance of the source and destination chains.
+- The protocol SHOULD guarantee Timeliness: A sent message is delivered at the destination within a bounded delivery time, which should be documented.
+- The above properties SHOULD NOT rely on trust in some centralized actor.
```
cjcobb23, line 191: Can we expand on this a little bit?
```suggestion
- The above properties SHOULD NOT rely on trust in some centralized actor. For example, safety should be guaranteed by some trustless mechanism such as a light client proof, or attestations by an open, decentralized validator set. Relaying should be decentralized or permissionless to ensure liveness; a centralized relayer can fail and thus halt the protocol.
```

ERCS/erc-7786.md
```diff
@@ -0,0 +1,222 @@
...
+
+## Motivation
+
+Cross-chain messaging systems (or bridges) allow communication between smart contracts deployed on different blockchains. There is a large diversity of such systems with multiple degrees of decentralization, with various components, that implement different interfaces, and provide different guarantees to the users.
+
+Because almost every protocol implementing a different workflow, using a specific interface, portability between bridges is basically impossible. This also forbid the development of generic contracts that rely on cross chain communication.
+
+The objective of the ERC is to provide a standard interface, and a corresponding workflow, for performing cross-chain communication between contracts. Existing cross-chain communication protocols, that do not nativelly implement this interface, should be able to adopt it using adapter gateway contracts.
+
```
norswap, line 24: There have been quite a few standards of these nature proposed, and none of them have seemingly achieved any traction.

Here is an overview of the state of these standards: https://github.com/0xFableOrg/xchain/blob/master/README.md#messaging-formats--interfaces

On the ERC front only, we have:
- [ERC-5164: Cross-Chain Execution](https://eips.ethereum.org/EIPS/eip-5164)
- [ERC-6170: Cross-Chain Messaging Interface](https://eips.ethereum.org/EIPS/eip-6170)
 - [ERC-7533: Public Cross Port](https://eips.ethereum.org/EIPS/eip-7533)
 
Do we need another one? What makes this one different / necessary? And most importantly, though I'm not sure that's in scope for the EIP motivation section: what will make it so that this one gets actually used?

Asfand78690 (approved):

ERCS/erc-7786.md
frangio: Thanks for the feedback! We should definitely address these points in the ERC. My views are:

ERC-7533 is very low level and seems to mandate a lot of details about the protocol. The goal of this new ERC is to be a viable interface for any underlying cross-chain messaging protocol.

ERC-5164's main drawback is lack of support for non-EVM chains, since it relies on the EVM's chain id. ERC-6170 technically supports any chain id format but doesn't specify which one, so it doesn't achieve portability. This new ERC mandates the use of CAIP-2 and CAIP-10 identifiers.

ERC-5164 also lacks the ability to expose other features/parameters of the underlying protocol such as setting a minimum gas limit and attaching tokens. ERC-6170 has a separate "bytes data" argument that could technically be used for these things but does not specify a format or way to query whether they're supported, so it would struggle with portability as is. This new ERC has an extensible mechanism for attributes that can expose more advanced bridge features, with a viable path to standardizing the different kinds of attributes.

> what will make it so that this one gets actually used?

The theory is that if the interface is simple and unopinionated yet expressive enough, it should be a no-brainer to adopt for the purpose of increasing interoperability and composability. We're working with Axelar and actively reaching out to developers of other cross-chain messaging protocols to get a critical mass.

frangio: @norswap I'd be interested to hear your thoughts on the spec!

ERCS/erc-7786.md
gusjavaz: Passing the full payload can be helpful in some cases for setting up validation logic. For instance, if the payload specifies an amount of tokens to mint, additional validation might be required if the amount exceeds a certain threshold.

ERCS/erc-7786.md
frangio: Going with `setMessageExecuted`.

ERCS/erc-7786.md
Amxx: IMO that is part of Safety. I would expend the safety part to mention that. WDYT?

ERCS/erc-7786.md
```diff
@@ -0,0 +1,222 @@
...
+A gateway MAY operate in both active and passive modes, or it MAY switch from operating exclusively in active mode to passive mode or vice versa.
+
+A receiver SHOULD support both active and passive modes for any gateway. This is accomplished by first checking whether the caller of `receiveMessage` is a known gateway, and only validating the message if it is not; the first case supports an active mode gateway, while the second case supports a passive mode gateway.
+
+### Properties
+
+The protocol underlying a pair of gateways is expected to guarantee a series of properties. For detailed definition and discussion we refer to XChain Research’s [Cross-chain Interoperability Report](https://github.com/0xFableOrg/xchain/blob/7789933bba24e2dd893cf515157af70474e7180b/README.md).
+
+- The protocol MUST guarantee Safety: A message is delivered at the destination if and only if it was sent at the source, and the sending transaction was finalized.
```
Amxx, line 188:
```suggestion
- The protocol MUST guarantee Safety: A message is delivered at the destination if and only if it was sent at the source, and with the guarantees that match the sending transaction. Therefore, a message should only be delivered once the sending transation is finalized, and a message should not be delivered more than once. Note here that a payload that is sent multiple time through multiple messages may be received more than once, but each message may only be received once.
```

ernestognw: I can't edit, but I'd adjust it to:

- The protocol MUST guarantee Safety: A message is delivered at the destination if and only if it was sent at the source. The delivery process must ensure a message is only delivered once the sending transaction is finalized and not delivered more than once. Note that a single payload may be sent and received multiple times through independent messages, but even in that case, each message is only received once.

ERCS/erc-7786.md
cjcobb23: In addition to what @frangio said, something else I would like to add is that neither ERC 5164 nor ERC 6170 standardize the interface that apps use to verify messages with the gateway/interop protocol. Which is a major component of any cross chain app, and without standardizing that layer, it is impossible to build a generic cross chain app, since the app is still tightly coupled with the verification of messages.

norswap: @frangio Thanks for the answer, that all makes sense.

@cjcobb23 I'm not sure I fully understand — EIP6170 has a `receiveMessage`. The difference seems to be that theirs live in their inbox (gateway), which then will call the app (as presumably the calldata is a message to the app). If I understand correctly, that's the same thing as the "active" mode in this EIP (with the difference is that the gateway function is unspecified and calls the `receiveMessage` on the destination contract instead of another function of the app's choosing).

While the passive mode is interesting, I don't think it really enables more use cases? The passive mode allows you to abstract the gateway (but they still need to be verified). This can equivalently be done in EIP6170 by having their inbox `receiveMessage` call a function in the app that checks the `msg.sender` (who then similarly checks against a list of known inboxes).

ERCS/erc-7786.md
```diff
@@ -0,0 +1,222 @@
...
+
+A Destination Gateway is a contract that implements a protocol to validate messages sent on other chains and have them received at their destination.
+
+The gateway can operate in Active or Passive Mode.
+
+In both modes, the destination account of a message, aka the receiver, MUST implement a `receiveMessage` function. The use of this function depends on the mode of the gateway as described in the following sections.
+
+```solidity
+interface IERC7786Receiver {
+    function executeMessage(
```
norswap, line 135:
```suggestion
    function receiveMessage(
```

ERCS/erc-7786.md
```diff
@@ -0,0 +1,222 @@
...
+        address gateway,
+        bytes calldata gatewayMessageKey,
+        string calldata sourceChain, // CAIP-2 chain identifier
+        string calldata sender, // CAIP-10 account address
+        bytes calldata payload,
+        bytes[] calldata attributes
+    ) external payable;
+}
+```
```
norswap, line 144: It's slightly confusing that the interface here is the interface of an app contract when it's under the "Destination Gateway" section. Reading the text does clarify though (esp. after fixing the mistake, it made me wonder if `executeMessage` was meant to be for the gateway and `receiveMessage` for the app receiver).

ERCS/erc-7786.md
```diff
@@ -0,0 +1,222 @@
...
+
+#### Active Mode
+
+The gateway directly invokes `receiveMessage`, and only does so with valid messages. The receiver MUST assume that a message is valid if the caller is a known gateway.
```
norswap, line 148: I was a bit puzzled about what "the gateway directly invokes" meant here.

I now understand that it means that somehow (unspecified)  the gateway control gets called by the messaging/bridging system, and then calls `receiveMessage` in turn.

ERCS/erc-7786.md
norswap: It allows supporting gateways that work in both modes, in case one wants to start using gateways in the other mode.

ERCS/erc-7786.md
norswap: Agreed, this should be an app-level concern (they could send a msg id on the calldata).

ERCS/erc-7786.md
```diff
@@ -0,0 +1,222 @@
...
+        bytes calldata messageKey,
+        string calldata sourceChain, // CAIP-2 chain identifier
+        string calldata sender, // CAIP-10 account address
+        bytes calldata payload,
+        bytes[] calldata attributes
+    ) external;
+}
+```
+
+##### `setMessageExecuted`
```
norswap, line 170: This is super confusing.

- It implies (but this is never stated explicitly) that messages have to be stored somewhere prior to this being called. Presumably by a previous call to `receiveMessage`?
      - This is super wasteful in terms of chain storage, and is expensive as well.
- I'm not entirely sure what the expected rationale is... a "gateway operator" wait for someone to call `receiveMessage` then actually executes it and calls `setMessageExecuted`?

ERCS/erc-7786.md
```diff
@@ -0,0 +1,222 @@
...
+}
+```
+
+#### Active Mode
+
+The gateway directly invokes `receiveMessage`, and only does so with valid messages. The receiver MUST assume that a message is valid if the caller is a known gateway.
+
+The arguments `gateway` and `gatewayMessageKey` are unused in active mode and SHOULD be zero and empty respectively.
+
+#### Passive Mode
```
norswap, line 152: I'm not a fan of the active / passive distinction, and don't really understand the passive mode's use case (cf. discussion on the comment thread in the motivation section above).

At first reading, I understood it as a way to distinguish between "permissioned" calls from a gateway operator, and "permissionless" calls for systems where proving can be done without permissionning things (light clients, zk). That doesn't seem to be the case, and even if it were, it could be achieve by letting people permissionlessly call a function on the gateway that then calls `receiveMessage`.

But the actual behaviour is ... not clear. The description makes it sound like it's validation only. This is a fair use case: a relayer may want to check if a message would succeed on the destination before relaying it. But can be achieved by either (1) an `eth_call`, even in permissioned cases (spoof the authorized sender, though this is a bit ugly/brittle and leaks implementation details), (2) a `verifyMessage` function (meant to be `eth_call`ed only) that verifies the non-permissioned parts (authenticity checks if any and app execution).

But maybe that's not what it does at all, as `setMessageExecuted` below seems to imply this is more than verification but actually stores an unexecuted message for later execution. This only seems useful insofar that the system is permissioned and the operator can eschew retrieving the source message from the source and only perform verification on it. But in a permissioned system, verification will require looking at the source anyway.

ERCS/erc-7786.md
```diff
@@ -0,0 +1,222 @@
...
+
+## Backwards Compatibility
+
+Existing cross-chain messaging protocols implement proprietary interfaces. We recommend that protocols natively implement the standard interface defined here, and propose the development of standard adapters for those that don't.
+
+## Security Considerations
+
+Unfortunately, CAIP-2 and CAIP-10 names are not unique. Using non-canonical strings may lead to undefined behavior, including message delivery failure and locked assets. While source gateways have a role to play in checking that user input are valid, we also believe that more effort should be put into standardizing and documenting what the canonical format is for each CAIP-2 namespace. This effort is beyond the scope of this ERC.
+
+Needs discussion.
```
norswap, line 218: Would add a note about replayability it, instructing that nonce management is an app-level concern.

ERCS/erc-7786.md
```diff
@@ -0,0 +1,222 @@
...
+- The above properties SHOULD NOT rely on trust in some centralized actor. For example, safety should be guaranteed by some trustless mechanism such as a light client proof, or attestations by an open, decentralized validator set. Relaying should be decentralized or permissionless to ensure liveness; a centralized relayer can fail and thus halt the protocol.
+
+### Pending discussion
+
+- How to "reply" to a message? Duplex gateway? Getter for reverse gateway address? Necessary for some applications, e.g., recovery from token bridging failure?
+- Should the destination and receiver inputs of `sendMessage` be kept as two separate strings, or merged as a single CAIP-10 string with a `:` separator? This has implication of the calldata length, which in some cases may be stored.
+- Do we want the gateway to have the ability to inform users of the address of the new version, similar to how `AccessManager` can update then authority trusted by an `AccessManaged`? This could be useful if a gateway is ever deprecated in favor of a new version.
+- Should data and metadata attributes be split in two? What are data attributes used for? Do we need them?
```
norswap, line 198: I would think that in the ERC-20 use case, it would be metadata, the gateway would look at it and either transfer the ERC-20 before calling the receiver or (arguably better), set an allowance to allow the destination to transfer the tokens to itself.

In this case the metadata can be stripped, or passed on to the receiver. (It stripped, might require the receiver to encode it in its own data format in some cases. In other cases, it's not needed (e.g. buying an NFT whose price is fixed — in that case it will try to transfer the requested amount and either succeed or fail).

Data on the other hand, seems unnecessary, since the sender already receives a payload it can use to encode whatever it needs. If it's useless to the gateways, it should be in the payload.

norswap: Btw, the ERC-20 use case is a nice illustration of the usefulness of metadata . A single-destination messaging system will really struggle to transfer ERC-20 and do a contract call at once (it would require something like making sure the ERC-20 contract itself be able to forward arbitrary calls to contracts).

Now, for that use case, a possibly simpler solution would be to enable multicalls in the messaging interface. This needs to be at the gateway level because simply deploying a "multicall receiver" would not work: the caller of the ERC-20 contract to mint a token would be the multicall receiver instead of the gateway.

I think it would be great to list off more potential use cases for metadata properties, but it seems potentially useful enough (for things we may not even foresee) that I would definitely keep it.

I think the multicall system is probably very generally useful as well, and you may want to consider including it. EIP-4337 (in which strangely I see many parallels to this despite their very different nature) has a similar capability as well.

ERCS/erc-7786.md
frangio: Good point let me split this off to a Receiver section.

ERCS/erc-7786.md
Asfand78690: UQAxBkv-DqOjrhJ7s5J6C9w3WqbSwB9H-kJc0cI-ElSbzi0l

ERCS/erc-7786.md
Amxx:
```suggestion
```

ERCS/erc-7786.md
Amxx: Would that be clearer ?
```suggestion
In active mode, the gateway calls `receiveMessage`, and only does so with valid messages. The receiver MUST assume that a message is valid if the caller is a known gateway.
```

ERCS/erc-7786.md
Amxx: In active mode the execution flows goes "someone" (relayer?) -> "gateway" -> "receiver", and the `executeMessage` is called by the gateway.

But some gateway refuse to have any "outbound calls" for security reasons. In that case, the execution flow goes "someone" (relayer?) -> "receiver" -> "gateway".

In boith cases we want to use the same entrypoint on the receiver (`executeMessage`). The distinction between the two mode is made by checking "is msg.sender" a known gateway? If no I'm in passive mode and I have to do the check.

ERCS/erc-7786.md
Amxx: This function used to be called `receiveMessage` and was renamed to `executeMessage`.

Personnaly, I prefered `receiveMessage` ... but at some point we have to stop going back and forward.
@frangio

ERCS/erc-7786.md
```diff
@@ -0,0 +1,212 @@
...
+```solidity
+interface IERC7786Receiver {
+    function executeMessage(
+        address gateway,
+        bytes calldata gatewayMessageKey,
+        string calldata sourceChain, // CAIP-2 chain identifier
+        string calldata sender, // CAIP-10 account address
+        bytes calldata payload,
+        bytes[] calldata attributes
+    ) external payable;
```
Amxx, line 131: IMO this should return a "magic value" (`IERC7786Receiver.executeMessage.selector`) that the gateway can check in active mode, to avoid any confusion with fallback messages.

```suggestion
    ) external payable returns (bytes4);
```

ERCS/erc-7786.md
Amxx: How are messages usually relayed/executed ?

There are cases such as mainnet->matic were the execution is part of the protocol, but in most cases, the execution involves some kind of ticket being passed and stored between chains.

For matic -> mainnet, we have a state root that is used to prove stuff that happened on matic. In that case the messageKey could be a merkle inclusion proof of some event in this state.

AFAIK, Arbitrum has a ticket system, which is basically the storage we need here. 

So yes its "wastefull", but unless the relayer has some specific authentication, it feels like its needed, and we already ahve this.

---

Now I get the possible confusion with the fact that the passive mode is optional, and only needs to be supported if the active mode has some failure mode. My personal opinion is that the passive mode can be hidden behind an "active adaptor". Many protocole already have a "retry mechanism" in the gateway. This "active adaptor" would turn a passive gateway into an active one with a similar "retry mechanism".

The reason we have this passive mode is that some protocole do work that way, and AFAIK they don't want to implement/maintain this active adaptor themselves. We could technically get rid of passive mode, and require someone to maintain that active adaptor ...

Amxx: Note: this passive mode is matching Axelar's workflow

"Someone/anyone" calls the destination contract with message details. This could use any endpoint, though there is a "standard" one that the axelar relayer networks uses, and that receiver that want use this network must have.

The receiver then calls axelar to confirm the details are correct. Axelar just says "yes or no". The receiver can fail, or do whatever they want.

Again the rational here is that the axelar gateway does never call any contract directly (for safety / reentrenacy protection)

ERCS/erc-7786.md
norswap: I think what I was mostly missing was "who calls the gateway"?

Somelike like this comment: https://github.com/ethereum/ERCs/pull/673#discussion_r1825655032 ought to be in the text I think!

ERCS/erc-7786.md
norswap: That probably doesn't need to be enshrined in the interface I think.

The ticketing system can be implementation-specific, a detail that only the gateway knows about. Then someone (whether the gateway in active mode or the user in passive mode) needs to call `receiveMessage` with the details (making it effectively similar to what `setMessageExecuted` currently does).

ERCS/erc-7786.md
frangio: >If it's useless to the gateways, it should be in the payload.

I like this formulation! I actually think it's kind of implied by the fact that gateways have to support the attributes you're including, otherwise they'll reject the message. 

Regarding multicalls, I'd reframe it as "atomic multi-messages" or something like that. I'm a bit reluctant given that most protocols I've seen don't support this (I only recall LayerZero having this). Additionally it could technically be implemented as an attribute. Definitely something to consider though.

ERCS/erc-7786.md
frangio: Not sure I see the potential attack vector. The application isn't handling any signatures or transactions manually. The underlying protocols should ensure non-replayability.

ERCS/erc-7786.md
frangio: I think we inconsistently applied the change and @norswap suggested this change to make it consistent. Our last proposal was to use `executeMessage`, but it seems like in most places we still have `receiveMessage`. 😅

ERCS/erc-7786.md
frangio: > The ticketing system can be implementation-specific, a detail that only the gateway knows about. Then someone (whether the gateway in active mode or the user in passive mode) needs to call `receiveMessage` with the details

Unless I'm misunderstanding this is exactly what the current spec says.

`setMessageExecuted` (which was at some point called `validateReceivedMessage`) is needed for when the message is submitted via `receiveMessage` by an account that isn't the gateway, so the message needs to be validated with the gateway before the receiver accepts it.

ERCS/erc-7786.md
```diff
@@ -0,0 +1,212 @@
...
+}
+```
+
+A **Destination Gateway** is a contract that implements a protocol to validate messages sent on other chains. The gateway can operate in Active or Passive Mode, depending on whether `executeMessage` is invoked by the gateway itself or some other acocunt.
+
+#### Active Mode
+
+The gateway directly invokes `receiveMessage`, and only does so with valid messages. The receiver MUST assume that a message is valid if the caller is a known gateway.
```
Amxx, line 139:
```suggestion
The gateway directly invokes `executeMessage`, and only does so with valid messages. The receiver MUST assume that a message is valid if the caller is a known gateway.
```

ERCS/erc-7786.md
```diff
@@ -0,0 +1,212 @@
...
+
+The arguments `gateway` and `gatewayMessageKey` are unused in active mode and SHOULD be zero and empty respectively.
+
+#### Passive Mode
+
+The gateway does not directly invoke `receiveMessage`, but provides a means to validate messages. The receiver allows any party to invoke `receiveMessage`, but if the caller is not a known gateway it MUST check that the gateway provided as an argument is a known gateway, and it MUST validate the message against it before accepting it, forwarding the message key.
```
Amxx, line 145:
```suggestion
The gateway does not directly invoke `executeMessage`, but provides a means to validate messages. The receiver allows any party to invoke `executeMessage`, but if the caller is not a known gateway it MUST check that the gateway provided as an argument is a known gateway, and it MUST validate the message against it before accepting it, forwarding the message key.
```

ERCS/erc-7786.md
```diff
@@ -0,0 +1,212 @@
...
+
+MUST revert if the message is invalid or has already been executed.
+
+TBD: Passing full payload or payload hash (as done by Axelar). Same question for attributes, possibly different answer depending on attribute malleability.
+
+#### Dual Active-Passive Mode
+
+A gateway MAY operate in both active and passive modes, or it MAY switch from operating exclusively in active mode to passive mode or vice versa.
+
+A receiver SHOULD support both active and passive modes for any gateway. This is accomplished by first checking whether the caller of `receiveMessage` is a known gateway, and only validating the message if it is not; the first case supports an active mode gateway, while the second case supports a passive mode gateway.
```
Amxx, line 173:
```suggestion
A receiver SHOULD support both active and passive modes for any gateway. This is accomplished by first checking whether the caller of `executeMessage` is a known gateway, and only validating the message if it is not; the first case supports an active mode gateway, while the second case supports a passive mode gateway.
```

ERCS/erc-7786.md
```diff
@@ -0,0 +1,212 @@
...
+
+Some protocols gateway support doing arbitrary direct calls on the receiver. In that case, the receiver must detect that they are being called by the gateway to properly identify cross-chain messages. Getters are available on the gateway to figure out where the cross-chain message comes from (source chain and sender address). This approach has the downside that it allows anyone to trigger any call from the gateway to any contract. This is dangerous if the gateway ever holds any assets ([ERC-20](./erc-20.md) or similar). The use of a dedicated `receiveMessage` function on the receiver protects any assets or permissions held by the gateway against such attacks. If the ability to perform direct calls is desired, this can be implemented as a wrapper on top of any gateway that implements this ERC.
```
Amxx, line 198:
```suggestion
Some protocols gateway support doing arbitrary direct calls on the receiver. In that case, the receiver must detect that they are being called by the gateway to properly identify cross-chain messages. Getters are available on the gateway to figure out where the cross-chain message comes from (source chain and sender address). This approach has the downside that it allows anyone to trigger any call from the gateway to any contract. This is dangerous if the gateway ever holds any assets ([ERC-20](./erc-20.md) or similar). The use of a dedicated `executeMessage` function on the receiver protects any assets or permissions held by the gateway against such attacks. If the ability to perform direct calls is desired, this can be implemented as a wrapper on top of any gateway that implements this ERC.
```

ERCS/erc-7786.md
```diff
@@ -0,0 +1,212 @@
...
+
+#### Post-processing
+
+After a sender has invoked `sendMessage`, further action MAY be required by the gateways to make the message effective. This is called *post-processing*. For example, some payment is typically required to cover the gas of executing the message at the destination.
+
+The exact interface for any such action is out of scope of this ERC. If the `postProcessingOwner` attribute is supported and present, such actions MUST be restricted to the specified account, otherwise they MUST be able to be performed by any party in a way that MUST NOT be able to compromise the eventual receipt of the message.
+
+### Reception Procedure
+
+The underlying protocol MUST ensure delivery of the message to the **receiver**, which MUST implement `IERC7786Receiver`. The requirements for `executeMessage` will be detailed further below.
```
Amxx, line 120:
```suggestion
The underlying protocol MUST ensure delivery of the message to the **receiver**, which MUST implement `IERC7786Receiver`. The requirements for `executeMessage` will be detailed further below. Upon successful processing of the message, the receiver MUST return `IERC7786Receiver.executeMessage.selector` (`0xc887c5da`).
```

Amxx: @frangio

ERCS/erc-7786.md
```diff
@@ -0,0 +1,212 @@
...
+}
+```
+
+A **Destination Gateway** is a contract that implements a protocol to validate messages sent on other chains. The gateway can operate in Active or Passive Mode, depending on whether `executeMessage` is invoked by the gateway itself or some other acocunt.
+
+#### Active Mode
+
+The gateway directly invokes `executeMessage`, and only does so with valid messages. The receiver MUST assume that a message is valid if the caller is a known gateway.
+
+The arguments `gateway` and `gatewayMessageKey` are unused in active mode and SHOULD be zero and empty respectively.
```
Amxx, line 141:
```suggestion
The arguments `gateway` and `gatewayMessageKey` are unused in active mode and SHOULD be zero and empty respectively.

The gateway MUST verify that `executeMessage` returns the correct value, and MUST revert if that is not the case.
```

ERCS/erc-7786.md
```diff
@@ -0,0 +1,222 @@
...
+
+#### Dual Active-Passive Mode
+
+A gateway MAY operate in both active and passive modes, or it MAY switch from operating exclusively in active mode to passive mode or vice versa.
+
+A receiver SHOULD support both active and passive modes for any gateway. This is accomplished by first checking whether the caller of `executeMessage` is a known gateway, and only validating the message if it is not; the first case supports an active mode gateway, while the second case supports a passive mode gateway.
+
+### Properties
+
+The protocol underlying a pair of gateways is expected to guarantee a series of properties. For detailed definition and discussion we refer to XChain Research’s [Cross-chain Interoperability Report](https://github.com/0xFableOrg/xchain/blob/7789933bba24e2dd893cf515157af70474e7180b/README.md).
```
Amxx, line 187:
```suggestion
The protocol underlying a pair of gateways is expected to guarantee a series of properties. For detailed definition and discussion we refer to XChain Research’s *Cross-chain Interoperability Report*.
```

ERCS/erc-7786.md
frangio: Added some diagrams.

ERCS/erc-7786.md
norswap: Yeah but if they don't it's a security concern: e.g. minting multiple times.

ERCS/erc-7786.md
cjcobb23: yes I was talking about passive mode. But even in active mode, does EIP 6170 standardize the interface between the gateway and the dApp the message is sent to? What interface must the dApp implement to receive cross chain messages? Is the `message_` field just calldata, and is the original sender and source chain propagated to the dApp and if so how? I think these things are all important to spell out.

ERCS/erc-7786.md
cjcobb23: I am ok with removing passive mode from the standard, to keep things simple

frangio:
We have some simplifications coming for this proposal, Passive mode will be removed

frangio:
We're looking into simplifying this proposal by removing "Passive Mode".

ERCS/erc-7786.md
```diff
@@ -0,0 +1,215 @@
...
+
+An **Source Gateway** is a contract that offers a protocol to send a message to a receiver on another chain. It MUST implement `IERC7786GatewaySource`.
+
+```solidity
+interface IERC7786GatewaySource {
+    event MessagePosted(bytes32 indexed outboxId, string sender, string receiver, bytes payload, uint256 value, bytes[] attributes);
+
+    error UnsupportedAttribute(bytes4 signature);
+
+    function supportsAttribute(bytes4 signature) external view returns (bool);
```
Amxx, line 69:
```suggestion
    error UnsupportedAttribute(bytes4 selector);

    function supportsAttribute(bytes4 selector) external view returns (bool);
```

ERCS/erc-7786.md
```diff
@@ -0,0 +1,184 @@
...
+
+Delivery of a message sent from another chain.
+
+The receiver MUST validate that the caller of this function is a **known gateway**, i.e., one whose underlying cross-chain messaging protocol it trusts.
+
+MUST return `IERC7786Receiver.executeMessage.selector` (`0x675b049b`). 
+
+#### Interaction Diagram
+
+![](../assets/erc-7786/send-execute.svg)
```
Amxx, line 143: HTMLProfer fails because it cannot find that file.

@abcoathup can you help us with that part ?

frangio: Looking at other ERCs it sounds like this might work? Don't know if you've tried it.

```suggestion
![](../assets/eip-7786/send-execute.svg)
```

ERCS/erc-7786.md
```diff
@@ -0,0 +1,184 @@
+---
+eip: 7786
+title: Cross-Chain Messaging Gateway
+description: An interface for contracts to send and receive cross-chain messages.
```
SamWilsn, line 4: Bit of a nitpick, but it might be nice to include what a "message" is in your description.

ERCS/erc-7786.md
```diff
@@ -0,0 +1,184 @@
...
+discussions-to: https://ethereum-magicians.org/t/erc-7786-cross-chain-messaging-gateway/21374
+status: Draft
+type: Standards Track
+category: ERC
+created: 2024-10-14
+---
+
+## Abstract
+
+This proposal describes an interface, and the corresponding workflow, for smart contracts to send arbitrary data through cross-chain messaging protocols. The end goal of this proposal is to have all such messaging protocols accessible via this interface (natively or using "adapters") to improve their composability and interoperability. That would allow a new class of cross-chain native smart contracts to emerge while reducing vendor lock-in. This proposal is modular by design, allowing users to leverage bridge-specific features through attributes while providing simple "universal" access to the simple feature of "just getting a simple message through".
```
SamWilsn, line 15: Your abstract reads more like an introduction. An ERC's abstract should give readers a high level technical overview of how the proposal accomplishes its goals, not just the why and what.

ERCS/erc-7786.md
```diff
@@ -0,0 +1,184 @@
...
+
+## Motivation
+
+Cross-chain messaging protocols (or bridges) allow communication between smart contracts deployed on different blockchains. There is a large diversity of such protocols with multiple degrees of decentralization, different architectures, implementing different interfaces, and providing different guarantees to users.
+
+Because almost every protocol implements a different workflow using a specific interface, portability between bridges is currently basically impossible. This also prevents the development of generic contracts that rely on cross chain communication.
+
+The objective of the ERC is to provide a standard interface, and a corresponding workflow, for performing cross-chain communication between contracts. Existing cross-chain communication protocols that do not natively implement this interface should be able to adopt it using adapter gateway contracts.
```
SamWilsn, line 23:
```suggestion
The objective of this ERC is to provide a standard interface, and a corresponding workflow, for performing cross-chain communication between contracts. Existing cross-chain communication protocols that do not natively implement this interface should be able to adopt it using adapter gateway contracts.
```

ERCS/erc-7786.md
```diff
@@ -0,0 +1,184 @@
...
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
+
+### Message Field Encoding
+
+A cross-chain message consists of a sender, receiver, payload, and list of attributes.
+
+#### Sender & Receiver
+
+The sender account (in the source chain) and receiver account (in the destination chain) MUST be represented using CAIP-10 account identifiers. Note that these are ASCII-encoded strings.
```
SamWilsn, line 37: "CAIP-10" is an external resource, and should be linked. Give me a bit and I'll add it to the linter.

ERCS/erc-7786.md
```diff
@@ -0,0 +1,184 @@
...
+
+- The protocol MUST guarantee Safety: A message is delivered at the destination if and only if it was sent at the source. The delivery process must ensure a message is only delivered once the sending transaction is finalized, and not delivered more than once. Note that there can be multiple messages with identical parameters that must be delivered separately.
+- The protocol MUST guarantee Liveness: A sent message is delivered at the destination eventually, assuming Liveness and censorship-resistance of the source and destination chains.
+- The protocol SHOULD guarantee Timeliness: A sent message is delivered at the destination within a bounded delivery time, which should be documented.
+- The above properties SHOULD NOT rely on trust in some centralized actor. For example, safety should be guaranteed by some trustless mechanism such as a light client proof, or attestations by an open, decentralized validator set. Relaying should be decentralized or permissionless to ensure liveness; a centralized relayer can fail and thus halt the protocol.
+
+### Pending discussion
```
SamWilsn, line 154: Please put this section in HTML-style comments. We don't want these draft sections making it into the final document, and the linter will enforce that.

ERCS/erc-7786.md
```diff
@@ -0,0 +1,184 @@
...
+
+## Backwards Compatibility
+
+Existing cross-chain messaging protocols implement proprietary interfaces. We recommend that protocols natively implement the standard interface defined here, and propose the development of standard adapters for those that don't.
+
+## Security Considerations
+
+Unfortunately, CAIP-2 and CAIP-10 names are not unique. Using non-canonical strings may lead to undefined behavior, including message delivery failure and locked assets. While source gateways have a role to play in checking that user input are valid, we also believe that more effort should be put into standardizing and documenting what the canonical format is for each CAIP-2 namespace. This effort is beyond the scope of this ERC.
+
+Needs discussion.
```
SamWilsn, line 180: HTML comment please!

SamWilsn (approved):
Please work on your description and abstract while in draft. Otherwise, this looks good to merge!

PR #679, opened 2024-10-21
Update ERC-6123: EIP-6123, align EIP with local development

Julius278:
added missing PaymentTriggered event


Julius278:
@cfries @pekola 
could you please check the changes and if the PR can be merged

Best regards,
Julius

cfries:
I need to have a look, because I thought that we discussed about theses events and I think the outcome would be that it is handled by the settlement token. 

I will check now.

cfries:
*I believe the change to ISDC is not needed / superfluous.*

**Remarks on `TradeSettlementPhase`:**

We started distiguishing events on the Trade and events on the Settlement. For an SDC that has multiple trades all trade may / should settle in a single settlement. So the name "TradeSettlementPhase" is wrong. It is a settlement phase. Hence the event should start with Settlement.

The 

```

event SettlementRequested(address initiator, string tradeData, string lastSettlementData);

```

marks the beginning of the settlement phase (from the perspective of the SDC - maybe the token has a finer granularity). So this should replace `TradeSettlementPhase`.

**Remarks on `TradeSettled `:**

There are already two events that mark the end of the settlement phase:

```
    /**
     * @dev Emitted when settlement process has been finished
     */
    event SettlementTransferred(string transactionData);

    /**
     * @dev Emitted when settlement process has been finished
     */
    event SettlementFailed(string transactionData);
```

Julius278:
Hi @cfries,
there is a difference between SettlementRequested, which marks the request to a service which will provide the new settlement data, and TradeSettlementPhase, which is emitted during "performSettlement".

TradeSettled is now more granular, that looks good, I will remove TradeSettled

cfries:
You are right. `performSettlement` is a different thing.

But there is already an event issued from performSettlement. The documentation in the comment already states that is is emitted when the settlement phase begins:

´``
 /**
     * @dev Emitted when Settlement has been valued and settlement phase is initiated
     * @param initiator the address of the requesting party
     * @param settlementAmount the settlement amount. If settlementAmount > 0 then receivingParty receives this amount from other party. If settlementAmount < 0 then other party receives -settlementAmount from receivingParty.
     * @param settlementData. the tripple (product, previousSettlementData, settlementData) determines the settlementAmount.
     */
    event SettlementEvaluated(address initiator, int256 settlementAmount, string settlementData);
```

The method `performSettlement` states that this event is emitted
```
    /**
     * @notice Called to trigger according settlement on chain-balances callback for initiateSettlement() event handler
     * @dev perform settlement checks, may initiate transfers and emits {SettlementEvaluated}
     * @param settlementAmount the settlement amount. If settlementAmount > 0 then receivingParty receives this amount from other party. If settlementAmount < 0 then other party receives -settlementAmount from receivingParty.
     * @param settlementData. the tripple (product, previousSettlementData, settlementData) determines the settlementAmount.
     */
    function performSettlement(int256 settlementAmount, string memory settlementData) external;
```

The name starts with "Settlement" due to the remark I made above. Settlement is indepent of Trade. I believe I made the renaming comments somewhere, but it was a time ago. Maybe in Teams? :-) 
Also the documentation has clear subsections: "Trade Events" and "Settlement Events". All the events in "Settlement Events" start with "Settlement".

**Comment by Julius278** [comment]
okay, SettlementEvaluated is also fine for me and has some more details.
Especially initiator and settlementData is nice cause you dont need to retrieve those information manually after receiving the event.

But I cant find any "PaymentTriggered" event currently and we will need it on one of the token interfaces cause it could/will be necessary on more than one possible token implemenation

**Comment by pekola** [comment]
Hi. @Julius278 - Why do we need to specify a separate interface for the offchain variant. One could include the PaymentTriggered Event into the ERC20TokenSettlement. Not sure whether we need a separate function "initParty".

@cfries - this might be a bit subtle but stumbling across the settlement event names: SettlementRequested, SettlementEvaluated, SettlementTransferred. Since we removed the "Valuation" out of the SettlementRequested, we could generalise the second event naming also to be even more generic: SettlementEvaluated --> SettlementDetermined.

**Comment by Julius278** [comment]
Hi @pekola,
I removed the offchain interface.

"initParty" is not directly required in case of onchain settlement but it can be used as an "who is this party / address" description.
If this is not desired, I can remove it

**Comment by cfries** [comment]
Can we make this PR draft to avoid risking merge. It is not required that the ERC staff looks at this before we are confident.

**Comment by Julius278** [comment]
> Can we make this PR draft to avoid risking merge. It is not required that the ERC staff looks at this before we are confident.

Hi @cfries, is there anything left to discuss after our call this morning? I guess we identified that we only need those two additions to the token interface (event PaymentTriggered and function initSDC).

**Comment by cfries** [comment]
The interface may need some changes.
1) In the even some parameters have underscores, others not. Why? Also what are these parameters? Hash?
2) In the implementation there is a setter

```
    function setSDCAddress(address _sdcAddress) public{
        sdcAddress = _sdcAddress;
    }

```

but the interface specifies "initSDC". I would prefer the setter.

I bleieve I cannot modify this PR here, I could try fork yours.

Julius278:
Hi Christian,

from my point of view setSDCAddress sounds like its a 1-to-1 relationship and I wouldnt specify it that way.
initSDC sounds a bit more general and opens options for 1-to-1 or 1-to-many relations.

Yeah, you probably cant edit it cause its from my fork, maybe try to fork that one so the change doesnt get lost, then I will close that PR.
I will put it into draft state now

PR #686, opened 2024-10-26
Add ERC: Tap to Pay

amhed:
See https://hackmd.io/VyxIMlk1SvCOpBpS6a_2uA?both for initial commentary on the standard


ERCS/erc-_____.md
```diff
@@ -0,0 +1,342 @@
...
+## Abstract
+This ERC defines a standard for contactless payment transactions that can allow for interoperable customer to merchant onchain payments, regardless of the wallets the customer and merchant are using.
+
+## Motivation
+Currently there is no standard mechanism in crypto to do contactless payment transactions via NFC. This ERC defines a standardized way of exchanging payment information between merchants and customers, or peer-to-peer individuals, so that efficient checkout can occur.
+
+## Specification
+
+Comprised of three parts:
+- A new web3Provider method called `requestContactlessPayment`
```
wilsoncusack, line 23: Is web3Provider an 1193 provider or something else? Also I think these generally have prefixes like eth_ or wallet_

ERCS/erc-_____.md
```diff
@@ -0,0 +1,342 @@
...
+
+
+### Use Cases
+
+#### Use Case 1: Customer purchases from a merchant
+1. A customer approaches a merchant and orders an item at their cash register (e.g. a cup of coffee).
+1. The merchant is running an app on their point-of-sale (PoS) system that allows initiating a checkout flow. 
+1. When the customer is done ordering, the merchant goes through the checkout flow up and triggers a payment request. 
+1. The merchant app calls the `requestContactlessPayment` RPC method with the expected payload. 
+1. The merchant's device emits an NFC signal via host card emulation ([HCE](https://developer.android.com/develop/connectivity/nfc/hce)) for the customer’s device to read.
```
wilsoncusack, line 35: Is this done via the Ethereum provider receiving the above call? So is it assumed the provider needs to be running on device?

ERCS/erc-_____.md
```diff
@@ -0,0 +1,342 @@
...
+1. The customer, running their wallet software of choice, reads the NFC signal. The customer’s wallet constructs the necessary transaction(s) to execute the onchain checkout.
```
wilsoncusack, line 36: Sorry I don't know much about this, why couldn't we just do a normal eth_sendTransaction or wallet_sendCalls via an NFC channel?

ERCS/erc-_____.md
```diff
@@ -0,0 +1,342 @@
...
+1. The merchant is running an app on their point-of-sale (PoS) system that allows initiating a checkout flow. 
+1. When the customer is done ordering, the merchant goes through the checkout flow up and triggers a payment request. 
+1. The merchant app calls the `requestContactlessPayment` RPC method with the expected payload. 
+1. The merchant's device emits an NFC signal via host card emulation ([HCE](https://developer.android.com/develop/connectivity/nfc/hce)) for the customer’s device to read.
+1. The customer, running their wallet software of choice, reads the NFC signal. The customer’s wallet constructs the necessary transaction(s) to execute the onchain checkout. 
+1. The wallet then signs and submits all transactions.
+1. Upon successful onchain execution of these transactions, the checkout is complete.
+
+#### Use Case 2: Peer-to-peer Sends
+1. Alice requests a payment on her wallet app by emitting an NFC signal. The contents of the signal is an EIP-681 URI
```
wilsoncusack, line 41: I think 681 is possibly over specified on things like gas values. Wonder if we could align on something more minimal.

ERCS/erc-_____.md
```diff
@@ -0,0 +1,342 @@
...
+### requestContactlessPayment
+The requestContactlessPayment method will be added as a standard method to the [EIP-1193 Ethereum Provider JavaScript API](https://eips.ethereum.org/EIPS/eip-1193). 
+
+This method takes in two parameters: `type` and `uri`.
+
+- `type` is an enum that represents the payload type:
+    - 1 means the URI is an EIP-681-compliant payload.
+    - 2 means the URI points to an HTTP relayer, where the response of the relayer is either a JSON object with the transaction data already encoded as a 0x string, or a message for the customer to sign. The customer wallet is responsible for constructing, signing and submitting the transaction/message.
+
+- `uri` represents either the EIP-681-compliant payload or the relayer endpoint to query for the JSON payload.
```
wilsoncusack, line 58: I feel like 681 URI is not widely adopted and is mostly useful when you can share a string with someone and not a whole payload. 

But here we control the channel so I'm again wondering why not just use existing wallet RPCs to describe what we want.

ERCS/erc-_____.md
amhed: Yup, it's the Ethereum Provider JavaScript API. Will update

ERCS/erc-_____.md
amhed: Yes

ERCS/erc-_____.md
amhed: We had some sluggish performance transmitting the NFC message with larger payloads and decided to use a relayer for a shorter URI.

ERCS/erc-_____.md
amhed: We're already using 681 heavily on send/receive flows on CB Wallet. We tried leaning into existing standards. What would you suggest could be better?

ERCS/erc-_____.md
amhed: You're right, it's not widely adopted. Open to any other suggestions 🙏

amhed:
Thanks for the quick review @wilsoncusack ❤️

ERCS/erc-_____.md
```diff
@@ -0,0 +1,342 @@
+---
+eip: ____
+title: "Contactless Crypto Payment Standard"
```
abcoathup, line 3:
```suggestion
title: Contactless Payment
```
Standard is superfluous & so is crypto for an ERC.

ERCS/erc-_____.md
```diff
@@ -0,0 +1,342 @@
...
+description: Standard for initializing contactless payment transactions from EVM wallets
+author: Amhed Herrera, Justin Lee, Arjun Dureja
+discussions-to:
```
abcoathup, line 6: Please create a discussions topic in Eth Magicians with a link to this PR.

amhed: Will do! I'm waiting for the DiscoBot to get me posting privileges. 

This is my user (tied to github username)
https://ethereum-magicians.org/u/amhed/summary

ERCS/erc-_____.md
```diff
@@ -0,0 +1,342 @@
+---
+eip: ____
```
abcoathup, line 2:
```suggestion
eip: 7798
```

ERCS/erc-_____.md
amhed: Done, thanks again
https://ethereum-magicians.org/t/erc-7798-contactless-payment/21501

ERCS/erc-_____.md
lukasrosario: you could still use a relayer by using wallet_sendCalls + something like a `nfcRelay` capability

ERCS/erc-7798.md
```diff
@@ -0,0 +1,361 @@
...
+  method: 'requestContactlessPayment', 
+  params: { 
+    type: 2,
+    uri: "https://nfcrelay.xyz/paymentTxParams?uuid=1234-abcd-56678",
+    verificationCode: "1234567890"
+}
+});
+```
+
+### Type 1
```
lukasrosario, line 95: are offchain signatures supported for type 1? i dont think it's covered by 681?

ERCS/erc-7798.md
```diff
@@ -0,0 +1,361 @@
...
+  }, 
+  rpcProxySubmissionParams?: {
+    submissionUrl?: string; // optional endpoint to submit the tx hash to
+  },
+  dappUrl?: string;
+  dappName?: string;
+}
+```
+
+**[EIP-712](https://eips.ethereum.org/EIPS/eip-712) Message (offline signature)**
```
lukasrosario, line 180: hm i guess sendCalls wouldnt work for getting offchain signatures

lukasrosario:
overall i agree with Wilson's feedback -- feel like it would be cleanest if we could just transmit the usual RPC request over the NFC channel. would be easier for devs to reason about and i think it would be easier to adopt than extending the 1193 provider.

seems like there were issues transmitting larger messages through nfc, and it makes me wonder if that should just be an issue for the provider to resolve if the message is to be transmitted over NFC. that way the app dev can use the same interface as usual, and then the wallet maintainers resolve how to compress & transmit to wallet.

amhed:
> seems like there were issues transmitting larger messages through nfc, and it makes me wonder if that should just be an issue for the provider to resolve if the message is to be transmitted over NFC. that way the app dev can use the same interface as usual, and then the wallet maintainers resolve how to compress & transmit to wallet.

When you mention the same interface you mend `wallet_sendCalls`? And send URIs instead of hashed strings representing txs?

ERCS/erc-7798.md
amhed: No, we covered them with the relayed URI. You can basically pass down any long-form payload / set of batched transactions for EIP-3009 or permit2

ERCS/erc-7798.md
```diff
@@ -0,0 +1,200 @@
...
+discussions-to: https://ethereum-magicians.org/t/erc-7798-contactless-payment/21501
+status: Draft
+type: Standards Track
+category: ERC
+created: 2024-10-25
+requires: 20, 681, 712
+---
+
+## Abstract
+This ERC defines a standard for contactless payment transactions that can allow for interoperable customer to merchant onchain payments, regardless of the wallets the customer and merchant are using.
```
SamWilsn, line 15: The abstract should contain enough detail to give the reader a high-level (but still technical) overview of _how_ your proposal accomplishes its goals.

ERCS/erc-7798.md
```diff
@@ -0,0 +1,200 @@
...
+
+## Specification
+
+Comprised of three parts:
+- A new Ethereum Provider JavaScript API method called `requestContactlessPayment`
+- An agreement on the payload for data exchange between the parties
+- An optional mechanism for relaying large JSON payloads using a backend relayer
+
+
+```mermaid
```
SamWilsn, line 28: Our renderer doesn't support mermaid. Please replace it with an image in your assets directory (preferably an SVG.)

ERCS/erc-7798.md
```diff
@@ -0,0 +1,200 @@
...
+1. When the customer is done ordering, the merchant goes through the checkout flow up and triggers a payment request. 
+1. The merchant app calls the `requestContactlessPayment` RPC method with the expected payload. 
+1. The merchant's device emits an NFC signal via host card emulation (HCE) for the customer’s device to read.
+1. The customer, running their wallet software of choice, reads the NFC signal. The customer’s wallet constructs the necessary transaction(s) to execute the onchain checkout. 
+1. The wallet then signs and submits all transactions.
+1. Upon successful onchain execution of these transactions, the checkout is complete.
+
+#### Use Case 2: Peer-to-peer Sends
+1. Alice requests a payment on her wallet app by emitting an NFC signal. The contents of the signal is an [ERC-681](erc-681.md) URI
+> []will this only work with 681? 681 has limitations, like receiver intent which would be great to solve too.
```
SamWilsn, line 57:
```suggestion
<!-- will this only work with 681? 681 has limitations, like receiver intent which would be great to solve too. -->
```

If you use HTML-style comments, the linter will make sure you replace them before going into Review.

ERCS/erc-7798.md
```diff
@@ -0,0 +1,200 @@
...
+
+1. [ERC-681](erc-681.md) URI Bob reads the NFC signal from Alice on his wallet app and constructs the transaction
+2. Bob holds his phone close to Alice's which triggers the NFC flow.
+1. Bob then signs and submits the transaction onchain.
+1. Send flow is complete.
+
+### requestContactlessPayment
+The requestContactlessPayment method will be added as a standard method to the [EIP-1193 Ethereum Provider JavaScript API](eip-1193). 
+
+This method takes in two parameters: `type` and `uri`.
```
SamWilsn, line 67: I usually recommend including a [JSON Schema](https://json-schema.org/) for any new API endpoints.

ERCS/erc-7798.md
```diff
@@ -0,0 +1,200 @@
...
+#### Use Case 2: Peer-to-peer Sends
+1. Alice requests a payment on her wallet app by emitting an NFC signal. The contents of the signal is an [ERC-681](erc-681.md) URI
+> []will this only work with 681? 681 has limitations, like receiver intent which would be great to solve too.
+
+1. [ERC-681](erc-681.md) URI Bob reads the NFC signal from Alice on his wallet app and constructs the transaction
+2. Bob holds his phone close to Alice's which triggers the NFC flow.
+1. Bob then signs and submits the transaction onchain.
+1. Send flow is complete.
+
+### requestContactlessPayment
```
chris13524, line 64: Shouldn't this have some prefix e.g. `wallet_`?

```suggestion
### wallet_requestContactlessPayment
```

ERCS/erc-7798.md
```diff
@@ -0,0 +1,200 @@
...
+    type: 1,
+    uri: "ethereum:0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913@8453/transfer?address=0xf7d07Ee99095FDC09001710Ec232162a788BB989&uint256=1e5",
+});
+
+window.ethereum.request({ 
+  method: 'requestContactlessPayment', 
+  params: { 
+    type: 2,
+    uri: "https://nfcrelay.xyz/paymentTxParams?uuid=1234-abcd-56678",
+    verificationCode: "1234567890"
```
chris13524, line 90: What is the purpose of `verificationCode` and why can't its value be included in the `uri`?

ERCS/erc-7798.md
```diff
@@ -0,0 +1,200 @@
...
+**Regular Send**
+```ts
+{
+  payloadType: 'eip681';
+  chainId: string;
+  contractAddress: string;
+  toAddress: string;
+  value: string;
+  dappUrl?: string;
+  dappName?: string;
```
chris13524, line 119: This is a phishing vector. Recommend not including it at all.

ERCS/erc-7798.md
```diff
@@ -0,0 +1,200 @@
...
+1. The merchant dapp makes a POST request to an NFC relayer endpoint with the information needed for completing a sale. 
+1. The merchant’s device then transmits the NFC relayer endpoint URI and verification code to the customer via any means
+1. The customer’s device then makes a GET request to the endpoint URI
+
+One of three payloads can be returned from the NFC relayer:
+
+**Regular Send**
+```ts
+{
+  payloadType: 'eip681';
```
chris13524, line 113: What makes this a eip681 payload? The fields don't seem to map to EIP-681 very well.

ERCS/erc-7798.md
```diff
@@ -0,0 +1,200 @@
...
+
+- `uri` represents either the ERC-681-compliant payload or the relayer endpoint to query for the JSON payload. 
+
+Example calls:
+
+```js 
+window.ethereum.request({ 
+  method: 'requestContactlessPayment', 
+  params: { 
+    type: 1,
```
chris13524, line 81: Suggest using strings for type instead of numbers. Specifically, type here should be `erc681` and then for the relay perhaps a separate ERC could be defined for the relay design. This allows easier future expansion to more types of in-NFC payloads and relayers.

ERCS/erc-7798.md
```diff
@@ -0,0 +1,200 @@
...
+    },
+  },
+  dappUrl?: string;
+  dappName?: string;
+  additionalPayload?: {
+    // ... any fields can go here, this is additional payload the RPC proxy may need but does not have to bee signed in the message
+  }
+}
+```
+Upon submitting the transaction for the [ERC-681](erc-681.md) and contract call cases, if `rpcProxySubmissionParams` is present, then the transaction hash can be optionally submitted via a POST request to the submission URL. The body of the POST would have the following structure:
```
chris13524, line 179: Optionally? Why not required?

ERCS/erc-7798.md
```diff
@@ -0,0 +1,200 @@
...
+
+### Type 2
+
+Handles cases where the payload necessary to construct, sign, and submit the transaction needed for payment is too large to emit over NFC. The intended flow is as follows: 
+
+1. The merchant dapp makes a POST request to an NFC relayer endpoint with the information needed for completing a sale. 
+1. The merchant’s device then transmits the NFC relayer endpoint URI and verification code to the customer via any means
+1. The customer’s device then makes a GET request to the endpoint URI
+
+One of three payloads can be returned from the NFC relayer:
```
chris13524, line 108: What about a payload type where any arbitrary wallet RPC call can be included? Whether that's `eth_sendTransaction` or more recent specs such as `wallet_sendCalls` or `wallet_grantPermissions`.

ganchoradkov:
Do you think expected error handling should be also included in the specs?
There are many ways something can go wrong before the checkout flow is completed e.g.
 - wallet not supporting `erc-7798`
 - transaction rejected by a node on submit e.g. not enough funds
 - user manually declining the transaction in the wallet
 - NFC relayer endpoint unreachable on either device

ERCS/erc-7798.md
ganchoradkov: I guess its a token/key that the wallet can use to verify the transaction received by the relayer endpoint is the same as the one send by the merchant

ERCS/erc-7798.md
```diff
@@ -0,0 +1,200 @@
...
+}
+```
+
+## Rationale
+
+The reason for having a relayer URI to pass data is due to limitations in the size of the data that can be transmitted wirelessly between devices directly. For example, QR codes have a maximum character size of 7,089 characters currently, but a user can attempt to buy an indefinite number of items in a single checkout transaction. By having a relayer URI, we can theoretically pass as much calldata as we desire to the customer’s wallet, as long as they fetch from the URI.
+
+[ERC-681](erc-681.md) URIs were also permitted to enable possible direct peer to peer payments. It’s more likely that they’ll be occurring between two end user wallets rather than between a customer and a merchant, but it would be good to include in the event the dapp only requires a simple send transaction to execute the checkout.
+
+## Security Considerations
```
llbartekll, line 192: NFC isn’t end-to-end encrypted, so I think there’s a potential vulnerability to a relay attack. An adversary intercepts and forwards communication between a user and a merchant, basically extending the NFC range. To mitigate this, wallets should have security precautions like transaction simulation, a trusted merchants list, or spending limits.

JamesSmartCell: > NFC isn’t end-to-end encrypted, so I think there’s a potential vulnerability to a relay attack. An adversary intercepts and forwards communication between a user and a merchant, basically extending the NFC range. To mitigate this, wallets should have security precautions like transaction simulation, a trusted merchants list, or spending limits.

NFC range is very limited; to clarify is the vulnerability you refer to changing the ERC-681 destination and physically inserting yourself in front of user before they tap?

Agree that wallet used in this mode should have at minimum a spending limit; trusted merchant list is problematic for most real world applications in that it may be a pre-calculated empty address for settlement.

In the real world NFC is very limited as if you're supporting iOS (which you will be) the NFC link is abstracted.

PR #687, opened 2024-10-26
Update ERC-7739: Update ERC-7739

Vectorized:

 - The build passes.

### DIFF
--- a/ERCS/erc-7739.md
+++ b/ERCS/erc-7739.md
@@ -130,7 +130,7 @@ typedDataSignTypehash =
                   "string version,",
                   "uint256 chainId,",
                   "address verifyingContract,",
-                  "bytes32 salt,"
+                  "bytes32 salt"
               ")",
               contentsType
           )
@@ -271,9 +271,9 @@ hash =
 
 The `PersonalSign` workflow does not require additional data to be appended to the `signature` passed into `isValidSignature`.
 
-### ERC-7739 support detection
+### Support detection
 
-Smart accounts SHOULD return `bytes4(0x77390001)` for `isValidSignature(0x7739773977397739773977397739773977397739773977397739773977397739, "Any Signature")`.
+Smart accounts SHOULD return `bytes4(0x77390001)` for `isValidSignature(0x7739773977397739773977397739773977397739773977397739773977397739, "Any Signature")` to indicate support for this standard.
 
 ### Signature verification workflow deduction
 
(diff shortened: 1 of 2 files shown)


ERCS/erc-7739.md
```diff
@@ -322,37 +315,45 @@ As many developers may not update their applications to support the nested EIP-7
 
 The `typedDataSignTypehash` must be constructed on-the-fly on-chain. This is to enforce that the signed contents will be visible in the signature request, by requiring that `contents` be a user defined type. 
 
-The structure is intentionally made flat with the fields of `eip712Domain` to make implementation feasible. Otherwise, smart accounts must implement on-chain lexographical sorting of strings for the struct type names when constructing `typedDataSignTypehash`.
+The fields of `eip712Domain` are flattened into the `TypedDataSign` structure to enable compatibility with certain Ethers JS versions.
```
frangio, line 318: Can you explain this?

frangio: Ok I understand.

```suggestion
The fields of `eip712Domain` are flattened into the `TypedDataSign` structure instead of being included as a field of type `EIP712Domain`, because the type could be different from and conflict with the domain type of the verifying contract.
```

ERCS/erc-7739.md
```diff
@@ -268,21 +271,11 @@ hash =
 
 The `PersonalSign` workflow does not require additional data to be appended to the `signature` passed into `isValidSignature`.
 
-### `supportsNestedTypedDataSign` function for detection
+### Support detection
 
-To facilitate automatic detection, smart accounts SHOULD implement the following function:
+Smart accounts SHOULD return `bytes4(0x77390001)` for `isValidSignature(0x7739773977397739773977397739773977397739773977397739773977397739, "Any Signature")` to indicate support for this standard.
```
frangio, line 276: I'd consider requiring the signature argument to be the empty string instead of accepting anything. Just to reduce the surface where `isValidSignature` behaves different than expected.

```suggestion
Smart accounts SHOULD return `bytes4(0x77390001)` for `isValidSignature(0x7739773977397739773977397739773977397739773977397739773977397739, "")` to indicate support for this standard.
```

faldrin084 (approved):

ERCS/erc-7739.md
```diff
@@ -322,37 +315,45 @@ As many developers may not update their applications to support the nested EIP-7
...
+### Rejecting `contentsName` beginning with any lowercase 7-bit ASCII character
+
+This recommendation is to keep the standard language agnostic and future-proof. Atomic types such as `uint256` may be named differently in other languages (e.g. `u256`).
 
 ## Backwards Compatibility
 
-No backwards compatibility issues.
+### Detection of previous draft
+
+In an earlier draft, we have designated a `supportsNestedTypedDataSign()` function for support detection, which returns `bytes4(0xd620c85a)`.
```
filmakarov, line 338: Can you please add something like:

"To ensure false negative detections, it is RECOMMENDED that dApps use `supportsNestedTypedDataSign()` as a fallback detection method in addition to the one described earlier in the spec."

to ensure dApps consider using `supportsNestedTypedDataSign()` and not just ignore it

faldrin084 (approved):

ERCS/erc-7739.md
```diff
@@ -322,37 +315,45 @@ As many developers may not update their applications to support the nested EIP-7
...
 
-### `supportsNestedTypedDataSign` for detection
+The `bytes1 fields` bitmap and `uint256[] extensions` array in [ERC-5267](./eip-5267.md) have been omitted. Differentiating between an absent field versus a zero field (e.g. `bytes32(0)`) offers no additional security benefits for on-chain defensive rehashing. The `extensions` parameter is a list of EIP numbers used for off-chain signaling.
 
-Without this function, this standard will not change the interface of the smart account, as it defines the behavior of `isValidSignature` without adding any new functions. As such, [ERC-165](./eip-165.md) cannot be used.
+### `contentsDescription` with implicit and explicit modes
 
-For future extendability, `supportsNestedTypedDataSign` is defined to return a bytes32 as the first word of its returned data. For bytecode compactness and to leave space for bit packing, only the leftmost 4 bytes are set to the function selector of `supportsNestedTypedDataSign`.
+When the `contents` structure contains nested types, EIP-712 lexicographical sorting can result in the `contentsName` not being positioned exactly at the start of the `contentsType`. As such, we need the explicit mode.
```
ernestognw, line 324: I understand the rationale behind these two modes but I feel the ERC should provide a general guideline for when to use each other. So far I've found the main benefit of using implicit mode when possible is that it reduces the calldata costs. If that's the case, I'd suggest to include it in this section

```suggestion
When the `contents` structure contains nested types, EIP-712 lexicographical sorting can result in the `contentsName` not being positioned exactly at the start of the `contentsType`. As such, we need the explicit mode.

Signing UIs may use implicit mode whenever possible to reduce the calldata size. Although the `explicit` mode works as a generalized approach.
```

PR #689, opened 2024-10-28
Update ERC-7683: Extending the standard

mrice32:
This PR proposes a few updates to ERC-7683. These updates are intended to make the language and functionality in the ERC capture more use cases than originally foreseen, require less (or no) protocol-specific code and casework for fillers, and provide visibility into the opaque bytes fields used in the standard.

See this [PR](https://github.com/across-protocol/ERCs/pull/1) for wider discussion of these changes.

Notable changes:

1. Terminology: initiate is now called open. swap has been replaced with transfer throughout.
1. The ERC now standardizes the fill function and process (i.e. the function on the destination chain).
1. Because protocols may require information or state from the origin transaction (open), the standard now requires that the resolved order includes FillInstructions, which tells the filler what data needs to be included in which destination transactions. The resolved order is now returned by open and emitted in an Open event.
1. The ERC now includes an onchain method as opposed to just a gasless one. This ensures that the standard can support both private mempool (gasless, usually) and public mempool order submission mechanisms.
1. The ERC now provides (currently stub) repos for sharing orderData subtypes and fillerData subtypes.
1. The ERC now explicitly supports flows where the fill happens before open.
1. Add a type parameter to identify the order subtypes.

### DIFF
--- a/ERCS/erc-7683.md
+++ b/ERCS/erc-7683.md
@@ -12,7 +12,7 @@ created: 2024-04-11
 
 ## Abstract
 
-The following standard allows for the implementation of a standard API for cross-chain trade execution systems. This standard provides a generic `CrossChainOrder` struct, as well as a standard `ISettlementContract` smart contract interface.
+The following standard allows for the implementation of a standard API for cross-chain value-transfer systems. This standard provides generic order structs, as well as a standard set of settlement smart contract interfaces.
 
 ## Motivation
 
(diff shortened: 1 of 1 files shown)


ERCS/erc-7683.md
```diff
@@ -24,153 +24,344 @@ By implementing a standard, cross-chain intents systems can interoperate and sha
...
-	uint32 chainId;
+	uint256 chainId;
 }
 
+/// @title FillInstruction type
+/// @notice Instructions to parameterize each leg of the fill
+/// @dev Provides all the origin-generated information required to produce a valid fill leg
+struct FillInstruction {
+	/// @dev The contract address that the order is meant to be settled by
+	uint64 destinationChainId;
```
nicholaspai, line 136: should this be uint256? @mrice32

PR #89, opened 2023-11-03
Add ERC: Sovereign Bridged Token

just-a-node:
This PR follows the [EIP -> ERC migration guide](https://github.com/ethereum/ERCs/issues/8) to migrate the [original PR for EIP-7281](https://github.com/ethereum/EIPs/pull/7281).

PR #692, opened 2024-10-30
Add ERC: Crosschain Token Interface

0xParticle:
This ERC introduces a minimal interface for tokens to communicate cross-chain. It allows bridges with mint and burn rights to send and relay token transfers with a standardized API.


ERCS/erc-0000.md
```diff
@@ -0,0 +1,176 @@
...
+
+MUST trigger when `crosschainMint` is successfully called.
+
+```solidity
+event CrosschainMint(address indexed _to, uint256 _amount)
+```
+
+Note: implementations might consider additionally emitting `Transfer(address(0), _to, _amount)`to be compliant with [ERC-5679](./eip-5679.md).
+
+**`CrosschainBurnt`**
```
blmalone, line 67:
```suggestion
**`CrosschainBurn`**
```

Can this be `CrosschainBurn` or change the `CrosschainMint` to `CrosschainMinted`? Whatever we land on the tense of the verb should be the same I think.

ERCS/erc-0000.md
```diff
@@ -0,0 +1,176 @@
+---
+eip: 0000
```
abcoathup, line 2:
```suggestion
eip: 7802
```

ERCS/erc-0000.md
```diff
@@ -0,0 +1,176 @@
...
+title: Crosschain Token Interface
+description: Minimal token interface for cross-chain transfers
+author: skeletor <skeletor@defi.sucks>, parti (@0xParticle), joxes <joxes@defi.sucks>, ng <ng@defi.sucks>, agus duha <agusduha@defi.sucks>, disco <0xdiscotech@defi.sucks>, gotzen <gotzen@defi.sucks>, 0age <0age@uniswap.org>, Mark Tyneway <mark@oplabs.co>, Zain Bacchus <zain@oplabs.co>, Matt Solomon <msolomon@oplabs.co>, Maurelian <maurelian@protonmail.ch>
+discussions-to: https://ethereum-magicians.org/t/erc-xxxx-crosschain-token-interface/21508
```
abcoathup, line 6:
```suggestion
discussions-to: https://ethereum-magicians.org/t/erc-7802-crosschain-token-interface/21508
```
Updated title on Eth Magicians with assigned number

ERCS/erc-0000.md
```diff
@@ -0,0 +1,176 @@
...
+```
+
+**`crosschainBurn`**
+
+Burns `_amount` of token from address `_account`.
+
+This function works as the burning entry point for bridge contracts. Each implementation is responsible for its access control logic.
+
+```solidity
+crosschainBurn(address _account, uint256 _amount)
```
blmalone, line 52:
```suggestion
function crosschainBurn(address _account, uint256 _amount) external;
```

Adding full function signature.

ERCS/erc-0000.md
```diff
@@ -0,0 +1,176 @@
...
+### Methods
+
+**`crosschainMint`**
+
+Mints `_amount` of token to address `_account`. 
+
+This function works as the minting entry point for bridge contracts. Each implementation is responsible for its access control logic.
+
+```solidity
+crosschainMint(address _account, uint256 _amount)
```
blmalone, line 42:
```suggestion
function crosschainMint(address _account, uint256 _amount) external;
```
Adding full function signature.

ERCS/erc-0000.md
```diff
@@ -0,0 +1,176 @@
...
+discussions-to: https://ethereum-magicians.org/t/erc-xxxx-crosschain-token-interface/21508
+status: Draft
+type: Standards Track
+category: ERC
+created: 2024-10-30
+---
+
+## Abstract
+
+The ERC-xxx standard introduces a minimal interface for tokens to communicate cross-chain. It allows bridges with mint and burn rights to send and relay token transfers with a standardized API.
```
blmalone, line 15:
```suggestion
The ERC-7802 standard introduces a minimal interface for tokens to communicate cross-chain. It allows bridges with mint and burn rights to send and relay token transfers with a standardized API.
```

ERCS/erc-0000.md
```diff
@@ -0,0 +1,176 @@
...
+
+This interface fragmentation is a massive issue in cross-chain communication among chains via third-party bridges or future canonical solutions. At this point, it is clear that every bridge would benefit from a standardized interface for minted/burnt tokens. 
+
+There have been different attempts in the past to standardize token-bridging interfaces. However, third-party providers are also developing cross-chain token frameworks, such as NTT by Wormhole and OFT by LayerZero. Each framework defines its features, like rate limits and fee switches, and implements its mint and burn versions. The resultant interfaces become highly specific, lacking naming conventions and structures.
+
+The proposed interface includes the most relevant and minimal set of actions used by most of these standards. These actions also do not require any governance or owner participation, in contrast, for instance, to set rate limits. xERC-20, OFT, NTT, and others can be built on top of this interface. 
+
+## Specification
+
+This ERC introduces the `IERCxxx` interface.
```
blmalone, line 31:
```suggestion
This ERC introduces the `IERC7802` interface.
```

ERCS/erc-0000.md
```diff
@@ -0,0 +1,176 @@
...
+
+Source: L2Beat (10/31/24)
+
+## Reference Implementation
+
+```solidity
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.25;
+
+abstract contract CrosschainERC20 is ERC20, IERCxxx {
```
blmalone, line 131:
```suggestion
abstract contract CrosschainERC20 is ERC20, IERC7802 {
```

ERCS/erc-0000.md
```diff
@@ -0,0 +1,176 @@
...
+
+**`CrosschainMint`**
+
+MUST trigger when `crosschainMint` is successfully called.
+
+```solidity
+event CrosschainMint(address indexed _to, uint256 _amount)
+```
+
+Note: implementations might consider additionally emitting `Transfer(address(0), _to, _amount)`to be compliant with [ERC-5679](./eip-5679.md).
```
blmalone, line 65:
```suggestion
Note: implementations might consider additionally emitting `Transfer(address(0), _to, _amount)` to be compliant with [ERC-5679](./eip-5679.md).
```

ERCS/erc-0000.md
```diff
@@ -0,0 +1,176 @@
+---
+eip: 0000
+title: Crosschain Token Interface
+description: Minimal token interface for cross-chain transfers
+author: skeletor <skeletor@defi.sucks>, parti (@0xParticle), joxes <joxes@defi.sucks>, ng <ng@defi.sucks>, agus duha <agusduha@defi.sucks>, disco <0xdiscotech@defi.sucks>, gotzen <gotzen@defi.sucks>, 0age <0age@uniswap.org>, Mark Tyneway <mark@oplabs.co>, Zain Bacchus <zain@oplabs.co>, Matt Solomon <msolomon@oplabs.co>, Maurelian <maurelian@protonmail.ch>
+discussions-to: https://ethereum-magicians.org/t/erc-xxxx-crosschain-token-interface/21508
+status: Draft
+type: Standards Track
+category: ERC
```
blmalone, line 9: I wonder does this fall into the Interface category as per [EIP-1](https://eips.ethereum.org/EIPS/eip-1).

blmalone (commented):
Some suggested ammendments and a question around whether this should fall under the interface category as per eip-1.

ERCS/erc-0000.md
```diff
@@ -0,0 +1,176 @@
...
+```
+
+Note: implementations might consider additionally emitting `Transfer(address(0), _to, _amount)`to be compliant with [ERC-5679](./eip-5679.md).
+
+**`CrosschainBurnt`**
+
+MUST trigger when `crosschainBurn` is successfully called.
+
+```solidity
+event CrosschainBurnt(address indexed _from, uint256 _amount)
```
blmalone, line 72:
```suggestion
event CrosschainBurn(address indexed _from, uint256 _amount)
```

blmalone: Change predicated on this [comment](https://github.com/ethereum/ERCs/pull/692/files#r1823583714).

ERCS/erc-0000.md
```diff
@@ -0,0 +1,176 @@
...
+        emit CrosschainMint(_to, _amount);
+    }
+
+    /// @notice Allows the TOKEN_BRIDGE to burn tokens.
+    /// @param _from   Address to burn tokens from.
+    /// @param _amount Amount of tokens to burn.
+    function crosschainBurn(address _from, uint256 _amount) external onlyTokenBridge {
+        _burn(_from, _amount);
+
+        emit CrosschainBurn(_from, _amount);
```
blmalone, line 163:
```suggestion
        emit CrosschainBurn(_from, _amount);
```

This is an error if the plan is to keep `CrosschainBurn` in this [comment](https://github.com/ethereum/ERCs/pull/692/files#r1823583714).

ERCS/erc-0000.md
```diff
@@ -0,0 +1,176 @@
...
+5. Mantle ($983M)
+6. LayerZero OFTs ($465M)
+7. Omnibridge ($449M)
+8. Base ($350M)
+9. Scroll ($128M)
+10. zkSync ($77M)
+
+Source: L2Beat (10/31/24)
+
+## Reference Implementation
```
blmalone, line 125: I personally think that we should include the actual interface as code in the ERC. e.g. 


```solidity
// SPDX-License-Identifier: MIT
pragma solidity ^0.8.0;

/// @title ICrosschainERC20
/// @notice Defines the interface for crosschain ERC20 transfers.
interface ICrosschainERC20 {
    /// @notice Emitted when a crosschain transfer mints tokens.
    /// @param to       Address of the account tokens are being minted for.
    /// @param amount   Amount of tokens minted.
    event CrosschainMint(address indexed to, uint256 amount);

    /// @notice Emitted when a crosschain transfer burns tokens.
    /// @param from     Address of the account tokens are being burned from.
    /// @param amount   Amount of tokens burned.
    event CrosschainBurn(address indexed from, uint256 amount);

    /// @notice Mint tokens through a crosschain transfer.
    /// @param _to     Address to mint tokens to.
    /// @param _amount Amount of tokens to mint.
    function crosschainMint(address _to, uint256 _amount) external;

    /// @notice Burn tokens through a crosschain transfer.
    /// @param _from   Address to burn tokens from.
    /// @param _amount Amount of tokens to burn.
    function crosschainBurn(address _from, uint256 _amount) external;
}
```
Retrieved from [here](https://github.com/ethereum-optimism/optimism/blob/4e13b20fe2f8b974089d73f565eb6ce6cc261c9b/packages/contracts-bedrock/src/L2/interfaces/ICrosschainERC20.sol#L4).

This reduces ambiguity imo.

ERCS/erc-0000.md
blmalone: Seems like this is meant to be `CrosschainBurn`, per [this comment](https://github.com/ethereum/ERCs/pull/692/files#r1823658960).

ERCS/erc-0000.md
blmalone: See comment above about also including the interface in this section.

ERCS/erc-0000.md
blmalone: So, for completeness we could add this full reference implementation (this can be copy and pasted into [remix](https://remix.ethereum.org/) and it works out of the box):

```solidity
// SPDX-License-Identifier: MIT
pragma solidity 0.8.25;

import "@openzeppelin/contracts/token/ERC20/ERC20.sol";

/// @title IERC7802
/// @notice Defines the interface for crosschain ERC20 transfers.
interface IERC7802 {
    /// @notice Emitted when a crosschain transfer mints tokens.
    /// @param to       Address of the account tokens are being minted for.
    /// @param amount   Amount of tokens minted.
    event CrosschainMint(address indexed to, uint256 amount);

    /// @notice Emitted when a crosschain transfer burns tokens.
    /// @param from     Address of the account tokens are being burned from.
    /// @param amount   Amount of tokens burned.
    event CrosschainBurn(address indexed from, uint256 amount);

    /// @notice Mint tokens through a crosschain transfer.
    /// @param _to     Address to mint tokens to.
    /// @param _amount Amount of tokens to mint.
    function crosschainMint(address _to, uint256 _amount) external;

    /// @notice Burn tokens through a crosschain transfer.
    /// @param _from   Address to burn tokens from.
    /// @param _amount Amount of tokens to burn.
    function crosschainBurn(address _from, uint256 _amount) external;
}

abstract contract CrosschainERC20 is ERC20, IERC7802 {
    /// @notice Address of the TOKEN_BRIDGE contract that is allowed to mint/burn tokens.
    address public immutable TOKEN_BRIDGE;

    /// @notice Custom error for unauthorized access.
    error Unauthorized();

    /// @notice Constructor to set the TOKEN_BRIDGE address.
    /// @param _tokenBridge Address of the TOKEN_BRIDGE.
    constructor(address _tokenBridge) {
        require(_tokenBridge != address(0), "Invalid TOKEN_BRIDGE address");
        TOKEN_BRIDGE = _tokenBridge;
    }

    /// @notice A modifier that only allows the TOKEN_BRIDGE to call
    modifier onlyTokenBridge() {
        if (msg.sender != TOKEN_BRIDGE) revert Unauthorized();
        _;
    }

    /// @notice Allows the TOKEN_BRIDGE to mint tokens.
    /// @param _to     Address to mint tokens to.
    /// @param _amount Amount of tokens to mint.
    function crosschainMint(address _to, uint256 _amount) external onlyTokenBridge {
        _mint(_to, _amount);
        emit CrosschainMint(_to, _amount);
    }

    /// @notice Allows the TOKEN_BRIDGE to burn tokens.
    /// @param _from   Address to burn tokens from.
    /// @param _amount Amount of tokens to burn.
    function crosschainBurn(address _from, uint256 _amount) external onlyTokenBridge {
        _burn(_from, _amount);
        emit CrosschainBurn(_from, _amount);
    }
}

```

ERCS/erc-0000.md
```diff
@@ -0,0 +1,176 @@
...
+    /// @notice Constructor to set the TOKEN_BRIDGE address.
+    /// @param _tokenBridge Address of the TOKEN_BRIDGE.
+    constructor(address _tokenBridge) {
+        require(_tokenBridge != address(0), "Invalid TOKEN_BRIDGE address");
+        TOKEN_BRIDGE = _tokenBridge;
+    }
+    
+    /// @notice A modifier that only allows the TOKEN_BRIDGE to call
+    modifier onlyTokenBridge() {
+        if (msg.sender != TOKEN_BRIDGE) revert Unauthorized();
```
blmalone, line 144: This `Unauthorized` error doesn't exist on this contract. I've added it in this comment: https://github.com/ethereum/ERCs/pull/692#discussion_r1823665479

ERCS/erc-0000.md
0xParticle: correct is `CrosschainBurn`. Will resolve

ERCS/erc-0000.md
0xParticle: good catch. Added to the latest commit

ERCS/erc-0000.md
0xParticle: Yes, this is way better. Added in the latest commit.

ERCS/erc-0000.md
0xParticle: I think this belongs to ERC, as it's an application-level standard. The interface category applies to language-level standards.

blmalone:
@0xParticle I left an additional comment on Discord: https://discord.com/channels/1244729134312198194/1297972645014536325/1301548267447779328 

It pertains to introducing IERC165 and committing to the `bytes4` interface for `IERC7802` (i.e. `0x33331994`).

ERCS/erc-0000.md
blmalone: Agreed 👍🏻

ERCS/erc-7802.md
```diff
@@ -0,0 +1,234 @@
...
+    /// @param _amount Amount of tokens to mint.
+    function crosschainMint(address _to, uint256 _amount) external;
+
+    /// @notice Burn tokens through a crosschain transfer.
+    /// @param _from   Address to burn tokens from.
+    /// @param _amount Amount of tokens to burn.
+    function crosschainBurn(address _from, uint256 _amount) external;
+}
+
+abstract contract CrosschainERC20 is ERC20, IERC7802 {
```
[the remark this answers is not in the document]
0xParticle: Just removed it. Also, made the constructor a bit more generic.

ERCS/erc-7802.md
```diff
@@ -0,0 +1,236 @@
...
+
+Burns `_amount` of token from address `_account`.
+
+This function works as the burning entry point for bridge contracts. Each implementation is responsible for its access control logic.
+
+```solidity
+function crosschainBurn(address _account, uint256 _amount) external;
+```
+
+### Events
```
MSilb7, line 75: Adding source/destination chain id & a unique message id to events would make crosschain data more easily usable. This ~may be a hard requirement for analytics, but hard to know without seeing further build out. This is necessary for joining messages across chains, to then see inflows/outflows by chain and routes.

This gap is a main reason why you don't see this data for 3P bridges. Most sites do ~1-3 L2<>L2 bridges, then give up due to the high overhead.

hildobby: +1

0xParticle: Hi @MSilb7 !

Thank you for your suggestion. We have discussed your idea internally and agree that it has value. However, we couldn't find a minimalistic set of parameters to add to the event without making the current standard more opinionated:

- **Identifier**: Incorporating an identifier would require defining an encoding standard within the ERC, which we believe is beyond the scope of this proposal.
- **Adding `from/to` and `chainIdFrom/chainIdTo`**: Including these parameters would require changing the inputs of `crosschainMint` and `crosschainBurn` to allow the bridge to pass them. This approach is opinionated because bridges might have multiple destinations or origins, and imposing single input parameters could limit their flexibility. Moreover, event-collisions would be possible.

We see significant value in adding `msg.sender` to the `CrosschainMint` and `CrosschainBurn` events. This addition would allow integrators to identify which bridge interacts with the tokens, and does not affect the interface's neutrality in anyway. 

I believe your idea has the potential to become a standard for bridge events.

MSilb7: On `msg.sender`:
- Is the use case of `msg.sender` that you would only look at transactions sent from that address on other chains? Or only function calls (internal txs) initiating from that address?
- Is this already derivable from the transaction / internal transaction? If so then the only benefit is just that you don't need to use traces, you can just use logs.

My overall opinion is like: Without any way to clearly identify how you identify the source burn for any destination mint, it's not useful. The only value it fulfills now is like: I know that this is intended to be a cross-chain transfer, versus a same-chain transfer. But it still gives me no information about where the transfer came from or goes to. Which then opens up other questions, like "How do applications know if the burn/mint is valid?"

Deluaney1 (changes requested):
03f247e2ac8061e11176c8d0526f8a8af3aebffe

yorhodes (commented):
Nothing about any of these interfaces is "crosschain" in nature so wondering why not just use an existing "mintable" or ["burnable" interface](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/extensions/ERC20Burnable.sol)? What about the existing [xERC20 standard](https://github.com/ethereum/ERCs/pull/89) is insufficient?

prohimone:
> Nothing about any of these interfaces is "crosschain" in nature so wondering why not just use an existing "mintable" or ["burnable" interface](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/extensions/ERC20Burnable.sol)? What about the existing [xERC20 standard](https://github.com/ethereum/ERCs/pull/89) is insufficient?

or compatibility to the existing xERC20 for that matter

Zodomo:
This interface should be made to match [xERC20](https://github.com/defi-wonderland/xERC20/blob/main/solidity/contracts/XERC20.sol), which is essentially an implementation of this, but with safety features such as rate limiting. When it is **very** easy for this standard to match that one, why doesn't it?

0xParticle:
> Nothing about any of these interfaces is "crosschain" in nature so wondering why not just use an existing "mintable" or ["burnable" interface](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/extensions/ERC20Burnable.sol)? What about the existing [xERC20 standard](https://github.com/ethereum/ERCs/pull/89) is insufficient?

`wondering why not just use an existing "mintable" or "burnable" interface`
Token issuers might want to implement different logic when minting and burning come from a bridge, such as additional security checks or events. Even though it is possible to use a standard mintable interface, having a separate function allows for a clear separation of concerns. 

It might feel unnatural for tokens that do not implement any logic different from a regular `mint` and `burn`, but we want to ensure the standardized API works for the general case.

`What about the existing https://github.com/ethereum/ERCs/pull/89 is insufficient?`
`xERC20` is not insufficient. In fact, adding mint/burn limits is a great security feature and can be built very easily on top of this ERC. Our proposal aims to complement standards like xERC20 by providing a minimalistic interface that includes only the common essential actions across various cross-chain token standards.

Features such as mint and burn limits should be optional rather than mandatory components of a cross-chain token interface. Adding limits also implies that the token should have an owner, which might not be the most general design.

Again, our proposed standard is not intended to compete with xERC20 in any way. It's a smaller Lego piece that xERC20 and other standards can use.

0xParticle:
> > Nothing about any of these interfaces is "crosschain" in nature so wondering why not just use an existing "mintable" or ["burnable" interface](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/extensions/ERC20Burnable.sol)? What about the existing [xERC20 standard](https://github.com/ethereum/ERCs/pull/89) is insufficient?
> 
> or compatibility to the existing xERC20 for that matter

Its compatible with xERC20. Only difference is the function naming for `crosschainMint` and `crosschainBurn`, which is not a closed call, but should be different from `mint` and `burn`. See https://github.com/ethereum/ERCs/pull/692#issuecomment-2465036224

yorhodes:
> Features such as mint and burn limits should be optional rather than mandatory components of a cross-chain token interface. It is possible to set the limits to max in xERC20, but you would still need a token owner, which might not be the most general design.
> 
> Again, our proposed standard is not intended to compete with xERC20 in any way. It's a smaller Lego piece that xERC20 and other standards can use.

Thanks for clarifying @0xParticle. I do think it would be in the spirit of the ERC process to mitigate fragmentation by reusing the xERC20 function interface (subset) but understand that you think the `crosschain` prefix is more explicit. My personal feeling is the opposite -- that there is no reason the token standard needs to distinguish between "crosschain" and "local" mint/burn roles, similar to how it is an antipattern to distinguish between EOAs and contracts.

ERCS/erc-7802.md
karankurbur: adding msg.sender to the event would make indexing significantly easier as we can now find all mints/burns associated with a specific bridge.

zainbacchus:
> I do think it would be in the spirit of the ERC process to mitigate fragmentation by reusing the xERC20 function interface (subset). This seems much more natural than encouraging existing tokens to upgrade/migrate to this new interface.

Definitely understand the spirit of reusing where possible, one concern in reusing mint/burn from my perspective working on interop for one of the L2 clusters is with generating a false sense of "interoperability" for tokens that are already deployed and have no way to deploying to the same address on other chains.

0xParticle:
> Thanks for clarifying @0xParticle. I do think it would be in the spirit of the ERC process to mitigate fragmentation by reusing the xERC20 function interface (subset). This seems much more natural than encouraging existing tokens to upgrade/migrate to this new interface.
> 
> I understand that you think the `crosschain` prefix is important to be explicit about. My personal feeling is the opposite -- that there is no reason the token standard needs to distinguish between "crosschain" and "local" mint/burn roles, similar to how it is an antipattern to distinguish between EOAs and contracts. Token contracts can still implement crosschain checks/logic behind the mint/burn functions if they so choose.

Thank you for your feedback! This has been the main concern so far, so I'll elaborate on why we think having separate functions makes a lot of sense. 

Local minting and burning are fundamentally different from cross-chain minting and burning. 
- In cross-chain operations, the total circulating supply across all chains is expected to remain constant, as tokens are transferred between chains rather than created or destroyed in isolation.
- Agents that mint and burn tokens in cross-chain transfer fundamentally differ from token owners. It make sense for the two actors to have different permissions.
Therefore, having different checks, access controls, and logic for cross-chain actions is reasonable. The mint/burn limits that xERC20 introduces are a great example.

Merging local and cross-chain minting/burning into the same functions can lead to complex implementations that intertwine different operational logic. By splitting into two, we separate concerns, making the codebase cleaner and more maintainable. 

This separation of concerns is particularly relevant for 
- Upgrades: Any changes in access control, limits, or logic will only affect the separate cross-chain functions (crosschainMint and crosschainBurn) without altering the standard local mint and burn implementations.
- Integrations with Different Chains: To make an ERC20 token cross-chain compatible, you simply need to implement the ERC-7802 extension with the corresponding access controls for each chain. For example, when integrating with Optimism, the ERC20 token would grant access to the Optimism bridge; when integrating with Arbitrum, it would grant access to the Arbitrum bridge. The local mint and burn functions remain unchanged. Using dedicated functions for cross-chain operations provides a more modular approach, avoiding the need to modify the base implementation for each chain.

A similar reasoning applies to having dedicated cross-chain-specific events. The separation significantly facilitates the work of indexers, analytics tools, and auditors. It allows for straightforward tracking of cross-chain activities, detecting anomalies, and monitoring bridge operations. If cross-chain and local events are indistinguishable, off-chain agents must implement complex logic to differentiate them, increasing the potential for errors and inefficiencies.

The analogy between distinguishing EOAs/contracts and local/cross-chain operations isn't directly applicable. As reflected in the protocol's roadmap, the Ethereum community is working towards account abstraction to unify EOAs and contract accounts. However, there's no equivalent movement to unify local and cross-chain token operations. Cross-chain actions inherently involve additional complexities and external dependencies that justify distinct handling in token contracts.

Finally, the standard is still compatible with xERC20 via custom adapters, even with a different naming. We will work to improve this compatibility, as we think xERC20 is a great standard and want to support it. As a matter of fact, my folks at Wonderland are coauthors to the xERC20 standard.

jacekv:
This looks very much like the Pantos token contract for native cross-chain tokens. Just the naming of the functions is different, yet concept is the same:
https://github.com/pantos-io/ethereum-contracts/blob/main/src/interfaces/IPantosToken.sol

ERCS/erc-7802.md
0xParticle: We just added `msg.sender`. It was very easy to implement and can help off-chain agents.

Daevisoon (approved):

0xParticle:
Regarding compatibility with xERC20 (ERC-7802), I wrote this document:
https://defi-wonderland.notion.site/xERC20-ERC7802-compatibility-14c9a4c092c780ca94a8cb81e980d813?pvs=4

ERCS/erc-7802.md
```diff
@@ -0,0 +1,270 @@
...
+**`CrosschainMint`**
+
+MUST trigger when `crosschainMint` is successfully called. 
+The `_sender` parameter MUST be set to the msg.sender at the time the function is called.
+
+```solidity
+event CrosschainMint(address indexed _to, uint256 _amount, address indexed _sender);
+```
+
+Note: implementations might consider additionally emitting `Transfer(address(0), _to, _amount)` to be compliant with [ERC-5679](./eip-5679.md).
```
frangio, line 86: I _strongly_ suggest making the Transfer events mandatory. 

ERC-20 is more relevant than ERC-5679 IMO, and in the former it's already specified as SHOULD (i.e., a strong recommendation):
>A token contract which creates new tokens SHOULD trigger a Transfer event with the _from address set to 0x0 when tokens are created.

SamWilsn:
Gentle reminder to keep technical discussion in the Ethereum Magicians thread. It's easy to lose context after the pull request is merged.

ERCS/erc-7802.md
```diff
@@ -0,0 +1,270 @@
+---
+eip: 7802
+title: Crosschain Token Interface
+description: Minimal token interface for cross-chain transfers
```
SamWilsn, line 4: > Minimal

Too many standards claim to be minimal. I'd recommend removing this, and using your description to further elaborate on the ideas introduced in your title.

ERCS/erc-7802.md
```diff
@@ -0,0 +1,270 @@
...
+status: Draft
+type: Standards Track
+category: ERC
+created: 2024-10-30
+requires: 165
+---
+
+## Abstract
+
+This standard introduces a minimal interface for tokens to communicate cross-chain. It allows bridges with mint and burn rights to send and relay token transfers with a standardized API.
```
SamWilsn, line 16: Not bad, but I'd like to see a bit more technical meat here. Could you sketch out _how_ your proposal operates, in addition to the description you already have.

ERCS/erc-7802.md
```diff
@@ -0,0 +1,270 @@
...
+
+However, each bridge implements a different interface for minting/burning on non-native chains. For example, L1 bridged tokens need to support the `OptimismMintableERC20` specification when bridged (minted) to the OP Stack, but the `IArbToken` interface when bridged to the Arbitrum Stack.
+
+This interface fragmentation is a massive issue in cross-chain communication among chains via third-party bridges or future canonical solutions. At this point, it is clear that every bridge would benefit from a standardized interface for minted/burnt tokens. 
+
+There have been different attempts in the past to standardize token-bridging interfaces. However, third-party providers are also developing cross-chain token frameworks, such as NTT by Wormhole and OFT by LayerZero. Each framework defines its features, like rate limits and fee switches, and implements its mint and burn versions. The resultant interfaces become highly specific, lacking naming conventions and structures.
```
SamWilsn, line 26:
```suggestion
There have been different attempts in the past to standardize token-bridging interfaces. However, third-party providers are also developing cross-chain token frameworks. Each framework defines its features, like rate limits and fee switches, and implements its mint and burn versions. The resultant interfaces become highly specific, lacking naming conventions and structures.
```

Avoid naming specific products or vendors.

ERCS/erc-7802.md
```diff
@@ -0,0 +1,270 @@
...
+
+```solidity
+function crosschainBurn(address _account, uint256 _amount) external;
+```
+
+### Events
+
+**`CrosschainMint`**
+
+MUST trigger when `crosschainMint` is successfully called.
```
SamWilsn, line 79: This leaves a lot unspecified (which is fine if it's intentional.) For example, should other non-standard functions that also effectively perform a crosschain mint also trigger this event, or is it specifically tied to this function?

ERCS/erc-7802.md
```diff
@@ -0,0 +1,270 @@
...
+eip: 7802
+title: Crosschain Token Interface
+description: Minimal token interface for cross-chain transfers
+author: skeletor (@skeletor-spaceman), parti (@0xParticle), joxes (@Joxess), ng (@0xng), agus duha (@agusduha), disco (@0xDiscotech), gotzen <gotzen@defi.sucks>, 0age <0age@uniswap.org>, Mark Tyneway <mark@oplabs.co>, Zain Bacchus <zain@oplabs.co>, Matt Solomon <msolomon@oplabs.co>, Maurelian <maurelian@protonmail.ch>, Blaine Malone (@blmalone)
+discussions-to: https://ethereum-magicians.org/t/erc-7802-crosschain-token-interface/21508
+status: Draft
+type: Standards Track
+category: ERC
+created: 2024-10-30
+requires: 165
```
SamWilsn, line 11:
```suggestion
requires: 165, 5679
```

ERCS/erc-7802.md
```diff
@@ -0,0 +1,270 @@
...
+
+Merging local and cross-chain minting/burning into the same functions can lead to complex implementations that intertwine different operational logic. 
+By splitting into two, concerns remain separate, making the codebase cleaner and more maintainable.
+
+This separation of concerns is particularly relevant for
+
+- Upgrades: Any changes in access control, limits, or logic will only affect the separate cross-chain functions (`crosschainMint` and `crosschainBurn`) without altering the standard local mint and burn implementations.
+- Integrations with Different Chains: To make an [ERC-20](./eip-20.md) cross-chain compatible,
+issuers simply need to implement the [ERC-7802](./eip-7802.md) extension with the corresponding access controls for each chain. 
+For example, when integrating with Optimism, the ERC-20 would grant access to the Optimism bridge; when integrating with Arbitrum, it would grant access to the Arbitrum bridge.
```
SamWilsn, line 145: Try to avoid specific vendors/products.

ERCS/erc-7802.md
```diff
@@ -0,0 +1,270 @@
...
+It is important to note that this check can only be performed locally on the chain where the token contract resides. There is no inherent guarantee that the token on the receiving chain also supports the `IERC7802` interface. Ensuring cross-chain consistency of interface support is the responsibility of the bridge implementation.
+
+## Backwards Compatibility
+
+This proposal is fully backwards compatible with [ERC-20](./eip-20.md).
+
+As discussed in the Motivation section, a minimal, flexible cross-chain standard interface is necessary. The problem becomes larger as more tokens are deployed without a standardized format.
+
+- Upgradable tokens can be upgraded to implement the new interface.
+- Non-upgradable tokens cannot implement the interface on the token itself. They can still migrate to a standard-compliant version using a lockbox mechanism, as proposed by xERC-20. The idea is to lock non-mintable tokens and mint the same amount of interface-compliant tokens. The bridge contract can act as a lockbox on the native chain.
```
SamWilsn, line 167: Does xERC-20 have an ERC? If so, you should link it.

ERCS/erc-7802.md
```diff
@@ -0,0 +1,270 @@
...
+
+Bridge contracts will also need an upgrade to integrate with the interface. Most popular bridges in terms of liquidity at the date are upgradable or use tokens that can be upgraded to point to new bridge contracts. 
+
+1. Arbitrum One ($5.23B)
```
SamWilsn, line 171: Would be best to remove this leaderboard.

ERCS/erc-7802.md
```diff
@@ -0,0 +1,270 @@
...
+8. Base ($350M)
+9. Scroll ($128M)
+10. zkSync ($77M)
+
+Source: L2Beat (10/31/24)
+
+## Reference Implementation
+
+```solidity
+// SPDX-License-Identifier: MIT
```
SamWilsn, line 187: Code inline in the document must be CC0-1.0 licensed. If you have to keep it MIT, we do permit that in the assets directory.

SamWilsn (changes requested):

ERCS/erc-7802.md
0xParticle: fixed

ERCS/erc-7802.md
0xParticle: yes! removed

ERCS/erc-7802.md
0xParticle: Agree. Just removed.

ERCS/erc-7802.md
0xParticle: Modified this section, let me know your thoughts now

ERCS/erc-7802.md
0xParticle: Added more technical details.

ERCS/erc-7802.md
0xParticle: Yes, this is intentionally unspecified. The `CrosschainMint` event MUST be triggered when the `crosschainMint` function is successfully called. However, implementers are not restricted from emitting this event when using other non-standard functions that effectively perform a crosschain mint.

If the implementation adheres to this standard, the full interface—including `crosschainMint` and `crosschainBurn`—will already be available, so it is expected that these functions are used for crosschain minting and burning.  That said, the event is not mandated for every action that triggers a crosschain operation, as this standard focuses specifically on the standardized interface.

ERCS/erc-7802.md
0xParticle: Very good catch. Just pushed to modify this invariant.

ERCS/erc-7802.md
skidipapo: Can someone here explain how to view/claim tokens or prizes that were claimed too late because I didn't understand? I hope for the information

skidipapo (changes requested):
Can someone here explain how to view/claim tokens or prizes that were claimed too late because I didn't understand? I hope for the information

Feroslav268 (approved):

ERCS/erc-7802.md
```diff
@@ -0,0 +1,259 @@
+---
+eip: 7802
+title: Crosschain Token Interface
```
SamWilsn, line 3: "Interface" in an ERC title doesn't add a ton of information. Most ERCs contain some kind of interface.

How about:

```suggestion
title: Permissioned Token Mint/Burn Across Chains
```

ERCS/erc-7802.md
```diff
@@ -0,0 +1,259 @@
...
+description: Token interface that allows authorized contracts to mint and burn token representations during cross-chain transfers
```
SamWilsn, line 4: Similarly, you can remove some nonspecific bits from your description (and hopefully expand it further):

```suggestion
description: Allow authorized contracts to mint and burn token representations during cross-chain transfers
```

ERCS/erc-7802.md
```diff
@@ -0,0 +1,259 @@
...
+category: ERC
+created: 2024-10-30
+requires: 165, 5679
+---
+
+## Abstract
+
+This standard introduces a minimal interface, `IERC7802`, for tokens to enable standardized cross-chain communication. The interface consists of two functions, `crosschainMint` and `crosschainBurn`, which allow authorized bridge contracts to mint and burn token representations during cross-chain transfers. These functions serve as the entry points for bridge logic, enabling consistent handling of token supply across chains.
+
+The interface also defines two standardized events, `CrosschainMint` and `CrosschainBurn`, which emit metadata, including the target address, token amount, and caller. These events facilitate deterministic indexing and monitoring of cross-chain activities by off-chain agents, such as indexers, analytics tools, and auditors.
```
SamWilsn, line 18: Please choose either cross-chain or crosschain and use it consistently throughout the document.

ERCS/erc-7802.md
```diff
@@ -0,0 +1,259 @@
...
+
+`IERC7802` is intentionally lightweight, ensuring minimal overhead for implementation. Its modular design enables extensibility, allowing additional features—such as mint/burn limits, transfer fees, or bridge-specific access control mechanisms—to be layered on top without modifying the base interface.
```
SamWilsn, line 20: This paragraph probably belongs (with some rewording) in the Rationale section, since it explains why this standard is designed the way it is.

ERCS/erc-7802.md
```diff
@@ -0,0 +1,259 @@
...
+
+`IERC7802` is intentionally lightweight, ensuring minimal overhead for implementation. Its modular design enables extensibility, allowing additional features—such as mint/burn limits, transfer fees, or bridge-specific access control mechanisms—to be layered on top without modifying the base interface. 
+
+## Motivation
+
+All rollups and multiple important sidechains (Gnosis and Polygon PoS, among others) implement canonical bridges that embed their security into some part of the network's core architecture. These bridges do not have mint/burn rights over original tokens, so they usually lock (unlock) liquidity on the native chain and then mint (burn) a non-equivalent representation on the other. Mint/burn is used because the native token is non-existent on that side, so they must create a new representation. However, each bridge implements a different interface for minting/burning on non-native chains.
```
SamWilsn, line 24: > Gnosis and Polygon PoS

I'd prefer if you didn't mention specific products. It unfairly advertises them over new projects that might appear after this proposal goes Final. This isn't really a hard rule though.

ERCS/erc-7802.md
```diff
@@ -0,0 +1,259 @@
...
+Implementors MUST ensure that the `supportsInterface` method of ERC-165 returns true for this interface ID to indicate support for `IERC7802`.
+
+
+### Methods
+
+**`crosschainMint`**
+
+Mints `_amount` of token to address `_account`. 
+
+This function works as the minting entry point for bridge contracts. Each implementation is responsible for its access control logic.
```
SamWilsn, line 61: You can move the access control notes on each method to the security considerations section.

SamWilsn (approved):
I think this is acceptable for draft. Do try to take care of these comments eventually. Also replace xerc-20 with links to that proposal when it's merged.

ERCS/erc-7802.md
0xParticle: Agree with the suggestion of removing "interface" from the title. I'm not convinced about using "Permissioned", as that is not inherent to the interface (it could be permissionless). What do you think about something like  "Token with Mint/Burn access Across Chains"?

0xParticle:
> I think this is acceptable for draft. Do try to take care of these comments eventually. Also replace xerc-20 with links to that proposal when it's merged.

Thanks for the feedback. I will open a new PR addressing the various comments.

PR #693, opened 2024-10-30
Add ERC: EIP-712 Extensions for Account Abstraction

frangio:
Moved from https://github.com/ethereum/EIPs/pull/8985


ERCS/erc-0000.md
```diff
@@ -0,0 +1,157 @@
+---
+eip: 0000
```
abcoathup, line 2:
```suggestion
eip: 7803
```

ERCS/erc-0000.md
```diff
@@ -0,0 +1,157 @@
...
+title: EIP-712 Extensions for Account Abstraction
+description: Improvements for EIP-712 to support smart contract accounts.
+author: Francisco Giordano (@frangio)
+discussions-to: https://ethereum-magicians.org/t/eip-eip-712-extensions-for-account-abstraction/21436
```
abcoathup, line 6:
```suggestion
discussions-to: https://ethereum-magicians.org/t/erc-7803-eip-712-extensions-for-account-abstraction/21436
```
Changed category, added ERC number on Eth Magicians.  Can you update Eth Magicians topic to link to this PR.

ERCS/erc-7803.md
```diff
@@ -0,0 +1,157 @@
...
+3. One of the signers uses a smart contract account controlled by an ECDSA key held in a hardware wallet, so their wallet requests an EIP-712 signature from the hardware wallet, prepending the domain of the smart contract account to `signingDomains`, which is now an array of length 2.
+4. The signer verifies the contents of the signature in their hardware wallet. They are able to see that they are signing a message intended for a particular dapp domain, on behalf of their smart contract account (closest signing domain), as a member of the multisig (furthest signing domain).
+
+#### Encoding of data to be signed
+
+In the presence of `signingDomains` the account should encode the message to be signed according to the following recursive procedure:
+
+- `encodeForSigningDomains(signingDomainSeparators : [𝔹²⁵⁶], verifyingDomainSeparator : 𝔹²⁵⁶, message : 𝕊) =`
+  - If `signingDomainSeparators = [first, ...others]`: `"\x19\x02" ‖ first ‖ encodeForSigningDomains(others, verifyingDomainSeparator, message)`
```
ernestognw, line 79: Is first the initial dapp domain? If so, I think `dapp` would be more explicit:

```suggestion
  - If `signingDomainSeparators = [dapp, ...others]`: `"\x19\x02" ‖ dapp ‖ encodeForSigningDomains(others, verifyingDomainSeparator, message)`
```

ERCS/erc-7803.md
```diff
@@ -0,0 +1,157 @@
...
+
+### `authMethods`
+
+This property is an array of supported signature authentication methods, listed in the order that the verifying domain tries them.
+
+Each member of the array is an object with the following keys:
+
+- `id`: An string that identifies the method. It may be one of:
+    - `ECDSA`: ECDSA signatures by Externally Owned Accounts.
+    - `ERC-{n}`: A standard type of signature specified by an ERC. `n` must not be padded with zeros.
```
ernestognw, line 92: Is there an authentication method other than ERC-1271? If 1271 is generalized enough I think this can be simplified to just `ECDSA` or `ERC-1271`.

In the case we want to make this ERC more broadly compatible, then it may make sense to define the signature algorithms supported by other networks (e.g. ed25519 in cosmos). I feel this is not the case

ERCS/erc-7803.md
```diff
@@ -0,0 +1,157 @@
...
+`signingDomainSeparators` is the array of hashes of the domains included in `signingDomains`, in the same order, computed according to EIP-712's `hashStruct`.
+
+### `authMethods`
+
+This property is an array of supported signature authentication methods, listed in the order that the verifying domain tries them.
+
+Each member of the array is an object with the following keys:
+
+- `id`: An string that identifies the method. It may be one of:
+    - `ECDSA`: ECDSA signatures by Externally Owned Accounts.
```
ernestognw, line 91: I feel we the spec should suggest priority for ERC-1271 if there's code in the address. Just as we do in [SignatureChecker](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/0034c302241c4b1a1685272d4df42ca5d64b8c34/contracts/utils/cryptography/SignatureChecker.sol#L23) since after EIP-7702, the validation mechanism should decide whether to prioritize ECDSA or ERC-1271.

ERCS/erc-7803.md
howydev: IIUC - `signingDomainSeparators` is a new addition and will only contain account domains. The dapp domain is the `domain` in L51

frangio: Yes these are distinct from the dapp domain.

ERCS/erc-7803.md
frangio: >Is there an authentication method other than ERC-1271? 

Yes, ERC-6492 is another standard that is mentioned earlier in the Motivation section. It covers a scenario that ERC-1271 doesn't (predeploy contracts). I'd rather keep this generic than listing the standards that exist now.

ERCS/erc-7803.md
frangio: I agree that if there's code ERC-1271 should take priority but not all contracts implement it that way (e.g., prior versions of OpenZeppelin Contracts didn't), so if a dapp needs a signature for one of those contracts it should be possible to express that order of preferences.

frangio: This is explained in the Motivation section too, let me know if it's not clear.

SamWilsn (dismissed):

SamWilsn (approved):

PR #698, opened 2024-11-04
Update ERC-7656: Move to Last Call

sullof:
This ERC is very stable and has been used in more than one project (for example in [Zazook NFT](https://github.com/andyscraven/zazook-nft), where 7656 is used to deploy a subscription service associated to the NFT).
Time to move to Last Call.


SamWilsn (commented):
You cannot move to Last Call until 6551 (as a dependency) also moves to last call. It's awkward, but if you only vaguely allude to 6551 (without using its number) you can advance to Last Call.

---

Please triple check your interface id before going to final.

sullof:
> You cannot move to Last Call until 6551 (as a dependency) also moves to last call. It's awkward, but if you only vaguely allude to 6551 (without using its number) you can advance to Last Call.

I did it.

> Please triple check your interface id before going to final.

Done, many times :-)

SamWilsn (dismissed):
I've taken the liberty of copying the security considerations into this proposal, and correcting the license on your reference implementation.

sullof:
> I've taken the liberty of copying the security considerations into this proposal, and correcting the license on your reference implementation.

That is great, thanks. However, those anti-fraud considerations are applicable only if we talk about linked accounts. So, I specified it in the title of the sub-section. I am afraid you have to approve it again :(

sullof:
Also, the link to the asset was wrong. I just fixed it

SamWilsn (approved):

PR #709, opened 2024-11-08
Add ERC: Wallet Asset Discovery

lukaisailovic:
JSON-RPC method for wallets to share a user’s complete asset list with Dapps, including assets not easily discoverable through on-chain information alone.

### DIFF
--- a/ERCS/erc-7811.md
+++ b/ERCS/erc-7811.md
@@ -30,6 +30,8 @@ The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL 
 #### Request schema
 
 ```ts
+type Hex = `0x${string}`;
+
 type WalletGetAssetsRequest = {
   account: Hex;
   assetFilter?: Record<Hex, Hex[]>;
@@ -48,7 +50,7 @@ If the `assetFilter` field is provided, but the asset array is empty, wallet sho
 
 ```json
 {
-  "account": "0x123",
+  "account": "0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045",
   "assetFilter": {
     "0x1": ["0x456", "native"],
     "0xa": ["0x789", "native"]
@@ -90,7 +92,7 @@ Asset fields:
 {
   "0x1": [
     {
-      address: "0x123",
+      address: "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48",
       balance: "0xcaaea35047fe5702",
       type: "ERC20",
       metadata: {
@@ -165,7 +167,7 @@ Example:
 
 ```json
 {
-  address: "0x123",
+  address: "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48",
   balance: "0xcaaea35047fe5702",
   type: "ERC20",
   metadata: {
--- a/ERCS/erc-7811.md
+++ b/ERCS/erc-7811.md
@@ -1,7 +1,7 @@
 ---
 eip: 7811
 title: Wallet Asset Discovery
-description: JSON-RPC method for wallets to share a user’s complete asset list with Dapps, including assets not easily discoverable through on-chain information alone.
+description: JSON-RPC method for wallets to share a user’s full asset list with Dapps, including assets not easily found through on-chain data alone
 author: Luka Isailovic (@lukaisailovic), Konrad Kopp (@kopy-kat)
 discussions-to: https://ethereum-magicians.org/t/erc-7811-wallet-asset-discovery/21639
 status: Draft
@@ -63,18 +63,18 @@ type Asset = {
 type WalletGetAssetsResponse = Record<Hex, Asset[]>;
 ```
 
-The key **SHOULD** be EIP-155 chainId
+The key **SHOULD** be [EIP-155](./eip-155.md) chainId
 
 Asset fields:
 
-`address` is the address of the asset as Hex. Native assets **MUST** use `0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE` address specified by ERC-7528
+`address` is the address of the asset as Hex. Native assets **MUST** use `0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE` address specified by [ERC-7528](./eip-7528.md)
 
 `balance` is the balance of the asset as Hex
 
 **`type`:** A string indicating the type of the asset. Common asset types include but **aren’t limited to**:
 
-- **`ERC20`:** For ERC20 tokens
-- **`ERC721`:** For ERC721 tokens (NFTs)
+- **`ERC20`:** For [ERC-20](./eip-20.md) tokens
+- **`ERC721`:** For [ERC-721](./eip-721.md) tokens (NFTs)
 - **`native`:** For the chain's native currency
 
 **`metadata`:** An optional object containing additional information about the asset. The specific fields within the metadata object may vary depending on the asset type and the wallet's implementation."
@@ -200,7 +200,7 @@ If the wallet is using CAIP-25 authorization, wallet **SHOULD** include `assetDi
 }
 ```
 
-If the wallet supports ERC-5792 wallet **SHOULD** respond on `wallet_getCapabilities` request using the `assetDiscovery` key. Value should be an object with `supported` key and value `true`
+If the wallet supports [ERC-5792](./eip-5792.md) wallet **SHOULD** respond on `wallet_getCapabilities` request using the `assetDiscovery` key. Value should be an object with `supported` key and value `true`
 Wallet **SHOULD** include this for every chainId.
 
 ```json
@@ -212,6 +212,10 @@ Wallet **SHOULD** include this for every chainId.
   }
 }
 ```
+
+## Rationale
+
+## Security Considerations
 
 ## Copyright
 
--- a/ERCS/erc-7811.md
+++ b/ERCS/erc-7811.md
@@ -1,9 +1,9 @@
 ---
-eip: xxxx
-title: Wallet Asset Discovery ERC
+eip: 7811
+title: Wallet Asset Discovery
 description: JSON-RPC method for wallets to share a user’s complete asset list with Dapps, including assets not easily discoverable through on-chain information alone.
 author: Luka Isailovic (@lukaisailovic), Konrad Kopp (@kopy-kat)
-discussions-to: TBD
+discussions-to: https://ethereum-magicians.org/t/erc-7811-wallet-asset-discovery/21639
 status: Draft
 type: Standards Track
 category: ERC
--- a/ERCS/erc-7811.md
+++ b/ERCS/erc-7811.md
@@ -2,7 +2,7 @@
 eip: 7811
 title: Wallet Asset Discovery
 description: JSON-RPC method for wallets to share a user’s full asset list with Dapps, including assets not easily found through on-chain data alone
-author: Luka Isailovic (@lukaisailovic), Konrad Kopp (@kopy-kat)
+author: Luka Isailovic (@lukaisailovic), Konrad Kopp (@kopy-kat), Derek Rein (@arein)
 discussions-to: https://ethereum-magicians.org/t/erc-7811-wallet-asset-discovery/21639
 status: Draft
 type: Standards Track
@@ -41,6 +41,8 @@ type WalletGetAssetsRequest = {
 `assetFilter` is an **OPTIONAL** field that specifies only the assets Dapp cares about on the specific chains. If it is provided, the response from the wallet **SHOULD** include those assets.
 
 If the `assetFilter` field is omitted, the wallet **SHOULD** return all available assets for the requested account. It is **RECOMMENDED** that the returned assets be ordered by estimated value in descending order, as determined by the wallet.
+
+If the `assetFilter` field is provided, but the asset array is empty, wallet should return all assets on the requested chains.
 
 #### Example request
 
--- a/ERCS/erc-7811.md
+++ b/ERCS/erc-7811.md
@@ -35,6 +35,8 @@ type Hex = `0x${string}`;
 type WalletGetAssetsRequest = {
   account: Hex;
   assetFilter?: Record<Hex, Hex[]>;
+  assetTypeFilter?: string[];
+  chainFilter?: Hex[];
 };
 ```
 
@@ -44,7 +46,9 @@ type WalletGetAssetsRequest = {
 
 If the `assetFilter` field is omitted, the wallet **SHOULD** return all available assets for the requested account. It is **RECOMMENDED** that the returned assets be ordered by estimated value in descending order, as determined by the wallet.
 
-If the `assetFilter` field is provided and the assets array is empty, the wallet **MUST** restrict results to the requested chains and **SHOULD** return all assets on those chains.
+`assetTypeFilter` is an **OPTIONAL** field that specifies an array of asset types, as defined in this ERC. If `assetTypeFilter` field is provided, wallet **MUST** include only those assets in the response.
+
+`chainFilter` is an **OPTIONAL** field that specifies an array of chain ids, where each value in the array **MUST** be a valid [EIP-155](./eip-155.md) chainId
 
 #### Example request
 
@@ -54,7 +58,9 @@ If the `assetFilter` field is provided and the assets array is empty, the wallet
   "assetFilter": {
     "0x1": ["0x456", "native"],
     "0xa": ["0x789", "native"]
-  }
+  },
+  "assetTypeFilter": ["ERC20", "native"],
+  "chainFilter": ["0x1"]
 }
 ```
 
--- a/ERCS/erc-7811.md
+++ b/ERCS/erc-7811.md
@@ -33,16 +33,27 @@ The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL 
 type Hex = `0x${string}`;
 type Address = Hex;
 type AssetType = "native" | "erc20" | "erc721" | string;
+type Address = Hex;
+type AddressOrNative = Address | "native";
+type Eip155ChainId = Hex;
 
 type WalletGetAssetsRequest = {
   account: Address;
-  assetFilter?: Record<Hex, string[]>;
+  assetFilter?: Record<
+    Eip155ChainId,
+    {
+      address: AddressOrNative;
+      type: AssetType;
+    }[]
+  >;
   assetTypeFilter?: AssetType[];
   chainFilter?: Hex[];
 };
 ```
 
 `account` is a **REQUIRED** field that indicates for which account assets are requested.
+
+`assetFiled` is **OPTIONAL** field that accepts a list of assets identifiers. Each asset identifier is an object that contains `address` and `type` fields and is scoped by `chainId`, where ChainId **MUST** be a valid [EIP-155](./eip-155.md) chainId.
 
 If the `assetFilter` field is provided, the wallet **MUST** only return the assets specified within it, even if `assetTypeFilter` or `chainFilter` could have further filtered the result. This effectively disregards the `assetTypeFilter` and `chainFilter` fields entirely. The reason for this is that they are already implicitly defined within the `assetFilter`.
 
@@ -80,7 +91,7 @@ type Asset = {
 type WalletGetAssetsResponse = Record<Hex, Asset[]>;
 ```
 
-The key **SHOULD** be [EIP-155](./eip-155.md) chainId
+The key **MUST** be [EIP-155](./eip-155.md) chainId
 
 Asset fields:
 
@@ -130,9 +141,9 @@ Asset fields:
 }
 ```
 
-### Asset definitions
+### Well-known asset types
 
-This section specifies a structure for the most commonly used asset types.
+Below are expansions of `metadata` for well-known asset types. Implementations that are compliant with this ERC and return these well-known asset types **MUST** return at least these fields in `metadata`. Implementations **MAY** return more fields than specified here.
 This ERC does not specify an exhaustive list of asset types.
 Since the type is a generic string, there could be a mismatch between the type Dapp expects and the one returned by the wallet.
 It’s important that no two assets share the same type.
@@ -198,24 +209,31 @@ type Erc721Asset = {
   metadata: {
     name: string;
     symbol: string;
+    tokenId: Hex;
+    tokenURI?: string;
   };
 };
 ```
 
-### Capabilities
-
-If the wallet is using [CAIP-25](https://github.com/ChainAgnostic/CAIPs/blob/0848f06f6cfc29ce619bccdd5035c1d500033b21/CAIPs/caip-25.md) authorization, wallet **SHOULD** include `assetDiscovery` key in the CAIP-25 `sessionsProperties` object. Value should be an object with `supported` key and value `true`
+Example:
 
 ```json
 {
-  //...
-  "sessionProperties": {
-    "assetDiscovery": {
-      "supported": true
-    }
+  "address": "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48",
+  "balance": "0x01",
+  "type": "erc721",
+  "metadata": {
+    "name": "Thor's hammer",
+    "symbol": "THOR",
+    "tokenId": "0x1",
+    "tokenURI": "ipfs://hash"
   }
 }
 ```
+
+### Capabilities
+
+If the wallet is using [CAIP-25](https://github.com/ChainAgnostic/CAIPs/blob/0848f06f6cfc29ce619bccdd5035c1d500033b21/CAIPs/caip-25.md) authorization, wallet **SHOULD** include `wallet_getAssets` in the `methods` array in `sessionScopes` of `eip155` namespace.
 
 If the wallet supports [ERC-5792](./eip-5792.md) wallet **SHOULD** respond on `wallet_getCapabilities` request using the `assetDiscovery` key. Value should be an object with `supported` key and value `true`
 Wallet **SHOULD** include this for every chainId.
--- a/ERCS/erc-7811.md
+++ b/ERCS/erc-7811.md
@@ -90,8 +90,8 @@ Asset fields:
 
 **`type`:** A string indicating the type of the asset. Common asset types include but **aren’t limited to**:
 
-- **`ERC20`:** For [ERC-20](./eip-20.md) tokens
-- **`ERC721`:** For [ERC-721](./eip-721.md) tokens (NFTs)
+- **`erc20`:** For [ERC-20](./eip-20.md) tokens
+- **`erc721`:** For [ERC-721](./eip-721.md) tokens (NFTs)
 - **`native`:** For the chain's native asset
 
 **`metadata`:** An **OPTIONAL** object containing additional information about the asset. The specific fields within the metadata object may vary depending on the asset type and the wallet's implementation.
--- a/ERCS/erc-7811.md
+++ b/ERCS/erc-7811.md
@@ -31,22 +31,24 @@ The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL 
 ```ts
 type WalletGetAssetsRequest = {
   account: Hex;
-  requiredAssets?: Record<Hex, Hex[]>;
+  assetFilter?: Record<Hex, Hex[]>;
 };
 ```
 
 `account` is a required field that indicates for which account assets are requested.
 
-`requiredAssets` is an optional field that specifies only the assets Dapp cares about on the specific chains. If it is provided, the response from the wallet SHOULD include those assets.
+`assetFilter` is an optional field that specifies only the assets Dapp cares about on the specific chains. If it is provided, the response from the wallet **SHOULD** include those assets.
+
+If the `assetFilter` field is omitted, the wallet **SHOULD** return all available assets for the requested account. It is **RECOMMENDED** that the returned assets be ordered by estimated value in descending order, as determined by the wallet.
 
 #### Example request
 
 ```json
 {
   "account": "0x123",
-  "requiredAssets": {
-    "0x1": ["0x456", "0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE"],
-    "0xa": ["0x789", "0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE"]
+  "assetFilter": {
+    "0x1": ["0x456", "native"],
+    "0xa": ["0x789", "native"]
   }
 }
 ```
@@ -55,7 +57,7 @@ type WalletGetAssetsRequest = {
 
 ```ts
 type Asset = {
-  address: Hex;
+  address: Hex | "native";
   balance: Hex;
   type: string;
   metadata: any;
@@ -95,7 +97,7 @@ Asset fields:
       }
     },
     {
-      address: "0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE",
+      address: "native",
       balance: "0xcaaea35047fe5702",
       type: "native"
     }
@@ -116,16 +118,18 @@ Asset fields:
 ### Asset definitions
 
 This section specifies a structure for the most commonly used asset types.
-This ERC does not specify an exhaustive list of asset types. Since the type is a generic string, there could be a mismatch between the type Dapp expects and the one returned by the wallet. It’s important that no two assets share the same type. Therefore, new asset types should be specified in ERCs, either in this ERC as an amendment or in another ERC.
+This ERC does not specify an exhaustive list of asset types.
+Since the type is a generic string, there could be a mismatch between the type Dapp expects and the one returned by the wallet.
+It’s important that no two assets share the same type.
+Therefore, new asset types should be specified in future ERCs.
 
 **Native**
 
 ```ts
 type NativeAsset = {
-  address: "0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE";
+  address: "native";
   balance: Hex;
   type: "native";
-  metadata: any;
 };
 ```
 
@@ -133,7 +137,7 @@ Example:
 
 ```json
 {
-  "address": "0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE",
+  "address": "native",
   "balance": "0xcaaea35047fe5702",
   "type": "native"
 }
@@ -150,7 +154,6 @@ type Erc20Asset = {
     name: string;
     symbol: string;
     decimals: number;
-    [key: string]: any;
   };
 };
 ```
@@ -180,14 +183,13 @@ type Erc721Asset = {
   metadata: {
     name: string;
     symbol: string;
-    [key: string]: any;
   };
 };
 ```
 
 ### Capabilities
 
-If the wallet is using CAIP-25 authorization, wallet **SHOULD** include `assetDiscovery` key in the CAIP-25 `sessionsProperties` object. Value should be an object with `supported` key and value `true`
+If the wallet is using [CAIP-25](https://github.com/ChainAgnostic/CAIPs/blob/0848f06f6cfc29ce619bccdd5035c1d500033b21/CAIPs/caip-25.md) authorization, wallet **SHOULD** include `assetDiscovery` key in the CAIP-25 `sessionsProperties` object. Value should be an object with `supported` key and value `true`
 
 ```json
 {
@@ -215,7 +217,11 @@ Wallet **SHOULD** include this for every chainId.
 
 ## Rationale
 
+<!-- TODO -->
+
 ## Security Considerations
+
+<!-- TODO -->
 
 ## Copyright
 
--- a/ERCS/erc-7811.md
+++ b/ERCS/erc-7811.md
@@ -185,7 +185,7 @@ type Erc721Asset = {
 };
 ```
 
-## Capabilities
+### Capabilities
 
 If the wallet is using CAIP-25 authorization, wallet **SHOULD** include `assetDiscovery` key in the CAIP-25 `sessionsProperties` object. Value should be an object with `supported` key and value `true`
 
--- a/ERCS/erc-7811.md
+++ b/ERCS/erc-7811.md
@@ -44,7 +44,7 @@ type WalletGetAssetsRequest = {
 
 If the `assetFilter` field is omitted, the wallet **SHOULD** return all available assets for the requested account. It is **RECOMMENDED** that the returned assets be ordered by estimated value in descending order, as determined by the wallet.
 
-If the `assetFilter` field is provided, but the asset array is empty, wallet should return all assets on the requested chains.
+If the `assetFilter` field is provided and the assets array is empty, the wallet **MUST** restrict results to the requested chains and **SHOULD** return all assets on those chains.
 
 #### Example request
 
--- a/ERCS/erc-7811.md
+++ b/ERCS/erc-7811.md
@@ -71,8 +71,22 @@ Consumers of `wallet_getAssets` SHOULD set `assetFilter`, `assetTypeFilter` and 
 {
   "account": "0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045",
   "assetFilter": {
-    "0x1": ["0x456", "native"],
-    "0xa": ["0x789", "native"]
+    "0x1": [
+      {
+        "address": "0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045",
+        "type": "erc20"
+      },
+      {
+        "address": "native",
+        "type": "native"
+      }
+    ],
+    "0xa": [
+      {
+        "address": "0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045",
+        "type": "erc20"
+      }
+    ]
   },
   "assetTypeFilter": ["ERC20", "native"],
   "chainFilter": ["0x1"]
--- a/ERCS/erc-7811.md
+++ b/ERCS/erc-7811.md
@@ -32,7 +32,7 @@ The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL 
 ```ts
 type Hex = `0x${string}`;
 type Address = Hex;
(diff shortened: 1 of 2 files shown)


ERCS/erc-wallet-asset-discovery.md
```diff
@@ -0,0 +1,218 @@
+---
+eip: xxxx
```
abcoathup, line 2:
```suggestion
eip: 7811
```

ERCS/erc-wallet-asset-discovery.md
```diff
@@ -0,0 +1,218 @@
...
+title: Wallet Asset Discovery ERC
```
abcoathup, line 3:
```suggestion
title: Wallet Asset Discovery
```
ERC is superfluose

ERCS/erc-wallet-asset-discovery.md
```diff
@@ -0,0 +1,218 @@
...
+description: JSON-RPC method for wallets to share a user’s complete asset list with Dapps, including assets not easily discoverable through on-chain information alone.
+author: Luka Isailovic (@lukaisailovic), Konrad Kopp (@kopy-kat)
+discussions-to: TBD
```
abcoathup, line 6: Please create a discussions topic on Eth Magicians with a link to this PR.

ERCS/erc-7811.md
```diff
@@ -0,0 +1,222 @@
...
+#### Request schema
+
+```ts
+type WalletGetAssetsRequest = {
+  account: Hex;
+  requiredAssets?: Record<Hex, Hex[]>;
+};
+```
+
+`account` is a required field that indicates for which account assets are requested.
```
chris13524, line 38: Is this an account address or CAIP-10?

ERCS/erc-7811.md
```diff
@@ -0,0 +1,222 @@
...
+type Asset = {
+  address: Hex;
+  balance: Hex;
+  type: string;
+  metadata: any;
+};
+type WalletGetAssetsResponse = Record<Hex, Asset[]>;
+```
+
+The key **SHOULD** be [EIP-155](./eip-155.md) chainId
```
chris13524, line 66: Why not `MUST`?

chris13524: What about cross-chain assets? E.g. Solana? Maybe this should be a CAIP-2?

ERCS/erc-7811.md
```diff
@@ -0,0 +1,222 @@
...
+```ts
+type Erc20Asset = {
+  address: Hex;
+  balance: Hex;
+  type: "ERC20";
+  metadata: {
+    name: string;
+    symbol: string;
+    decimals: number;
+    [key: string]: any;
```
jxom, line 153: What extra would go here? If this is different per wallet, might be a bit hard/unrealistic for applications to coerce. Can we keep it strict?

ERCS/erc-7811.md
```diff
@@ -0,0 +1,222 @@
...
+This ERC does not specify an exhaustive list of asset types. Since the type is a generic string, there could be a mismatch between the type Dapp expects and the one returned by the wallet. It’s important that no two assets share the same type. Therefore, new asset types should be specified in ERCs, either in this ERC as an amendment or in another ERC.
+
+**Native**
+
+```ts
+type NativeAsset = {
+  address: "0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE";
+  balance: Hex;
+  type: "native";
+  metadata: any;
```
jxom, line 128: should this be omitted for native?

ERCS/erc-7811.md
```diff
@@ -0,0 +1,222 @@
...
+        name: "Token",
+        symbol: "TOK",
+        decimals: 18,
+      }
+    },
+    {
+      address: "0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE",
+      balance: "0xcaaea35047fe5702",
+      type: "native"
+    }
```
jxom, line 101: Wonder if we should have a rule that native asset must be the first item in the array. Would feel a bit weird to have a linear time complexity just to extract a native balance (and there could be many many tokens).

ERCS/erc-7811.md
```diff
@@ -0,0 +1,222 @@
...
+  metadata: any;
+};
+type WalletGetAssetsResponse = Record<Hex, Asset[]>;
+```
+
+The key **SHOULD** be [EIP-155](./eip-155.md) chainId
+
+Asset fields:
+
+`address` is the address of the asset as Hex. Native assets **MUST** use `0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE` address specified by [ERC-7528](./eip-7528.md)
```
jxom, line 70: Kind of odd. Why do we need to include a stub address?

ERCS/erc-7811.md
```diff
@@ -0,0 +1,222 @@
...
+**`metadata`:** An optional object containing additional information about the asset. The specific fields within the metadata object may vary depending on the asset type and the wallet's implementation."
+
+#### Example response
+
+```json
+{
+  "0x1": [
+    {
+      address: "0x123",
+      balance: "0xcaaea35047fe5702",
```
chris13524, line 89: Suggest using `address` as the key to simplify lookups

```suggestion
  "0x1": [
    "0x123": {
      balance: "0xcaaea35047fe5702",
```

jxom: ah yeah, then you could have `native` key too

ERCS/erc-7811.md
lukaisailovic: Its account address, however I named it account to keep the naming consistent with other calls.

ERCS/erc-7811.md
lukaisailovic: Good point, will update!

ERCS/erc-7811.md
lukaisailovic: This is an interesting point, would simplify lookups a lot for the clients!

ERCS/erc-7811.md
chris13524:
```typescript
address: Hex | "native";
```

?

ERCS/erc-7811.md
chris13524: Would [this suggestion](https://github.com/ethereum/ERCs/pull/709/files/ff4b3eb3ac8e275d5f055546caed66919ab0d3d0#r1853008255) resolve your concern?

ERCS/erc-7811.md
```diff
@@ -0,0 +1,222 @@
...
+discussions-to: https://ethereum-magicians.org/t/erc-7811-wallet-asset-discovery/21639
+status: Draft
+type: Standards Track
+category: ERC
+created: 2024-11-07
+---
+
+## Abstract
+
+This ERC introduces a new RPC call, wallet_getAssets, for wallets to declare to the Dapp what assets are owned by the user. This allows for more accurate asset discovery and the use of assets that aren’t available on-chain but can be provided by the wallet
```
SamWilsn, line 15:
```suggestion
This ERC introduces a new RPC call, `wallet_getAssets`, for wallets to declare to the Dapp what assets are owned by the user. This allows for more accurate asset discovery and the use of assets that aren’t available on-chain but can be provided by the wallet
```

Please put code in backticks, both here and throughout the rest of the document.

ERCS/erc-7811.md
```diff
@@ -0,0 +1,222 @@
...
+
+The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL NOT”, “SHOULD”, “SHOULD NOT”, “RECOMMENDED”, “MAY”, and “OPTIONAL” in this document are to be interpreted as described in RFC 2119.
+
+### Method: `wallet_getAssets`
+
+#### Request schema
+
+```ts
+type WalletGetAssetsRequest = {
+  account: Hex;
```
SamWilsn, line 33: I'm not super familiar with TypeScript. Is `Hex` a predefined type?

ERCS/erc-7811.md
SamWilsn, line 38:
```suggestion
`account` is a REQUIRED field that indicates for which account assets are requested.
```

You should uppercase the RFC 2119 keywords.

ERCS/erc-7811.md
```diff
@@ -0,0 +1,222 @@
...
+  requiredAssets?: Record<Hex, Hex[]>;
+};
+```
+
+`account` is a required field that indicates for which account assets are requested.
+
+`requiredAssets` is an optional field that specifies only the assets Dapp cares about on the specific chains. If it is provided, the response from the wallet SHOULD include those assets.
```
SamWilsn, line 40:
```suggestion
`requiredAssets` is an OPTIONAL field that specifies only the assets Dapp cares about on the specific chains. If it is provided, the response from the wallet SHOULD include those assets.
```

ERCS/erc-7811.md
```diff
@@ -0,0 +1,222 @@
...
+      }
+    }
+  ],
+};
+```
+
+### Asset definitions
+
+This section specifies a structure for the most commonly used asset types.
+This ERC does not specify an exhaustive list of asset types. Since the type is a generic string, there could be a mismatch between the type Dapp expects and the one returned by the wallet. It’s important that no two assets share the same type. Therefore, new asset types should be specified in ERCs, either in this ERC as an amendment or in another ERC.
```
SamWilsn, line 119:
```suggestion
This ERC does not specify an exhaustive list of asset types. Since the type is a generic string, there could be a mismatch between the type Dapp expects and the one returned by the wallet. It’s important that no two assets share the same type. Therefore, new asset types should be specified in future ERCs.
```

You cannot amend a Final ERC, so I'd recommend removing this bit.

ERCS/erc-7811.md
```diff
@@ -0,0 +1,222 @@
+---
+eip: 7811
+title: Wallet Asset Discovery
+description: JSON-RPC method for wallets to share a user’s full asset list with Dapps, including assets not easily found through on-chain data alone
+author: Luka Isailovic (@lukaisailovic), Konrad Kopp (@kopy-kat)
+discussions-to: https://ethereum-magicians.org/t/erc-7811-wallet-asset-discovery/21639
+status: Draft
+type: Standards Track
+category: ERC
+created: 2024-11-07
```
SamWilsn, line 10:
```suggestion
created: 2024-11-07
requires: 20, 155, 721, 5792, 7528
```

ERCS/erc-7811.md
```diff
@@ -0,0 +1,222 @@
...
+    name: string;
+    symbol: string;
+    [key: string]: any;
+  };
+};
+```
+
+### Capabilities
+
+If the wallet is using CAIP-25 authorization, wallet **SHOULD** include `assetDiscovery` key in the CAIP-25 `sessionsProperties` object. Value should be an object with `supported` key and value `true`
```
SamWilsn, line 190:
```suggestion
If the wallet is using [CAIP-25](https://github.com/ChainAgnostic/CAIPs/blob/0848f06f6cfc29ce619bccdd5035c1d500033b21/CAIPs/caip-25.md) authorization, wallet **SHOULD** include `assetDiscovery` key in the CAIP-25 `sessionsProperties` object. Value should be an object with `supported` key and value `true`
```

ERCS/erc-7811.md
```diff
@@ -0,0 +1,222 @@
...
+  "0xa": {
+    "assetDiscovery": {
+      "supported": true
+    }
+  }
+}
+```
+
+## Rationale
+
```
SamWilsn, line 217:
```suggestion

<!-- TODO -->

```

ERCS/erc-7811.md
```diff
@@ -0,0 +1,222 @@
...
+## Security Considerations
+
```
SamWilsn, line 219:
```suggestion

<!-- TODO -->

```

Allenqel:
Ethereum/ ERCs #709

ERCS/erc-7811.md
```diff
@@ -0,0 +1,222 @@
+---
+eip: 7811
+title: Wallet Asset Discovery
+description: JSON-RPC method for wallets to share a user’s full asset list with Dapps, including assets not easily found through on-chain data alone
```
xinbenlv, line 4:
```suggestion
description: JSON-RPC method for wallets to share the full asset list owned by a given account address with Dapps, including assets not easily found through on-chain data alone
```

Could you clarify Is it an user or a account (wallet) address? 

For example many traditional wallet app user have a seed phrase that enables them to have multiple account addresses by BIP32 plus on different chain.

lukaisailovic: This is specific to one account. What I think you are asking is if there is a way to return all assets owned by the user across different accounts?

ERCS/erc-7811.md
lukaisailovic: @SamWilsn Its not actually, the TS native way to do this would be `0x${string}` but it looks ugly IMO. Hex type ships with all the clients folks usually use like Wagmi, Ethers etc

lukaisailovic:
@SamWilsn Pushed an update based on your comments and resolved them on the PR. TY for the review!

ERCS/erc-7811.md
lukaisailovic: @chris13524 Was thinking more about this, I agree that it simplifies lookups, however the API is less intuitive and we complicate the most common use case which is just looping through and displaying all assets. For objects that would be super annoying in js

ERCS/erc-7811.md
xinbenlv: Since it's in specification section, if you are introducing a data type please define them

ERCS/erc-7811.md
```diff
@@ -0,0 +1,231 @@
...
+
+If the `assetFilter` field is omitted, the wallet **SHOULD** return all available assets for the requested account. It is **RECOMMENDED** that the returned assets be ordered by estimated value in descending order, as determined by the wallet.
+
+If the `assetFilter` field is provided, but the asset array is empty, wallet should return all assets on the requested chains.
+
+#### Example request
+
+```json
+{
+  "account": "0x123",
```
xinbenlv, line 51: Can you use full length example?

lukaisailovic: Done!

ERCS/erc-7811.md
lukaisailovic: Addressed!

ERCS/erc-7811.md
jxom: If the goal is to optimize for API clarity, wouldn't it make sense to group by token instead of chain id? I am not sure how `Object.values` is any less intuitive (esp. if consumers are using libraries that abstract over JSON-RPC).

lukaisailovic: @jxom I was not a fan of grouping per chain, and then per asset. It seems like a lot of nesting which is why I don't find it intuitive. 
Agree that this will prob be abstracted. 
Do you think it makes sense to group on per asset and specify a list of chains instead? 
We run into an issue of one asset having different addresses across chains then, which is why I grouped it per chain initially.

ERCS/erc-7811.md
```diff
@@ -0,0 +1,239 @@
...
+  metadata: {
+    name: string;
+    symbol: string;
+  };
+};
+```
+
+### Capabilities
+
+If the wallet is using [CAIP-25](https://github.com/ChainAgnostic/CAIPs/blob/0848f06f6cfc29ce619bccdd5035c1d500033b21/CAIPs/caip-25.md) authorization, wallet **SHOULD** include `assetDiscovery` key in the CAIP-25 `sessionsProperties` object. Value should be an object with `supported` key and value `true`
```
ignaciosantise, line 203:
```suggestion
If the wallet is using [CAIP-25](https://github.com/ChainAgnostic/CAIPs/blob/main/CAIPs/caip-25.md) authorization, wallet **SHOULD** include `assetDiscovery` key in the CAIP-25 `sessionsProperties` object. Value should be an object with `supported` key and value `true`
```

ERCS/erc-7811.md
```diff
@@ -0,0 +1,239 @@
...
+  "assetTypeFilter": ["ERC20", "native"],
+  "chainFilter": ["0x1"]
+}
+```
+
+#### Response schema
+
+```ts
+type Asset = {
+  address: Hex | "native";
```
kopy-kat, line 71: is there an argument to use the zero address for native and thus only have a single type?

kopy-kat: especially given we already have type

ERCS/erc-7811.md
```diff
@@ -0,0 +1,239 @@
...
+};
+```
+
+`account` is a **REQUIRED** field that indicates for which account assets are requested.
+
+If the `assetFilter` field is provided, the wallet **MUST** only return the assets specified within it, effectively disregarding the `assetTypeFilter` and `chainFilter` fields as they are already implicitly defined within the `assetFilter`.
+
+If the `assetFilter` field is omitted, the wallet **SHOULD** return all available assets for the requested account. It is **RECOMMENDED** that the returned assets be ordered by estimated value in descending order, as determined by the wallet.
+
+`assetTypeFilter` is an **OPTIONAL** field that specifies an array of asset types, as defined in this ERC. If `assetTypeFilter` field is provided, wallet **MUST** include only those assets in the response.
```
llbartekll, line 49: what are the types? should it be also standardised?

ERCS/erc-7811.md
llbartekll: agree 👍 or optional?

lukaisailovic: Yeah, dapps would have to do an if check anyways

ERCS/erc-7811.md
lukaisailovic: Do you mean the asset types? They are defined in the ERC

ERCS/erc-7811.md
```diff
@@ -0,0 +1,239 @@
...
+### Method: `wallet_getAssets`
+
+#### Request schema
+
+```ts
+type Hex = `0x${string}`;
+
+type WalletGetAssetsRequest = {
+  account: Hex;
+  assetFilter?: Record<Hex, string[]>;
```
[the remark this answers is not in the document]
lukaisailovic: I think this was lost during the updates, an oversight, will add it back!

ERCS/erc-7811.md
lukaisailovic: Using CAIP-2 would mean that the assets would also have to be defined in a cross-chain way. Currently there isn't a nice way to do that so thats why it was skipped

ERCS/erc-7811.md
llbartekll: 👌I see now

ERCS/erc-7811.md
lukaisailovic: Also you can use Solana/non-evm as a source of funds. This ERC just specifies how much can you get on certain EVM chains, it does not say where from

ERCS/erc-7811.md
```diff
@@ -0,0 +1,239 @@
...
+```json
+{
+  "0x1": [
+    {
+      "address": "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48",
+      "balance": "0xcaaea35047fe5702",
+      "type": "ERC20",
+      "metadata": {
+        "name": "Token",
+        "symbol": "TOK",
```
llbartekll, line 106: what if wallet won't provide token TOK in the response? should the dapp fallback with it's own on chain discovery?

chris13524: It's a required field, so the wallet _must_ provide "TOK" in the response

ERCS/erc-7811.md
bumblefudge: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1.md#chain-agnostic-improvement-proposals-caips
Because only commit-specific links are allowed by EIP-1 (I was lobbying for links of the type proposed by ignacio but the EIP editors preferred immutable referents rather than jekyll-rendered ones; i've seen similar for W3C specs in other places, FWIW)

ERCS/erc-7811.md
bumblefudge: normrefing CAIP-2 might add too much complexity but adding an example with funds from Solana or BTC referenced by CAIP-2 might make the point?

ERCS/erc-7811.md
```diff
@@ -0,0 +1,239 @@
...
+  "sessionProperties": {
+    "assetDiscovery": {
+      "supported": true
+    }
+  }
+}
+```
+
+If the wallet supports [ERC-5792](./eip-5792.md) wallet **SHOULD** respond on `wallet_getCapabilities` request using the `assetDiscovery` key. Value should be an object with `supported` key and value `true`
+Wallet **SHOULD** include this for every chainId.
```
chris13524, line 217: > every chainId

This could create excessive bandwidth. Should we consider a way to indicate support for all chains? 🤔

ERCS/erc-7811.md
chris13524, line 49:
```suggestion
`assetTypeFilter` is an **OPTIONAL** field that specifies an array of asset types, as defined in this ERC. If `assetTypeFilter` field is provided, wallet **MUST** include only assets with those types in the response
```

ERCS/erc-7811.md
```diff
@@ -0,0 +1,239 @@
...
+`account` is a **REQUIRED** field that indicates for which account assets are requested.
+
+If the `assetFilter` field is provided, the wallet **MUST** only return the assets specified within it, effectively disregarding the `assetTypeFilter` and `chainFilter` fields as they are already implicitly defined within the `assetFilter`.
+
+If the `assetFilter` field is omitted, the wallet **SHOULD** return all available assets for the requested account. It is **RECOMMENDED** that the returned assets be ordered by estimated value in descending order, as determined by the wallet.
+
+`assetTypeFilter` is an **OPTIONAL** field that specifies an array of asset types, as defined in this ERC. If `assetTypeFilter` field is provided, wallet **MUST** include only those assets in the response.
+
+`chainFilter` is an **OPTIONAL** field that specifies an array of chain ids, where each value in the array **MUST** be a valid [EIP-155](./eip-155.md) chainId
+
```
chris13524, line 52:
```suggestion

Callers of `wallet_getAssets` SHOULD provide `assetFilter`, `assetTypeFilter` and `chainFilter` with as granular as reasonably possible values. For example, if an app is only interested in interacting with USDC on Base, it should provide filters for this. This ensures that wallets and the underlying infrastructure do not incur excessive cost.

```

ERCS/erc-7811.md
```diff
@@ -0,0 +1,239 @@
...
+`address` is the address of the asset as `Hex` or `native` string for native assets.
+
+`balance` is the balance of the asset as `Hex`
+
+**`type`:** A string indicating the type of the asset. Common asset types include but **aren’t limited to**:
+
+- **`ERC20`:** For [ERC-20](./eip-20.md) tokens
+- **`ERC721`:** For [ERC-721](./eip-721.md) tokens (NFTs)
+- **`native`:** For the chain's native asset
```
chris13524, line 91: Nit: suggest using lowercase. Similar to CAIP-2 and other keys in-general

```suggestion
- **`erc20`:** For [ERC-20](./eip-20.md) tokens
- **`erc721`:** For [ERC-721](./eip-721.md) tokens (NFTs)
- **`native`:** For the chain's native asset
```

ERCS/erc-7811.md
```diff
@@ -0,0 +1,239 @@
...
+
+```ts
+type Erc20Asset = {
+  address: Hex;
+  balance: Hex;
+  type: "ERC20";
+  metadata: {
+    name: string;
+    symbol: string;
+    decimals: number;
```
chris13524, line 167: TIL technically these are optional fields. Shouldn't we follow that? https://github.com/ethereum/ERCs/blob/master/ERCS/erc-20.md#token

chris13524, line 167: TIL technically these are optional fields. Shouldn't we follow that? https://github.com/ethereum/ERCs/blob/master/ERCS/erc-20.md#token

```suggestion
    name?: string;
    symbol?: string;
    decimals?: number;
```

ERCS/erc-7811.md
```diff
@@ -0,0 +1,239 @@
...
+```json
+{
+  //...
+  "sessionProperties": {
+    "assetDiscovery": {
+      "supported": true
+    }
+  }
+}
+```
```
chris13524, line 214:
```suggestion

CAIP-25 wallets MAY use [CAIP-345](https://github.com/ChainAgnostic/CAIPs/pull/345) to provide an alternative RPC endpoint to send `wallet_getAssets` to.

```

ERCS/erc-7811.md
```diff
@@ -0,0 +1,239 @@
...
+  "type": "ERC20",
+  "metadata": {
+    "name": "Token",
+    "symbol": "TOK",
+    "decimals": 18
+  }
+}
+```
+
+**ERC-721 Token**
```
chris13524, line 187: Can we put an example for this? I wonder if `balance` should be part of `metadata` since it doesn't make sense for NFTs?

ERCS/erc-7811.md
chris13524, line 71: `address` is not sufficient to identify the asset. For example ERC-720 and ERC-1155 tokens have an additional `tokenId`. This problem is also addressed in [CAIP-19](https://github.com/ChainAgnostic/namespaces/blob/main/eip155/caip19.md)

Furthermore, `address` doesn't make sense for native tokens. We handled that with a special case, but I'd argue `address` should be moved into `metadata` instead.

chris13524: I suggest moving this `address` into `metadata` to avoid the special case. https://github.com/ethereum/ERCs/pull/709/files#r1958733655

ERCS/erc-7811.md
chris13524: We should make this `MUST`

ERCS/erc-7811.md
```diff
@@ -0,0 +1,239 @@
...
+      "type": "ERC721",
+      "metadata": {
+        //...
+      }
+    }
+  ]
+}
+```
+
+### Asset definitions
```
chris13524, line 129:
```suggestion
### Well-known asset types

Below are expansions of `metadata` for well-known asset types. Implementations that are ERC-7811-compliant that return these well-known asset types MUST return at least these fields in `metadata`. Implementations MAY return more fields than specified here.
```

ERCS/erc-7811.md
```diff
@@ -0,0 +1,239 @@
...
+  account: Hex;
+  assetFilter?: Record<Hex, string[]>;
+  assetTypeFilter?: string[];
+  chainFilter?: Hex[];
+};
+```
+
+`account` is a **REQUIRED** field that indicates for which account assets are requested.
+
+If the `assetFilter` field is provided, the wallet **MUST** only return the assets specified within it, effectively disregarding the `assetTypeFilter` and `chainFilter` fields as they are already implicitly defined within the `assetFilter`.
```
chris13524, line 45:
```suggestion
If the `assetFilter` field is provided, the wallet **MUST** only return the assets specified within it, even if `assetTypeFilter` or `chainFilter` could have further filtered the result. This effectively disregards the `assetTypeFilter` and `chainFilter` fields entirely as they are already implicitly defined within the `assetFilter`.
```

ERCS/erc-7811.md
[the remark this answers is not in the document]
chris13524: We should use [CAIP-19](https://chainagnostic.org/CAIPs/caip-19) asset IDs here instead. This allows us to filter more granularly such as to the specific token ID.

```suggestion
  assetFilter?: string[];
```

ERCS/erc-7811.md
```diff
@@ -0,0 +1,239 @@
...
+## Specification
+
+The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL NOT”, “SHOULD”, “SHOULD NOT”, “RECOMMENDED”, “MAY”, and “OPTIONAL” in this document are to be interpreted as described in RFC 2119.
+
+### Method: `wallet_getAssets`
+
+#### Request schema
+
+```ts
+type Hex = `0x${string}`;
```
chris13524, line 33: Suggest defining more type names such as Address and AssetType so it's more clear the semantics in other places.

```suggestion
type Hex = `0x${string}`;
type Address = Hex;
type AssetType = "native" | "erc20" | "erc720" | string;
```

ERCS/erc-7811.md
```diff
@@ -0,0 +1,239 @@
...
+
+type WalletGetAssetsRequest = {
+  account: Hex;
+  assetFilter?: Record<Hex, string[]>;
+  assetTypeFilter?: string[];
```
chris13524, line 38:
```suggestion
  assetTypeFilter?: AssetType[];
```

ERCS/erc-7811.md
```diff
@@ -0,0 +1,239 @@
...
+
+### Method: `wallet_getAssets`
+
+#### Request schema
+
+```ts
+type Hex = `0x${string}`;
+
+type WalletGetAssetsRequest = {
+  account: Hex;
```
chris13524, line 36: I think we should make `account` an array. In some apps you may need to interact with multiple accounts at once. For example if an app asked you which account you want to use to pay for the doughnut, you could select from which account and at the same time easily see the balance of each account. Currently this is `O(n)` for each account.

ERCS/erc-7811.md
```diff
@@ -0,0 +1,239 @@
...
+  assetFilter?: Record<Hex, string[]>;
+  assetTypeFilter?: string[];
+  chainFilter?: Hex[];
+};
```
chris13524, line 40: Suggest adding a way to specify how much metadata you want. For example, Dune [requires you to specify](https://docs.dune.com/echo/evm/balances#parameter-metadata) additional fields to return (e.g. logo) but we could extend this into this spec. Reason is this could have cost-savings if resolving certain information is more costly.

Below is a mapping of asset types to an array of fields that the requester would like to be in the response:

```suggestion
  metadata: { [key: AssetType]: string[] },
};
```

ERCS/erc-7811.md
chris13524: We should use CAIP-19. So this would be 0.000001 USDC:

```suggestion
  "eip155:1/erc20:0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48": {
      balance: "0x1",
```

ERCS/erc-7811.md
llbartekll: as discussed `assetDiscovery` no longer needed

ERCS/erc-7811.md
lukaisailovic: We would need the response to be an array as well in that case. Do you think its worth it as its not a common use case?

ERCS/erc-7811.md
lukaisailovic: The current structure of 5792 capabilities requires this

ERCS/erc-7811.md
lukaisailovic: I think the CAIP needs to be a specific hash (merged)

ERCS/erc-7811.md
lukaisailovic: Well technically balance would be the amount. There can be more than one in the wallet in some cases, depends on the NFT

ERCS/erc-7811.md
lukaisailovic: I think the wallet should handle that and we should keep it as required in the interface

ERCS/erc-7811.md
lukaisailovic: > Implementations MAY return more fields than specified here.

I think we should remove this part

ERCS/erc-7811.md
chris13524: What is the drawback? It _should_ be a common use case.

ERCS/erc-7811.md
chris13524: How should the wallet handle it?

lukaisailovic: I currently don't know of a case where token doesn't have a symbol or decimals. It has to be displayed in the wallet somehow, so if it doesn't have those values, what does wallet do with it? How does it display the balance to the user? The assumption is that they would handle that case if it exists and would apply it to 7811 as well

ERCS/erc-7811.md
chris13524: > how would the dapp know about that and ultimately utilize it in the implementation

Probably some standard. But I still think it's useful to allow arbitrary data. You could ask the same question about capabilities or scopeProperties

ERCS/erc-7811.md
```diff
@@ -0,0 +1,243 @@
...
+```
+
+Example:
+
+```json
+{
+  "address": "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48",
+  "balance": "0xcaaea35047fe5702",
+  "type": "ERC20",
```
chris13524, line 182:
```suggestion
  "type": "erc20",
```

ERCS/erc-7811.md
```diff
@@ -0,0 +1,243 @@
...
+}
+```
+
+**ERC-721 Token**
+
+```ts
+type Erc721Asset = {
+  address: Hex;
+  balance: Hex;
+  type: "ERC20";
```
chris13524, line 197:
```suggestion
  type: "erc721";
```

ERCS/erc-7811.md
```diff
@@ -0,0 +1,243 @@
...
+}
+```
+
+**ERC-20 Token**
+
+```ts
+type Erc20Asset = {
+  address: Hex;
+  balance: Hex;
+  type: "ERC20";
```
chris13524, line 167:
```suggestion
  type: "erc20";
```

ERCS/erc-7811.md
```diff
@@ -0,0 +1,243 @@
...
+The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL NOT”, “SHOULD”, “SHOULD NOT”, “RECOMMENDED”, “MAY”, and “OPTIONAL” in this document are to be interpreted as described in RFC 2119.
+
+### Method: `wallet_getAssets`
+
+#### Request schema
+
+```ts
+type Hex = `0x${string}`;
+type Address = Hex;
+type AssetType = "native" | "erc20" | "erc720" | string;
```
chris13524, line 35:
```suggestion
type AssetType = "native" | "erc20" | "erc721" | string;
```

ERCS/erc-7811.md
```diff
@@ -0,0 +1,243 @@
...
+};
+```
+
+Example:
+
+```json
+{
+  "address": "native",
+  "balance": "0xcaaea35047fe5702",
+  "type": "native"
```
chris13524, line 157: Native tokens should have this metadata too. E.g. "Etheruem", "ETH", and 18?

```suggestion
  "type": "native"
  "metadata": {
    name: string;
    symbol: string;
    decimals: number;
  }
```

ERCS/erc-7811.md
chris13524: For example, I think we should include `price` and `icon_url` in the metadata.

ERCS/erc-7811.md
chris13524: Noting that _we_ (wallet service devs) have to figure this out and provide the values, not wallets.

lukaisailovic: 7811 is Dapp <> Wallet Communication. Wallet service should be the extension of the wallet

ERCS/erc-7811.md
lukaisailovic: @chris13524 Why? 
I don't think it should be required. Adds unnecessary complexity, specifically for icon_url which the wallets potentially do not have if they store the icons locally. 
If we make it optional, then the dapp can't rely on it and needs to have a fallback. If it must have a fallback, whats the value add in the first place. I think its okay that we make the metadata extensible to allow future asset types tho.

ERCS/erc-7811.md
lukaisailovic: Added unnecessary complexity. One more level of nesting in the response

ERCS/erc-7811.md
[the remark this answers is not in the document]
chris13524: Avoiding CAIP-19, we could filter by token ID in this way:

```suggestion
  assetFilter?: Record<Hex, {
    address: Address | "native",
    tokenId?: string,
    [key: string]: any,
  }[]>;
```

What are folk's appetite to do additional filtering beyond just the asset address?

ERCS/erc-7811.md
```diff
@@ -0,0 +1,243 @@
...
+
+The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL NOT”, “SHOULD”, “SHOULD NOT”, “RECOMMENDED”, “MAY”, and “OPTIONAL” in this document are to be interpreted as described in RFC 2119.
+
+### Method: `wallet_getAssets`
+
+#### Request schema
+
+```ts
+type Hex = `0x${string}`;
+type Address = Hex;
```
chris13524, line 34:
```suggestion
type Address = Hex;
type AddressOrNative = Address | "native";
type Eip191ChainId = Hex;
```

ERCS/erc-7811.md
lukaisailovic: @chris13524 Added **MAY** for extensibility

ERCS/erc-7811.md
lukaisailovic: Isn't this standardized across all EIP-155 chains?

ERCS/erc-7811.md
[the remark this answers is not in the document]
lukaisailovic: @chris13524 Wdyt about the update I pushed for the filter?

ERCS/erc-7811.md
bumblefudge: yup, the EIP linter requires a link like `github.com/chainagnostic/.../{commit}/.../CAIP-345.md`. Use this for now but remember to update to the newest commit before publishing to EIP:
https://github.com/ChainAgnostic/CAIPs/blob/4c90de67ad87d2fca2c6b58c3a01ecebf928e008/CAIPs/caip-345.md

ERCS/erc-7811.md
bumblefudge: I'm not sure what you mean by standardized? Each registration in ethereum-lists/chains has to [specify it](https://github.com/ethereum-lists/chains?tab=readme-ov-file#example)

ERCS/erc-7811.md
```diff
@@ -0,0 +1,275 @@
...
+      address: AddressOrNative;
+      type: AssetType;
+    }[]
+  >;
+  assetTypeFilter?: AssetType[];
+  chainFilter?: Hex[];
+};
+```
+
+`account` is a **REQUIRED** field that indicates for which account assets are requested.
```
chris13524, line 54:
```suggestion
`accounts` is a **OPTIONAL** field that indicates for which accounts assets are requested.
```

ERCS/erc-7811.md
```diff
@@ -0,0 +1,275 @@
...
+#### Response schema
+
+```ts
+type Asset = {
+  address: AddressOrNative;
+  balance: Hex;
+  type: string;
+  metadata: any;
+};
+type WalletGetAssetsResponse = Record<Hex, Asset[]>;
```
chris13524, line 105:
```suggestion
type WalletGetAssetsResponse = Record<Address, Record<Hex, Asset[]>>;
```

ERCS/erc-7811.md
```diff
@@ -0,0 +1,275 @@
...
+- **`erc721`:** For [ERC-721](./eip-721.md) tokens (NFTs)
+- **`native`:** For the chain's native asset
+
+**`metadata`:** An **OPTIONAL** object containing additional information about the asset. The specific fields within the metadata object may vary depending on the asset type and the wallet's implementation.
+
+#### Example response
+
+```json
+{
+  "0x1": [
```
chris13524, line 128: TODO scope by address

ERCS/erc-7811.md
joelythe: M'y accountedge

ERCS/erc-7811.md
joelythe: M'y accountedge

joelythe: M'y accountedge

ERCS/erc-7811.md
joelythe: M'y accountedge

ERCS/erc-7811.md
```diff
@@ -0,0 +1,275 @@
...
+  "chainFilter": ["0x1"]
+}
+```
+
+#### Response schema
+
+```ts
+type Asset = {
+  address: AddressOrNative;
+  balance: Hex;
```
joelythe, line 101: M'y accountedge

ERCS/erc-7811.md
joelythe: M'y accountedge

ERCS/erc-7811.md
```diff
@@ -0,0 +1,275 @@
...
+eip: 7811
+title: Wallet Asset Discovery
+description: JSON-RPC method for wallets to share a user’s full asset list with Dapps, including assets not easily found through on-chain data alone
+author: Luka Isailovic (@lukaisailovic), Konrad Kopp (@kopy-kat), Derek Rein (@arein), Chris Smith (@chris13524)
+discussions-to: https://ethereum-magicians.org/t/erc-7811-wallet-asset-discovery/21639
+status: Draft
+type: Standards Track
+category: ERC
+created: 2024-11-07
+requires: 20, 155, 721, 5792
```
joelythe, line 11: M'y accountedge

ERCS/erc-7811.md
```diff
@@ -0,0 +1,275 @@
+---
+eip: 7811
+title: Wallet Asset Discovery
+description: JSON-RPC method for wallets to share a user’s full asset list with Dapps, including assets not easily found through on-chain data alone
```
joelythe, line 4: M'y accountedge

ERCS/erc-7811.md
```diff
@@ -0,0 +1,275 @@
...
+  assetFilter?: Record<
+    Eip155ChainId,
+    {
+      address: AddressOrNative;
+      type: AssetType;
+    }[]
+  >;
+  assetTypeFilter?: AssetType[];
+  chainFilter?: Hex[];
+};
```
joelythe, line 51: M'y accountedge

PR #399, opened 2024-04-26
Add ERC: ERC-20 Transfer Reference Extension

radeksvarz:

PR #716, opened 2024-11-12
Update ERC-7699: Update erc-7699.md - suggested minor fixes

radeksvarz:
as suggested by Sam during finalization of initial PR #399


Magken91:
Need some help with a few things

PR #735, opened 2024-11-27
Add ERC: Readable Interoperable Addresses using ENS

jrudolf:


ERCS/erc-chain_specific_addresses.md
```diff
@@ -0,0 +1,235 @@
...
+
+### A special case for ENS resolution
+
+Any ENS name today can be resolved to a chain identifier as in [ENSIP-11](https://docs.ens.domains/ensip/11) or to an address as in [ENSIP-9](https://docs.ens.domains/ensip/9).
+We could imagine having a name `user.eth` that points to a record of the form `{address ; chain_id}`. Given such an address a wallet can verify it resolves to a valid account.
+The advantage of this format is that it is very flexible and can accommodate a number of use cases, however it can also lead to confusion for users because a name does not necessarily resolve to a valid account. The same `user.eth` could lead to a website, a NFT or multiple addresses.
+
+The resolution of a `address@chain` on the contrary, imposes that the left-hand resolves to an address and the right-hand to a chain identifier.
+
+When given a `user@rollup.eth`, the wallet can resolve `rollup.eth` to get a chain identifier and `user.rollup.eth` to get an address. In any other case it fails.
```
niran, line 128: This doesn't require a new syntax: When the wallet is resolving `user.rollup.eth`, the standard the resolution process should be resolve to resolve the way you described: resolve `rollup.eth` to get a chain identifier and `user.rollup.eth` to get an address.

This approach also allows `user.rollup.eth` to override this behavior by setting a record for their own preferred chain identifier to use when their ENS name is resolved. `*.rollup.eth` names will likely never override this, but I'd expect almost every `user.eth` to want to receive funds on some chain that isn't L1.

ERCS/erc-chain_specific_addresses.md
```diff
@@ -0,0 +1,235 @@
...
+
+In the above proposal the ENS TLD and chain name coincide which may be confusing or incorrect in some cases. A more explicit approach could be to have an additional suffix for the chain name.
+Example:
+```
+name.eth.mainnet -> {L1_chain_id : 0x1;      L1_ens_address : <ENS-address-on-mainnet>}
+name.eth.sepolia -> {L1_chain_id : 0xaa36a7; L1_ens_address : <ENS-address-on-sepolia>}
+```
+### Note: default fallbacks
+
+If a user receives a legacy address without chain name, the wallet can:
```
niran, line 165: I don't think any of these fallbacks should be our target behavior. Users should be able to control on which chain assets sent to their ENS name arrive. I think ENS experts should lead the way on this resolution process via an ENSIP.

ERCS/erc-chain_specific_addresses.md
```diff
@@ -0,0 +1,235 @@
+---
```
abcoathup, line 1:
```suggestion
---
eip: 7828
```

ERCS/erc-chain_specific_addresses.md
```diff
@@ -0,0 +1,235 @@
...
+title: Chain-specific addresses using ENS
+description: A unified chain-specific address format that allows specifying the account as well as the chain on which that account intends to transact. 
+author: Sam Kaufman (@SampkaML), Marco Stronati (@paracetamolo), Yuliya Alexiev (@yuliyaalexiev), Jeff Lau (@jefflau), Sam Wilson (@samwilsn), Vitalik Buterin (@vbuterin)
+discussions-to: <URL>
```
abcoathup, line 5: Please create a discussions topic in Eth Magicians.  

It just needs to be a link to this PR.

ERCS/erc-chain_specific_addresses.md
```diff
@@ -0,0 +1,235 @@
...
+```
+L1-TLD ::= eth | sepolia | …
+chain_id ::= 0x[a-fA-F0-9]{1,64}
+address ::= 0x[a-fA-F0-9]{40}
+chain ::= <chain_id> | <L1-TLD> | <ens-name> . <L1-TLD>
+user ::= <address> | <ens-subdomain>
+account ::= <user>@<chain>
+```
+
+Note the difference between `ens-name`, which is a full ENS name, and `ens-subdomain` that is just a segment of a name between dots. E.g. `user.app.eth` is a name, `user` and `app` are subdomains.
```
vbuterin, line 61: In the below examples, isn't the user also just `<address> | <ens-name>`?

ERCS/erc-7828.md
```diff
@@ -0,0 +1,236 @@
+---
+eip: 7828
+title: Chain-specific addresses using ENS
+description: A unified chain-specific address format that allows specifying the account as well as the chain on which that account intends to transact. 
+author: Sam Kaufman (@SampkaML), Marco Stronati (@paracetamolo), Yuliya Alexiev (@yuliyaalexiev), Jeff Lau (@jefflau), Sam Wilson (@samwilsn), Vitalik Buterin (@vbuterin)
+discussions-to: https://ethereum-magicians.org/t/erc-7828-chain-specific-addresses/21930
```
abcoathup, line 6:
```suggestion
discussions-to: https://ethereum-magicians.org/t/erc-7828-chain-specific-addresses-using-ens/21930
```
Updated Eth Magicians title to match ERC title

lukepuplett:
My apologies if I'm in the wrong place, I've not done this before. I'm here to comment on this proposal. My feedback is simply to think about how it might map to blockchain agnostic representations, notably this one:

https://github.com/ChainAgnostic/CAIPs/blob/main/CAIPs/caip-10.md

Thank you.

ERCS/erc-chain_specific_addresses.md
yoavw: In the examples below it's indeed an <ens-name> but if I understand correctly it could also use an `ens-subdomain` such as  `user@arbitrum.eth` which might mean something different from `user.eth@arbitrum.eth`.

The former would resolve `user` locally on arbitrum, and the latter (being a fully-qualified name) would resolve `user.eth` on L1. Is that not the intention?

ERCS/erc-7828.md
```diff
@@ -0,0 +1,236 @@
...
+
+### A special case for ENS resolution
+
+Any ENS name today can be resolved to a chain identifier or to an address.
+We could imagine having a name `user.eth` that points to a record of the form `{address ; chain_id}`. Given such an address a wallet can verify it resolves to a valid account.
+The advantage of this format is that it is very flexible and can accommodate a number of use cases, however it can also lead to confusion for users because a name does not necessarily resolve to a valid account. The same `user.eth` could lead to a website, a NFT or multiple addresses.
+
+The resolution of a `address@chain` on the contrary, imposes that the left-hand resolves to an address and the right-hand to a chain identifier.
+
+When given a `user@rollup.eth`, the wallet can resolve `rollup.eth` to get a chain identifier and `user.rollup.eth` to get an address. In any other case it fails.
```
iJaack, line 129: We'd need to add that the resolution would be possible thanks to ERC-7785 (#669)

ERCS/erc-7828.md
```diff
@@ -0,0 +1,236 @@
...
+Otherwise the user can be more explicit and give as recipient `rollup1.user.eth`.
+
+### Note: Speculative
+
+Alternatively we could store multiple addressed under the same domain as
+```
+user.eth -> { rollup1 : address * chain_id ; 
+              rollup2 : address * chain_id } 
+```
+if a syntax to access ENS records could be standardized, the user could be asked to be paid at
```
iJaack, line 192: This could be possible with Namechain

ERCS/erc-chain_specific_addresses.md
iJaack: This actually opens up a very big issue that I believe it’s not discussed here: in theory, `user` should be`user` across the whole Ethereum ecosystem at least, but in fact there are many Domain Name Services beyond ENS that are local to some L2s or even just one.

For example, I have jaack.eth, and also registered jaack.avax on Avalanche C-Chain and jaack.blast on Blast.

First, I believe we should consider ENS as the ‘main router’ for DNS resolutions, because it’s stated as a requirement in the ERC. Since this is the case, then resolution should be global across Ethereum, including rollups. So `user` needs to be `user.eth on any chain, especially considering that this ERC would likely be implemented around the time that Namechain will be at least on testnet, so the ENS registry is itself on a ‘service’ L2.

@yoavw the case where an address is local only to its chain is the one where most addresses owned by users are smart wallets / accounts that are not the same across chains (unlike EOAs, that have the same pubkey / privkey across any EVM).

iJaack:
@lucemans I believe ENSIP-9 doesn’t actually solve the issue, but gives redundancy in case this only gets applied to the stricter Ethereum ecosystem. The whole reason to do this multi-chain resolution is to allow for ‘clean names’ based on a widely used primitive: `chainID`. ENSIP-9 builds on top of this with another identification system, while this ERC is an alternative.

BTW thank you for mentioning, it’s actually a great feature of ENS that could be publicized more!

SamWilsn:
Please keep technical discussion on the forum: https://ethereum-magicians.org/t/erc-7828-chain-specific-addresses-using-ens/21930

ERCS/erc-7828.md
```diff
@@ -0,0 +1,236 @@
...
+chain ::= <chain_id> | <L1-TLD> | <ens-name> . <L1-TLD>
+user ::= <address> | <ens-subdomain>
+account ::= <user>@<chain>
+```
+
+Note the difference between `ens-name`, which is a full ENS name, and `ens-subdomain` that is just a segment of a name between dots. E.g. `user.app.eth` is a name, `user` and `app` are subdomains.
+
+A few examples below. 
+
+Option 1: using @ to separate address and chain
```
tynes, line 66: I think option 1 is better both aesthetically and verbally, its intuitive to say "at" a location which will make it easier for people with less context to understand

ERCS/erc-7828.md
```diff
@@ -0,0 +1,236 @@
...
+
+Two desired properties:
+1) checksums are MANDATORY
+2) checksum hash goes over the entire address, so users can't just go and replace a component and expect it to stay valid
+
+
+### A special case for ENS resolution
+
+Any ENS name today can be resolved to a chain identifier or to an address.
+We could imagine having a name `user.eth` that points to a record of the form `{address ; chain_id}`. Given such an address a wallet can verify it resolves to a valid account.
```
tynes, line 124: What exactly does valid account in this context mean?

ERCS/erc-7828.md
```diff
@@ -0,0 +1,236 @@
...
+
+### Note: Speculative
+
+Alternatively we could store multiple addressed under the same domain as
+```
+user.eth -> { rollup1 : address * chain_id ; 
+              rollup2 : address * chain_id } 
+```
+if a syntax to access ENS records could be standardized, the user could be asked to be paid at
+`user.eth/rollup1`
```
tynes, line 193: [RFC 8615]( https://en.wikipedia.org/wiki/Well-known_URI) defines something similar to this, an HTTP server would serve from `user.eth/.well-known/addresses.json` (we can come up with an exact filename). I suppose it depends on if we want this legible onchain

ERCS/erc-7828.md
iJaack: I believe a valid account is one that had at least one transaction made, either IN or OUT?

Meaning that is an active address in some form.

ERCS/erc-7828.md
```diff
@@ -0,0 +1,236 @@
...
+We could imagine having a name `user.eth` that points to a record of the form `{address ; chain_id}`. Given such an address a wallet can verify it resolves to a valid account.
+The advantage of this format is that it is very flexible and can accommodate a number of use cases, however it can also lead to confusion for users because a name does not necessarily resolve to a valid account. The same `user.eth` could lead to a website, a NFT or multiple addresses.
+
+The resolution of a `address@chain` on the contrary, imposes that the left-hand resolves to an address and the right-hand to a chain identifier.
+
+When given a `user@rollup.eth`, the wallet can resolve `rollup.eth` to get a chain identifier and `user.rollup.eth` to get an address. In any other case it fails.
+
+### L2 resolution
+
+In case an address is not registered on L1, but only on a L2, the resolution can be processed using [ERC-3668](./eip-3668.md) and [ERC-2544](./eip-2544.md)'s  Wildcard Resolution.
```
clowestab, line 133: Broken links. 

2544 is more commonly known as ENSIP-10 => https://docs.ens.domains/ensip/10

ERCS/erc-7828.md
```diff
@@ -0,0 +1,236 @@
...
+Valid addresses MUST include the identifier of the chain they belong to.
+
+```
+L1-TLD ::= eth | sepolia | …
+chain_id ::= 0x[a-fA-F0-9]{1,64}
+address ::= 0x[a-fA-F0-9]{40}
+chain ::= <chain_id> | <L1-TLD> | <ens-name> . <L1-TLD>
+user ::= <address> | <ens-subdomain>
+account ::= <user>@<chain>
+```
```
SamWilsn, line 60: It would be nice to note what format these snippets are in.

ERCS/erc-7828.md
```diff
@@ -0,0 +1,236 @@
...
+- alice.eth@arbitrum.eth
+
+Rollup Sepolia
+- 0x12345...6789@arbitrum.sepolia
+
+My ENS name is registered on rollup1, but I want to receive funds on rollup2
+- alice.rollup1.eth@rollup2.eth
+```
+
+Option 2: using : instead of @
```
SamWilsn, line 90: You should write proposals as if they are already Final. It gets confusing otherwise. If you have multiple options, choose one for now and `<!-- comment out -->` the rest.

ERCS/erc-7828.md
```diff
@@ -0,0 +1,236 @@
...
+
+My ENS name is registered on rollup1, but I want to receive funds on rollup2
+- alice.rollup1.eth:rollup2.eth
+```
+### CHECKSUM
+<!-- TODO: add more explanation here -->
+
+Two desired properties:
+1) checksums are MANDATORY
+2) checksum hash goes over the entire address, so users can't just go and replace a component and expect it to stay valid
```
SamWilsn, line 118: This type of content belongs in the Rationale section, as it is describing why you made particular choices. Keep the Specification section as close to just requirements as possible so it's easy for an implementer to follow.

ERCS/erc-7828.md
SamWilsn, line 129: This section is _really_ mixing rationale and specification. I can't tell if [line 129](https://github.com/ethereum/ERCs/pull/735/files#diff-c4fc873c59b4ff9320791b6b72317f03b6a87c64e4de2aaad171ba775f690e14R129) is a requirement ("In any other case it fails.") or just context.

ERCS/erc-7828.md
SamWilsn: We don't allow links to ENSIPs, but if you'd like to request an exception the process is documented in [EIP-5757](https://eips.ethereum.org/EIPS/eip-5757).

ERCS/erc-7828.md
```diff
@@ -0,0 +1,236 @@
...
+### L2 resolution
+
+In case an address is not registered on L1, but only on a L2, the resolution can be processed using [ERC-3668](./eip-3668.md) and [ERC-2544](./eip-2544.md)'s  Wildcard Resolution.
+
+In the previous example `user@rollup.eth`, `user` would not be registered on L1.
+In this case the wallet can resolve `rollup.eth` to get a chain identifier as before and when attempting to resolve `user.rollup.eth` to get an address, it would fail and be redirected to the L2 gateway. Any answer from the gateway needs to be verified as explained in the EIP.
+
+#### Note: avoiding the http gateway
+
+In order to avoid contacting an external http gateway, we could define the gateway to be a ENS contract on the L2. In this way a wallet operator would need to only rely on a node following the L1 and a node following the L2.
```
SamWilsn, line 140: Who is this an instruction for (like who is "we" here)? If this is a note to the authors, wrap it in a TODO comment:

```suggestion
<!-- TODO:
In order to avoid contacting an external http gateway, we could define the gateway to be a ENS contract on the L2. In this way a wallet operator would need to only rely on a node following the L1 and a node following the L2.
-->
```

ERCS/erc-7828.md
```diff
@@ -0,0 +1,236 @@
...
+Rollup
+- 0x12345...6789@chainId
+- 0x12345...6789@arbitrum.eth
+- alice.eth@arbitrum.eth
+
+Rollup Sepolia
+- 0x12345...6789@arbitrum.sepolia
+
+My ENS name is registered on rollup1, but I want to receive funds on rollup2
+- alice.rollup1.eth@rollup2.eth
```
u59149403, line 87: Too quirky. It is too hard to understand that `alice.eth` part of `alice.eth@arbitrum.eth` should be resolved on `arbitrum.eth` chain, but `alice.rollup1.eth` part of `alice.rollup1.eth@rollup2.eth` should be resolved on `rollup1.eth`

ERCS/erc-7828.md
u59149403, line 90: `:` implies totally different hierarchy. I. e. `a:b` means that `b` is located inside of `a`

ERCS/erc-7828.md
```diff
@@ -0,0 +1,236 @@
...
+In the previous example `user@rollup.eth`, `user` would not be registered on L1.
+In this case the wallet can resolve `rollup.eth` to get a chain identifier as before and when attempting to resolve `user.rollup.eth` to get an address, it would fail and be redirected to the L2 gateway. Any answer from the gateway needs to be verified as explained in the EIP.
+
+#### Note: avoiding the http gateway
+
+In order to avoid contacting an external http gateway, we could define the gateway to be a ENS contract on the L2. In this way a wallet operator would need to only rely on a node following the L1 and a node following the L2.
+
+### L1 resolution
+
+Ethereum Mainnet and its testnets can be resolved to their corresponding chain identifiers using a centralized list, which remains unchanged from how it works today. Other L1 registrations are out of scope for this EIP.
```
u59149403, line 144: > Other L1 registrations are out of scope for this EIP

I totally don't like this. Other L1s should be supported somehow

ERCS/erc-7828.md
```diff
@@ -0,0 +1,236 @@
...
+```
+Example:
+```
+eth     -> {L1_chain_id : 0x1;      L1_ens_address : <ENS-address-on-mainnet>}
+sepolia -> {L1_chain_id : 0xaa36a7; L1_ens_address : <ENS-address-on-sepolia>}
+```
+
+### Note: clashes of L1 and TLD
+
+In the above proposal the ENS TLD and chain name coincide which may be confusing or incorrect in some cases. A more explicit approach could be to have an additional suffix for the chain name.
```
u59149403, line 158: I totally agree with [SamWilsn](https://github.com/SamWilsn) comment above: ERCs should be written as if they are already final. So, just stick to some particular approach instead of writing "A more explicit approach could be..."

u59149403:
@lucemans

> Have a look at ENSIP-9 (defining different records for different chains)
> https://docs.ens.domains/ensip/9

ENSIP-9 tries to support not only EVM chains, but also bitcoin, etc. This is good thing. But it invents its own way to refer to various chains. This is bad. New version of ENSIP-9 should be created, which refers to caip-10 ( https://chainagnostic.org/CAIPs/caip-10 ) instead. Everybody should stick to CAIP-10. See also this very good comment: https://ethereum-magicians.org/t/chain-specific-addresses/6449/28 by rekmarks

ERCS/erc-7828.md
```diff
@@ -0,0 +1,284 @@
...
+In the same spirit, the address could be a human-readable name as well, which is already a use case for ENS. By coupling the TLD to the resolving method used, this standard could leverage current and future features of ENS and other naming registries.
+
+Moreover, the above can be leveraged to allow for a name to represent the same entity corresponding to different addresses on different chains, mitigating the risk of sending funds to an address the intended recipient doesn't actually control.
+
+Desired properties:
+- A unified format to represent any address on L1, L2 or beyond.
+- The chain portion can be an ERC-7785 domain name when that standard is ready, or use solutions similar to ethereum-lists/chains in the immediate future.
+- The address portion can be either the appropriate type of address for the chain, or a domain name.
+- The address portion and the chain portion should be resolved separately.
+- Checksum hash goes over the entire address & chain, so users can't just replace a component and expect it to stay valid.
```
u59149403, line 41: Checksum is needed for cases, when user manually types address, which does not have any intrinsic meaning for the user. For example, `0xdAC17F958D2ee523a2206206994597C13D831ec7`. If the user types such address manually, then they can easily make mistake. And checksum here helps: it minimizes probability of mistake.

But if we have meaningful address, such as `vitalik.eth`, then we don't need checksum. Because checksum here makes typing of this address harder, not easier!

`vitalik.eth#2cb81d22c001` is harder both to type and memorize than `vitalik.eth`. We don't have checksums in meaningful URLs, right? We type `github.com` to URL bar, not `github.com#2cb81d22c001`, right?

Okay, so, as well as I understand, ERC-7828 is designed for addresses, meaningful for humans, as opposed to CAIP-10, CAIP-50, CAIP-350 and ERC-7930. So it should not have checksums at all

ERCS/erc-7828.md
```diff
@@ -0,0 +1,284 @@
...
+
+Assuming ERC-7785 starts listing chains at l2.eth
+- alice.eth@arb-nova.l2.eth#AAAAAAAA
+
+My ENS name is registered on rollup1 (via ENSIP-10 & ENSIP-16), but I want to receive funds on rollup2
+- alice.rollup1.eth@rollup2.short#BBBBBBBB
+
+Non-evm chain
+- bc1..23@bip122:000000000019d6689c085ae165831e93#CCCCCCCC
+- alice.eth@bip122:000000000019d6689c085ae165831e93#CCCCCCCC
```
u59149403, line 97: Where this `alice.eth` is resolved? Please, specify this. It it meant to be resolved on Ethereum, then I totally don't understand why in `alice.eth@bip122:000000000019d6689c085ae165831e93#CCCCCCCC` the `alice.eth` part should be resolved on Ethereum and in `alice.eth@arb-nova.short#AAAAAAAA` the `alice.eth` part should be resolved in Arbitrum Nova. In any case, all this should be specified

ERCS/erc-7828.md
```diff
@@ -0,0 +1,284 @@
...
+1. Let the user input an address name. Assume it's `alice.eth`
+2. Let the user select a chain. Assume it's `oeth.short`
+3. Look up the TLD (characters from the last dot up to the end of the string) of the result of step 2. For this example it's `.short`.
+4. Look up the result from the step above on the [name resolver registry] to know what to do with the entirety of the result from step 2. In this case, the hardcoded table says it corresponds to
+    - chain namespace: `eip155`
+    - chain reference: `10` (chainid of OP Mainnet)
+5. Look up the TLD (characters from the last dot up to the end of the string) of the result of step 1. For this example it's `.eth`, meaning we should attempt resolution via ENS
+6. As per ENSIP-11, convert the eip155 chainid into an ENSIP-11-specific `coinType`: `0x80000000 & 0x0000000A == 0x8000000A` [^1]
+7. Compute the namehash of the result of step 1, according to ENSIP-1.
+8. Query the ENS registry for the appropiate resolver by calling `resolver(bytes32 node)` with the result of the step above
```
u59149403, line 214: Which ENS registry I should call? Ethereum's one or oeth's one?

ERCS/erc-7828.md
0xteddybear: >  Because checksum here makes typing of this address harder, not easier!

it's not necessary for users to type the checksum, the idea would be for the users to type `name.tld@chain.tld` and the wallet to generate the checksum based on what address said name resolves to.

> Okay, so, as well as I understand, ERC-7828 is designed for addresses, meaningful for humans, as opposed to CAIP-10, CAIP-50, CAIP-350 and ERC-7930. So it should not have checksums at all

I believe checksums have a place in human-readable names, since they could help prevent:
- unicode glyph collision attacks (`vitаlik.eth`)
- squatting of domains with common typos (`vitslik.eth`)

And they could also provide a link between addresses representing the same on-chain entity (e.g, `alice.eth@eth.short#FFFFFFFF` might display as `0x12345...6789@eth.short#FFFFFFFF` on a wallet which has trouble resolving the ENS name) which gives user the idea of the addresses _potentially_ being the same, which would make simple failures less obscure

ERCS/erc-7828.md
0xteddybear: I tried to make this part more legible as a repository of possible name formats than a normative specification on how to resolve each one, which in the scope for later sections.

In particular, the core of this proposal is for the `.eth` part to determine how should the name be resolved. By using `.eth`, this name implies it should be resolved via ENS, and what contract to call is an ENS detail (in particular, ENS resolution always *starts* on ethereum mainnet, but might jump somewhere else via CCIP calls)

ERCS/erc-7828.md
0xteddybear: in the address naming section, we specify the following:

> Resolution and reverse-resolution of chain names is to be executed as defined in ENSIP-19, ENSIP-16, ENSIP-11 and ENSIP-10.

in particular (and as described in step 6) we are using ENSIP-11, which instructs resolvers to use the mainnet ENS registry

ERCS/erc-7828.md
0xteddybear: taken care of :sunglasses:

ERCS/erc-7828.md
```diff
@@ -0,0 +1,284 @@
...
+- By coupling the domain's TLD to a known resolution method, we avoid the case where the same human-readable name is registered on different naming registries pointing to different addresses, which would only have the address' 4-byte checksum as a mitigation on users sending funds to unintended addresses.
+- Maintaining a list of chain/name registries instead of a list of chains means it'd have to be updated less frequently, allowing for more extensive review of new entries.
+
+## Copyright
+
+Copyright and related rights waived via [CC0](../LICENSE.md).
+
+[ERC-7930]: ./ERC-7930.md
+[ERC-7785]: ./ERC-7785.md
+[name resolver registry]: TODO
```
mm-zk, line 284: do we have any link / draft doc for name resolver registry?

ERCS/erc-7828.md
```diff
@@ -0,0 +1,284 @@
...
+> This list does not define an example for the case where no resolution is desired, as in that case an Interoperable Address v2 can be used instead
+<!-- TODO: should we actually define it so we can produce an Interoperable Address without a chainid, but with ENS resolution for the address? -->
+
+#### `.short`
+Binary id: `0x0000`
+
+##### Chain naming
+
+###### Supported CASA namespaces
+Only those explicitly listed here
```
mm-zk, line 147: you might want to mention that we want to include all the stuff from ethereum-lists/chains (otherwise the reader thinks that we will have only 5 entries here ;-)

ERCS/erc-7828.md
```diff
@@ -0,0 +1,284 @@
...
+- 0x12345...6789@eip155:1#FFFFFFFF
+- 0x12345...6789@eth.short#FFFFFFFF
+- alice.eth@eth.short#FFFFFFFF
+
+Testnet (Sepolia)
+- 0x12345...6789@eip155:11155111#00000000
+- alice.testeth@sep.short#00000000
+
+Rollup
+- 0x12345...6789@eip155:4270#AAAAAAAA
```
mm-zk, line 85: what would be the 'chain resolver' set to in this example? 
(as we have only 0x0000 (which is for .short) and 0x0001 (for .eth))

ERCS/erc-7828.md
0xteddybear: that's a good point, my idea is to remove this list and have that information in the name resolver registry, similar to what I did on ERC-7930 with CAIP-350

ERCS/erc-7828.md
```diff
@@ -0,0 +1,284 @@
...
+- A unified format for accounts that specifies, together with the address, the chain where the address lives.
+- The use of human-readable chain names and how they can be resolved to chain identifiers.
+- The use of human-readable account names and how they can be resolved to addresses.
+- A centralized index of chain and account resolution methods, way smaller in scope than ethereum-lists/chains
+
+## Motivation
+The current Ethereum address landscape is leading to an ecosystem that will have hundreds and eventually thousands of L2s that use the same address format as Ethereum mainnet. This means an address by itself is not enough information to know which chain the address is related to. This can be problematic if funds are sent to an unreachable address on the incorrect chain. From the user account it should be possible to obtain the right chain identifier (chainID) to include in a transaction. 
+
+The mapping from chain names to identifiers has, since [EIP-155](./eip-155.md), been maintained off chain using a centralized list. This solution has a few shortcomings:
```
kelemeno, line 24: Deprecating/decentralising this list has a danger, as now the chainIds of different chains could clash.

ERCS/erc-7828.md
```diff
@@ -0,0 +1,284 @@
...
+<human readable name>: := <address>@<chain>#<checksum>
+<address>:             := <raw-address> | <scoped-name>
+<chain>:               := <raw-address> | <scoped-name>
+<checksum>:            := [0-9A-F]{8}
+
+<raw-address>:         := [.-:_%a-zA-Z0-9]*
+<raw-chain>:           := [.-:_a-zA-Z0-9]*
+<scoped-name>:         := <subdomain> . <L0-TLD>
+<L0-TLD>:              := [a-zA-Z][-a-zA-Z0-9]{0,9}
+<subdomain>:           := [a-zA-Z][-a-zA-Z0-9]*
```
nategraf, line 60: One thing to clarify about this grammar, which I think is true based on your comment below: multiple levels of subdomains are allowed, ya? So in the `user.app.eth` example you have the layers of `user.app.eth` -> `user.app` -> `.eth`. I think this currently would not parse, if I read this right.

nategraf (commented):
Thanks for putting this together!

ERCS/erc-7828.md
0xteddybear: > would be better to not imply that this ERC completely deprecates it.

what wording in particular implies this standard deprecates it? My intention with this standard has always been to be a building block to _eventually_ deprecate it, but this ERC alone is not enough to deprecate it. That's a concern of ERC-7785, with which am I not involved

ERCS/erc-7828.md
0xteddybear: good find, thanks! I'll re-define it so there can be multiple submdomain levels

ERCS/erc-7828.md
```diff
@@ -0,0 +1,293 @@
...
+discussions-to: https://ethereum-magicians.org/t/erc-7828-chain-specific-addresses-using-ens/21930
+status: Draft
+type: Standards Track
+category: ERC
+created: 2024-11-27
+requires: 55, 137, 155, 7785
+---
+
+## Abstract
+This proposal builds off of [ERC-7930] (Interoperable Addresses & Names) to provide a standard and human-readable format for chain-specific addresses which provides:
```
lucemans, line 15:
```suggestion
## Abstract

This proposal builds off of [ERC-7930] (Interoperable Addresses & Names) to provide a standard and human-readable format for chain-specific addresses which provides:
```

ERCS/erc-7828.md
```diff
@@ -0,0 +1,293 @@
...
+- A unified format for accounts that specifies, together with the address, the chain where the address lives.
+- The use of human-readable chain names and how they can be resolved to chain identifiers.
+- The use of human-readable account names and how they can be resolved to addresses.
+- A centralized index of chain and account resolution methods, way smaller in scope than ethereum-lists/chains
+
+## Motivation
+The current Ethereum address landscape is leading to an ecosystem that will have hundreds and eventually thousands of L2s that use the same address format as Ethereum mainnet. This means an address by itself is not enough information to know which chain the address is related to. This can be problematic if funds are sent to an unreachable address on the incorrect chain. From the user account it should be possible to obtain the right chain identifier (chainID) to include in a transaction.
```
lucemans, line 22:
```suggestion
## Motivation

The current Ethereum address landscape is leading to an ecosystem that will have hundreds and eventually thousands of L2s that use the same address format as Ethereum mainnet. This means an address by itself is not enough information to know which chain the address is related to. This can be problematic if funds are sent to an unreachable address on the incorrect chain. From the user account it should be possible to obtain the right chain identifier (chainID) to include in a transaction. 
```

ERCS/erc-7828.md
```diff
@@ -0,0 +1,293 @@
...
+Desired properties:
+- A unified format to represent any address on L1, L2 or beyond.
+- The chain portion can be an ERC-7785 domain name when that standard is ready, or use solutions similar to ethereum-lists/chains in the immediate future.
+- The address portion can be either the appropriate type of address for the chain, or a domain name.
+- The address portion and the chain portion should be resolved separately.
+- Checksum hash goes over the entire address & chain, so users can't just replace a component and expect it to stay valid.
+- Addresses can be serialized to binary and then de-serialized back to text retaining the same human-readable name, even if serialization and de-serialization happen on different wallets.
+
+## Specification
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
```
lucemans, line 45:
```suggestion
## Specification

The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
```

ERCS/erc-7828.md
```diff
@@ -0,0 +1,293 @@
...
+
+### Format
+This standard defines a sub-syntax with extra semantics on top of the syntax defined in [ERC-7930].
```
lucemans, line 48:
```suggestion
### Format

This standard defines a sub-syntax with extra semantics on top of the syntax defined in [ERC-7930].
```

ERCS/erc-7828.md
```diff
@@ -0,0 +1,293 @@
...
+
+Assuming the shortcode list adds a few other CAIP namespaces
+- alice.eth<>btc.short#CCCCCCCC
+```
+
+> [!NOTE]
+> This standard does not explicitly define a hierarchical name structure for anything other than the top level domain, however it allows the underlying naming scheme to be hierarchical as in the ENS-in-rollup example.
+
+### Checksum
+Addresses must be serialized to the `ChainType, ChainReferenceLength, ChainReference, AddressLength, Address` format proposed in [ERC-7930] as Interoperable Addresses v1, and the 4-byte checksum MUST be displayed as part of the address as described in the syntax above.
```
lucemans, line 108:
```suggestion
### Checksum

Addresses must be serialized to the `ChainType, ChainReferenceLength, ChainReference, AddressLength, Address` format proposed in [ERC-7930] as Interoperable Addresses v1, and the 4-byte checksum MUST be displayed as part of the address as described in the syntax above.
```

ERCS/erc-7828.md
```diff
@@ -0,0 +1,293 @@
...
+
+This ensures that the checksum changes if any part of the address or chain ID is modified, turning dangerous manual edits into detectable errors for wallet users.
+
+### The Name Resolver Registry
+We propose for this ERC to strive to reach Final status instead of becoming a Living document, and maintain a separate dictionary for the different name resolving protocols or methods, similar to ethereum-lists/chains.
```
lucemans, line 113:
```suggestion
### The Name Resolver Registry

We propose for this ERC to strive to reach Final status instead of becoming a Living document, and maintain a separate dictionary for the different name resolving protocols or methods, similar to ethereum-lists/chains.
```

ERCS/erc-7828.md
```diff
@@ -0,0 +1,293 @@
...
+It is worth noting that:
+- The same `<L0-TLD>` can map to different algorithms and/or data for chains and addresses. e.g. if ERC-7785 is published as an ENS product, using `.eth` on a chain name will use a different algorithm for resolution than if the same TLD were used on an address.
+- A given protocol may have more than one TLD assigned to it, to avoid having to store the chainid & address of the name resolver on the Interoperable Address itself. An example of this would be to use different addresses for the testnet & mainnet ENS deployments.
+
+This is an easier list to maintain than ethereum-lists/chains, since it does not have to list every single chain, and the problem of knowing which contracts are trustworthy as naming registries is one wallets would have to tackle anyway.
+
+### Example entries for the `L0-TLD` list
+
+#### No resolution
+Binary id: `0x0000`
```
lucemans, line 138:
```suggestion
#### No resolution

Binary id: `0x0000`
```

ERCS/erc-7828.md
```diff
@@ -0,0 +1,293 @@
...
+
+To be used in v2 Interoperable Addresses to indicate that no resolution is desired, and the 7930-style raw chain identifier used in the Interoperable Name instead.
+
+> [!NOTE]
+> In the case where both `ChainResolver` and `AddressResolver` are set to this value, the resulting Interoperable Name would match an Interoperable Name from ERC-7930
+
+#### `.short`
+Binary id: `0x0001`
```
lucemans, line 146:
```suggestion
#### `.short`

Binary id: `0x0001`
```

ERCS/erc-7828.md
```diff
@@ -0,0 +1,293 @@
...
+
+##### Chain naming
+
+###### Supported CASA namespaces
+Only those explicitly listed here
```
lucemans, line 151:
```suggestion
###### Supported CASA namespaces

Only those explicitly listed here
```

ERCS/erc-7828.md
```diff
@@ -0,0 +1,293 @@
...
+
+###### Resolution algorithm
+Simply refer to the table below
```
lucemans, line 154:
```suggestion
###### Resolution algorithm

Simply refer to the table below
```

ERCS/erc-7828.md
```diff
@@ -0,0 +1,293 @@
...
+| `eth`      | `eip155`         | `1`                                |
+| `oeth`     | `eip155`         | `10`                               |
+| `arb-nova` | `eip155`         | `4270`                             |
+| `btc`      | `bip122`         | `000000000019d6689c085ae165831e93` |
+| `solana`   | `solana`         | `5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp` |
+
+<!-- TODO: we could fill it programmatically from ethereum-lists/chains or instead refer to it for eip155 -->  
+
+##### Address naming
+Not supported
```
lucemans, line 167:
```suggestion
##### Address naming

Not supported
```

ERCS/erc-7828.md
```diff
@@ -0,0 +1,293 @@
...
+
+#### `.eth`
+Binary id: `0x0002`
+
+###### Supported CASA namespaces
+`eip155` and anything supported by SLIP-0044 for forward resolution (Interoperable Address -> human-readable name)
```
lucemans, line 173:
```suggestion
###### Supported CASA namespaces

`eip155` and anything supported by SLIP-0044 for forward resolution (Interoperable Address -> human-readable name)
```

ERCS/erc-7828.md
```diff
@@ -0,0 +1,293 @@
...
+
+`eip155` chains explicitly supported by ENS by deployment of `L2ReverseRegistrar` and registration on ENS registry for reverse resolution (human-readable name -> Interoperable Address)
+
+##### Chain naming
+Not currently supported, TBD in ERC-7785
```
lucemans, line 178:
```suggestion
##### Chain naming

Not currently supported, TBD in ERC-7785
```

ERCS/erc-7828.md
```diff
@@ -0,0 +1,293 @@
...
+
+##### Address naming
+<!-- TODO: specify this better -->
+Resolution and reverse-resolution of chain names is to be executed as defined in ENSIP-19, ENSIP-16, ENSIP-11 and ENSIP-10.
```
lucemans, line 182:
```suggestion
##### Address naming

<!-- TODO: specify this better -->
Resolution and reverse-resolution of chain names is to be executed as defined in ENSIP-19, ENSIP-16, ENSIP-11 and ENSIP-10.
```

ERCS/erc-7828.md
```diff
@@ -0,0 +1,293 @@
...
+
+Some caveats for ENS support are:
+- ENSIP-6 and ENSIP-17 are explicitly not supported, as they define names on different TLDs.
+- While ENS could add support for other TLDs for forward resolution, this standard would require another entry on this list to support them. This does not apply to _reverse_, resolution, since those names are not ever displayed.
+- For addresses on chains that are both supported by SLIP-0044 and the special scheme for the eip155 namespace defined in ENSIP-11, the latter should be used.
+<!-- TODO: probably a few more edge cases -->
+
+### Interoperable Address v2 definition
+To allow for addresses to be serialized from human-readable names to an Interoperable Address, and then back to _the same_ human-readable name we define a new version of Interoperable Addresses which includes information on the desired naming registry.
```
lucemans, line 191:
```suggestion
### Interoperable Address v2 definition

To allow for addresses to be serialized from human-readable names to an Interoperable Address, and then back to _the same_ human-readable name we define a new version of Interoperable Addresses which includes information on the desired naming registry.
```

ERCS/erc-7828.md
```diff
@@ -0,0 +1,293 @@
...
+Where:
+<!-- TODO: should we go back to ERC-7930 to define that a given bit should mean it can also be parsed as v1? because it'd be the case for this one -->
+- `Version` : A 2-byte version identifier. must be big-endian unsigned integer `2`.
+- `ChainType`, `ChainReferenceLength`, `ChainReference`, `AddressLength`, `Address`: Same as in [ERC-7930]
+- `ChainResolver`: 2-byte identifier of the method used to resolve the chain to a human-readable string, from the [^name resolver registry].
+- `AddressResolver`: 2-byte identifier of the method used to resolve the address to a human-readable string, from the [^name resolver registry].
+<!-- TODO: should we skip them if ChainReferenceLength / AddressLength are zero? -->
+
+### Forward resolution step-by-step example
+1. Let the user input an address name. Assume it's `alice.eth`
```
lucemans, line 209:
```suggestion
### Forward resolution step-by-step example

1. Let the user input an address name. Assume it's `alice.eth`
```

ERCS/erc-7828.md
```diff
@@ -0,0 +1,293 @@
...
+10. Begin serialization into an Interoperable Address v2 by appending the version: `[0002]`
+11. Serialize `ChainType`, `ChainReferenceLength`, `ChainReference`, `AddressLength` and `Address` according to [ERC-7930]: `[ 0002 0000 01 0A 14 AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA ]`
+12. Append the binary id for the `ChainResolver` and `AddressResolver` from the Name Resolver Registry:  `[ 0002 0000 01 0A 14 AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA 0000 0001]`
+13. Interoperable Address is complete: `0x00020000010A14AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA00000002`
+14. Compute the checksum as described in [ERC-7930]: `0xC69BEB13` and display it to the user.
+
+[^1]: While this sees production usage, it is not consistent with the design principles behind [ERC-7930] and assumes 32-bit chainids, which are not compatible with an [ERC-7785] future. We recommend using something closer to what is described in footnote 2.
+
+### Reverse Resolution step-by-step example
+Starting from the Interoperable Address serialized above: `0x00020000010A14AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA00010002`
```
lucemans, line 230:
```suggestion
### Reverse Resolution step-by-step example

Starting from the Interoperable Address serialized above: `0x00020000010A14AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA00010002`
```

ERCS/erc-7828.md
```diff
@@ -0,0 +1,293 @@
...
+14. Check forward resolution of the name as described in ENSIP-11 and repeated in the section above.
+    - If it resolves to the same address, then proceed normally
+    - If it resolves to a different address, including an empty bytes array, display the address raw and optionally show a warning to the user.
+15. Compute the checksum as described in [ERC-7930]: `0xC69BEB13`.
+16. Format the address in the format `<address><><chain>#<checksum>` and display it to the user: `alice.eth<>oeth.short#C69BEB13`
+
+[^2]: This is a clarification & recommendation on top of ENSIP-19, which is in Draft status and is somewhat ambiguous.
+
+## Rationale
+- We chose to use a centralized list of name registries, aka a [^name resolver registry], since mapping of chain names to chain ids, aka on-chain configs (as proposed in [ERC-7785]) are not yet a solved problem. Therefore there is a need to maintain a list of possible implementations, including stopgap solutions such as relying on ethereum-lists/chains.
```
lucemans, line 264:
```suggestion
## Rationale

- We chose to use a centralized list of name registries, aka a [^name resolver registry], since mapping of chain names to chain ids, aka on-chain configs (as proposed in [ERC-7785]) are not yet a solved problem. Therefore there is a need to maintain a list of possible implementations, including stopgap solutions such as relying on ethereum-lists/chains.
```

ERCS/erc-7828.md
```diff
@@ -0,0 +1,293 @@
...
+
+## Open Discussions
```
lucemans, line 266:
```suggestion
## Open Discussions

```

ERCS/erc-7828.md
```diff
@@ -0,0 +1,293 @@
...
+- Further constrain the syntax of human-readable names to minimize:
+    - Collisions on names by e.g. UTS46 case folding
+    - Addresses that are valid ERC-7828 but cant be squeezed into ENS or comparable standards
+- Conflicts with existing standards:
+    - ENSIP-11 defines custom serialization logic for a few address types and it'd be very convenient if it used the same CASA profile as [ERC-7930]
+    - ENSIP-11 defines the bitmask to use for `eip155` `coinType` assuming chainids are 31 bits or shorter, which is in conflict with ERC-7785 and the addresses representable in [ERC-7930].
+    - ENSIP-19 does not extensively define address or name serialization for the reverse registrar node name
+
+## Backwards Compatibility
+The naming scheme herein defined can represent all names supported by [ERC-7930] by displaying raw addresses without resolution
```
lucemans, line 276:
```suggestion
## Backwards Compatibility

The naming scheme herein defined can represent all names supported by [ERC-7930] by displaying raw addresses without resolution
```

lucemans (changes requested):
Not approval, just natural amendment fixing markdown formatting

ERCS/erc-7828.md
```diff
@@ -0,0 +1,225 @@
...
+
+### Resolving address names
+The entirety of address name resolving of all TLDs is delegated to ENS, as specified in current & future ENSIPs.
+
+Some caveats for ENS support are:
+- For addresses on chains that are both supported by SLIP-0044 and the special scheme for the eip155 namespace defined in ENSIP-11, the latter should be used.
+<!-- TODO: find a few more edge cases or get rid of the ul -->
+
+#### Supported CASA namespaces
+Forward resolution (Interoperable Address -> Interoperable Name): `eip155` (via ENSIP-11) and anything supported by SLIP-0044, as defined in ENSIP-9.
```
SamWilsn: ENSIPs are not currently permitted for external links. Instead use the ERC (might be eip-2304?)

ERCS/erc-7828.md
```diff
@@ -0,0 +1,225 @@
...
+## Rationale
+- Using ENS as the only resolving method means it's enough to use the same Interoperable Addresses v1 for this standard, without extending it to also store the name resolver used. This means however that wallets are free to show users both resolved and unresolved (raw) Interoperable Names.
+- Explicitly reserving a place in the syntax of the Interoperable Name for a planned on-chain chain naming standard ensures forward compatibility.
+
+## Open Discussions
+- Further constrain the syntax of human-readable names to minimize:
+    - Collisions on names by e.g. UTS46 case folding
+    - Addresses that are valid ERC-7828 but cant be squeezed into ENS or comparable standards
+- Conflicts with existing standards:
+    - ENSIP-11 defines the bitmask to use for `eip155` `coinType` assuming chainids are 31 bits or shorter, which is in conflict with [ERC-7785] and the addresses representable in [ERC-7930].
```
SamWilsn: These kind of sections end up being problematic in Final proposals. They tend to get frozen in time, and cannot be updated. Instead, put this in an `<!-- HTML-style comment -->` and the linter will ensure it gets removed before advancing.

0xtiti: Fixed [c4edfff](https://github.com/ethereum/ERCs/pull/735/commits/c4edfff070491d428140122e0e37df771a3b955e)

ERCS/erc-7828.md
0xtiti: Fixed [e54908b](https://github.com/ethereum/ERCs/pull/735/commits/e54908b619bda54037a366e115750a45219c4718)

ERCS/erc-7828.md
```diff
@@ -0,0 +1,248 @@
...
+#### Supported CASA namespaces
+
+While currently only `eip155` chains are supported, the on-chain registry could start listing other chains in order to make human-readable names for non-evm chains possible.
+
+### Resolving address names
+
+The entirety of address name resolving of all TLDs is delegated to ENS, as specified in current & future ENSIPs.
+
+Some caveats for ENS support are:
+- For addresses on chains that are both supported by SLIP-0044 and the special scheme for the eip155 namespace defined in ENSIP-11, the latter should be used.
```
g11tech:
```suggestion
- For addresses on chains that are both supported by SLIP-0044 and the special scheme for the [EIP-155](./eip-155.md) namespace defined in ENSIP-11, the latter should be used.
```

ERCS/erc-7828.md
```diff
@@ -0,0 +1,248 @@
...
+
+#### Supported CASA namespaces
+
+Forward resolution (Interoperable Address -> Interoperable Name): Supported via the multichain address resolution standard defined in [ERC-2304], which uses coin types from [SLIP44] to resolve names for various chains, including the `eip155` namespace.
```
g11tech:
```suggestion
Forward resolution (Interoperable Address -> Interoperable Name): Supported via the multichain address resolution standard defined in [ERC-2304], which uses coin types from [SLIP44] to resolve names for various chains, including the [EIP-155](./eip-155.md) namespace.
```

ERCS/erc-7828.md
```diff
@@ -0,0 +1,248 @@
...
+
+Reverse resolution (Interoperable Name -> Interoperable Address): `eip155` chains explicitly supported by ENS by deployment of `L2ReverseRegistrar` on them and registration of that contract's address on the ENS registry.
```
g11tech:
```suggestion
Reverse resolution (Interoperable Name -> Interoperable Address): [EIP-155](./eip-155.md) chains explicitly supported by ENS by deployment of `L2ReverseRegistrar` on them and registration of that contract's address on the ENS registry.
```

ERCS/erc-7828.md
```diff
@@ -0,0 +1,248 @@
...
+This section is non-normative, as future ENSIPs can modify the way in which ENS is used. Its purpose is to demonstrate how human-readable interoperable names are resolved to the [ERC-7930] binary format used at the protocol level. The following examples assume that chain names are registered as subdomains under `l2.eth`, and that resolution is performed via `L2Resolver` capable of mapping those subdomains to [ERC-7930] compliant chain identifiers.
+
+### Forward resolution step-by-step example
+
+1. Let the user input an address name. Assume it's `alice.eth`.
+2. Let the user select a chain. Assume it's `optimism`. The `l2.eth` suffix is implicit and hidden from end users.
+3. Append `l2.eth` and compute the ENS namehash as per ENSIP-1: `namehash("optimism.l2.eth")`.
+4. Call `L2Resolver.chainId(bytes32 _node)` with the result of step 3 as input, which returns the chain ID in [ERC-7930] byte format.
+5. Convert the returned chain identifier into the CAIP-2 format, as specified by [ERC-7930].
+6. As per ENSIP-11, convert the eip155 `chainId` into an ENSIP-11-specific `coinType`: `0x80000000 & 0x0000000A == 0x8000000A`.
```
g11tech:
```suggestion
6. As per ENSIP-11, convert the [EIP-155](./eip-155.md) `chainId` into an ENSIP-11-specific `coinType`: `0x80000000 & 0x0000000A == 0x8000000A`.
```

ERCS/erc-7828.md
```diff
@@ -0,0 +1,248 @@
...
+### Resolving address names
+
+The entirety of address name resolving of all TLDs is delegated to ENS, as specified in current & future ENSIPs.
+
+Some caveats for ENS support are:
+- For addresses on chains that are both supported by SLIP-0044 and the special scheme for the [EIP-155](./eip-155.md) namespace defined in ENSIP-11, the latter should be used.
+
+#### Supported CASA namespaces
+
+Forward resolution (Interoperable Address -> Interoperable Name): Supported via the multichain address resolution standard defined in [ERC-2304], which uses coin types from [SLIP44] to resolve names for various chains, including the [EIP-155](./eip-155.md) namespace.
```
g11tech:
```suggestion
Forward resolution (Interoperable Address -> Interoperable Name): Supported via the multichain address resolution standard defined in [ERC-2304](./erc-2304.md), which uses coin types from [SLIP44] to resolve names for various chains, including the [EIP-155](./eip-155.md) namespace.
```

ERCS/erc-7828.md
```diff
@@ -0,0 +1,248 @@
...
+
+Reverse resolution (Interoperable Name -> Interoperable Address): [EIP-155](./eip-155.md) chains explicitly supported by ENS by deployment of `L2ReverseRegistrar` on them and registration of that contract's address on the ENS registry.
+
+## Step-by-step examples
```
g11tech:
```suggestion
### Step-by-step examples
```
just guessing if this resolves the "extra section" error marked by bot

ERCS/erc-7828.md
```diff
@@ -0,0 +1,248 @@
...
+### Resolving address names
+
+The entirety of address name resolving of all TLDs is delegated to ENS, as specified in current & future ENSIPs.
+
+Some caveats for ENS support are:
+- For addresses on chains that are both supported by SLIP-0044 and the special scheme for the [EIP-155](./eip-155.md) namespace defined in ENSIP-11, the latter should be used.
+
+#### Supported CASA namespaces
+
+Forward resolution (Interoperable Address -> Interoperable Name): Supported via the multichain address resolution standard defined in [ERC-2304](./erc-2304.md), which uses coin types from [SLIP44] to resolve names for various chains, including the [EIP-155](./eip-155.md) namespace.
```
g11tech:
```suggestion
Forward resolution (Interoperable Address -> Interoperable Name): Supported via the multichain address resolution standard defined in [ERC-2304](./erc-2304.md), which uses coin types from [SLIP44](https://github.com/satoshilabs/slips/blob/master/slip-0044.md) to resolve names for various chains, including the [EIP-155](./eip-155.md) namespace.
```

ERCS/erc-7828.md
```diff
@@ -0,0 +1,248 @@
...
+- ENSIP-19 introduces trust assumptions when resolving reverse records offchain:
+    - ZK-rollups may offer trust-minimized resolution via on-chain proofs.
+    - Sidechains may rely on centralized CCIP-read gateways or signature authorities.
+
+    Wallets MUST surface whether reverse resolution was trust-minimized or trusted.
+
+## Copyright
+
+Copyright and related rights waived via [CC0](../LICENSE.md).
+
```
g11tech:
```suggestion

[EIP-155]: ./EIP-155.md
```

ERCS/erc-7828.md
```diff
@@ -0,0 +1,249 @@
...
+### Resolving address names
+
+The entirety of address name resolving of all TLDs is delegated to ENS, as specified in current & future ENSIPs.
+
+Some caveats for ENS support are:
+- For addresses on chains that are both supported by SLIP-0044 and the special scheme for the [EIP-155](./eip-155.md) namespace defined in ENSIP-11, the latter should be used.
+
+#### Supported CASA namespaces
+
+Forward resolution (Interoperable Address -> Interoperable Name): Supported via the multichain address resolution standard defined in [ERC-2304](./erc-2304.md), which uses coin types from [SLIP44](https://github.com/satoshilabs/slips/blob/master/slip-0044.md) to resolve names for various chains, including the [EIP-155](./eip-155.md) namespace.
```
g11tech:
```suggestion
Forward resolution (Interoperable Address -> Interoperable Name): Supported via the multichain address resolution standard defined in [ERC-2304], which uses coin types from [SLIP44] to resolve names for various chains, including the [EIP-155] namespace.
```

ERCS/erc-7828.md
```diff
@@ -0,0 +1,249 @@
...
+
+## Copyright
+
+Copyright and related rights waived via [CC0](../LICENSE.md).
+
+[EIP-155]: ./eip-155.md
+[ERC-7930]: ./eip-7930.md
+[ERC-7785]: ./eip-7785.md
+[ERC-2304]: ./eip-2304.md
+[SLIP44]: https://github.com/satoshilabs/slips/blob/master/slip-0044.md
```
SamWilsn: SLIPs aren't am approved external resource. You'll need to do one of:

- remove the reference
- duplicate the information here or in `assets/`
- follow EIP-5757 to get SLIPs allowed

ERCS/erc-7828.md
```diff
@@ -0,0 +1,241 @@
...
+status: Draft
+type: Standards Track
+category: ERC
+created: 2024-11-27
+requires: 155, 7930
+---
+
+## Abstract
+
+This proposal extends [ERC-7930] (Interoperable Addresses & Names) by standardizing a human-readable format for chain-specific addresses in the form `<address>@<chain>#<checksum>`. It introduces:
```
g11tech:
```suggestion
This proposal extends [ERC-7930](./eip-7930.md) (Interoperable Addresses & Names) by standardizing a human-readable format for chain-specific addresses in the form `<address>@<chain>#<checksum>`. It introduces:
```

ERCS/erc-7828.md
```diff
@@ -0,0 +1,241 @@
...
+
+Instead of using non-human-readable numeric chain identifiers, this specification **SHALL** require a human-readable chain name resolved on-chain via ENS wildcard resolver. The centralized chain list maintained in `ethereum-lists/chains` **SHALL** be superseded by an on-chain registry. This registry provides a single source of truth for mapping chain names to chain identifiers and enables decentralized, extensible chain metadata resolution.
+
+In the same spirit, the address could be a human-readable name as well, which is already a use case for ENS. By coupling the TLD to the resolving method used, this standard could leverage current and future features of ENS as well as other naming registries.
+
+Moreover, the above can be leveraged to allow for a name to represent the same entity corresponding to different addresses on different chains, mitigating the risk of sending funds to an address the intended recipient doesn't actually control.
+
+Desired properties:
+- Backwards compatibility with [ERC-7930]
+- The chain portion can be an [ERC-7785] domain name when that standard is productive.
```
g11tech:
```suggestion
- The chain portion can be an [ERC-7785](./eip-7785.md) domain name when that standard is productive.
```

ERCS/erc-7828.md
euler0x: solved c33c60f0f0ffdae14def0f6ba83eb8a15bfa47eb

g11tech (approved):

PR #736, opened 2024-11-28
Update ERC-7579: specify executeFromExecutor return data type

kopy-kat:


kopy-kat:
thanks to @ernestognw for the suggestion

PR #738, opened 2024-11-28
Update ERC-7743: Move to Review

jamesavechives:

 - The build passes.


SamWilsn (changes requested):
You have a lot of `private` members in your specification section. Those are not externally visible, and so should not appear in an ERC. Instead, try to only document the externally visible behaviours (like events, public functions, etc.) and leave the implementation details up to the developers building the contracts.

jamesavechives:
> You have a lot of `private` members in your specification section. Those are not externally visible, and so should not appear in an ERC. Instead, try to only document the externally visible behaviours (like events, public functions, etc.) and leave the implementation details up to the developers building the contracts.

Done!

PR #749, opened 2024-12-03
Add ERC: Multi-Chain Addressing

SamWilsn:


ERCS/erc-7831.md
```diff
@@ -0,0 +1,367 @@
...
+
+More formally, valid chain-specific addresses MUST adhere to the following grammar:
+
+```abnf
+address         = OPEN bare-address CLOSE /
+                  bare-address
+
+bare-address    = local-part SEP chain-part [SEP root-part]
+
+OPEN            = '('
```
g11tech, line 66: just a nit:

would prefer if you just use `(` ... etc directly in grammar to have a shorter specification and bundle all the grammar rules together for various parts

ERCS/erc-7831.md
```diff
@@ -0,0 +1,367 @@
...
+
+The `root-part` identifies the root chain against which other names are resolved. When present, the `root-part` SHALL be the [EIP-155] `chainid` of the root chain in decimal format. `root-part` SHOULD NOT be present when `chainid == 1`, and MUST be present when `chainid != 1`.
+
+```abnf
+root-part = 1*DIGIT
+```
+
+### Resolution
+
+Resolving a chain-specific address begins on the right, and moves leftward.
```
g11tech, line 131: its left to right grammar resolution right?

g11tech: `At a high level, a chain-specific address is made of three components separated by colons (`:`) ordered from most specific on the left to most general on the right:`

ERCS/erc-7831.md
```diff
@@ -0,0 +1,367 @@
...
+[ERC-137]: ./eip-137.md
+[EIP-155]: ./eip-155.md
+[ERC-165]: ./eip-165.md
+[ERC-2304]: ./eip-2304.md
+
+## Rationale
+
+### Component Order
+
+The components are ordered from most specific to most general because... <!-- TODO -->
```
g11tech, line 273: so the question being answered here is: why address first and then chain specification next instead of other way around where we have been writing chain:address format

ERCS/erc-7831.md
```diff
@@ -0,0 +1,367 @@
...
+  -->
+
+### ENS Configuration
+
+#### Mainnet (1)
+
+| Name          | Coin Type       | Record                                       |
+| ------------- | --------------- | -------------------------------------------- |
+| `ethereum.tbd.eth`<!-- TODO --> | - | a bridge contract to mainnet (1)         |
+| `rollup1.tbd.eth`<!-- TODO -->  | - | a bridge contract to rollup1 (1608)      |
```
g11tech, line 316: why can't we specify `tbd.eth` as `REGISTRY_SUBDOMAIN` and use later everywhere else

ERCS/erc-7831.md
SamWilsn: Like instead of:

> ```abnf
> address         = OPEN bare-address CLOSE /
>                  bare-address
> ```

Use:

```abnf
address         = '(' bare-address ')' /
                  bare-address
```

ERCS/erc-7831.md
SamWilsn: Updated! Resolution does begin on the right.

ERCS/erc-7831.md
SamWilsn: We could. Would that be better?

g11tech: yes thats more easy for eyes to identiy ... it took me a while to see tbd.eth was a TBD lol

g11tech (approved):
lgtm, the nits are take it or leave it based on author preference

PR #758, opened 2024-12-06
Add ERC: Wallet Call Preparation API

lukasrosario:
This proposal defines complementary JSON-RPC methods to [EIP-5792](./eip-5792.md)'s `wallet_sendCalls`. While `wallet_sendCalls` is used for an app to submit `calls` to be signed and submitted in a wallet's interface, the methods in this proposal are for an application to request prepared calls (where "prepared" depends on the wallet implementation) to be signed and submitted by the application itself.


ERCS/erc-draft-wallet-call-preparation-api.md
```diff
@@ -0,0 +1,200 @@
+---
```
abcoathup, line 1:
```suggestion
---
eip: 7836
```

ERCS/erc-draft-wallet-call-preparation-api.md
```diff
@@ -0,0 +1,200 @@
...
+title: Wallet Call Preparation API
+description: Adds JSON-RPC methods for requesting prepared EIP-5792 calls.
+author: Lukas Rosario (@lukasrosario), Conner Swenberg (@ilikesymmetry), Adam Hodges (@ajhodges), Paaras Bhandari (@paarasbhandari)
+discussions-to: TODO
```
abcoathup, line 5: Please create a discussions topic in Eth Magicians with a link to this PR: https://ethereum-magicians.org/c/ercs/57

ERCS/erc-draft-wallet-call-preparation-api.md
```diff
@@ -0,0 +1,200 @@
...
+Two new JSON-RPC methods are defined. If the user is interacting with an application on the browser, they can be used by sending the requests to the wallet's [EIP-1193](./eip-1193.md) provider. However, this proposal also describes how an application might interact with a "wallet server" for background processes.
+
+### `wallet_prepareCalls`
+
+Accepts an EIP-5792 `wallet_sendCalls` request, and returns the prepared calls according to the account's implementation.
+
+#### Parameters
+
+```typescript!
+type PrepareCallsParams = [{
```
ajhodges, line 32: @lukasrosario should we add a 'version' field to the params for these two methods similar to the one in 5792?

ERCS/erc-7836.md
```diff
@@ -0,0 +1,205 @@
...
+    }
+  },
+  preparedCapabilities: {
+    paymasterService: {
+      sponsor: 'My app'
+    }
+  }
+}]
+```
+
```
ajhodges, line 106:
```suggestion
##### Example Return Value

```typescript!
{
  type: 'user-operation-v06',
  userOp: {
    sender: '0x...',
    ...
  },
  chainId: '0x01',
  signatureRequest: {
    hash: '0x...'
  }
  capabilities: {
    paymasterService: {
      sponsor: 'My app'
    }
  }
}
```

```

File: ERCS/erc-7836.md [hunk 5]
```diff
@@ -0,0 +1,205 @@
...
+    chainId: `0x${string}`
+    signatureRequest: {
+      type: string
+      hash: `0x${string}`
+    }
+  }[]
+  preparedCapabilities: Record<string, any>
+}]
+```
+
```
ajhodges, line 82:
```suggestion

Extra `type`-specific fields may be specified inside `PrepareCallsReturnValue` in the response.
```

ERCS/erc-7836.md
```diff
@@ -0,0 +1,205 @@
...
+### `wallet_sendPreparedCalls`
+
+Accepts prepared calls from the response to a `wallet_prepareCalls` request along with a signature, and returns an EIP-5792 call bundle ID.
+
+#### Parameters
+
+```typescript!
+type SendPreparedCallsParams = [{
+  version: string
```
ajhodges, line 117: Does it make sense to have `version` in each preparedCall in the request? Maybe we should wrap the request like:
{
  version: "1.0",
  preparedCalls: [...]
}

ERCS/erc-7836.md
```diff
@@ -0,0 +1,205 @@
...
+  type: string
+  preparedCalls: {
+    data: any
+    chainId: `0x${string}`
+    signature: {
+      type: string
+      data: `0x${string}`
+    }
+  }[]
+}]
```
ajhodges, line 127:
```suggestion
type SendPreparedCallsParams = [{
  version: string
  type: string
  data: any
  chainId: `0x${string}`
  signature: {
    type: string
    signature: `0x${string}`
  }
}]
```

ERCS/erc-7836.md
```diff
@@ -0,0 +1,205 @@
...
+      sender: '0x...',
+      ...
+    },
+    chainId: '0x01',
+    signature: {
+      type: 'secp256k1',
+      data: '0x...'
+    }
+  }]
+}])
```
ajhodges, line 147:
```suggestion
wallet_sendPreparedCalls([{
  version: '1.0',
  type: 'user-operation-v06',
  data: {
    sender: '0x...',
    ...
  },
  chainId: '0x01',
  signature: {
    type: 'secp256k1',
    data: '0x...'
  }
}])
```

ajhodges (commented):
@lukasrosario Proposed some changes to the schema based on our current internal working prototype. Example requests/responses: https://gist.github.com/ajhodges/6ebcbc73e75afac3b3ecf0906e136ae2

ERCS/erc-7836.md
lukasrosario: yeah that makes sense

ERCS/erc-7836.md
```diff
@@ -0,0 +1,188 @@
...
+
+```typescript!
+type SendPreparedCallsParams = [{
+  version: string
+  type: string
+  data: any
+  chainId: `0x${string}`
+  signature: {
+    type: string
+    signature: `0x${string}`
```
montycheese, line 112:
```
signature: {
   type: string
   data : {
   // e.g. for webauthn
     publicKey: string,
     signature: string
   }
}
```

ERCS/erc-7836.md
```diff
@@ -0,0 +1,188 @@
...
+requires: 1193, 5792
+---
+
+## Abstract
+
+This proposal defines complementary JSON-RPC methods to [EIP-5792](./eip-5792.md)'s `wallet_sendCalls`. While `wallet_sendCalls` is used for an app to submit `calls` to be signed and submitted in a wallet's interface, the methods in this proposal are for an application to request prepared calls (where "prepared" depends on the wallet implementation) to be signed and submitted by the application itself.
+
+## Motivation
+
+With more recent developments in account abstraction and session keys, there is an increasing need for applications to be able to sign and submit operations without switching to a wallet interface (e.g. browser extension). This can be tricky because different account implementations might have different call data encoding, signature formats, etc. To remedy this, we need a way for apps to know how to submit an operation for any account implementation.
```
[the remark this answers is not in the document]
jakubuid: I can give one that corresponds to mobile flow. `wallet_getCallsStatus` doesn't require any user interaction; it just requires the connection between a wallet and a dapp, which is tricky on mobile apps because the mobile wallet can be in the background, aka offline. In this case, wallet_getCallsStatus would not work, as the wallet wouldn't have access to the Internet.

In the mobile flow, the wallet would be foregrounded in the moment of receiving the request, and the user would be automatically redirected to the dapp with `wallet_getCallsStatus` result without any user interaction, which is a bad UX. 

Dapp, in this scenario, would just call an endpoint to get the status without needing the wallet to be online or without needing redirecting to the wallet.

ERCS/erc-7836.md
```diff
@@ -0,0 +1,188 @@
...
+Two new JSON-RPC methods are defined. If the user is interacting with an application on the browser, they can be used by sending the requests to the wallet's [EIP-1193](./eip-1193.md) provider. However, this proposal also describes how an application might interact with a "wallet server" for background processes.
+
+### `wallet_prepareCalls`
+
+Accepts an EIP-5792 `wallet_sendCalls` request, and returns the prepared calls according to the account's implementation.
+
+#### Parameters
+
+```typescript!
+type PrepareCallsParams = [{
```
ajhodges: We should add app-provided IDs ([as proposed to be added to 5792](https://github.com/ethereum/EIPs/pull/8826/files#diff-7a494500c37dc4b57613207d62874322198a4ced215f4a0457a0af908bb6364eR70)).

ERCS/erc-7836.md
```diff
@@ -0,0 +1,188 @@
...
+
+Accepts prepared calls from the response to a `wallet_prepareCalls` request along with a signature, and returns an EIP-5792 call bundle ID.
+
+#### Parameters
+
+```typescript!
+type SendPreparedCallsParams = [{
+  version: string
+  type: string
+  data: any
```
V00D00-child, line 108: Should we define `data` as a `Record<string, any>`  where the values of `any` are specific to the `type` and `version`?

ERCS/erc-7836.md
```diff
@@ -0,0 +1,188 @@
...
+  chainId: `0x${string}`
+  signature: {
+    type: string
```
V00D00-child, line 111: Why does the "wallet server" need to know the signature type? Couldn't the signature be absorbed into the top-level `data`?

ajhodges: Taking a smart contract wallet as an example, it's possible that the raw signature might need to be ABI-encoded before being considered valid onchain. This should be the responsibility of the wallet, as only it would know what format the contracts expect.

ERCS/erc-7836.md
```diff
@@ -0,0 +1,188 @@
...
+
+![walletserver](../assets/eip-7836/0.svg)
+
+## Backwards Compatibility
+
+TBD
+
+## Security Considerations
+
+Needs discussion.
```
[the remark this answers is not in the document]
howydev: For security reasons, the dapp should reconstruct `signatureRequest` in `PrepareCallsReturnValue` and compare it against the received value, and throw an error if it doesn't match. This prevents malicious wallets from phishing signatures on digests from the dapp 

I recommend we remove the `signatureRequest` field - thoughts?

ERCS/erc-7836.md
```diff
@@ -0,0 +1,188 @@
...
+      "url": "https://.../0x2105"
+    }
+  },
+  "0x14A34": {
+    "walletServer": {
+      "supported": true,
+      "url": "https://.../0x14A34"
+    }
+  }
+}
```
chris13524, line 171: Suggest referencing standardized [CAIP-345](https://github.com/ChainAgnostic/CAIPs/pull/345) as the example. We would appreciate feedback on CAIP-345 and how it interops with ERC-7836. This is what we are using for WalletConnect connections.

```suggestion
{
  "0x2105": {
    "walletService": {
      "url": "https://.../0x2105",
      "methods": ["wallet_getCallsStatus"]
    }
  },
  "0x14A34": {
    "walletService": {
      "url": "https://.../0x14A34",
      "methods": ["wallet_getCallsStatus"]
    }
  }
}
```

ERCS/erc-7836.md
```diff
@@ -0,0 +1,188 @@
...
+```
+
+If a wallet's [EIP-1193](./eip-1193.md) provider receives these requests, it might choose to reroute the request to its own wallet server. Alternatively, the [EIP-1193](./eip-1193.md) provider can handle the requests directly.
```
chris13524, line 174: This discussion/specification would be best moved into CAIP-345.

ERCS/erc-7836.md
```diff
@@ -0,0 +1,188 @@
...
+{
+  type: 'user-operation-v06',
+  userOp: {
+    sender: '0x...',
+    ...
+  },
+  chainId: '0x01',
+  signatureRequest: {
+    hash: '0x...'
+  }
```
moldy530, line 89: @adamegyed and I were chatting about this field, it'd be nice if this was something like:
```ts
signatureRequest: {
  type: "personal_sign" | "eth_signTypedData_v4" | ...,
  data: any, // this would match the params for the above type
}
```

As an example: for EP 0.8 where the signature of a UO is a 712 signature this could allow a client to compare the signature request to the returned `data` field as well to make it less of a blind sign.

It might depend on the client, so maybe webauthn or other signature types are required

jxom: Why is this neccessary? You can detect the type via the [EIP-191](https://eips.ethereum.org/EIPS/eip-191) prefix (`0x19 <1 byte version> <version specific data> <data to sign>`).

ERCS/erc-7836.md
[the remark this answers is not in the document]
SamWilsn: This belongs in the document itself. Comments get lost in the shuffle.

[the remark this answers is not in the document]
azf20: I am not sure `getCallsStatus` is the relevant example here (I think this is about getting prepared calls)?

ERCS/erc-7836.md
```diff
@@ -0,0 +1,189 @@
...
+      url: 'https://...'
+    }
+  }
+}])
+```
+
+#### Return Value
+
+```typescript!
+type PrepareCallsReturnValue = {
```
azf20: should this include the calls themselves, in whatever format? (userop / regular txn etc)?

azf20: As ideally the application then has all the information it needs to make the transaction itself, as well as calling the subsequent method?

ERCS/erc-7836.md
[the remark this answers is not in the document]
azf20: I think this is a valid consideration, but I wonder if that's possible without requiring the wallet to leak a lot of the complexity about their particular account implementation which is what this ERC is looking to avoid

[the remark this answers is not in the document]
ajhodges: > I wonder if that's possible without requiring the wallet to leak a lot of the complexity about their particular account implementation which is what this ERC is looking to avoid

+1 - I think the design of prepareCalls should account for varying levels of trust and sophistication of apps. If the app has full trust in the preparation API, it can just sign the signatureRequest. If it wants to verify that the transaction does what was intended, it can simulate (possibly via [ERC-7882](https://github.com/ethereum/ERCs/pull/892)). If it does not trust it at all and is equipped to do so, it can reconstruct the signatureRequest, perform static analysis of the preparedCall context, and simulate.

SamWilsn (approved):

Fool256 (approved):

PR #775, opened 2024-12-16
Update ERC-7562: AA-486 add creation rules OP-31, EREP-60, EREP-61

drortirosh:


yoavw (commented):
LGTM.  Seems to cover what we discussed.

PR #779, opened 2024-12-17
Add ERC: Wallet Connection API

ilikesymmetry:
This proposal defines a new RPC for wallet connection with an emphasis on extensibility. Builds on the notion of optional “capabilities” defined in [ERC-5792](https://eips.ethereum.org/EIPS/eip-5792#wallet_getcapabilities) to add new functionality modularly. This proposal defines one capability to reduce separate interactions for connection and authentication, but otherwise seeks to leave capability definitions open-ended.


ERCS/erc-draft-wallet-connection-api.md
```diff
@@ -0,0 +1,180 @@
+---
```
abcoathup, line 1:
```suggestion
---  
eip: 7846
```

ERCS/erc-draft-wallet-connection-api.md
```diff
@@ -0,0 +1,180 @@
+---  
+title: Wallet Connection API  
+description: Adds JSON-RPC method for requesting wallet connection with modular capabilities.  
+author: Conner Swenberg (@ilikesymmetry).
+discussions-to: https://ethereum-magicians.org/t/erc-xxxx-wallet-connection-api/22245
```
abcoathup, line 5:
```suggestion
discussions-to: https://ethereum-magicians.org/t/erc-7846-wallet-connection-api/22245
```
Added assigned number

ERCS/erc-draft-wallet-connection-api.md
```diff
@@ -0,0 +1,180 @@
...
+  chains?: `0x${string}`[]; // optional chain IDs (EIP-155 hex)
+  capabilities?: Record<string,any>; // optional connection capabilities
+}]
+
+type WalletConnectResult = {
+  account: {
+    address: `0x${string}`; // connected account address
+    supportedChainsAndCapabilities: Record<`0x${string}`,any>; // chain-specific capabilities, mirrors ERC-5792 wallet_getCapabilities
+  },
+  capabilityResults: Record<string,any>; // results of this connection request's connection capabilities
```
jxom:
```suggestion
  capabilities: Record<string,any>; // results of this connection request's connection capabilities
```

`capabilities` for consistency? "result" seems implied as the consumer will be extracting it from a response anyway, and we don't call it `capabilitiesRequest`.

ERCS/erc-draft-wallet-connection-api.md
```diff
@@ -0,0 +1,180 @@
...
+  nonce: string,
+  issuedAt: string
+  expirationTime?: string,
+  notBefore?: string,
+  requestId?: string,
+  resources?: string[]
+}
+
+type SignInWithEthereumCapabilityResult = {
+  message: string, // formatted SIWE message
```
jxom:
```suggestion
```

Do we need the message? I thought the wallet would be the one displaying the formatted message.

ilikesymmetry: Was thinking that the app needs to verify that the message the wallet prompted signing for matches their requested arguments. For instance a malicious wallet could attempt to reuse an old expiry/nonce if it had access to a previously signed SIWE message. The most succinct way to send back all fields and verify against the signature is to pass back the final `message`

ERCS/erc-draft-wallet-connection-api.md
```diff
@@ -0,0 +1,180 @@
...
+type WalletConnectParams = [{
+  version: string;
+  chains?: `0x${string}`[]; // optional chain IDs (EIP-155 hex)
+  capabilities?: Record<string,any>; // optional connection capabilities
+}]
+
+type WalletConnectResult = {
+  account: {
+    address: `0x${string}`; // connected account address
+    supportedChainsAndCapabilities: Record<`0x${string}`,any>; // chain-specific capabilities, mirrors ERC-5792 wallet_getCapabilities
```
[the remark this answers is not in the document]
ilikesymmetry: The intent was to remove the need to make a separate `wallet_getCapabilities` request. That may not be necessary at API layer though because SDKs could do this automatically for apps.

ERCS/erc-draft-wallet-connection-api.md
jxom: Got it. Makes sense!

ERCS/erc-7846.md
```diff
@@ -0,0 +1,168 @@
...
+  statement?: string
+  issuedAt?: string
+  expirationTime?: string,
+  notBefore?: string,
+  requestId?: string,
+  resources?: string[]
+}
+
+type SignInWithEthereumCapabilityResult = {
+  message: string, // formatted SIWE message
```
jxom, line 110: Might also be nice to add a `payload` property so consumers don't have to parse the `message` string to figure out the inferred parameters (e.g. `address`, `domain`, etc)?
```suggestion
  message: string, // formatted SIWE message
  payload: {
    address: `0x{string}`,
    nonce: string,
    ...
  },
```

atimark (approved):

ERCS/erc-7846.md
```diff
@@ -0,0 +1,168 @@
...
+### `wallet_connect`
+
+Request the user to connect a single account and optionally confirm chain support and add capabilities.
+
+#### RPC Specification
+
+For each chain requested by the app, wallets MUST return a mapped capabilities object if the chain is supported and SHOULD return an empty object if no capabilities exist. Wallets MUST NOT return mapped capabilities objects for chains they do not support. If an app does not declare chains they would like to confirm support for, the wallet can return any chains it wishes to declare support for.
+
+```typescript
+type WalletConnectParams = [{
```
GregTheGreek, line 32:
```suggestion
type WalletConnectRequest = [{
```

Nit: But more in-line with how 1993 handles things. I know 5792 didn't take that approach.

Jakeshellybase (changes requested):
Need changes

pedrouid:
While I understand the motivation for creating EVM-specific standards for Wallets... there is nothing new or extra that ERC-7846 provides that isn't already covered by CAIP-222

https://chainagnostic.org/CAIPs/caip-222

Why restrict such an important pattern like "Connect + Sign" to only EVM wallets?

johanneskares (approved):

johanneskares:
> This proposal defines a new RPC for wallet connection with an emphasis on extensibility. Builds on the notion of optional “capabilities” defined in [ERC-5792](https://eips.ethereum.org/EIPS/eip-5792#wallet_getcapabilities) to add new functionality modularly. This proposal defines one capability to reduce separate interactions for connection and authentication, but otherwise seeks to leave capability definitions open-ended.

one huge problem still:

For passkey wallets, e.g. Signing Up with Coinbase Smart Wallet, you would still need 2 interactions with this ERC. First TouchID to get the Ethereum Address, second one to sign the message. Sign In With Ethereum is not a good solution. The address should not be part of the signed message. It's implicit from the signature.

I'd suggest changing this ERC to allow signing an arbitrary message, makes it simpler and leaves the kind of signature to the dApp.

bumblefudge:
> [](https://chainagnostic.org/CAIPs/caip-222)
> 
> Why restrict such an important pattern like "Connect + Sign" to only EVM wallets?

I would tend to agree that having a multi-VM/multi-L1 syntax for every context _except_ EVM and an EVM-specific context is needlessly inharmonious, when using the CAIP-222 RPC message would be functionally identical to this bespoke one.

That said, we have an ERC here that seems to work, and are comparing it to the hypothetical ERC which achieves the same properties with a CAIP-222-conformant message. Would it help to write a competing ERC that has all the same properties and additional can be parsed by a multi-chain/CAIP-222 system like WalletConnect's? I'd rather not be the one to write it, but if no one else is offering AND evaluators of this ERC would appreciate an apples-to-apples ERC, I can write one...

ERCS/erc-7846.md
bumblefudge: devil's advocate, consumers not using the ABNF to parse it are risking all kinds of security problems, payload and message could fall out of sync (which is itself just an additional surface to secure), etc etc

jxom: Wdym? If they are out of sync, then signature verification will fail.

ERCS/erc-7846.md
```diff
@@ -0,0 +1,233 @@
+--- 
+eip: 7846
+title: Wallet Connection API
```
SamWilsn, line 3: I know there isn't a ton of room in the title (44 characters IIRC), but it would be nice if you could cram a bit more detail in here to help disambiguate.

ERCS/erc-7846.md
```diff
@@ -0,0 +1,233 @@
...
+discussions-to: https://ethereum-magicians.org/t/erc-7846-wallet-connection-api/22245
+status: Draft
+type: Standards Track
+category: ERC
+created: 2024-12-15
+---
+
+## Abstract
+
+This ERC introduces a new wallet connection JSON-RPC method focused on extensibility. It leverages the modular capabilities approach defined in [ERC-5792](https://eips.ethereum.org/EIPS/eip-5792#wallet_getcapabilities) to streamline connections and authentication into a single interaction.
```
SamWilsn, line 15: You should mention the specific endpoint in the abstract, and have maybe a bit more technical detail here. An abstract is not an introduction.

ERCS/erc-7846.md
```diff
@@ -0,0 +1,233 @@
...
+await provider.request({
+  method: 'wallet_disconnect',
+})
+```
+
+### Capabilities
+
+#### `signInWithEthereum`
+
+Adds support for offchain authentication using [ERC-4361: Sign-In with Ethereum](https://eips.ethereum.org/EIPS/eip-4361).
```
SamWilsn, line 118: You'll need to use a relative link. Something like:

```suggestion
Adds support for offchain authentication using [ERC-4361](./eip-4361.md).
```

ERCS/erc-7846.md
```diff
@@ -0,0 +1,233 @@
...
+
+##### Parameters
+
+Same as ERC-4361 specification with minor modifications: 
+* The casing of multi-word fields has been adjusted to camelCase instead of kebab-case. Resources are an array field. 
+* The account address returned by `wallet_connect` MUST match the address inferred in the SIWE message.
```
SamWilsn, line 124: You should expand the abbreviation the first time it is used:

```suggestion
* The account address returned by `wallet_connect` MUST match the address inferred in the Sign-In with Ethereum (SIWE) message.
```

ERCS/erc-7846.md
```diff
@@ -0,0 +1,233 @@
...
+This ERC introduces a new wallet connection JSON-RPC method focused on extensibility. It leverages the modular capabilities approach defined in [ERC-5792](https://eips.ethereum.org/EIPS/eip-5792#wallet_getcapabilities) to streamline connections and authentication into a single interaction.
+
+## Motivation
+
+With applications beginning to require support for more sophisticated functionality in wallet connection flows, the need for a unified and extensible wallet connection JSON-RPC method has become more apparent.
+
+This is especially evident in the case of attempting to batch connection with authentication, where existing methods like `eth_requestAccounts` and `personal_sign` lack extensibility and require at least two separate user interactions (ie. connect and then sign).
+
+## Specification
+
```
SamWilsn, line 24: Since you're using the uppercase keywords (MUST), you should include the paragraph explaining them from the EIP template.

ERCS/erc-7846.md
```diff
@@ -0,0 +1,233 @@
...
+Applies [ERC-4361 security principles](https://eips.ethereum.org/EIPS/eip-4361#security-considerations). As more capabilities are added, care must be taken to avoid unpredictable interactions.
+
+## Privacy Considerations
+
+Wallet addresses and any shared capabilities must be handled securely to avoid data leaks or man-in-the-middle attacks.
+
+## Copyright
+
+Copyright and related rights waived via CC0.
```
SamWilsn, line 233: Please use the exact waiver from the template.

ERCS/erc-7846.md
wilsoncusack: agree!

Genebond (approved):

ERCS/erc-7846.md
```diff
@@ -0,0 +1,233 @@
...
+ *       }
+ *     }
+ *   ]
+ * }
+ */
+```
+
+## Rationale
+
+### Multiple Accounts
```
glitch-txs, line 205: Should we add a method to request an accounts switch? To change the targeted account. 
In such case the wallet should emit accountsChanged with the order of the accounts array altered.

glitch-txs (commented):
I really like this EIP, it removes friction from the users that need to auth as well as adding a `disconnect` method which has been always needed in the current standard (9913).

I like that it mentions that future EIP's might extend capabilities. Which makes the EIP flexible without introducing breaking changes.

The DX looks nice and simple!

In the future I would personally like to be added an option to submit transaction requests right away together with the "wallet_connect" request. Even batching, which I think the wallet could internally handle in its own way (in case it supports it or just running them "synchronously"). I think this would be particularly useful for "one action" (and not one transaction) user needs, like Swapping or payments.

ERCS/erc-7846.md
stephancill: Would prefer to keep it as-is to be consistent with [5792 (Wallet Call API)](https://eip.tools/eip/5792)

But if you feel strongly that we should disambiguate further then I'd suggest "Wallet Connection API with Capabilities"

stephancill:
Hi @SamWilsn and @glitch-txs thanks for the feedback. I've worked with @ilikesymmetry to update the PR

ERCS/erc-7846.md
```diff
@@ -0,0 +1,233 @@
+---
+eip: 7846
+title: Wallet Connection API
+description: Adds JSON-RPC method for requesting wallet connection with modular capabilities.
```
SamWilsn, line 4: Minor nit, but you repeat "wallet connection" from the title in your description. It might be better to reword it so the reader knows what a "wallet connection" _is_.

SamWilsn (approved):

PR #784, opened 2024-12-20
Update ERC-4337: Improve sections

frangio:
It's hard to navigate ERC-4337. In particular, it's very hard to find the definition of PackedUserOperation. This change should improve that.

I think it would also be nice to have the entire EntryPoint interface in one section. Currently it's scattered around.


frangio:
Closing as stale.

PR #786, opened 2024-12-21
Update ERC-7821: Fix return type of `supportsExecutionMode` in interface definition.

Amxx:
The ERC refers to `supportsExecutionMode` as returning a boolean:
> Only returns true for:

And the reference implementation returns a boolean. Yet the interface specify returning a uint256. This looks like a mistake.


Vectorized:
Nice catch! Thank you!

Vectorized (approved):

PR #798, opened 2024-12-25
Update ERC-1185: ERC-1185

xinbenlv:
Update ERC-1185 to new ERC format requirements

### DIFF
--- a/ERCS/erc-1185.md
+++ b/ERCS/erc-1185.md
@@ -21,7 +21,7 @@ ENS is a highly desirable store for DNS information.  It provides the distribute
 
 ## Specification
 
-The resolver profile to support DNS on ENS follows the resolver specification as defined in [ERC-137](./eip-137.md).
+The resolver profile to support DNS on ENS follows the resolver specification as defined in [ERC-137](./erc-137.md).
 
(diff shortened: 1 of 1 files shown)


xinbenlv:
A few early comments

## Editorial Suggestions
- Update EIP-137 to ERC-137 per https://github.com/ethereum/EIPs/pull/5273
- Consider adding in the specification the ERC-165 (supportsInterface), which you already using in ref-impl
- Consider making `setZonehash`/`zonehash` optional interfaces in the requirement
- In Security Consideration, consider mentioning the importance of `authorised(node)` and its implication if done wrong.
- Since CoreDNS ENS Plugin supports these implementation, have you considered mentioning the CoreDNS ENS plugin and its reading side of code in GoLang?

## Technical discussions
- How do you envision  the DNS TTL being set and handled? Could you ellaborate it in the ref-impl section
- Have you considered specific way to supply DNSSEC keys when it has more implications?

ERCS/erc-1185.md
```diff
@@ -21,7 +21,7 @@ ENS is a highly desirable store for DNS information.  It provides the distribute
 
 ## Specification
 
-The resolver profile to support DNS on ENS follows the resolver specification as defined in [ERC-137](./eip-137.md).
```
SamWilsn, line 24: Changing these to `erc-...` will break the renderer.

PR #799, opened 2024-12-26
Fix broken image link (currently 404)

prototypo:
In the ERC version of EIP-1, the EIP-process-update.jpg image is 404 because the asset path is missing in the ERC repo.

This copies the image to the ERC repo instead of just fixing the link in the assumption that EIP-1 may be forked in future as the ERC process evolves.


SamWilsn:
This is also mixing in a change to the content of EIP-1.

PR #802, opened 2024-12-27
Propose sequencing header

prototypo:
# Rationale

Ethereum fragmentation via L2 sequencing silos has arguably been a primary cause of ETH’s divergence from BTC since the significant rise of L2s in the last two years. This is because the value of transactions at the L2 layer has devalued the value of the L1; in other words, we currently risk the market taking the L1 layer for granted, which would create an existential threat to the proof of stake consensus system. Thus, this proposal suggests that new ERCs should be required to state whether and how they align with the goal of Shared Sequencing/Universal Synchronous Composability. Such mandatory statements during the ERC proposal process can assist both editors and Core Devs with the review of new ERCs in relation to the Ethereum roadmap.

# Summary of changes:
- Add fragmentation status metadata and a description of its use to the EIP Header Preamble.
- Add review responsibility to EIP editor to ensure fragmentation status metadata is present.

# Notes:
- The only ERC of `type: Meta` is currently EIP-1. EIPs have made more heavy use of Meta documents than ERCs, with 36 Meta EIPs so far. Therefore, this change is proposed to EIP-1 instead of introducing a new Meta ERC.
- The use of `other` as a possible value for the `sequencing` header is considered suboptimal, but was included to future-proof the proposal in the case that other sequencing solutions beyond `based` and `shared` become available. Alternatives to this approach would be very welcome.

# Dependencies:
- If accepted, this change would require an update to https://github.com/ethereum/EIPs/blob/master/eip-template.md or a fork of that template to the ERC repository.
- If accepted, this change would require either the same change to be made to the version of EIP-1 in the EIPs repository, or the creation of a fork of EIP-1 in the ERC repository.

### DIFF
--- a/ERCS/eip-1.md
+++ b/ERCS/eip-1.md
@@ -142,6 +142,8 @@ Each EIP must begin with an [RFC 822](https://www.ietf.org/rfc/rfc822.txt) style
 
 `withdrawal-reason`: *A sentence explaining why the EIP was withdrawn.* (Optional field, only needed when status is `Withdrawn`)
 
+`sequencing`: One of `based`, `shared`, `other` or `N/A`.
+
 Headers that permit lists must separate elements with commas.
 
 Headers requiring dates will always do so in the format of ISO 8601 (yyyy-mm-dd).
@@ -191,6 +193,14 @@ The `created` header records the date that the EIP was assigned a number. Both h
 EIPs may have a `requires` header, indicating the EIP numbers that this EIP depends on. If such a dependency exists, this field is required.
 
 A `requires` dependency is created when the current EIP cannot be understood or implemented without a concept or technical element from another EIP. Merely mentioning another EIP does not necessarily create such a dependency.
+
+### `sequencing` header
+
+The `sequencing` header specifies whether an ERC relates to `based` sequencing, non-based `shared` sequencing, `other` forms of sequencing, or does not invovle sequencing at all (which is indicated by the use of `N/A`).
+
+EIPs MUST include a `sequencing` header if and only if the `category` is `ERC` and MUST NOT include one otherwise.
+
+For example, those applications using Ethereum mainnet's validators to sequence transactions would be noted as `based`. Those using [EIP-7727](./eip-7727.md) style sequencing would be considered `shared` because they may not choose to sequence transactions via the L1 validators.
 
 ## Linking to External Resources
 
@@ -458,6 +468,7 @@ For each new EIP that comes in, an editor does the following:
 - Read the EIP to check if it is ready: sound and complete. The ideas must make technical sense, even if they don't seem likely to get to final status.
 - The title should accurately describe the content.
 - Check the EIP for language (spelling, grammar, sentence structure, etc.), markup (GitHub flavored Markdown), code style
+- Check the EIP for required headers.
 
 If the EIP isn't ready, the editor will send it back to the author for revision, with specific instructions.
 


fulldecent:
The proposed extra metadata collection is very underspecified. Extensive examples should be provided since you are expecting 1000+ people to classify their own EIPs based.

prototypo:
@fulldecent Thanks for engaging. That's a good and correct thought. Please do note that **_only new EIPs within the ERC category_** would be expected to provide this metadata. That would be much less than 1000+ for the foreseeable future.

The original examples paragraph says this:

> For example, those applications using Ethereum mainnet's validators to sequence transactions would be noted as `based`. Those using [EIP-7727](./eip-7727.md) style sequencing would be considered `shared` because they may not choose to sequence transactions via the L1 validators.

Would the following be more acceptable? What else would you consider to be necessary?

> Examples of correctly applied sequencing headers are:
> 1. Proposals for changes to based applications would have a sequencing header value of `based`. Such proposals would include those aiming to improve based rollup interactions, such as validity proof systems.
> 2. Proposals for changes to applications using non-based sequencing shared or shareable by multiple L2s would have a sequencing header value of `shared`. Such proposals would include [EIP-7727](./eip-7727.md) style sequencing or use of the Espresso sequencer because they may not choose to sequence transactions via the L1 validators.
> 3. In the case where some future shared sequencing techniques are developed that are neither based nor shared as defined above, they would have a sequencing header value of `other`.
> 4. Proposals that do not relate to sequencing in any way would have a sequencing header value of `N/A`.

It is worth highlighting that the purpose of this metadata is not technical; it is social engineering to facilitate alignment with the Ethereum roadmap and to help focus discussions on developments that are likely to eventuate in universal synchronous composability. I believe such hacks are especially useful in decentralised standards development.

prototypo:
In full disclosure, this pull request was submitted in partial fulfillment of the application process for the role of ERC & Application Standards Coordinator at the EF. That does _not_ mean that I don't think it is a useful social hack as part of the difficult job of standards development. I do, in fact, think it would be useful. However, I doubt I would have proposed it in the absence of the job application, so reviewers should understand that motivation.

fulldecent:
If we will have required metadata here then we would also use that to sort EIPs. 

For example, there's another pull request here for updating the website to improve filtering. And this only makes sense if all of the existing publications are classified.

At a minimum, the final ones. 

Would you like to propose classifications for the existing EIP's? Actually, you could probably make a good first draft using a neural network.

prototypo:
@fulldecent Yes, I can propose classifications for existing EIPs. Good idea.

prototypo:
Here are some rough numbers to scope the classification of existing ERCs:
- There are 482 ERCs of type: Standards Track and category: ERC currently in the repo.
- Only 1 references a sequencer.
- 180 reference some kind of composability, but most of those are red herrings.

So, at first glance, I suspect most of the existing ERCs should be assigned a sequencing header of 'N/A'.

fulldecent:
So this proposed mandatory field would not have had an impact on most of the existing publications. And, if implemented, it would require every new publisher to understand and implement the new classifications.

---

Are you expecting a change of the future publications that would make this proposal more relevant for the new publications versus the prior ones?

prototypo:
Excellent question, and yes, I agree with your assessment.

I would, in fact, expect new proposals under the ERC category to be (much) more relevant to shared sequencing than in the past. That is due to the [rapid increase in L2 activity](https://l2beat.com/scaling/activity) (select the "all time" option to see the recent rise) in the past two years. It is that rise in activity that has put downward pressure on ETH and has caused shared sequencing to become the preferred solution to Ethereum fragmentation.

Do you agree or disagree with that assessment?

fulldecent:
So far I'm seeing plenty of normal application standards.

prototypo:
@fulldecent One possible reason that L2s would continue to propose "normal" application standards would be that they don't see the problems that fragmentation causes at layer 1.

Are you saying that you disagree with the hypothesis or the approach to addressing it? Either would be defensible, IMO. Predicting the future is hard.

I'm hardly the only one to suggest that fragmentation is a key issue for Ethereum, though, so I'm interested in your opinion.

SamWilsn:
I'm not sure how generally applicable this is? Are there any examples of current proposals that would have a value other than `based`? If not, I think this putting the cart before the horse.

PR #819, opened 2024-12-31
Update ERC-7578: Move to Last Call

SamWilsn:


SamWilsn:
This is me changing the `last-call-deadline` of https://github.com/ethereum/ERCs/pull/668

PR #851, opened 2025-01-14
Add ERC: ERC-721 Verifiable Credential Extension

vmc-crossmint:
This standard is an extension of [ERC-721](https://eips.ethereum.org/EIPS/eip-721) that adds verifiable credential properties at the contract (collection) level. It allows ERC-721 tokens to natively reference and manage on-chain credentials, including W3C Verifiable Credentials, by defining mechanisms for storage, encryption, revocation, and verification.


ERCS/erc-7861.md
```diff
@@ -0,0 +1,205 @@
...
+
+This standard addresses key gaps in the W3C framework by defining methods for storage, revocation, retrieval, and presentation, facilitating interoperability within a flexible, standardized structure. It supports custom cryptographic methods for signatures and proofs—from basic issuer signatures to advanced zero-knowledge (ZK) proofs and selective disclosure—allowing adaptability for diverse security requirements.
+
+Additionally, the standard natively supports encryption, crucial for managing sensitive data. It also provides flexibility in credential storage, referencing off-chain data via credentialURI to accommodate any storage solution, from decentralized storage (like IPFS) to centralized systems.
+
+Originally designed with W3C credentials in mind, this standard is broad enough to support any kind of credential, for example content credentials. More generally this framework can be used to link any off-chain, private payload to an NFT.
+
+### Overview
```
SamWilsn, line 31: This overview would be more at home in the specification section, since it deals with the "what" instead of the "why" of this proposal.

ERCS/erc-7861.md
```diff
@@ -0,0 +1,205 @@
...
+    /// @notice Check if a credential has been revoked 
+    /// @param tokenId The credential ID to check 
+    /// @return false if the credential has not been revoked
+    function isRevoked(uint256 tokenId) external view returns (bool);
+}
+```
+
+The issuer(), credentialType(), encryptionMethod(), verificationMethod() and credentialURI(uint256 tokenId) functions MUST be implemented as view functions.
+
+The encryptionMethod() function MUST return an empty string if and only if the credential is unencrypted. In case of encrypted credentials the method SHOULD return only the encryption method; Any additional encryption details required for decryption MAY be retrieved separately (a separate function or via the contract metadata).
```
SamWilsn, line 89: This bit is insufficiently defined. What methods are supported? If it's entirely implementation defined, you might want to note that.

ERCS/erc-7861.md
```diff
@@ -0,0 +1,205 @@
...
+
+The verificationMethod() function MUST return the method to verify the credential proof.
+
+The credentialURI(uint256 tokenId) function MAY be used to retrieve the credential location. In case a given implementation uses a different method to retrieve the credential, credentialURI(uint256 tokenId) MUST return an empty string.
+
+The supportsInterface method MUST return true when called with 0x7861069.
```
SamWilsn, line 95: You should reference ERC-165 here.

ERCS/erc-7861.md
```diff
@@ -0,0 +1,205 @@
...
+
+The isRevoked(uint256 tokenId) MUST return false if the credential has not been revoked. MAY return true if the credential does not exist.
```
SamWilsn, line 97: You don't specify when this function MUST return true

ERCS/erc-7861.md
```diff
@@ -0,0 +1,205 @@
...
+
+The CredentialIssued and CredentialRevoked events MAY be emitted to allow off-chain systems to monitor credential status efficiently. Credential revocation is not mandated, but MAY be implemented via burning the associated NFT.
```
SamWilsn, line 99: When are they emitted?

ERCS/erc-7861.md
```diff
@@ -0,0 +1,205 @@
...
+
+## Rationale
+This extension aims to provide a standardized way to represent verifiable credentials within the ERC-721 framework. The collection-wide properties (issuer and type) allow for efficient management of credentials where these properties are shared across all tokens in the collection.
```
SamWilsn, line 102: This paragraph is justifying the proposal as a whole, and so belongs in the Motivation section. My favorite analogy is:

> Motivation: we need to build a shed because...
> Rationale: we painted the shed red because...

ERCS/erc-7861.md
```diff
@@ -0,0 +1,205 @@
...
+
+## Security Considerations
+Implementers of the [ERC-7861](./eip-7861.md) standard must enforce strict access control to ensure only authorized parties can issue or revoke credentials. Implementers must also carefully consider access control for setting credential properties, protecting the integrity of the collection's credential information.
+
+Credentials should be non-transferable to maintain authenticity and prevent unauthorized transfers. Override transfer functions to enforce this if needed.
+
+The credentialURI should not contain sensitive information, as it is publicly accessible. Sensitive data should be encrypted and stored off-chain, with the credentialURI pointing to the location of this encrypted data; encryption is essential for any sensitive data stored on public storage solutions like IPFS.
+
+## Copyright
+Copyright and related rights waived via CC0.
```
SamWilsn, line 205: Please use the exact waiver from the template.

ERCS/erc-7861.md
```diff
@@ -0,0 +1,205 @@
...
+
+Once decrypted, the credential's authenticity and integrity can be verified using the verificationMethod, ensuring it remains untampered. For ZK proofs, specific queries can also be verified without disclosing the entire credential. The standard also accounts for credential revocation, which can be implemented in various ways: via burning the associated NFT, an additional structure in the contract or even an off-chain revocation list; in the first case, checking that the NFT is unburned confirms the credential's validity.
+
+While originally intended for verifiable credentials, this standard's flexibility allows it to link any private or external data payload to a user's wallet through an NFT, expanding its utility across various data-reference use cases.
+
+## Specification
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119.
+
+The verifiable credential extension is OPTIONAL for [ERC-721](./eip-721.md) contracts.
```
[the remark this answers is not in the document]
vmc-crossmint: Fair hahah

ERCS/erc-7861.md
vmc-crossmint: good point, the w3c itself has it super loosly defined

ERCS/erc-7861.md
vmc-crossmint: Moved it in the motivation part in s separate subsection

PR #852, opened 2025-01-15
Update ERC-7786: Extend ERC 7786 to support PULL communication

philippecamacho:

 - The build passes.

### DIFF
--- a/ERCS/erc-7786.md
+++ b/ERCS/erc-7786.md
@@ -111,7 +111,19 @@ The exact interface for any such action is out of scope of this ERC. If the `pos
 
 ### Reception Procedure
 
-A **Destination Gateway** is a contract that implements a protocol to validate messages sent on other chains. The interface of the destination gateway and how it is invoked is out of scope of this ERC.
+A **Destination Gateway** is a contract that implements a protocol to validate messages sent on other chains. The interface optionally specifies how an application can read message from the gateway.
+
+```solidity
+interface IERC7786GatewayDestination {
+
+    // Optional
+    function receiveMessage(
+        bytes messageId,
+        string calldata srcChainId, // [CAIP-2] chain identifier
+        string calldata srcAddress  // [CAIP-10] account address
+    ) external returns (bytes payload);
+}
+```
 
 The protocol MUST ensure delivery of a sent message to its **receiver** using the `IERC7786Receiver` interface (specified below), which the receiver MUST implement.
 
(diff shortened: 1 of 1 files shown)


ERCS/erc-7786.md
```diff
@@ -103,6 +103,12 @@ This event signals that a would-be sender has requested a message to be sent.
 
 If `outboxId` is present, post-processing MAY be required to send the message through the cross-chain channel.
 
+#### Mandatory and optional attributes
+
+The `attributes` parameter MUST contain as its first element the value `read(true)` or `read(false)`. `read(true)` means the message can be read by the application on the destination gateway. `read(false)` means the message by pushed by the destination gateway on the application via `executeMessage`(see below).
```
alxiong, line 108: 1. if mandatory, why not just explicitly enforce at the function signature level? i.e. `sendMessage(dest, recv, payload, bool pullable, attributes)` 
2. personally, i think `read(true)` is not intelligible. I prefer one of `pullable`, `supports_pull`, `pull_based`, `push_based`. 

wdyt? @frangio @elliedavidson

ERCS/erc-7786.md
```diff
@@ -103,6 +103,12 @@ This event signals that a would-be sender has requested a message to be sent.
...
+
+In the case `read(true)` is set as teh first element of the `attributes` parameters, the next element MUST be `messageId(val)` where val is a sequence of 32 bytes.
```
alxiong, line 110: - typo "teh" -> "the"
- didn't Francisco and CJ said they would want to add `msgId` as mandatory for all?

philippecamacho: > typo "teh" -> "the"

Thanks, fixed in 6ad57be0522f5bed7be627b1051335710d6b3376.

> didn't Francisco and CJ say they would want to add msgId as mandatory for all?

That is possible indeed.
cc @frangio  @cjcobb23

ERCS/erc-7786.md
```diff
@@ -103,6 +103,12 @@ This event signals that a would-be sender has requested a message to be sent.
 
 If `outboxId` is present, post-processing MAY be required to send the message through the cross-chain channel.
 
+#### Mandatory and optional attributes
```
frangio, line 106: All attributes should be optional. `read(false)` would be the default.

ERCS/erc-7786.md
```diff
@@ -111,7 +117,19 @@ The exact interface for any such action is out of scope of this ERC. If the `pos
...
+
+```solidity
+interface IERC7786GatewayDestination {
+
+    // Optional
+    function receiveMessage(
+        bytes messageId,
+        string calldata srcChainId, // [CAIP-2] chain identifier
+        string calldata srcAddress  // [CAIP-10] account address
+    ) external returns (bytes payload);
```
frangio, line 130: I think this is the more protocol-agnostic version:

```suggestion
    function receiveMessage(
        bytes calldata messageKey,
    ) external returns (Message memory);
```

ERCS/erc-7786.md
philippecamacho: Makes sense. I use the term "Default attributes" instead, in 15c1f1b7dea821026f012451b6191e1c93093c5a. Also following @alxiong's [suggestion](https://github.com/ethereum/ERCs/pull/852#discussion_r1917627298) I used the name `pushBased` instead of `read`.

ERCS/erc-7786.md
philippecamacho: `messageId` has been defined as a default attribute in 15c1f1b7dea821026f012451b6191e1c93093c5a.

ERCS/erc-7786.md
philippecamacho: `pushBased` used in 15c1f1b7dea821026f012451b6191e1c93093c5a.

ERCS/erc-7786.md
```diff
@@ -20,9 +20,9 @@ Cross-chain messaging protocols (or bridges) allow communication between smart c
...
+The objective of this ERC is to provide a standard interface, and corresponding workflows, for performing cross-chain communication between contracts. Existing cross-chain communication protocols that do not natively implement this interface should be able to adopt it using adapter gateway contracts.
 
-Compared to previous ERCs in this area, this ERC offers compatibility with chains outside of the Ethereum/EVM ecosystem, and it is extensible to support the different feature sets of various protocols while offering a shared core of standard functionality.
+Compared to previous ERCs in this area, this ERC offers compatibility with chains outside of the Ethereum/EVM ecosystem, and it is extensible to support the different feature sets of various protocols while offering a shared core of standard functionality. Moreover, this standard supports both PUSH and PULL communication modes, capturing a larger class of interoperability protocols.
```
cjcobb23, line 25: I actually have concerns about using push and pull in this way, based on our conversations in the working group. I think these terms are too broad, and mean different things to different people, and so adding this here could lead to confusion.

ERCS/erc-7786.md
```diff
@@ -111,11 +118,21 @@ The exact interface for any such action is out of scope of this ERC. If the `pos
 
 ### Reception Procedure
 
-A **Destination Gateway** is a contract that implements a protocol to validate messages sent on other chains. The interface of the destination gateway and how it is invoked is out of scope of this ERC.
+A **Destination Gateway** is a contract that allows the application to obtain the message. Optionally the Gateway can provide an interface to let the application on the destination chain fetch a message stored by the Gateway providing only the unique message identifier.
```
cjcobb23, line 121:
```suggestion
A **Destination Gateway** is a contract that receives the sent messages. The gateway either passes the message to the app via `executeMessage` (in the push based model), or stores the message for later reading (in the pull based model).
```

cjcobb23 (commented):
To be totally honest, I think the addition of this either push or pull operating mode has complicated the standard more than is optimal. I would actually prefer for all gateways to support both push and pull, and I think that would be cleaner. If an app doesn't want push, it can just choose to either not implement `executeMessage`, or implement it as a no-op. And for the storage concern (forcing all gateways to store all messages so they are queryable), gateways can still delete messages after some time if they choose to, with a mechanism to repopulate if neccessary; most messages will be consumed fairly quickly anyways, and repopulation can be triggered by retrying the relaying process from source.

Just my thoughts. I just think a big advantage of this ERC is it's simplicity, and I don't really like the added complexity of this switching behavior.

ERCS/erc-7786.md
philippecamacho: I agree, also because PUSH/PULL communication modes are not defined in this section. The sentence was removed in 39159f76.

PR #855, opened 2025-01-18
Update ERC-7744: Move to Review

peersky:
This PR concludes discussion that was facilitated on the Magician forums, particularly: 

- Addresses self-destruct concerns
- Addresses EIP7702 delegated address concern 
- Changes solidity file names according to standard number 
- Changes salt and deployment address to accommodate bytecode changes
- New artifact was compiled with no metadata included to avoid bytecode changes if somone tries to re-compile it locally


peersky:
@SamWilsn not sure how to fix the error, since required link for EIP7702 is not in this repo and AFAIK we cannot have external links as well

SamWilsn (approved):

PR #875, opened 2025-01-30
Update ERC-6900: Prefix ERC-6900 interfaces names to avoid conflicts

Amxx:
ERC-6900 defines multiples interfaces (`IModularAccount`, `IModule`, ...), some of which may clash with different interfaces defined in other ERCs. For example, both ERC-6900 and ERC-7579 define a `IModule` interfaces (and they are not compatibles).

Prefixing all interfaces declared in ERC-6900 with the name of the ERC avoid that kind of conflict. 

This is common practice: for example, ERC-721 and ERC-1155 defined `ERC721TokenReceiver` and `ERC1155TokenReceiver` respectivelly.

[A similar PR is targetting erc-7579.md with similar changes](https://github.com/ethereum/ERCs/pull/876)

PR #876, opened 2025-01-30
Update ERC-7579: Prefix ERC-7579 interfaces names to avoid conflicts

Amxx:
ERC-7579 defines multiples interfaces (`IExecution`, `IModule`, ...), some of which may clash with different interfaces defined in other ERCs. For example, both ERC-6900 and ERC-7579 define a `IModule` interfaces (and they are not compatibles).

Prefixing all interfaces declared in ERC-7579 with the name of the ERC avoid that kind of conflict. 

This is common practice: for example, ERC-721 and ERC-1155 defined `ERC721TokenReceiver` and `ERC1155TokenReceiver` respectivelly.

[A similar PR is targetting erc-6900.md with similar changes](https://github.com/ethereum/ERCs/pull/875)

### DIFF
--- a/ERCS/erc-7579.md
+++ b/ERCS/erc-7579.md
@@ -53,7 +53,7 @@ The smart account's validation function SHOULD return the return value of the va
 To comply with this standard, smart accounts MUST implement the execution interface below:
 
 ```solidity
-interface IExecution {
+interface IERC7579Execution {
     /**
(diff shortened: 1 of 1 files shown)


epoon-circle (commented):
Fwiw, I think the prefix makes sense, but I'll defer to someone actively working on AA for approval.

kopy-kat (approved):

Issue #877, opened 2025-01-31
Add Warnings in ERC-6492 Implementation


pcaversaccio:
### Summary

> Since the exploit vector is publicly known, I consider the risk of discussing this matter openly to be low. Therefore, I'm opening a standard issue.

Last week a contract was [exploited](https://x.com/Phalcon_xyz/status/1882630151583981787) using the [ERC-6492 reference implementation](https://eips.ethereum.org/EIPS/eip-6492#reference-implementation). An exemplary exploit contract can be retrieved [here](https://github.com/0xGreed/odos_exploit/blob/6b683cc676aa62797fb2870302b070446b60a2e0/test/OdosExploit.t.sol#L44-L68). So the major issue (apart from inheriting the implementation, which you should not do) is the combination of the identity precompile located at address `0x04` in combination with ERC-6492. Please note that there is an [EIP proposal](https://eips.ethereum.org/EIPS/eip-7666) to replace the identity precompile with EVM code which we might can leverage in the future.

### Actions to Discuss

- Add a warning that the universal verifier should not be inherited.
- We need to add a warning to the current [ERC-6492 reference implementation](https://eips.ethereum.org/EIPS/eip-6492#reference-implementation) regarding the usage of `_signer = address(0x04)` (and the combination with arbitrary calls). We could even go a step further, and either disallow that specific precompile address `0x04` or the full [precompile address range](https://eips.ethereum.org/EIPS/eip-1352): `0x00` - `address(2**16-1)`.
- Do we see any impacts of the newly discovered attack vector for [ERC-1271](https://eips.ethereum.org/EIPS/eip-1271) implementations?

Cc: @Ivshti @Agusx1211

h/t goes to @0xkarmacoma for raising this issue with me.

Agusx1211:
If I understand correctly, the exploit happened because "Odos Protocol" inherited from the `UniversalSigValidator` contract instead of deploying it as an independent contract and calling it. This is a bit of a footgun because the ERC does not explicitly mention that you DON'T need to inherit from the `UniversalSigValidator` or use it as a library internally.  

I think we should perform two different mitigations:  

1) Add a big warning that the `UniversalSigValidator` MUST NOT reside within any contracts of the protocol, and instead MUST be an external contract that is called when signatures need to be verified.  

[...]

PR #878, opened 2025-01-31
Add proxy call to 6492 reference implementation

Agusx1211:
Fixes one of the concerns raised here: https://github.com/ethereum/ERCs/issues/877

It forces all contract deployment calls to be made using a proxy, ensuring that if `UniversalSigValidator` is inherited then the `isValidSigImpl` method can't be used to make arbitrary calls.


SamWilsn:
Closing as per ethcatherders/EIPIP#381

PR #879, opened 2025-01-31
Update ERC-7573: added keyEncryptedSeller to ILockingContract TransferConfirmed event

Julius278:
added the already existing keyEncryptedSeller param to the ILockingContract TransferConfirmed event.

`/**
     * @dev Emitted  when the transfer for the token is incepted
     * @param id the trade identifier of the trade.
     * @param amount the number of tokens to be transferred.
     * @param from The address of the seller.
     * @param to The address of the buyer.
     * @param keyEncryptedSeller Encryption of the key that can be used by the seller to reclaim the token in case of cancellation or failed payment.
     * @param keyEncryptedBuyer Encryption of the key that can be used by the buyer to claim the token in case of a successful payment.
     */
    event TransferConfirmed(bytes32 id, int amount, address from, address to, string keyEncryptedSeller, string keyEncryptedBuyer);`

In this case its easier to call transferAndDecrypt (buyer) or cancelAndDecrypt (seller) without further interaction with the ILockingContract

### DIFF
--- a/ERCS/erc-7573.md
+++ b/ERCS/erc-7573.md
@@ -62,6 +62,7 @@ function confirmTransfer(bytes32 id, int amount, address to, string memory keyEn
 
 Called from the seller of the token to confirm token transfer. Emits a `TransferConfirmed` event.
 The parameter `id` is an identifier of the trade. The parameter `to` is the address of the buyer (the address of the seller is `msg.sender`).
+The parameter `keyEncryptedSeller` is an encryption of the key that can be used by the seller to reclaim the token.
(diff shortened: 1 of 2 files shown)


Julius278:
@cfries 
Hi Christian, could you please check that Pull Request?
I guess I dont have the correct access rights or something like that for the automated checks to be successful.
Best regards,
Julius

cfries:
I wanted to make changes to this PR in a review, but I do not have the rights... (or I do not know how to do it :).

cfries:
I do not understand why the tests are failing. This looks like a GitHub issue.

Julius278:
not needed anymore

PR #883, opened 2025-02-03
Update ERC-7821: Remove Call Struct from Specification

McOso:
Update (Feb 7, 2025): Based on convo, this PR just removes the `Call` struct from the Specification section since it is not needed to standardize the execution interface. The Reference Implementation has been left unchanged.

---

ERC-7821 implies using a `Call` struct. ERC-7579 already leverages an `Execution` struct in their reference implementation.

This is to change the `Call` struct to the referenced ERC-7579 `Execution` struct.

---

*Another option is to remove the struct from this standard altogether since you are really trying to standardize the execute interface*

### DIFF
--- a/ERCS/erc-7821.md
+++ b/ERCS/erc-7821.md
@@ -38,13 +38,6 @@ The minimal batch executor interface is defined as follows:
 ```solidity
 /// @dev Interface for minimal batch executor.
 interface IERC7821 {
-    /// @dev Call struct for the `execute` function.
-    struct Call {
-        address target; // Replaced with `address(this)` if `address(0)`.
-        uint256 value; // Amount of native currency (i.e. Ether) to send.
-        bytes data; // Calldata to send with the call.
-    }
-
     /// @dev Executes the calls in `executionData`.
     /// Reverts and bubbles up error if any call fails.
     ///


jxom:
Why does it need to match the 7579 struct?

McOso:
> Why does it need to match the 7579 struct? It's just a naming thing, no?

I just think it is helpful to match the same struct since the ERC is already referencing 7579. Really, call struct could just be removed from this ERC since the goal isn't to standardize the call struct, it is to standardize the execution interface.

jxom:
It references 7579 on the `execute` & `supportsExecutionMode` signatures specifically – I don't think other aspects of this ERC necessarily need to reference/be completely aligned with 7579.

Vectorized:
Won’t merge. 

I prefer `Call`, since it is shorter.

McOso:
>  I don't think other aspects of this ERC necessarily need to reference/be completely aligned with 7579 semantics.
Yeah, I'm leaning to it being just unnecessary to define any struct in this standard.

@Vectorized what if we remove the Call struct from the specification section, since you are trying to standardize the execution interface and not the call struct. But then you can use the Call struct in the reference implementation?

jxom:
Yeah, don't think it would hurt to remove L41-45 as it's not being used in the Overview. We can leave the Reference Implementation as-is.

McOso:
Updated the PR based on the convo. The change now just removes the `Call` struct from the Specification section but keeps the reference implementation as is.

PR #887, opened 2025-02-06
Update ERC-7291: Move to Review

veenkumarr:

 - The build passes.


wongtsejian (approved):

ERCS/erc-7291.md
```diff
@@ -4,7 +4,7 @@ title: Purpose bound money
 description: An interface extending ERC-1155 to implement purpose-bound money, a hybrid form of programmable payment and money.
```
SamWilsn, line 4:
```suggestion
description: An extension to ERC-1155 implementing a hybrid form of programmable payment and money.
```

This has a bit less fluff, but you can probably expand a little more, since you're most familiar with this proposal. The goal of the description is to add detail over what is introduced in the title.

SamWilsn (commented):
- On line 25: `s/taught/thought/`
- Your abstract could use a really short and simple definition of PBM right in the first line. Maybe something like: A PBM token is a wrapper around an ERC-1155 token that can be used for a limited activity. What you have is decent, but not succinct. Keep the technical stuff you have now.
- Your motivation section doesn't really explain why we need purpose bound money. Instead it explains why we need a standard. Your motivation could use one or two motivating use cases where PBMs are useful.
- Why define "PBM Wallet" as a term if it is exactly equivalent to just wallet?
- Line 49: `s/specify/specified/`
- Line 71: How is infinity encoded?
- You should separate the mandatory fields (eg. `name`, `faceValue`, etc.) into an interface that must be implemented. We generally don't want `contract` (abstract or not) in the specification section.
- Line 650: This is generally true for all smart contracts. Keep the security considerations section limited to considerations unique to your proposal (otherwise it'd be _huge_).

SamWilsn (approved):

PR #900, opened 2025-02-19
Update ERC-7812: Update ERC-7812 SMT circuit

Arvolear:


SamWilsn:
This is a Final proposal, so we only allow changes for clarifications and errors. Can you give me a summary of these changes and why they are necessary?

Arvolear:
> This is a Final proposal, so we only allow changes for clarifications and errors. Can you give me a summary of these changes and why they are necessary?

Hey @SamWilsn, it's definitely not Final. To be honest, I'll probably change the registry code as well due to some integration issues we encountered. This means the deployment address will also change. 

If possible, could you please merge this minor circuit update? For the changes described above I'll open a separate PR. 

image

PR #907, opened 2025-02-23
Update ERC-165: fix typo

0xdapper:


0xdapper:
All unrelated errors

ERCS/erc-165.md
```diff
@@ -88,7 +88,7 @@ Implementation note, there are several logical ways to implement this function.
 1. The source contract makes a `STATICCALL` to the destination address with input data: `0x01ffc9a701ffc9a700000000000000000000000000000000000000000000000000000000` and gas 30,000. This corresponds to `contract.supportsInterface(0x01ffc9a7)`.
 2. If the call fails or return false, the destination contract does not implement ERC-165.
 3. If the call returns true, a second call is made with input data `0x01ffc9a7ffffffff00000000000000000000000000000000000000000000000000000000`.
-4. If the second call fails or returns true, the destination contract does not implement ERC-165.
+4. If the second call fails or returns false, the destination contract does not implement ERC-165.
```
SamWilsn, line 91: Are you sure about this? From my limited understanding, it is unlikely that an ERC-165 contract would return `true` for an interface ID of `0xffffffff`.

0xdapper: Seems like I got confused. You are right. It is weird there is an explicit check for NOT supporting `0xffffffff`.

PR #918, opened 2025-02-25
Update ERC-7743: Move to Last Call

jamesavechives:
This ERC proposes a new standard for non-fungible tokens (NFTs) that supports multiple owners. The MO-NFT standard allows a single NFT to have multiple owners, reflecting the shared and distributable nature of digital assets. This model incorporates mechanisms for provider-defined transfer fees and ownership burning, enabling flexible and collaborative ownership structures. It maintains compatibility with the existing [ERC-721](https://eips.ethereum.org/EIPS/eip-721) standard to ensure interoperability with current tools and platforms.


SamWilsn (commented):
On lines 80/81 and 83/84 you talk about these functions being disabled or overridden. Later you mention "restricted". You don't specify what "restricted" means or how these functions should be overridden. Should `safeTransferFrom` not check if the owner is a contract? Should they always revert?

The rationale given (not aligning with the MO-NFT's ownership structure) doesn't make sense. The safe variants to exactly the same thing as the non-safe variants, except they check for contract support. Essentially, if `transferFrom` aligns with the ownership model, then so must `safeTransferFrom`.

---

Disabling the approval functions is similarly ambiguous, and severely limits the compatibility of the standard. I, as an owner, should be able to approve a contract to act on my behalf.

---

I expect a lot more detail in the Rationale section about _why_ the above choices were made. They very heavily modify the ERC-721 behaviour and so require heavy justification.

jamesavechives:
> On lines 80/81 and 83/84 you talk about these functions being disabled or overridden. Later you mention "restricted". You don't specify what "restricted" means or how these functions should be overridden. Should `safeTransferFrom` not check if the owner is a contract? Should they always revert?
> 
> The rationale given (not aligning with the MO-NFT's ownership structure) doesn't make sense. The safe variants to exactly the same thing as the non-safe variants, except they check for contract support. Essentially, if `transferFrom` aligns with the ownership model, then so must `safeTransferFrom`.
> 
> Disabling the approval functions is similarly ambiguous, and severely limits the compatibility of the standard. I, as an owner, should be able to approve a contract to act on my behalf.
> 
> I expect a lot more detail in the Rationale section about _why_ the above choices were made. They very heavily modify the ERC-721 behaviour and so require heavy justification.

OK, done!

ERCS/erc-7743.md
```diff
@@ -58,14 +59,15 @@ Traditional NFTs enforce a single-ownership model, which does not align with the
 
    - During a transfer, the specified `transferValue` fee is transferred to the provider. The contract must have sufficient balance to cover this fee.
```
SamWilsn, line 60: Just to check my understanding here, the MO-NFT contract itself transfers to the provider, so the expected pattern is:

- owner sends fee to the MO-NFT contract
- owner calls `transfer`
- MO-NFT contract sends fee from its own balance to the provider
- MO-NFT contract updates the active owner

ERCS/erc-7743.md
```diff
@@ -58,14 +59,15 @@ Traditional NFTs enforce a single-ownership model, which does not align with the
...
 
-3. **Burning Ownership**:
+3. **Archiving Ownership**:
 
-   - Owners can remove themselves from the ownership list using the `burn` function.
+   - Owners can mark themselves as archived for a specific token, meaning they can no longer transfer that token. Once set, this status cannot be reversed. And for Digital Asset platform, this status means the owner can download the real digital asset directly; if it is not archived, then it must be changed to archived first, then owner can download the real digital asset.
 
      ```solidity
-     function burn(uint256 tokenId) external;
+     function setArchivedStatus(uint256 tokenId, bool archived) external;
```
SamWilsn, line 67: If it cannot be unset, why do you need the bool? Wouldn't this be sufficient:

```suggestion
     function archive(uint256 tokenId) external;
```

ERCS/erc-7743.md
```diff
@@ -96,20 +98,21 @@ Traditional NFTs enforce a single-ownership model, which does not align with the
 
 - `function setTransferValue(uint256 tokenId, uint256 newTransferValue) external;`
 
-**Burn Function**
+**Archived Status Function**:
```
SamWilsn, line 101:
```suggestion
**Archived Status Function**
```

SamWilsn (approved):

ERCS/erc-7743.md
jamesavechives: Yes, you can also say like that. And finally the owner sends fee to the provider.

ERCS/erc-7743.md
jamesavechives: Yes, it seems to be right, I think I will take it into next pull request. Thanks!

PR #919, opened 2025-02-25
Update ERC-7750: Move to Last Call

jamesavechives:
This ERC proposes a Decentralized Employment System (DES) built on the Ethereum blockchain. The DES facilitates the creation and management of companies, records comprehensive employment histories through unique employee tokens, enables the formation and execution of labor contracts, automates salary payments via an escrow mechanism, incorporates a robust moderation system for dispute resolution, and implements a reputation-based review system for both employers and employees. By leveraging blockchain’s transparency and immutability, the DES ensures accountability and trust throughout the employment lifecycle, from company creation and hiring to contract fulfillment and termination.

The system operates post employee testing and prior to the final hiring and contract signing. Employees possess a Soulbound Token (SBT) representing their employment history, which companies review before finalizing labor contracts. This token-based approach ensures a secure and verifiable employment record that enhances the hiring process’s integrity.


Rafazzzz (approved):

SamWilsn:
You mentioned on the discussion thread that you "intend to build" an implementation of this standard. I would recommend holding off going into Last Call until you've gone through most of the implementation work. It's _extremely_ common for standard to change during the implementation process, and you cannot update Final proposals. Any changes would have to be made in a new proposal, effectively making this one obsolete.

I won't stop you from going to Last Call, but I wanted to raise this issue first.

SamWilsn:
@jamesavechives Just wondering how you'd like to proceed on this.

jamesavechives:
> @jamesavechives Just wondering how you'd like to proceed on this.

I think you are right! It is really hard to form a common decentralized employment standard, the real purpose of mine is to build a system to record employment history on the blockchain.

jamesavechives:
> Judging by the `registerCompany` function, this contract looks like it's intended to be shared between multiple employers/employees. Put another way, it looks like there would only ever be a single instance of this contract deployed on chain. Is that the case?
> 
> If so, you don't need an ERC. ERCs are useful when different parties need to coordinate without prior communication. ERC-20 and wallets are a great example: every token wants to work in every wallet and every wallet wants to support every token, but doing T*W integrations would be an impossible amount of work. Standardizing the interface makes all tokens interoperable.
> 
> Here, however, it looks like there's only one DES contract. Anyone who wants to integrate with it can just read your project's documentation.

OK, I will try it.

PR #981, opened 2025-03-19
Update ERC-7786: Add `requestRelay`

frangio:


ERCS/erc-7786.md
```diff
@@ -74,6 +74,13 @@ interface IERC7786GatewaySource {
         bytes calldata payload,
         bytes[] calldata attributes
     ) external payable returns (bytes32 outboxId);
+
+    function requestRelay(
+        bytes32 outboxId,
+        uint256 gasLimit,
+        uint256 relayerFee,
+        string calldata refundReceiver
```
cjcobb23, line 82: why is this calldata as opposed to just an address?

ERCS/erc-7786.md
```diff
@@ -74,6 +74,13 @@ interface IERC7786GatewaySource {
         bytes calldata payload,
         bytes[] calldata attributes
     ) external payable returns (bytes32 outboxId);
+
+    function requestRelay(
+        bytes32 outboxId,
+        uint256 gasLimit,
+        uint256 relayerFee,
```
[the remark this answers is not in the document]
cjcobb23: as opposed to just letting the relayer deduct the fee from the attached value

ERCS/erc-7786.md
```diff
@@ -109,6 +116,14 @@ After a sender has invoked `sendMessage`, further action MAY be required by the
 
 The exact interface for any such action is out of scope of this ERC. If the `postProcessingOwner` attribute is supported and present, such actions MUST be restricted to the specified account, otherwise they MUST be able to be performed by any party in a way that MUST NOT be able to compromise the eventual receipt of the message.
 
+#### `requestRelay`
+
+Requests that the message identified by `outboxId` be delivered to the receiver (i.e., relayed).
```
cjcobb23, line 121:
```suggestion
Requests that the message identified by `outboxId` be delivered to the receiver (i.e., relayed). This includes execution on the destination chain, as well as any intermediate processing necessary prior to execution, such as proof generation, verification, relaying to the interop provider, etc.
```

ERCS/erc-7786.md
```diff
@@ -109,6 +116,14 @@ After a sender has invoked `sendMessage`, further action MAY be required by the
...
+
+Pre-pays via `msg.value` for the execution of the message on the receiver up to `gasLimit`. A `relayerFee` should be deducted from this to compensate the relayer.
+
+Any surplus after execution at destination MUST be refunded to `refundReceiver` on the destination chain.
+
```
cjcobb23, line 126: We probably want a blurb about how relaying should be permissionless. Something like:

Relaying should be permissionless. Callers should be free to self relay or use a different relayer service, in which case they should not need to call `requestRelay` on the gateway. Messages should be considered valid and ready for execution without `requestRelay` being called. Relaying can be a competitive marketplace, and no relayer should be relied on for liveness of the protocol.

ERCS/erc-7786.md
```diff
@@ -109,6 +116,14 @@ After a sender has invoked `sendMessage`, further action MAY be required by the
 
 The exact interface for any such action is out of scope of this ERC. If the `postProcessingOwner` attribute is supported and present, such actions MUST be restricted to the specified account, otherwise they MUST be able to be performed by any party in a way that MUST NOT be able to compromise the eventual receipt of the message.
 
+#### `requestRelay`
+
```
cjcobb23, line 120: we probably want a blurb about what relaying even is

ERCS/erc-7786.md
```diff
@@ -74,6 +74,13 @@ interface IERC7786GatewaySource {
         bytes calldata payload,
         bytes[] calldata attributes
     ) external payable returns (bytes32 outboxId);
+
+    function requestRelay(
```
cjcobb23, line 78: are there any bridges that don't actually have a concept of a relayer? Maybe something in the l2 environment

ERCS/erc-7786.md
```diff
@@ -109,6 +116,14 @@ After a sender has invoked `sendMessage`, further action MAY be required by the
 
 The exact interface for any such action is out of scope of this ERC. If the `postProcessingOwner` attribute is supported and present, such actions MUST be restricted to the specified account, otherwise they MUST be able to be performed by any party in a way that MUST NOT be able to compromise the eventual receipt of the message.
 
+#### `requestRelay`
+
+Requests that the message identified by `outboxId` be delivered to the receiver (i.e., relayed).
+
+Pre-pays via `msg.value` for the execution of the message on the receiver up to `gasLimit`. A `relayerFee` should be deducted from this to compensate the relayer.
```
cjcobb23, line 123: there may be additional fees besides just the relayer fee. Such as, do any protocols just charge protocol fees at the gateway itself? I feel like it might be better to let all of the fees be sort of opaque, instead of specifying the fees separately from the gas needed for execution. I'd argue the user doesn't really care about the breakdown; just really care how much they need to pay to get the message executed at the destination.

cjcobb23:
> 3. Should it be possible to pay with a different token?

I don't think this should be part of the standard. I think gas services will implement this, but it shouldn't be required.

cjcobb23:
> 2. Should we define a quoting/estimation mechanism for gas price?

I think so. I think more generally there should just be an estimate fee mechanism, that returns exactly what the user should pay to get the message delivered and executed. So it would also include any sort of fixed costs, in addition to the gas price.

cjcobb23:
> 1. Should this functionality be in a contract separate from the gateway?

I think the gateway should include these functions, so that way everything a user needs to send a cross chain message is encapsulated in one contract. But it should be possible for a separate contract to implement just the gas service part of the interface, so that way users can use a different gas service than whatever is tied to the gateway.

ERCS/erc-7786.md
frangio: Because it's an address on the destination chain it has to be in CAIP-10 format or whatever format we end up using. It's not an EVM address necessarily.

ERCS/erc-7786.md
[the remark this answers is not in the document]
frangio: Because if the gas price at destination turns out lower than estimated, that would result in surplus that we want to have refunded to the sender, not turned into additional fees for the relayer.

An equivalent alternative to specifying `relayerFee` would be to specify a `maxGasPrice` and deduce `relayerFee = msg.value - (gasLimit * maxGasPrice)`.

ERCS/erc-7786.md
frangio:
```suggestion
Requests that the message identified by `outboxId` be delivered to the receiver (i.e., relayed). This includes execution on the destination chain, and MAY include other intermediate processing necessary prior to execution, such as proof generation, verification, relaying to the interop provider, etc.
```

ERCS/erc-7786.md
frangio: In L2 environments, as far as I've seen, relaying is present but in a different form where it's just built into the protocol and doesn't rely on third parties. But other than that it's very similar in the sense that you need to pay for execution.

ERCS/erc-7786.md
frangio: Replying to this in https://github.com/ethereum/ERCs/pull/981#discussion_r2004117940.

ERCS/erc-7786.md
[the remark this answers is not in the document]
frangio: From https://github.com/ethereum/ERCs/pull/981#discussion_r2003695846:

>there may be additional fees besides just the relayer fee.

This could be a good reason to specify `maxGasPrice` instead of `relayerFee`. But that has a similar problem, where it could be that gas price at destination is multi-dimensional, so a single `maxGasPrice` is not a good fit either. 

I'm not sure if either of this options is a clear winner.

ERCS/erc-7786.md
kelemeno, line 123: what if the destination chain has a different custom base token?

frangio: This is another open question for me. It could have to be paid directly in the destination chain gas token, but this may not be the UX we want. It may need to be swapped for the destination chain token, either at source or destination, though this sounds like it may leave the sender vulnerable to attacks if not designed correctly.

ERCS/erc-7786.md
Amxx: Going from mainnet to matic (polygon) I'm not sure I used to pay anything (other than the high gas cost of sending the message on mainnet)

ERCS/erc-7786.md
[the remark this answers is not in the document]
frangio: Resolved to remove `relayerFee` and handle this out of band.

ERCS/erc-7786.md
frangio: I went with the simpler version below because I think most of the above becomes tricky with L2s.

> Note that relaying SHOULD be permissionless. No relayer should be relied on for liveness of the protocol.

ERCS/erc-7786.md
```diff
@@ -74,6 +74,21 @@ interface IERC7786GatewaySource {
...
+    function quoteRelay(
+        string calldata destinationChain,
+        string calldata receiver,
+        bytes calldata payload,
+        bytes[] calldata attributes
+        uint256 gasLimit,
+        string[] calldata refundReceivers
+    ) external returns (uint256);
+
+    function requestRelay(
```
jboetticher, line 87: In a protocol where the sending & the relay requesting are not separated, what should this function do?

frangio: In this case, `sendMessage` would not actually interact with the protocol, the message would be stored until `requestRelay` is invoked. Alternatively, the `requestRelay` parameters are included as an attribute and it can all be done in one go.

jboetticher: Sounds like the `sendMessage` function would be quite the misnomer then 😄 

Why separate the sendMessage & relay functionality in this standard at all? I know this closely follows the Axelar workflow so maybe @cjcobb23 can comment

ERCS/erc-7786.md
```diff
@@ -74,6 +74,21 @@ interface IERC7786GatewaySource {
         bytes calldata payload,
         bytes[] calldata attributes
     ) external payable returns (bytes32 outboxId);
+
+    function quoteRelay(
+        string calldata destinationChain,
+        string calldata receiver,
+        bytes calldata payload,
+        bytes[] calldata attributes
```
kelemeno, line 82: missing comma after "attributes"

ERCS/erc-7786.md
```diff
@@ -87,27 +102,49 @@ A gateway MAY be upgraded with support for additional attributes. Once present s
...
 
-#### Post-processing
+Returns an estimated amount of native token required to relay a message with the provided parameters and with a certain gas limit.
 
-After a sender has invoked `sendMessage`, further action MAY be required by the gateways to make the message effective. This is called _post-processing_. For example, some payment is typically required to cover the gas of executing the message at the destination.
+#### `requestRelay`
+
+Requests that the message identified by `outboxId` be relayed and delivered to the receiver.
+
+Pays via `msg.value` for the execution of the message on the receiver up to `gasLimit`. A relayer fee MAY be deducted, the specifics of which are unspecified and SHOULD be available out of band (e.g., as gateway documentation).
```
kelemeno, line 135: This does not support custom base tokens if I understand correctly. Or should the conversion be handled by the protocol? It cannot be handled by native bridges though.

frangio:
This PR now builds on https://github.com/ethereum/ERCs/pull/1050 so until that PR is merged the diff in the Files tab is not helpful. I recommend looking at the [diff here](https://github.com/frangio/ERCs/compare/7786-wg-updates..7786-request-relay#diff-07848a22c16b11582927b37d76b2c54a0617b1444b30d11bca783ac98eaed01f) instead.

LucasGrasso (approved):

ERCS/erc-7786.md
```diff
@@ -153,10 +153,60 @@ MUST return `IERC7786Recipient.receiveMessage.selector` (`0x2432ef26`).
...
+interface IERC7786Relayer {
+    // OPTIONAL
+    function quoteRelay(
+        bytes calldata recipient, // Binary Interoperable Address
+        bytes calldata payload,
+        bytes[] calldata attributes,
+        uint256 value,
+        uint256 gasLimit,
+        address refundRecipient
+    ) external returns (uint256);
```
Amxx:
```suggestion
    ) external view returns (uint256);
```

ERCS/erc-7786.md
```diff
@@ -153,10 +153,60 @@ MUST return `IERC7786Recipient.receiveMessage.selector` (`0x2432ef26`).
...
+        bytes calldata recipient, // Binary Interoperable Address
+        bytes calldata payload,
+        bytes[] calldata attributes,
+        uint256 value,
+        uint256 gasLimit,
+        address refundRecipient
+    ) external view returns (uint256);
+
+    function requestRelay(
+        bytes32 sendId,
```
cjcobb23, line 179: I think we should pass the full message params here, so the interface will work better with existing protocols. I.e. pass the recipient and payload.

cjcobb23: we could even keep this version to use for top ups (adding more payment to an existing call).

frangio:
Will propose as separate ERC in the future.

PR #986, opened 2025-03-21
Add ERC: Signature Verifiers

Amxx:


ERCS/erc-xxxx.md
```diff
@@ -0,0 +1,111 @@
...
+
+```solidity
+interface IERCXXXXKeyVerifier {
+  /**
+   * @dev Helper returning wether a key is properly formatted. For example if the verifier verifies elliptic curve
+   * cryptography such as P256, this function will check that the key corresponds to a point on the curve.
+   */
+  function isValidKey(bytes calldata key) external view returns (bool);
+
+  /**
```
Amxx, line 52:
```suggestion
  /**
```

ERCS/erc-xxxx.md
```diff
@@ -0,0 +1,105 @@
...
+        } catch {
+          return false;
+        }
+      }
+  }
+}
+```
+
+## Rationale
+
```
Amxx, line 92:
```suggestion

TODO

```

ERCS/erc-xxxx.md
```diff
@@ -0,0 +1,107 @@
+---
+eip: XXXX
+title: Key Verifiers
+description: Standard  way to verify signatures by address-less key
+author: Hadrien Croubois (@Amxx), Ernesto García (@ernestognw), Francisco Giordano (@frangio)
```
Amxx, line 5:
```suggestion
author: Hadrien Croubois (@Amxx), Ernesto García (@ernestognw), Francisco Giordano (@frangio), Aryeh Greenberg (@ar00)
```

ERCS/erc-7913.md
```diff
@@ -0,0 +1,107 @@
+---
+eip: 7913
+title: Key Verifiers
+description: Standard  way to verify signatures by address-less key
```
Amxx, line 4:
```suggestion
description: Interface for the verification of signatures by address-less keys
```

ERCS/erc-7913.md
```diff
@@ -0,0 +1,107 @@
...
+author: Hadrien Croubois (@Amxx), Ernesto García (@ernestognw), Francisco Giordano (@frangio), Aryeh Greenberg (@ar00)
+discussions-to: https://ethereum-magicians.org/t/<todo>
```
Amxx, line 6:
```suggestion
discussions-to: https://ethereum-magicians.org/t/erc-7913-key-verifiers/23262
```

ERCS/erc-7913.md
```diff
@@ -0,0 +1,107 @@
...
+
+```solidity
+interface IERCXXXXKeyVerifier {
+  /**
+   * @dev Main function that verifies if `signature` is a valid signature by `key` of `hash`.
+   *
+   * MUST return the bytes4 magic value IERCXXXXKeyVerifier.isValidKeySignature.selector if the signature is valid.
+   * SHOULD return 0xffffffff or revert if the signature is not valid.
+   */
+  function isValidKeySignature(bytes calldata key, bytes32 hash, bytes calldata signature) external view returns (bytes4);
```
frangio, line 52: We're using the verb "verification" so I would go with `verifySignature` or even just `verify` since this is a single-purpose contract we don't need to be maximally unambiguous.

ERCS/erc-7913.md
```diff
@@ -0,0 +1,107 @@
...
+eip: 7913
+title: Key Verifiers
+description: Interface for the verification of signatures by address-less keys
+author: Hadrien Croubois (@Amxx), Ernesto García (@ernestognw), Francisco Giordano (@frangio), Aryeh Greenberg (@ar00)
+discussions-to: https://ethereum-magicians.org/t/erc-7913-key-verifiers/23262
+status: Draft
+type: Standards Track
+category: ERC
+created: 2025-03-21
+requires: 1271
```
frangio, line 11: This shouldn't be needed. If we remove the backwards-compatible verification procedure from the specification it will be possible to remove it.

ERCS/erc-7913.md
```diff
@@ -0,0 +1,107 @@
...
+- Keys are described as a `bytes` object of arbitrary length. For example a P256 or RSA public key.
+- Verifiers are smart contracts in charge of signature verification for a given key. They have an ethereum address.
+- A signer is a `bytes` object that is the concatenation of the verifier's address, and the key: `verifier || key`. A signer is at least 20 bytes long.
+
+### Key Verifier interface
+
+Verifiers MUST implement the following interface
+
+```solidity
+interface IERCXXXXKeyVerifier {
```
frangio, line 45: Here I would go with `SignatureVerifier` (`IERC7913SignatureVerifier`).

ERCS/erc-7913.md
```diff
@@ -0,0 +1,108 @@
...
+
+TODO
+
+## Backwards Compatibility
+
+This ERC proposed a new contract interface, with no consequences to the existing system.
+
+The signer format described here is backward compatible with EOA and ERC-1271 signers. Existing signers will be compatible with systems that adopt this new signer representation model.
+
+## Reference implementation
```
Amxx, line 77:
```suggestion
## Reference Implementation
```

ERCS/erc-7913.md
```diff
@@ -0,0 +1,107 @@
...
+   * @dev Main function that verifies if `signature` is a valid signature by `key` of `hash`.
+   *
+   * MUST return the bytes4 magic value IERCXXXXKeyVerifier.isValidKeySignature.selector if the signature is valid.
+   * SHOULD return 0xffffffff or revert if the signature is not valid.
+   */
+  function isValidKeySignature(bytes calldata key, bytes32 hash, bytes calldata signature) external view returns (bytes4);
+}
+```
+
+### Signature verification
```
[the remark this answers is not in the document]
Amxx: Would you agree with keeping the logic here, and moving only the code to the Reference Implementation section ?

ERCS/erc-7913.md
```diff
@@ -0,0 +1,108 @@
...
+status: Draft
+type: Standards Track
+category: ERC
+created: 2025-03-21
+---
+
+## Abstract
+
+Externally Owned Accounts (EOA) can sign messages with their associated private keys. Additionally [ERC-1271](./eip-1271.md) defines a method for signature verification by smart accounts such as multisig. In both cases the identity of the signer is an ethereum address. We propose a standard to extend this concept of signer description, and signature verification, to keys that do not have an ethereum identity of their own, in the sens that they don't have their own address to represent them.
+
```
arr00, line 16:
```suggestion
Externally Owned Accounts (EOA) can sign messages with their associated private keys. Additionally [ERC-1271](./eip-1271.md) defines a method for signature verification by smart accounts such as multisig. In both cases the identity of the signer is an ethereum address. We propose a standard to extend this concept of signer description, and signature verification, to keys that do not have an ethereum identity of their own, in the sense that they don't have their own address to represent them.

```

ERCS/erc-7913.md
```diff
@@ -0,0 +1,108 @@
...
+The `(verifier, key)` pairs can be given permission to control a smart account, perform social recovery, or do any other operation without ever having a dedicated on-chain address. Systems that want to adopt this approach need to transition away from the model were signers are identified by their address to a new model where signers may not have an address, and are identified by `bytes` object.
```
arr00, line 25:
```suggestion
The `(verifier, key)` pairs can be given permission to control a smart account, perform social recovery, or do any other operation without ever having a dedicated on-chain address. Systems that want to adopt this approach need to transition away from the model were signers are identified by their address to a new model where signers may not have an address, and are identified by a `bytes` object.
```

Amxx:
This is ready for review @g11tech, @SamWilsn, @xinbenlv

ERCS/erc-7913.md
```diff
@@ -0,0 +1,112 @@
...
+          return false;
+        }
+      }
+  }
+}
+```
+
+## Security Considerations
+
+TODO
```
xinbenlv, line 108: Author made a explict choice that "Verifiers SHOULD be stateless.", in which case `verify` will be `view/pure` and not causing state change. I'd suggest we remind compliant contract developer to add logic to prevent replay attack: such as if their application rely on the state such as fund availablity, nonce, or whether another internal TX has been performed.

xinbenlv (approved):
Leaving technical feedback on Ethereum Magician

PR #987, opened 2025-03-21
Add ERC: An Interface for Transfer From Native

snreynolds:

 - The build passes.


ERCS/erc-7914.md
```diff
@@ -0,0 +1,157 @@
+---
+eip: 7914
+title: An Interface for Transfer From Native
+description: This proposal introduces a method for approved spenders to execute native token transfers on behalf of users.
+author: Sara Reynolds (@snreynolds), Eric Zhong (@zhongeric), Daniel Gretzke (@gretzke), Diana Kocsis (@dianakocsis), Mark Toda (@marktoda), and Ryan McPeck (@McOso)
+discussions-to: <TODO>
+status: Draft
+type: Standards
+category: ERC
+created: <TODO>
```
bomanaps, line 10: This should contain the date on which the proposal was initiated, regardless of its state.

ERCS/erc-7914.md
```diff
@@ -0,0 +1,157 @@
...
+
+- `from` : The account that the ETH transfer should originate from. If this function is implemented on the spender’s contract account, then it should check `from` is set to `address(this)`. If `from` is not `address(this)` the contract COULD forward the call. Note that specifically parameterizing a `from` address allows for flexibility, and other types of integrations with this standard.
+- `recipient`: The recipient of the Native transfer.
+- `amount`: The amount of Native to transfer.
+
+SHOULD early return when amount == 0, otherwise anyone can execute any fallback function on any contract on behalf of the user without explicit approval.
+
+MUST verify whether `msg.sender` (the caller) is authorized to execute the Native transfer.
+
+MUST handle the Native transfer and return false on failure.
```
Vectorized, line 40: Could clarify on revert or return false. Add rationale for choice.

ERCS/erc-7914.md
```diff
@@ -0,0 +1,157 @@
...
+interface ERC165 {
+    /// @notice Query if a contract implements an interface
+    /// @param interfaceID The interface identifier, as specified in ERC-165
+    /// @dev Interface identification is specified in ERC-165. This function
+    ///  uses less than 30,000 gas.
+    /// @return `true` if the contract implements `interfaceID` and
+    ///  `interfaceID` is not 0xffffffff, `false` otherwise
+    function supportsInterface(bytes4 interfaceID) external view returns (bool);
+}
+```
```
Vectorized, line 102: imo, to avoid inheritance headaches, would prefer if there's some way to recycle one of the existing functions.

I think if we have some `nativeAllowance(address spender)`, we can use it as a signaling mechanism.

Technically, `transferFromNative` with a zero amount could be used to signal too.

ERCS/erc-7914.md
```diff
@@ -0,0 +1,157 @@
...
+
+The following event MUST be emitted when approvals are updated:
+
+```solidity
+event ApproveNative(address indexed owner, address indexed spender, uint256 value);
+```
+
+### Optional Transient Approval
+
+For cases where approval should only persist within a single transaction, a transient approval mechanism MAY be used. This allows a user to grant temporary approval to a `spender` for a single transaction, without modifying persistent storage.
```
Vectorized, line 71: I think need clarity on which native allowance to deduct first. Would prefer transient to be deducted first.

snreynolds: Ah that is interesting. Yeah definitely needs some outline for order of operations. My gut impl would be to default first to transient like you said. 

I'm not sure how we should handle the case then if there is more requested than the transient amount allowed. Do you then pull from the persisted allotment? It feels like that may be user error of the transient allotment if the caller didn't request enough to be allowed transiently?

dianakocsis: i think it would be cleaner if we just separated them - `transferFromNative` and `transferFromNativeTransient`. that way it's easier to know which allowance to subtract from

dianakocsis:
would also be good to include `allowance(address spender)` and `transientAllowance(addressSpender)` in the interface

ERCS/erc-7914.md
```diff
@@ -0,0 +1,157 @@
...
+
+```solidity
+function transferFromNative(address from, address recipient, uint256 amount) public returns (bool);
+```
+
+- `from` : The account that the ETH transfer should originate from. If this function is implemented on the spender’s contract account, then it should check `from` is set to `address(this)`. If `from` is not `address(this)` the contract COULD forward the call. Note that specifically parameterizing a `from` address allows for flexibility, and other types of integrations with this standard.
+- `recipient`: The recipient of the Native transfer.
+- `amount`: The amount of Native to transfer.
+
+SHOULD early return when amount == 0, otherwise anyone can execute any fallback function on any contract on behalf of the user without explicit approval.
```
zhongeric, line 36: Should early return with `true`, same for transient function

ERCS/erc-7914.md
```diff
@@ -0,0 +1,157 @@
...
+```
+
+### Persistent Approval
+
+An approval system similar to the ERC-20 `approve` mechanism MUST be implemented to allow users to specify trusted accounts that can perform Native transfers on their behalf.
+
+To modify this approval, the following function MUST be implemented.
+
+```solidity
+function approveNative(address spender, uint256 amount) external returns (bool);
```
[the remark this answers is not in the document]
0xmonsoon: I think it would be better to rename the functions: `nativeAllowance(address spender)` and `transientNativeAllowance(address spender)`

0xvanbeethoven:
It would be really useful to add a `permitNative` that works in the same way as `permit` so users can approve native eth transfers without having to execute a tx.

ERCS/erc-7914.md
```diff
@@ -0,0 +1,160 @@
+---
+eip: 7914
+title: An Interface for Transfer From Native
```
SamWilsn, line 3: > An Interface for

This bit doesn't really add much to your title. Because space is so limited (~44 characters), I'd recommend removing it and trying to make your title more information dense.

For example:

```suggestion
title: Approved Spenders for Native Ether
```

ERCS/erc-7914.md
```diff
@@ -0,0 +1,160 @@
...
+description: This proposal introduces a method for approved spenders to execute native token transfers on behalf of users.
```
SamWilsn, line 4: Similarly, here in your description, there's a lot of text that doesn't actually add any information. You could trim this down like so:

```suggestion
description: Approved spenders for native token transfers from EIP-7702 EOAs and ERC-4337 smart accounts on behalf of users.
```

ERCS/erc-7914.md
```diff
@@ -0,0 +1,160 @@
...
+discussions-to: https://ethereum-magicians.org/t/erc-7914-an-interface-for-transfer-from-native/24523
+status: Draft
+type: Standards Track
+category: ERC
+created: 2025-03-21
+---
+
+## Abstract
+
+This proposal introduces a method for native transfers that can originate from an approved spender on a smart contract wallet or a [ERC-7702](./eip-7702.md) delegated EOA account. For wallets that can execute code, other contracts can be authorized to pull native tokens via a new callback mechanism.
```
SamWilsn, line 15: Your abstract does a good job of sketching out the goals of the proposal, but is lacking a high-level but still technical overview of _how_ the proposal accomplishes those goals. Essentially, your abstract should additionally answer the question: what is the method?

ERCS/erc-7914.md
```diff
@@ -0,0 +1,160 @@
...
+
+## Motivation
+
+Because Native transfers must be initiated by the calling account, spenders cannot pull Native tokens without executing code on a user's wallet. This presents a problem in intent-based systems, where transactions are expected to be sent on behalf of users rather than initiated by the users themselves.
```
SamWilsn, line 19: Why are you capitalizing native?

ERCS/erc-7914.md
```diff
@@ -0,0 +1,160 @@
...
+
+## Specification
+
+The contract code MUST implement the following function:
+
+```solidity
+function transferFromNative(address from, address recipient, uint256 amount) public returns (bool);
+```
+
+- `from` : The account that the ETH transfer should originate from. If this function is implemented on the spender’s contract account, then it should check `from` is set to `address(this)`. If `from` is not `address(this)` the contract COULD forward the call. Note that specifically parameterizing a `from` address allows for flexibility, and other types of integrations with this standard.
```
SamWilsn, line 31: > Note that specifically parameterizing a `from` address allows for flexibility, and other types of integrations with this standard.

This type of content belongs in the Rationale section.

ERCS/erc-7914.md
```diff
@@ -0,0 +1,160 @@
...
+
+MUST emit the following event on success:
+
+```solidity
+event TransferFromNative(address indexed from, address indexed to, uint256 value);
+```
+
+### Persistent Approval
+
+An approval system similar to the [ERC-20](./eip-20.md) `approve` mechanism MUST be implemented to allow users to specify trusted accounts that can perform Native transfers on their behalf.
```
SamWilsn, line 49: > An approval system similar [...]

In this line, "similar" here is too vague. You should either exactly copy the approval from ERC-20 or define the system in this document.

ERCS/erc-7914.md
```diff
@@ -0,0 +1,160 @@
...
+
+Note that a spender could be approved for funds BOTH transiently and persistently and that users of both of these approval mechanisms should be aware that the total sum is approved.
+
+
+Similar to the persistent approval mechanism, a view function to read the transient allowance given a spender can be implemented:
+```solidity
+function transientNativeAllowance(address spender) public view returns (uint256)
+```
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
```
SamWilsn, line 107: This should be moved to the top of the Specification section.

ERCS/erc-7914.md
```diff
@@ -0,0 +1,160 @@
...
+
+## Rationale
+
+- Enables seamless ETH transfers without requiring users to submit raw transactions themselves.
+- Leverages the capabilities introduced by ERC-7702, which allows EOAs to execute code.
+- Adopts a standardized approval mechanism similar to ERC-20 for developer experience.
+- Outlines both persistent and transient approval mechanisms to minimize hanging approvals for tighter security on the account.
```
SamWilsn, line 114: All of this belongs in the Motivation section because it's justifying the proposal as a whole. The Rationale section should be used to explain choices made within your document. My pet analogy for the two sections are:

> Rationale: We decided to paint our shed red because...

PR #998, opened 2025-03-30
Update ERC-7738: Move to Final

JamesSmartCell:


SamWilsn:
You need to follow the process defined in EIP-1: https://eips.ethereum.org/EIPS/eip-1#eip-process

PR #1002, opened 2025-04-03
feat: ERC initial draft

0xteddybear:

 - The build passes.

### DIFF
--- a/ERCS/erc-X.md
+++ b/ERCS/erc-X.md
@@ -1,0 +1,220 @@
+---
+eip: X
+title: Interoperable Addresses
+description: An extensible binary format to refer to an address specific to one chain.
+author: Teddy (@0xteddybear), Joxes (@0xJoxess), Nick Johnson (@Arachnid)
+status: Draft
+discussions-to: 
+type: Standards Track
+category: ERC
+created: 2025-02-02
+requires: 
+---
+
+## Abstract
+Interoperable Addresses is a binary format to describe an address specific to one chain with an optional human-readable name representation.
+It is designed in a manner that allows the community extend it to support naming schemes (or other not-yet-imagined features) in the future, as well as to support arbitrary-length addresses and chainids.
+
+## Motivation
+The Ethereum ecosystem is rapidly expanding into a multi-chain environment encompassing L1s, L2s, sidechains, and rollups—some EVM‐compatible, others not. A simple address alone is not sufficient to describe a blockchain account, since it does not identify which chain the address belongs to, creating ambiguity for wallets, dApps, and users. At the same time, it is important for the format to be extensible to support both current standards for address naming and future standards for chain naming.
+
+The current standard ethereum address representation defined in EIP-55, with _optional_ retrofitted checksums implemented via selective capitalization of `a-f` characters, poses the following challenges:
+
+- Verification of the checksum by a client is optional, meaning that mistyped or erroneous addresses may be accepted by these clients. There is no way for a user to distinguish a correct address from an incorrect one processed by a client that does not do validation.
+- Addresses do not include any indication of the chain to which the address applies. This means chain information must be expressed out of band, and introduces the risk of an address being mistakenly used on a chain it is not valid for. This risk is particularly pronounced for addresses that represent smart contracts.
+- The existing address format provides no mechanism for extension, meaning that there is no natural method to add these features to the existing address format.
+
+Interoperable Addresses build on insights from ERC-7828, CAIP-10 and CAIP-50, offering a unified format which combines:
+- Binding chain specificity (via explicit chain identifiers) to the raw address.
+- Compact & canonical binary format for use on cross-chain message passing and intent declaration.
+- Checksums for name collision mitigation & user error prevention.
+
+Furthermore, we have taken this as an opportunity to add a few features that will make addresses interoperable with infrastructure at the edges of and beyond the ethereum ecosystem, by supporting the representation of addresses of non-evm chains as well, for the benefit of cross-chain liquidity movements and cross-chain message passing.
+
+## Concepts
+Interoperable Address
+: A binary payload which unambiguously identifies a target address and allows conversion to a human-readable name.
+
+Human-readable name
+: A string representation of an interoperable address, meant to be used by humans.
+
+Target address
+: The (address, chain ID) pair a particular Interoperable Address points to.
+
+## Guarantees provided by the standard
+- Any future Interoperable Address will be trivially convertible to Interoperable Address v1 (specified below), which, together with the point below, makes them a canonical representation for users who need to treat them opaquely.
+- Two Interoperable Addresses, converted to this canonical version, will be bitwise-equal iff they represent the same _target address_.
+- Checksums are short enough to be compared by humans, and allow for easy differentiation of _target addresses_ independently of any extra data the Interoperable Address may contain.
+
+## Specification
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
+
+### Interoperable Address binary format definition
+In binary format addresses have the following encoding:
+
+```
+┌─────────┬────────────┬─────────┬─────────┬─────────┐
+│ Version │ Chainidlen │ Chainid │ Addrlen │ Address │
+└─────────┴────────────┴─────────┴─────────┴─────────┘
+```
+
+Where:
+
+Version
+: a 2-byte version identifier. In the current standard this must always be the big-endian unsigned int 1. Other versions should be defined in future ERCs.
+
+Chainidlen
+: a 1-byte integer encoding the length of Chainid in bytes.
+
+Chainid
+: Variable length, binary representation of CAIP-2 chain namespace & reference serialized as explained in [Appendix A](#appendix-a-binary-encoding-of-caip-2-blockchain-id) encoding the chain ID.
+
+Addrlen
+: 1-byte integer encoding the length of Address in bytes.
+
+Address
+: Variable length field containing the binary format of the address component. For EVM addresses this is always 20 bytes, for serialization details of other types of addresses see [Appendix B](#appendix-b-binary-encoding-of-addresses)
+
+#### Restrictions for all Interoperable Address versions
+- The Interoperable Address MUST include an address and specify the chain it belongs to, fully defining a target address. Any version adding an indirection layer between payload data and these two fields would violate this restriction.
+    - Furthermore, future versions MUST be serialized to v1 and use the same input data to the checksum algorithm as described below, so an Interoperable Address' checksum only represents its target address.
+- Interoperable Address versions MAY only be able to represent a subset of the CAIP-2 namespaces.
+- Interoperable Address versions MAY assign extra syntactic restrictions on the human-readable name, coupled to novel semantic meaning (e.g. use of ENS or other naming registries)
+- Interoperable Address versions MAY define extra fields for purposes such as storing information on how to display the addresses.
+
+### Human-readable name format definition
+This section is considered OPTIONAL.
+
+#### Syntax
+```bnf
+<human readable name> ::= <account>@<chain>#<checksum>
+<account>             ::= [-:_%a-zA-Z0-9]*
+<chain>               ::= [-:_a-zA-Z0-9]*
+<checksum>            ::= [0-9A-F]{8}
+```
+
+Where:
+
+Chain
+: String representation of CAIP-2 blockchain identifier, recovered from the binary representation described in [Appendix A](#appendix-a-binary-encoding-of-caip-2-blockchain-id)
+
+Address
+: Chain namespace specific text representation of the address from the binary representation. Mapping between the two described in [Appendix B](#appendix-b-binary-encoding-of-addresses)
+
+Checksum
+: 4-byte checksum calculated by computing the keccak256 hash of the concatenated `Chainidlen`, `Chainid`, `Addrlen` and `Address` fields of the binary representation (that is, the v1 binary representation skipping the `Version` field), and truncating all but the first 4 bytes of the output. Represented as a base16 string as defined in RFC-4648
+
+#### Rationale
+- Chain and account fields' syntax is deliberately chosen to be able to express CAIP-2 namespaces (by using the `@` symbol for the separator, freeing up `:`) and CAIP-10 account IDs, with the caveat that no length restriction is placed, so chains with longer address formats or full 256-bit EVM chainids can be represented.
+- Similarly, the account field includes `%` as a valid character to allow for url-encoding of any other characters.
+- While a fresh start on text-encoding schemes could e.g. make use of more characters with base58 or base64 to achieve greater information efficiency and/or abstract which chain the address exists on as an implementation detail, we chose a more familiar approach since:
+    - It makes the Interoperable Address <-> raw address relationship evident to the user, which will make the cases where the former are not (yet) supported by a dapp or hardware wallet easier to reason with, as users will evidently see the address is the same, but it's the chain&checksum that is being trimmed.
+    - Its extra human-readability will make it more useful as a a stopgap solution until a standard focused on chain and address names is finalized.
+
+## Appendix A: Binary encoding of CAIP-2 blockchain ID
+The first two bytes are the binary representation of CAIP-2 namespace (see table below), while the remaining bytes correspond to the CAIP-2 reference, whose encoding is namespace-specific and defined below.
+
+### CAIP-2 namespaces' binary representation table
+
+| Namespace | binary key |
+| ---       | ---        |
+| eip155    | `0x0000`   |
+| solana    | `0x0002`   |
+
+#### eip155
+The bare `chainid` encoded as a big-endian unsigned integer of the minimum necessary amount of bytes will be used [^1], and leading zeroes will be prohibited. 
+
+[^1]: This makes it possible to represent some chains using the full word as their chainid, which CAIP-2 does not support since the set of values representable with 32 `a-zA-Z0-9` characters has less than `type(uint256).max` elements. This is done in an effort to support chains whose ID is the output of `keccak256`, as proposed in ERC-7785.
+
+##### Examples
+Ethereum Mainnet: `0x000001` (1, encoded as uint8)
+
+Optimism Mainnet: `0x00000A` (10, encoded as uint8)
+
+Ethereum Sepolia: `0x0000aa36a7` (11155111, encoded as uint24)
+
+#### solana
+Solana networks are distinguished by its genesis blockhash, which is to be decoded from base58btc and stored raw in the binary representation and without removing any leading zeroes:
+
+In the human-readable name, it should be displayed in full and base58btc-encoded, as returned by the node.
+
+Note: CAIP-2 limits chain references to 32 characters and instructs to only use the first 32 characters of the base58btc-encoded genesis blockhash, therefore converting to it will require truncating the reference, so converting _from_ actual CAIP-2 to this standard is potentially ambiguous
+
+##### Examples
+Solana Mainnet
+: `0x000245296998a6f8e2a784db5d9f95e18fc23f70441a1039446801089879b08c7ef0`
+: `solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdpKuc147dw2N9d`
+
+## Appendix B: Binary encoding of addresses
+
+### eip155
+20 bytes of evm addresses trivially stored as the payload. For text representation, the EIP-55 format MUST be used.
+
+### solana
+base-58 encoded public keys should be decoded and stored as a 32 byte payload
+
+#### Examples
+`7S3P4HxJpyyigGzodYwHtCxZyUQe9JiBMHyRWXArAaKv` -> `0x5F90554BB3D8C2FC82B6EE59C49AAA143E77F7D49A83E956CE1DBEF17A43F805`
+
+`DYw8jCTfwHNRJhhmFcbXvVDTqWMEVFBX6ZKUmG5CNSKK` -> `0xBA7A74F374AB05B70D114A78112EF0D3F0695A819572C79710B5372000D81AE2`
+
+## Test cases
+
+### Example 1: Ethereum mainnet
+Chain: Ethereum Mainnet
+Address: `0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045`
+
+Human-readable representation: `0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045@eip155:1#82255c07`
+
+Binary representation:
+```
+0x00010300000114D8DA6BF26964AF9D7EED9E03E53415D37AA96045
+  ^^^^-------------------------------------------------- Version:    decimal 1
+      ^^------------------------------------------------ Chainidlen: decimal 3
+        ^^^^^^------------------------------------------ Chainid:    2 bytes of CAIP-2 namespace + 1 byte to store uint8(1)
+              ^^---------------------------------------- Addrlen:    decimal 20
+                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Address:    20 bytes of ethereum address
+```
+keccak256 input for checksum: `0x0300000114d8da6bf26964af9d7eed9e03e53415d37aa96045`
+note the version field is removed before hashing
+
+### Example 2: Solana mainnet
+Chain: Solana Mainnet
+Address: `MJKqp326RZCHnAAbew9MDdui3iCKWco7fsK9sVuZTX2`
+
+Human-readable representation: `MJKqp326RZCHnAAbew9MDdui3iCKWco7fsK9sVuZTX2@solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdpKuc147dw2N9d#aa7e9354`
+
+Binary representation:
+
+```
+0x000122000245296998a6f8e2a784db5d9f95e18fc23f70441a1039446801089879b08c7ef0205333498d5aea4ae009585c43f7b8c30df8e70187d4a713d134f977fc8dfe0b5
+  ^^^^--------------------------------------------------------------------------------------------------------------------------------------- Version:     decimal 1
+      ^^------------------------------------------------------------------------------------------------------------------------------------- Chainidlen:  decimal 34
+        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^----------------------------------------------------------------- Chainid:     2 bytes of CAIP-2 namespace + 32 bytes of solana genesis block
+                                                                            ^^--------------------------------------------------------------- Addrlen:     decimal 32
+                                                                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Address:     32 bytes of solana address
+```
+keccak256 input for checksum: `0x22000245296998a6f8e2a784db5d9f95e18fc23f70441a1039446801089879b08c7ef02005333498d5aea4ae009585c43f7b8c30df8e70187d4a713d134f977fc8dfe0b5`
+note the version field is removed before hashing
+
+## Rationale
+
+### Comparisons with other standards
+
+#### CAIP-10
+CAIP-10 proposes a standard text format to represent what in this document is defined as target addresses. 
+Since it's a text-only standard, it does not concern itself with the serialization/deserialization of the various chains' address formats, simply using the text representation of addresses already defined by the respective chains. This means it is trivial to add support for chains to CAIP-10, while doing so with this standard involves defining a serialization scheme.
+
+The above has the drawback of not being able to guarantee canonicity. That is, there could be multiple valid CAIP-10 addresses pointing to the same target address.
+
+Also, the text format is of little use for smart contracts involved with cross-chain liquidity transfers and message passing due to its reduced data efficiency and lack of canonicity.
+
+Human-readable names are convertible to CAIP-10 without even going through the binary representation, so backwards-compatibility with actors expecting CAIP-10 identifiers should not be an issue.
+
+#### ERC-7828
+ERC-7828 is a text-only representation of addresses with optional resolution into more human-readable strings by use of existing ENS infrastructure for address names, and both the centralized ethereum-lists/chains list and ERC-7785 for resolution of chain names, while this aims to be a simple but extendable binary-first format.
+
+We believe the interop roadmap is better served by having a standardized binary format for addresses first, which allows the message passing and intents verticals to move forward on a common interface, with a good-enough text representation which is familiar to users and useful for developers, and as a next step develop a chain & address name resolving standard on top of it, leveraging its uniformity and extensibility.
+
+## Copyright
+Copyright and related rights waived via [CC0](../LICENSE.md).


ERCS/erc-X.md
```diff
@@ -0,0 +1,220 @@
...
+Optimism Mainnet: `0x00000A` (10, encoded as uint8)
+
+Ethereum Sepolia: `0x0000aa36a7` (11155111, encoded as uint24)
+
+#### solana
+Solana networks are distinguished by its genesis blockhash, which is to be decoded from base58btc and stored raw in the binary representation and without removing any leading zeroes:
+
+In the human-readable name, it should be displayed in full and base58btc-encoded, as returned by the node.
+
+Note: CAIP-2 limits chain references to 32 characters and instructs to only use the first 32 characters of the base58btc-encoded genesis blockhash, therefore converting to it will require truncating the reference, so converting _from_ actual CAIP-2 to this standard is potentially ambiguous
```
frangio, line 141: I've already commented on this in https://github.com/ethereum/L2-interop/pull/30#discussion_r2014406077 but I'm raising it again because we still haven't resolved what to do about it.

I think "ambiguity" is an understatement. If I have a CAIP-2 id I can't convert it to a chain id of this kind without additional information.

ERCS/erc-X.md
```diff
@@ -0,0 +1,220 @@
...
+- Interoperable Address versions MAY assign extra syntactic restrictions on the human-readable name, coupled to novel semantic meaning (e.g. use of ENS or other naming registries)
+- Interoperable Address versions MAY define extra fields for purposes such as storing information on how to display the addresses.
+
+### Human-readable name format definition
+This section is considered OPTIONAL.
+
+#### Syntax
+```bnf
+<human readable name> ::= <account>@<chain>#<checksum>
+<account>             ::= [-:_%a-zA-Z0-9]*
```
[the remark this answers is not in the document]
0xteddybear: Reasonable. I wanted to use 'account' so the 'address' concept remains coupled to Interoperable Address, but 'account' is never really defined in this document so I'll stick to 'address' for consisntency I think

ERCS/erc-X.md
0xteddybear: thanks for bringing this up again. I'll define anoter CAIP _profile_ which will take care of: 
- defining how to serialize addresses
- choosing the two-byte identifier for the namespace
- defining how to serialize the chain reference

I imagine the mapping from short/text to long/binary ids can be defined as part of the last item? that'd mean moving choosing between potential hash collisions to the social layer.
Alternatively we could just not solve it, I don't think it'd be too problematic since users are very unlikely to type out caip-2 identifiers directly, and I find it hard to imagine a context where you start from the CAIP-2 string, have to produce an Interoperable Address, and can't afford a dictionary to map between the two.

What other solutions do you have in mind?

0xteddybear: update: given that this has to be defined for every chain namespace, there's the obvious solution of specifying how to truncate the blockhashes in this document :grinning: 
Since solana does not actively mine collisions on the first bytes as bitcoin does, and AFAIK does not have competing forks, the first few bytes are likely to remain meaningful. wdyt? Am I missing something?

ERCS/erc-X.md
```diff
@@ -88,21 +93,21 @@ This section is considered OPTIONAL.
...
+<address>             ::= [.-:_%a-zA-Z0-9]*
+<chain>               ::= [.-:_a-zA-Z0-9]*
 <checksum>            ::= [0-9A-F]{8}
 ```
 
 Where:
 
 Chain
-: String representation of CAIP-2 blockchain identifier, recovered from the binary representation described in [Appendix A](#appendix-a-binary-encoding-of-caip-2-blockchain-id)
+: String representation of CAIP namespace, recovered from the binary representation described in [Appendix A](#appendix-a-binary-encoding-of-caip-2-blockchain-id). In the case where `ChainReferenceLength` is zero, it should be the empty string.
```
frangio, line 104:
```diff
-CAIP-2 blockchain identifier
+CAIP namespace
```

Is this right? Namespace sounds wrong to me. A namespace would be for example `eip155`. I do think we want the CAIP-2 blockchain identifier, e.g. `eip155:1`.

ERCS/erc-X.md
```diff
@@ -112,47 +117,101 @@ Checksum
...
 
-Optimism Mainnet: `0x00000A` (10, encoded as uint8)
+Optimism Mainnet: `0x0A` (10, encoded as uint8)
 
-Ethereum Sepolia: `0x0000aa36a7` (11155111, encoded as uint24)
+Ethereum Sepolia: `0xaa36a7` (11155111, encoded as uint24)
 
 #### solana
-Solana networks are distinguished by its genesis blockhash, which is to be decoded from base58btc and stored raw in the binary representation and without removing any leading zeroes:
+Solana networks are distinguished by its genesis blockhash, which is normally represented as 44 base58btc characters, corresponding to 32 bytes. We chose to perform the same truncation as CAIP-2[^2] to keep representations trivially convertible between the two, making them also visually identical.
```
frangio, line 156: I think we should be consistent in the decision to truncate ids following CAIP-2 or not...

I'm not fond of truncation, so unless we know of scenarios where it will be necessary to convert CAIP-2 ids to interoperable addresses I think I would avoid truncation altogether.

ERCS/erc-X.md
```diff
@@ -112,47 +117,101 @@ Checksum
...
+Note that Solana Mainnet's blockhash is:
+- base58btc `5eykt4UsFv8P8NJdTREpY1vzqKqZKvdpKuc147dw2N9d`
+- base16 `45296998a6f8e2a784db5d9f95e18fc23f70441a1039446801089879b08c7ef0`
 
 ## Appendix B: Binary encoding of addresses
 
 ### eip155
-Bytes of evm addresses are trivially stored as the payload. For text representation, the EIP-55 format MUST be used.
+Bytes of EVM addresses are trivially stored as the payload. 
 It's worth noting that addresses are currently 20 bytes, but that might change in the future, most likely to 32 bytes, if the EVM Object Format is ever adopted.
```
frangio, line 188:
```suggestion
It's worth noting that addresses are currently 20 bytes, but that might change in the future, most likely to 32 bytes, if Address Space Expansion is ever implemented. If that happens, expanded addresses may be represented in 32 bytes, but pre-expansion addresses must remain 20 bytes in order to preserve a consistent address.
```

frangio: For context: EVM Object Format doesn't increase addresses to 32 bytes, it only lays the groundwork to make that possible in the future.

0xteddybear:
hello @SamWilsn , I have some editorial questions: 

- What should I do about the extra sections? the EIP Walidator [reports on them](https://github.com/ethereum/ERCs/actions/runs/14517629842/job/40730439882?pr=1002) but I believe they make a more complete document. Should I ignore the warning? Fit the information somewhere else, such as existing sections (not sure which would make sense) or an auxiliary file?

- Is the fact that the bot does not recognize reference-style links (`[EIP-55]` inline, and `[EIP-55]: url` at the bottom of the document) intended?

- And also, does it make sense to define a standard & allowed way to link to CAIPs similar to what is defined in EIP-1 for BIPs?

ERCS/erc-X.md
0xteddybear: Thanks for coming back to this! 
Clarified this as a footnote in f1c77147 , lmk what you think

ERCS/erc-X.md
0xteddybear: you're right, good catch! I've rewritten this part to display the chain type (which actually is the chain namespace :upside_down_face: ) in the case where the chain reference is not actually defined, which came up as an inconsistency when prototyping the wallet

ERCS/erc-X.md
```diff
@@ -0,0 +1,220 @@
...
+: a 1-byte integer encoding the length of Chainid in bytes.
+
+Chainid
+: Variable length, binary representation of CAIP-2 chain namespace & reference serialized as explained in [Appendix A](#appendix-a-binary-encoding-of-caip-2-blockchain-id) encoding the chain ID.
+
+Addrlen
+: 1-byte integer encoding the length of Address in bytes.
+
+Address
+: Variable length field containing the binary format of the address component. For EVM addresses this is always 20 bytes, for serialization details of other types of addresses see [Appendix B](#appendix-b-binary-encoding-of-addresses)
```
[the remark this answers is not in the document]
0xteddybear: 6925eb37

SamWilsn:
> What should I do about the extra sections?

Cram them where they are most appropriate. Many proposals introduce concepts in the first subsection of Specification. The appendix should probably go there as well. Comparisons to other proposals belong in Motivation. Guarantees can either go in Rationale or Specification depending on how it's written (eg. if the guarantee explains why you made a choice, it's Rationale—on my phone at the moment or I'd give you more specific advice). Test Cases just needs proper capitalization.

> Is the fact that the bot does not recognize reference-style links [...]

You've found an interesting bug. The missing reference is actually on the preceding line (in the heading). Reference-style links should definitely work.

> allowed way to link to CAIPs

We have one in fact:

https://github.com/ethereum/EIPs/blob/c4dda7f016ce2ee676d6a7b93545d5d843bb834c/config/eipw.toml#L155

I've just consistently forgotten to update EIP-1 for... a while.

ERCS/erc-7930.md
```diff
@@ -0,0 +1,223 @@
...
+Interoperable Addresses build on insights from [ERC-7828](./erc-7828.md), CAIP-10 and CAIP-50, offering a unified format which combines:
+- Binding chain specificity (via explicit chain identifiers) to the raw address.
+- Compact & canonical binary format for use on cross-chain message passing and intent declaration.
+- Checksums for name collision mitigation & user error prevention.
+
+Furthermore, we have taken this as an opportunity to add a few features that will make addresses interoperable with infrastructure at the edges of and beyond the ethereum ecosystem, by supporting the representation of addresses of non-evm chains as well, for the benefit of cross-chain liquidity movements and cross-chain message passing.
+
+### Guarantees this standard aims to provide
+- Any future Interoperable Address will be trivially convertible to Interoperable Address v1 (specified below), which, together with the point below, makes them a canonical representation for users who need to treat them opaquely.
+- Two Interoperable Addresses, converted to this canonical version, will be bitwise-equal if and only if they represent the same _target address_.
```
u59149403, line 36: As well as I understand, this is not true as of now. caip-19 ( https://chainagnostic.org/CAIPs/caip-19 ) mentions chain `cosmos:Binance-Chain-Tigris` in test cases. As well as I understand, this is Binance chain. But, as well as I understand, it can be also encoded as `eip155:56`. So, caip-2 chain ids are not unique. (Note: I'm not expert on Binance chain, so I can be wrong. It is possible this is simply error in caip-19 test cases.)

Also caip-288 ( https://chainagnostic.org/CAIPs/caip-288 ) "Best Practices for using CAIP 2" says: "As mentioned in [CAIPs#22](https://github.com/ChainAgnostic/CAIPs/issues/22), a given network is assumed to be unique within a namespace, but not unique across all namespaces".

So, even CAIPs themselves say that CAIP-2 ids are not globally unique

0xteddybear: 1- thanks, good catch!
2- _oh no_

will come back to this, would appreciate input from @frangio in the meantime, since he sort of represents the users needing canonicity :grin:

u59149403: > BSC Smart Chain

This is [RAS syndrome](https://en.wikipedia.org/wiki/RAS_syndrome). But okay, I understand you

ERCS/erc-7930.md
```diff
@@ -0,0 +1,223 @@
...
+
+## Abstract
+Interoperable Addresses is a binary format to describe an address specific to one chain with an optional human-readable name representation.
+
+It is designed in a manner that allows the community extend it to support naming schemes (or other not-yet-imagined features) in the future, as well as to support arbitrary-length addresses and chainids.
+
+## Motivation
+The Ethereum ecosystem is rapidly expanding into a multi-chain environment encompassing L1s, L2s, sidechains, and rollups—some EVM‐compatible, others not. A simple address alone is not sufficient to describe a blockchain account, since it does not identify which chain the address belongs to, creating ambiguity for wallets, dApps, and users. At the same time, it is important for the format to be extensible to support both current standards for address naming and future standards for chain naming.
+
+The current standard ethereum address representation defined in [EIP-55], with _optional_ retrofitted checksums implemented via selective capitalization of `a-f` characters, poses the following challenges:
```
oxlumi, line 21: EIP-55 should be linked to [https://eips.ethereum.org/EIPS/eip-55](https://eips.ethereum.org/EIPS/eip-55
)

ERCS/erc-7930.md
```diff
@@ -0,0 +1,223 @@
...
+- The existing address format provides no mechanism for extension, meaning that there is no natural method to add these features to the existing address format.
+
+Interoperable Addresses build on insights from [ERC-7828](./erc-7828.md), CAIP-10 and CAIP-50, offering a unified format which combines:
+- Binding chain specificity (via explicit chain identifiers) to the raw address.
+- Compact & canonical binary format for use on cross-chain message passing and intent declaration.
+- Checksums for name collision mitigation & user error prevention.
+
+Furthermore, we have taken this as an opportunity to add a few features that will make addresses interoperable with infrastructure at the edges of and beyond the ethereum ecosystem, by supporting the representation of addresses of non-evm chains as well, for the benefit of cross-chain liquidity movements and cross-chain message passing.
```
oxlumi, line 32: To maintain an impersonal tone, it would be better to rephrase this sentence as: _“Furthermore, this proposal additionally adds features to make addresses interoperable with infrastructure…”_

ERCS/erc-7930.md
```diff
@@ -0,0 +1,223 @@
...
+
+Checksum
+: 4-byte checksum calculated by computing the keccak256 hash of the concatenated `ChainType`, `ChainReferenceLength`, `ChainReference`, `AddressLength` and `Address` fields of the binary representation (that is, the v1 binary representation skipping the `Version` field), and truncating all but the first 4 bytes of the output. Represented as a base16 string as defined in RFC-4648.
+
+## Rationale
+- Chain and address fields' syntax is deliberately chosen to be able to express CAIP-2 namespaces (by using the `@` symbol for the separator, freeing up `:`) and CAIP-10 account IDs, with the caveat that no length restriction is placed, so chains with longer address formats or full 256-bit EVM chainids can be represented.
+- Similarly, the address field includes `%` as a valid character to allow for url-encoding of any other characters.
+- While a fresh start on text-encoding schemes could e.g. make use of more characters with base58 or base64 to achieve greater information efficiency and/or abstract which chain the address exists on as an implementation detail, we chose a more familiar approach since:
```
oxlumi, line 115: Here as well, it would improve clarity and maintain an impersonal tone to rephrase _“we chose a more familiar approach”_ as _“a familiar approach was selected because…”._

ERCS/erc-7930.md
```diff
@@ -0,0 +1,223 @@
...
+
+Also, the text format is of little use for smart contracts involved with cross-chain liquidity transfers and message passing due to its reduced data efficiency and lack of canonicity.
+
+Human-readable names are convertible to CAIP-10 without even going through the binary representation, so backwards-compatibility with actors expecting CAIP-10 identifiers should not be an issue.
+
+#### [ERC-7828]
+[ERC-7828] is a text-only representation of addresses with optional resolution into more human-readable strings by use of existing ENS infrastructure for address names, and both the centralized ethereum-lists/chains list and ERC-7785 for resolution of chain names, while this aims to be a simple but extendable binary-first format.
+
+We believe the interop roadmap is better served by having a standardized binary format for addresses first, which allows the message passing and intents verticals to move forward on a common interface, with a good-enough text representation which is familiar to users and useful for developers, and as a next step develop a chain & address name resolving standard on top of it, leveraging its uniformity and extensibility.
```
oxlumi, line 135: Rephrase here as well: instead of _“We believe the interop roadmap is better served…”_ consider _“It is expected that the interop roadmap is better served…”_ or simply _“The interop roadmap is better served…”._

ERCS/erc-7930.md
```diff
@@ -0,0 +1,223 @@
...
+  ^^^^------------------------------------------------------------------------ Version:              decimal 1
+      ^^^^-------------------------------------------------------------------- ChainType:            2 bytes of CAIP namespace
+          ^^------------------------------------------------------------------ ChainReferenceLength: decimal 32
+            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^-- ChainReference:       32 bytes of solana genesis block
+                                                                            ^^ AddressLength:        zero, indicating no address
+```
+keccak256 input for checksum: `0x00022045296998a6f8e2a784db5d9f95e18fc23f70441a1039446801089879b08c7ef000`.
+Note the version field is removed before hashing.
+
+## Security Considerations
```
oxlumi, line 214: It might be worth discussing some risks here. For example, could checksums fail or be improperly validated? Could misinterpreted chain identifiers lead to funds loss? Are there upgrade or compatibility risks introduced by the versioning scheme?

oxlumi (commented):
I’ve added some comments around:
– Rephrase subjective phrases like “we believe,” “we chose.”
– Add a bit more detail in Security Considerations.
– Small clarity and consistency tweaks.

ERCS/erc-7930.md
0xteddybear: this is linked :sparkles:  it's a reference-style link, with the destination on the bottom

ERCS/erc-7930.md
0xteddybear: > could checksums fail or be improperly validated?

I refrained from writing UX guidelines in this document given that it's mostly concerned with binary addresses, but imagine wallets not giving enough prominence to checksums so they are missed by users could be a possible failure mode.
I didn't add them on ERC-7828 either, since:
- Wallet's UX developers will probably be better at choosing how to display information than what I can come up with (@skeletor-spaceman came up with the example of using blockies instead of text checksums, for example)
- There's not an adversarial relationship between a given wallet's UX designer and anybody else which could be mitigated by a standard's normative.

Collisions on 4-byte checksums are feasible to mine, but it's unlikely the impersonator's address (or in the case of ERC-7828, name) is also a lookalike of the intended recipient's

...well, that hinges on the wallet displaying the name also having sane warnings around unicode glyph collisions, in the case if 7828. Will add it as a security consideration of that standard.

I also considered the possibility of hash collisions happening because of there not being a delimiter between address and chainid, but that seems harder to exploit than the fact that checksums are only 4`bytes long

> Could misinterpreted chain identifiers lead to funds loss?

Most definetely, but for non-malicious usecases (such as an attacker mining an address so that `(minedAddres, badChain)` has a 4-byte collision with `(whateverAddress, intendedChain)`) the checksum should be a good mitigation.

> Are there upgrade or compatibility risks introduced by the versioning scheme?

I tried to lay them out in the 'Restrictions for all Interoperable Address versions' section, basically we define version bit to mean the Interoperable Address is backwards-compatible with v1, and parseable with the same algorithm.

The main risk would be for a future Interoperable Address to define an associated Interoperable Name which can cause collisions with previous versions, e.g. having a v3 name be equivalent to a v2 name while representing a different (address, chain) pair.

In the past we explored including the address' version in the Interoperable Name, but that felt like a lazy (albeit robust) solution, which unnecessarily leaked implementation details onto users. The way we actually intend to avoid such a scenario is:
- Review of other ERCs defining further Interoperable Address versions
- Implementing ERC-7828 in a flexible enough way (by coupling the resolution method to a TLD) that lowers the need to define new versions in the near future.

wdyt?

btw, I'm posting this in the forum since this discussion should outlive this PR. Feel free to continue the discussion there.

0xteddybear: > ...well, that hinges on the wallet displaying the name also having sane warnings around unicode glyph collisions, in the case if 7828. Will add it as a security consideration of that standard.

pushed!

ERCS/erc-7930.md
oxlumi: Solved in [this commit](https://github.com/ethereum/ERCs/pull/1002/commits/cfad98a9cb459b9b20ed9cba7003947ba7151311).

0xteddybear:
hi @SamWilsn ! Implemented all feedback that I could from eip walidator, but there are a few things I couldn't get around:

- [ ] I got myself into a sort of circular dependency :grinning: since 7930 and 7828 reference each other. I believe the most reasonable thing to do would be to merge this PR with the warnings and then be strict about fixing them all on 7828's PR. The other alternatives I could think of are significantly more cumbersome (but I'm of course open to input):
    - [ ] reword this ERC to talk about 7828 as a 'possible future standard', ignoring 7828 already has a number assigned and is also close to reaching 'draft' status -> makes ERCs less navigable
    - [ ] move both ERCs to the same github PR and merge them atomically -> would mean losing track of lots of feedback since there's lot of it that happened in PRs instead of eth magicians' forum :sob:
- [ ] I'm in the process of getting CAIP-350 as an official draft, but in this PR I referenced the commit of the latest version so the link resolves as soon as the PR for CAIP-350 is merged
- [ ] I don't seem to have been able to make eipw understand the reference-style link in line 134

What would you advise to do regaring these three items above?

Is there anything else you see as preventing merging?

0xteddybear:
update: since it's not required to link to CAIPs, I removed links to CAIP-350, in order to minimize the amount of warnings. I'll link to CAIP-350 once we have a definitive commit hash of it

SamWilsn:
On my phone at the moment, so forgive my brevity.

Putting both in the same PR, but generally speaking we prefer an acyclic graph of dependencies. If you feel strongly about it, I'll make an exception. All technical discussion should take place in Ethereum Magicians. There will be a number of pull requests regardless of whether both proposals are in the same initial PR.

SamWilsn:
There seems to be a bug in `eipw`. Try making "ERC-7785" a link.

0xteddybear:
> Putting both in the same PR, but generally speaking we prefer an acyclic graph of dependencies. If you feel strongly about it, I'll make an exception.

I do feel strongly about this since, in order to have all technical discussion happen on ethereum magicians it'd be best to have a page on eips.ethereum.org as the main means to share the standard instead of having an open PR where people can comment. I believe we can get this standard to an official 'Draft' state very quicky but ERC-7828 is a few days away from that.

> There seems to be a bug in eipw. Try making "ERC-7785" a link.

Solved, thanks!

ERCS/erc-7930.md
```diff
@@ -0,0 +1,245 @@
...
+          ^^------------------------------------------------------------------ ChainReferenceLength: decimal 32
+            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^-- ChainReference:       32 bytes of solana genesis block
+                                                                            ^^ AddressLength:        zero, indicating no address
+```
+keccak256 input for checksum: `0x00022045296998a6f8e2a784db5d9f95e18fc23f70441a1039446801089879b08c7ef000`.
+Note the version field is removed before hashing.
+
+
+### Example 5: Arbitrum One address
+Chain: Ethereum Mainnet
```
0xmoebius, line 216:
```suggestion
### Example 5: Arbitrum One address
Chain: Arbitrum One
```

ERCS/erc-7930.md
```diff
@@ -0,0 +1,245 @@
...
+
+### Example 5: Arbitrum One address
+Chain: Arbitrum One
+Address: `0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045`
+
+Interoperable Name: `0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045@eip155:42161#38DDF570`
+
+Interoperable Address:
+```
+0x0001000001A4B114D8DA6BF26964AF9D7EED9E03E53415D37AA96045
```
Amxx, line 223:
```suggestion
0x0001000002A4B114D8DA6BF26964AF9D7EED9E03E53415D37AA96045
```

0xteddybear: good find, thanks!

ERCS/erc-7930.md
```diff
@@ -0,0 +1,245 @@
...
+
+It is therefore advised for implementers requiring canonicity of addresses (e.g by using them as keys in smart contract mappings or other key-value stores), to thoroughly review the the CAIP-350 profile of a chain namespace for the possibility of a lack of canonicity of addresses (which should be noted in the profile's 'Extra Considerations' section) as well as collisions with other already-supported namespaces.
+
+## Copyright
+Copyright and related rights waived via [CC0](../LICENSE.md).
+
+[EIP-55]: ./erc-55.md
+[ERC-7828]: ./erc-7828.md
+[ERC-7785]: ./erc-7785.md
+<!-- TODO: link to CAIP-350-->
```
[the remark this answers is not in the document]
0xtiti: Fixed [401b533](https://github.com/ethereum/ERCs/pull/1002/commits/401b533c7f4e85fdf56082bf729e6d5de34f2b46)

ERCS/erc-7930.md
```diff
@@ -0,0 +1,248 @@
...
+
+## Abstract
+Interoperable Addresses is a binary format to describe an address specific to one chain with a companion 'Interoperable Names' specification for human-readable display.
+
+It is designed in a manner that allows the community to extend it to support naming schemes (or other not-yet-imagined features) in the future, as well as to support arbitrary-length addresses and chainids.
+
+## Motivation
+The Ethereum ecosystem is rapidly expanding into a multi-chain environment encompassing L1s, L2s, sidechains, and rollups—some EVM‐compatible, others not. A simple address alone is not sufficient to describe a blockchain account, since it does not identify which chain the address belongs to, creating ambiguity for wallets, dApps, and users. At the same time, it is important for the format to be extensible to support both current standards for address naming and future standards for chain naming.
+
+The current standard ethereum address representation defined in [ERC-55], with _optional_ retrofitted checksums implemented via selective capitalization of `a-f` characters, poses the following challenges:
```
wei3erHase, line 21: [ERC-55] link is broken

ERCS/erc-7930.md
```diff
@@ -0,0 +1,245 @@
...
+                                                                            ^^---------------------------------------------------------------- AddressLength:        decimal 32
+                                                                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^--- Address:              32 bytes of solana address
+```
+keccak256 input for checksum: `0x00022045296998a6f8e2a784db5d9f95e18fc23f70441a1039446801089879b08c7ef02005333498d5aea4ae009585c43f7b8c30df8e70187d4a713d134f977fc8dfe0b5`.
+Note the version field is removed before hashing.
+
+### Example 3: EVM address without chainid
+Chain: `eip155` namespace, chainid not specified.
+Address: `0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045`
+Interoperable Name: `0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045@eip155:#144A4B21`
```
Amxx, line 181: could we do without the colon ? What would be the issue if we did not use it ?

My understanding is by parsing the `@` and the `#` we would get the 3 blocks (address, chain, checksum). The chain would be `eip155`. From here, the absence of column makes it clear that there is only a namespace without reference.

Absence of colon is only an issue if the system wants to support both "namespace without reference" and "reference without namespace" ... but I don't this the second case makes any sens, does it?

ERCS/erc-7930.md
```diff
@@ -0,0 +1,248 @@
...
+```
+keccak256 input for checksum: `0x00022045296998a6f8e2a784db5d9f95e18fc23f70441a1039446801089879b08c7ef000`.
+Note the version field is removed before hashing.
+
+
+### Example 5: Arbitrum One address
+Chain: Arbitrum One
+Address: `0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045`
+
+Interoperable Name: `0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045@eip155:42161#38DDF570`
```
Amxx, line 219: incorrect hash

```suggestion
Interoperable Name: `0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045@eip155:42161#D2E02854`
```

ERCS/erc-7930.md
```diff
@@ -0,0 +1,248 @@
...
+```
+keccak256 input for checksum: `0x000100000014D8DA6BF26964AF9D7EED9E03E53415D37AA96045`
+Note the version field is removed before hashing.
+
+### Example 4: Solana mainnet network, no address
+Chain: Solana Mainnet.
+
+Address: Not specified.
+
+Interoperable Name: `@solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdpKuc147dw2N9d#2eb18670`
```
Amxx, line 200: specs says checksum should use upercase

```suggestion
Interoperable Name: `@solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdpKuc147dw2N9d#2EB18670`
```

ERCS/erc-7930.md
```diff
@@ -0,0 +1,248 @@
...
+```
+keccak256 input for checksum: `0x0000010114D8DA6BF26964AF9D7EED9E03E53415D37AA96045`
+note the version field is removed before hashing
+
+### Example 2: Solana mainnet address
+Chain: Solana Mainnet
+
+Address: `MJKqp326RZCHnAAbew9MDdui3iCKWco7fsK9sVuZTX2`
+
+Interoperable Name: `MJKqp326RZCHnAAbew9MDdui3iCKWco7fsK9sVuZTX2@solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdpKuc147dw2N9d#88835c11`
```
Amxx, line 163: checksum should be upercase

```suggestion
Interoperable Name: `MJKqp326RZCHnAAbew9MDdui3iCKWco7fsK9sVuZTX2@solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdpKuc147dw2N9d#88835C11`
```

ERCS/erc-7930.md
0xteddybear: I believe it's correct since it's a [reference-style link](https://github.github.com/gfm/#reference-link) whose reference is defined in the bottom of the document. It points to `eip-55.md` instead of `erc-55.md` due to changes performed in later stages of the eip website's generation.  Or are you referring to something else?

Errors in link definitions would be caught by the [htmlproofer ci step](https://github.com/ethereum/ERCs/actions/runs/15288589020/job/43003777808?pr=1002)

ERCS/erc-7930.md
0xteddybear: thanks! fixed in f664c938897e63e9a3354d37190d3633b49f3548

ERCS/erc-7930.md
0xteddybear: thanks! fixed in f664c938897e63e9a3354d37190d3633b49f3548

ERCS/erc-7930.md
0xteddybear: thanks! fixed in f664c938897e63e9a3354d37190d3633b49f3548

ERCS/erc-7930.md
```diff
@@ -0,0 +1,223 @@
...
+Human-readable names are convertible to CAIP-10 without even going through the binary representation, so backwards-compatibility with actors expecting CAIP-10 identifiers should not be an issue.
+
+#### [ERC-7828]
+[ERC-7828] is a text-only representation of addresses with optional resolution into more human-readable strings by use of existing ENS infrastructure for address names, and both the centralized ethereum-lists/chains list and ERC-7785 for resolution of chain names, while this aims to be a simple but extendable binary-first format.
+
+We believe the interop roadmap is better served by having a standardized binary format for addresses first, which allows the message passing and intents verticals to move forward on a common interface, with a good-enough text representation which is familiar to users and useful for developers, and as a next step develop a chain & address name resolving standard on top of it, leveraging its uniformity and extensibility.
+
+## Test Cases
+
+### Example 1: Ethereum mainnet address
```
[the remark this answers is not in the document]
0xteddybear: did so with an Arbitrum One example below

ERCS/erc-7930.md
0xteddybear: you're correct, there's no need to include the colon in cases where there's no chain reference. Updated the spec & this example at 59ace6bd76d0ad2954beac901b44a60160872c6a

ERCS/erc-7930.md
ernestognw: imo this section is a bit convoluted, I would just keep it to:

```suggestion
Current Ethereum addresses ([ERC-55]) lack chain specificity and have optional checksums, creating several challenges:
```

ERCS/erc-7930.md
```diff
@@ -0,0 +1,248 @@
...
+
+## Motivation
+The Ethereum ecosystem is rapidly expanding into a multi-chain environment encompassing L1s, L2s, sidechains, and rollups—some EVM‐compatible, others not. A simple address alone is not sufficient to describe a blockchain account, since it does not identify which chain the address belongs to, creating ambiguity for wallets, dApps, and users. At the same time, it is important for the format to be extensible to support both current standards for address naming and future standards for chain naming.
+
+The current standard ethereum address representation defined in [ERC-55], with _optional_ retrofitted checksums implemented via selective capitalization of `a-f` characters, poses the following challenges:
+
+- Verification of the checksum by a client is optional, meaning that mistyped or erroneous addresses may be accepted by these clients. There is no way for a user to distinguish a correct address from an incorrect one processed by a client that does not do validation.
```
ernestognw: This bullet point could be simplified for better readability:

```suggestion
- Optional checksum validation allows mistyped addresses to be accepted without user awareness.
```

The current version covers the same concept but is a bit lengthy. This shorter version captures both the acceptance issue and the user awareness problem in one clear statement.

ERCS/erc-7930.md
```diff
@@ -0,0 +1,248 @@
...
+
+### Concepts
+Interoperable Address
+: A binary payload which unambiguously identifies a target address and allows conversion to a human-readable name.
+
+Interoperable Name
+: A string representation of an interoperable address, meant to be used by humans.
+
+Target address
+: The (address, chain ID) pair a particular Interoperable Address points to.
```
ernestognw, line 63: The definitions have a circular reference issue where "Interoperable Address" references "target address" before it's defined. Consider this revision to maintain the same order while fixing the circularity:

```suggestion
Interoperable Address
: Binary encoding of a (chain, address) pair with version and checksum.

Interoperable Name
: Human-readable text representation of an interoperable address.

Target Address
: A (chain, address) pair identifying a specific account on a specific blockchain.
```

This way each definition builds on concepts already established rather than forward-referencing.

ERCS/erc-7930.md
```diff
@@ -0,0 +1,248 @@
...
+
+### Interoperable Address binary format definition
+In binary format addresses have the following encoding:
```
ernestognw: The standard references V1 throughout but never formally introduces it. This section should clearly establish the V1 specification and use formal language:

```suggestion
### Interoperable Address V1 binary format definition
In binary format, addresses MUST have the following encoding:
```

Adding "V1" to the title provides the formal introduction, and "MUST" establishes the normative requirement.

ERCS/erc-7930.md
```diff
@@ -0,0 +1,248 @@
...
+```
+┌─────────┬───────────┬──────────────────────┬────────────────┬───────────────┬─────────┐
+│ Version │ ChainType │ ChainReferenceLength │ ChainReference │ AddressLength │ Address │
+└─────────┴───────────┴──────────────────────┴────────────────┴───────────────┴─────────┘
+```
+
+Where:
+
+Version
+: A 2-byte version identifier. In the current standard this must always be the big-endian unsigned int 1. Other versions should be defined in future ERCs.
```
ernestognw: The phrasing could be more direct and concise:

```suggestion
: A 2-byte version identifier. For version 1, this MUST be `0x0001` (big-endian). Future versions SHOULD be standardized in separate ERCs.
```

This removes "In the current standard" redundancy and makes the future versioning guidance clearer.

ERCS/erc-7930.md
```diff
@@ -0,0 +1,248 @@
...
+
+Address
+: Variable length field containing the binary format of the address component. For serialization details of different types of addresses see the CAIP-350 profile for the chain type.
+
+#### Restrictions for all Interoperable Address versions
+- Future Interoperable Address versions whose `ChainType`, `ChainReference` and `Address` can't be extracted in the same way as a `v1` address MUST set its most-significant version bit to 1, to allow for parser forward-compatibility.
+- The Interoperable Address MUST define the mechanism to both include an address and specify the chain it belongs to directly in the payload, fully defining a target address. Any version adding an indirection layer between payload data and these two fields would violate this restriction.
+    - While it is possible that a given instance of an Interoperable Address to not include an address or a chain reference, every version must allow for both to be specified at the same time.
+    - Furthermore, future versions MUST be serialized to v1 and use the same input data to the checksum algorithm as described below, so an Interoperable Address' checksum only represents its target address.
```
ernestognw: This section would benefit from a brief introduction explaining its purpose, plus clearer formatting of the rules:

```suggestion
#### Version Compatibility Rules
These rules ensure that future versions of Interoperable Addresses maintain backwards compatibility and consistent behavior across implementations:

- Future versions incompatible with v1 parsing MUST set the most significant bit of the version field to 1
- All versions MUST support both address and chain specification simultaneously  
- All versions MUST serialize to v1 format for checksum calculation, excluding the version field from the hash input
```

The intro paragraph clarifies why these rules exist, and the bullet points are more consistently formatted with clearer language.

ERCS/erc-7930.md
```diff
@@ -0,0 +1,248 @@
...
+
+Also, the text format is of little use for smart contracts involved with cross-chain liquidity transfers and message passing due to its reduced data efficiency and lack of canonicity.
+
+Interoperable Names are convertible to [CAIP-10] without even going through the binary representation, so backwards-compatibility with actors expecting [CAIP-10] identifiers should not be an issue.
+
+#### [ERC-7828]
+[ERC-7828] is a text-only representation of addresses with optional resolution into more human-readable strings by use of existing ENS infrastructure for address names, and both the centralized ethereum-lists/chains list and [ERC-7785] for resolution of chain names, while this aims to be a simple but extendable binary-first format.
+
+The interop roadmap is better served by having a standardized binary format for addresses first, which allows the message passing and intents verticals to move forward on a common interface, with a good-enough text representation which is familiar to users and useful for developers, and as a next step develop a chain & address name resolving standard on top of it, leveraging its uniformity and extensibility.
```
ernestognw: This condensed section is much better than the previous lengthy comparisons, but could use slight expansion for clarity:

```suggestion
### Relationship to Existing Standards
ERC-7930 provides binary encoding for the CAIP-10 semantic model with guaranteed canonicity, addressing CAIP-10's limitation of potential multiple representations for the same target address. The Interoperable Names format maintains backwards compatibility with CAIP-10 text format for seamless integration with existing tooling.
```

This adds just enough context about why binary encoding matters (canonicity) and what backwards compatibility means (existing tooling) without returning to the verbose comparisons. Also, I think this section should be moved to the "Backwards Compatibility" section regardless

ERCS/erc-7930.md
```diff
@@ -0,0 +1,248 @@
...
+
+## Test Cases
```
ernestognw, line 134: I think the examples 3 and 5 don't demonstrate new serialization concepts beyond what 1 and 2 already cover.

- **Example 3** (chain-only): While the spec allows zero-length addresses, this edge case adds complexity without showing practical usage. The parsing mechanics are identical to Example 2, just with `AddressLength = 0`.

- **Example 5** (Arbitrum): This is essentially a variation of Example 1 (EVM chain) with different chain ID encoding length. It doesn't demonstrate new serialization patterns.

I would consider either removing them or repurposing them

ERCS/erc-7930.md
```diff
@@ -0,0 +1,248 @@
...
+                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Address:              20 bytes of ethereum address
+```
+keccak256 input for checksum: `0x000002A4B114D8DA6BF26964AF9D7EED9E03E53415D37AA96045`
+note the version field is removed before hashing
+
+## Security Considerations
+While this standard aims to be a foundation to be able to canonically refer to addresses on different chains, that guarantee is going to be a leaky abstraction in the real world, given that e.g. a particular chain namespace might define a serialization scheme that can't guarantee canonicity of addresses, or a given network might have two valid [CAIP-2] ids referring to it.
+
+It is therefore advised for implementers requiring canonicity of addresses (e.g by using them as keys in smart contract mappings or other key-value stores), to thoroughly review the the CAIP-350 profile of a chain namespace for the possibility of a lack of canonicity of addresses (which should be noted in the profile's 'Extra Considerations' section) as well as collisions with other already-supported namespaces.
+
```
ernestognw, line 215: I think the security considerations should be divided in sections so that we can accomodate the following.

```suggestion
### 4-byte checksum collisions
Feasible to mine with current hardware. Implementers should display full addresses alongside checksums.

### Version collision risk
Future versions may create name collisions. See versioning restrictions for mitigation strategies.
```

Purposely recommending small sections but feel free to expand.

ERCS/erc-7930.md
```diff
@@ -0,0 +1,248 @@
...
+status: Draft
+type: Standards Track
+category: ERC
+created: 2025-02-02
+---
+
+## Abstract
+Interoperable Addresses is a binary format to describe an address specific to one chain with a companion 'Interoperable Names' specification for human-readable display.
+
+It is designed in a manner that allows the community to extend it to support naming schemes (or other not-yet-imagined features) in the future, as well as to support arbitrary-length addresses and chainids.
```
SamWilsn: Your abstract is missing a technical overview of _how_ it accomplishes its goals. So far it does a good job of explaining _what_ those goals are.

ERCS/erc-7930.md
```diff
@@ -0,0 +1,248 @@
...
+Where:
+
+Chain
+: String representation of [CAIP-2] blockchain identifier, recovered from the binary representation described in the CAIP-350 profile for the chain type. In the case where `ChainReferenceLength` is zero, it should be the CAIP namespace alone, and no trailing colon. Note that it's not possible to specify a reference without a namespace.
+
+Address
+: Chain namespace specific text representation of the address from the binary representation. Mapping between the two described in the CAIP-350 profile for the chain type. In the case where `AddressLength` is zero, it should be the empty string.
+
+Checksum
+: 4-byte checksum calculated by computing the keccak256 hash of the concatenated `ChainType`, `ChainReferenceLength`, `ChainReference`, `AddressLength` and `Address` fields of the binary representation (that is, the v1 binary representation skipping the `Version` field), and truncating all but the first 4 bytes of the output. Represented as a base16 string as defined in RFC-4648.
```
SamWilsn: Should link to RFC 4648 (details in EIP-1)

ERCS/erc-7930.md
```diff
@@ -0,0 +1,248 @@
...
+                                                                            ^^---------------------------------------------------------------- AddressLength:        decimal 32
+                                                                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^--- Address:              32 bytes of solana address
+```
+keccak256 input for checksum: `0x00022045296998a6f8e2a784db5d9f95e18fc23f70441a1039446801089879b08c7ef02005333498d5aea4ae009585c43f7b8c30df8e70187d4a713d134f977fc8dfe0b5`.
+Note the version field is removed before hashing.
+
+### Example 3: EVM address without chainid
+Chain: `eip155` namespace, chainid not specified.
+Address: `0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045`
+Interoperable Name: `0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045@eip155#144A4B21`
```
Amxx, line 181: incorrect hash
```suggestion
Interoperable Name: `0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045@eip155#B26DB7CB`
```

ERCS/erc-7930.md
```diff
@@ -0,0 +1,248 @@
...
+Interoperable Address:
+```
+0x000100000014D8DA6BF26964AF9D7EED9E03E53415D37AA96045
+  ^^^^------------------------------------------------ Version:              decimal 1
+      ^^^^-------------------------------------------- ChainType:            2 bytes of CAIP namespace
+          ^^------------------------------------------ ChainReferenceLength: zero, indicating no chainid
+            ^^---------------------------------------- AddressLength:        decimal 20
+              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Address:              20 bytes of ethereum address
+```
+keccak256 input for checksum: `0x000100000014D8DA6BF26964AF9D7EED9E03E53415D37AA96045`
```
Amxx: version was not removed 

```suggestion
keccak256 input for checksum: `0x00000014D8DA6BF26964AF9D7EED9E03E53415D37AA96045`
```

ERCS/erc-7930.md
Amxx: fix hash (do not include version in its computation
```suggestion
Interoperable Name: `0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045@eip155#B26DB7CB`
```

0xtiti: fixed [e48cc67](https://github.com/ethereum/ERCs/pull/1002/commits/e48cc67859ae755deae1cc39b442c9401cc17a18)

ERCS/erc-7930.md
0xtiti: fixed [e48cc67](https://github.com/ethereum/ERCs/pull/1002/commits/e48cc67859ae755deae1cc39b442c9401cc17a18)

frangio:
@SamWilsn Following up on https://github.com/ethereum/ERCs/pull/1002#pullrequestreview-2871834594. I believe links are fixed now with the exception of ERC-7828 which was [previously](https://github.com/ethereum/ERCs/pull/1002#issuecomment-2878098440) [agreed](https://github.com/ethereum/ERCs/pull/1002#issuecomment-2881658681) to allow as an exception.

We have two PRs that are blocked by getting this ERC merged (https://github.com/ethereum/ERCs/pull/1050, https://github.com/ethereum/ERCs/pull/981).

ernestognw:
> We have two PRs that are blocked by getting this ERC merged (https://github.com/ethereum/ERCs/pull/1050, https://github.com/ethereum/ERCs/pull/981).

I'm personally waiting to this PR to also update #1072 to require interoperable addresses 😄

ERCS/erc-7930.md
```diff
@@ -0,0 +1,224 @@
...
+└─────────┴───────────┴──────────────────────┴────────────────┴───────────────┴─────────┘
+```
+
+Where:
+
+Version
+: A 2-byte version identifier. For version 1, this must be `0x0001` (big-endian). Future versions SHOULD be standardized in separate ERCs.
+
+ChainType
+: A 2-byte value as defined in CAIP-350, corresponding to a [CAIP-2] _namespace_, which allows users to know how to interpret & display the other two fields.
```
bumblefudge:
```suggestion
: A 2-byte value as defined in CAIP-350, corresponding to a [CAIP-2] _namespace_, which allows users to know how to interpret & display the other two fields. For all EVM chains, the `chainType` is `0x0000`.
```
It seems like you can save all a substantial portion of your readers a click and a pageload by telling them they can just hardcode [0x0000](https://github.com/ChainAgnostic/namespaces/pull/140/files#diff-1defbab2825adf2d63ad93b31fc7710cd437a250045ae0bcb20f180bc3d6f634R4) 😄

bumblefudge:
@0xteddybear hey sorry was i blocking this getting merged? the spec outlining how to profile ERC-7930 for non-EVM chains is now here:
https://github.com/ChainAgnostic/CAIPs/blob/29762ef99a6ffea1e07e3f796c0d1a5a95e89b88/CAIPs/caip-350.md
and the profile for evm is here:
https://github.com/ChainAgnostic/namespaces/blob/1160b41fbd5dd5b1a062b759054932b523ca185f/eip155/caip350.md
and the template for future profiles is here:
https://github.com/ChainAgnostic/namespaces/blob/1160b41fbd5dd5b1a062b759054932b523ca185f/_template/caip350.md
note that these kinds of commit-specific github links are the ONLY format allowed, don't link to the [mutable] URLs of rendered jekyll (if you need/want to add links at all).

ERCS/erc-7930.md
```diff
@@ -0,0 +1,224 @@
...
+- While a fresh start on text-encoding schemes could e.g. make use of more characters with base58 or base64 to achieve greater information efficiency and/or abstract which chain the address exists on as an implementation detail, a more familiar approach was chosen since:
+    - It makes the Interoperable Address <-> raw address relationship evident to the user, which will make the cases where the former are not (yet) supported by a dapp or hardware wallet easier to reason with, as users will evidently see the address is the same, but it's the chain&checksum that is being trimmed.
+    - Its extra human-readability will make it more useful as a a stopgap solution until a standard focused on chain and address names is finalized.
+- We chose to support zero-length addresses and chainids to be able to use just one standard to represent both potentially foreign _target addresses_ (e.g. the recipient of a cross-chain message) and also plain addresses or chainids (e.g. to specify the origin-chain hook address of a cross-chain message), for greater uniformity across implementations.
+
+### Comparisons with other standards
```
SamWilsn: This subsection belongs under Motivation, since it's justifying the proposal as a whole.

ERCS/erc-7930.md
euler0x: solved bbb18bca68f6f113718f78fb670d7df49ccc2efb

ERCS/erc-7930.md
euler0x: solved 7f2246241299c687196cba5d8af7d5ff0b1b50c6

ERCS/erc-7930.md
euler0x: solved 13e2441be07693b31648689b9a8653a6cf5b2e12

ERCS/erc-7930.md
```diff
@@ -0,0 +1,226 @@
...
+Where:
+
+Chain
+: String representation of [CAIP-2] blockchain identifier, recovered from the binary representation described in the [CAIP-350] profile for the chain type. In the case where `ChainReferenceLength` is zero, it should be the CAIP namespace alone, and no trailing colon. Note that it's not possible to specify a reference without a namespace.
+
+Address
+: Chain namespace specific text representation of the address from the binary representation. Mapping between the two described in the [CAIP-350] profile for the chain type. In the case where `AddressLength` is zero, it should be the empty string.
+
+Checksum
+: 4-byte checksum calculated by computing the keccak256 hash of the concatenated `ChainType`, `ChainReferenceLength`, `ChainReference`, `AddressLength` and `Address` fields of the binary representation (that is, the v1 binary representation skipping the `Version` field), and truncating all but the first 4 bytes of the output. Represented as a base16 string as defined in [RFC 4648].
```
SamWilsn, line 124: Should the "should"s here be uppercased?

SamWilsn (approved):

skeletor-spaceman:
## Reference Implementation
- [Interop-SDK by Wonderland](https://github.com/defi-wonderland/interop-sdk/tree/dev/packages/addresses): A TypeScript/JavaScript library for encoding, decoding, and manipulating Interoperable Addresses.
- [OpenZeppelin draft implementation](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/draft-InteroperableAddress.sol) Solidity library contract for Interoperable Addresses.

PR #1005, opened 2025-04-07
Update ERC-7631: Move to Last Call

Vectorized:
The implementations in the wild do not have any request for changes for more than a year, so moving to last call.


xinbenlv (approved):

xinbenlv:
Don't forget to update deadline

PR #1013, opened 2025-04-16
Update ERC-3009: Replace EOA v,r,s signature with generic bytes signature

ilikesymmetry:
ERC-3009's external functions support limited signature types by enforcing a `v,r,s` interface. To support a more generic interface, parity with [ERC-1271](https://eips.ethereum.org/EIPS/eip-1271) is a simple change that expands potential users to include smart contract accounts. Note that USDC, the leading example of ERC-3009, actually implements copies of `transferWithAuthorization`, `receiveWithAuthorization`, and `cancelAuthorization` with the proposed interface here for the exact same reason.

This PR's goal is to formalize this approach to simplify integration of smart contract accounts and more.


atimark (approved):

atimark (approved):

SamWilsn:
I'm not sure if the authors of ERC-3009 are interested in this pull request, but you're always welcome to make your own proposal (with a new number) and advocate for your version.

PR #1035, opened 2025-05-10
Add ERC: Encode chain id with transaction hash

microbecode:
Suggestion for a new standard.

This standard proposes a way to encode the combination of a chain ID and a transaction hash into one string.


microbecode:
The filename seems to be failing validation, but I fail to understand why. As far as I can tell, the only requirements are prefix "erc-" and suffix ".md" and those requirements are fulfilled.

PR #1039, opened 2025-05-14
Update ERC-7677: "Update ERC-7677: chain id update"

jxom:
Reverts ethereum/ERCs#571 as chain ids are no longer per-call.


lukasrosario (approved):
ty

PR #1043, opened 2025-05-14
Add ERC: Unidirectional Wallet Uplink aka UWULink

moodysalem:
@tinaszheng


ERCS/eip-uwulink.md
```diff
@@ -0,0 +1,540 @@
+---
```
lightclient, line 1:
```suggestion
---
eip: 7946
```

ERCS/erc-7946.md
```diff
@@ -0,0 +1,541 @@
...
+Protocol Buffers suitable for low-bandwidth channels such as QR codes or NFC tags. Two modes of operation are
+supported: **Static Mode**, where the request payload directly contains the list of calls, and **Programmable Mode**,
+where the payload references a contract that generates the list of calls.
+
+## Motivation
+
+dApps often require users to perform multistep interactions, such as approving a token and then executing a
+swap. Traditionally, accomplishing this has required multiple user confirmations or complex wallet connectivity.
+Recent standards like EIP-5792 and EIP-7702 introduced ways to batch multiple calls into one atomic operation via
+JSON-RPC (e.g. `wallet_sendCalls`). However, those solutions assume an active connection
```
wilsoncusack, line 28: sendCalls does not require a `from` address and no prior connection is required

moodysalem: To what interface would you send the `wallet_sendCalls` RPC? In practice you connect a wallet first, check capabilities, and use `wallet_sendCalls` only if it's available. It seems like connecting a wallet is a prerequisite to know where to send the RPC.

This would be very different from the dApp having all actions available in the unconnected state, and when you trigger an action (e.g. a swap) it could pop up a modal with the QR code to perform the action, and/or a link with the `uwulink` scheme, a la magnet link, that triggers the transaction for any extension that registered itself to handle the scheme

SamWilsn (commented):
I'm reasonably happy with this proposal's formatting so far. The only outstanding issues is the protocol buffers reference. You can't exactly implement this proposal without understanding protobufs, but referring to its standard is likely to be contentious among editors.

moodysalem:
> I'm reasonably happy with this proposal's formatting so far. The only outstanding issues is the protocol buffers reference. You can't exactly implement this proposal without understanding protobufs, but referring to its standard is likely to be contentious among editors.

Thank you for fixing the errors! I was a bit stumped on that.

What makes protobuf different than JSON + a JSON schema that it would need to be referenced?

SamWilsn:
JSON is defined in [RFC 8259](https://www.rfc-editor.org/rfc/rfc8259) and JSON Schema isn't strictly required to build a compatible implementation. I do see your point though.

I'll see if we can allow Protobufs.

SamWilsn (approved):

ERCS/eip-uwulink.md
natalejoe: 64

PR #1038, opened 2025-05-12
Update ERC-7786: Remove Attributes, Add ExtraData; Support Hooks and Interoperable Addresses

Joxess:
This PR is based on the previous discussion of the Ethereum Interop working group. The major changes include support for Interoperable Addresses, removing attributes, adding extraData, supporting hooks, and other changes in semantics.

PR #1050, opened 2025-05-22
Update ERC-7786: Use interoperable binary addresses, improve attributes

frangio:
This PR includes the change to interoperable binary addresses from https://github.com/ethereum/ERCs/pull/1038, but preserves attributes with a small change, and does not add hooks.

To make clearer the fact that attributes are metadata strictly intended for the gateway, the attributes argument was removed from `receiveMessage` and metadata will not be forwarded to the receiver.

The rationale for preserving a structured form of metadata rather than opaque `extraData` is:
- The use of opaque `extraData` would rely entirely on off-chain consensus and coordination on encoding and semantics between the sender and the gateway. This prevents on-chain constructions that are able to detect errors and adapt to gateway functionality.
  - Tagging or versioning metadata facilitates on-chain coordination.
  - This is the `bytes4` selector that is prepended to the actual metadata values. The gateway can use this to safely detect that it supports the metadata that was passed in, which could otherwise successfully decode but be misinterpreted.
  - A getter on the gateway to check for support allows senders to adapt their behavior to what the gateway supports. This is `supportsMetadata(bytes4)`. For an example, consider a sender contract that supports multiple ways of specifying gas payment for the call and can choose which one to use.
- The possibility of providing multiple pieces of metadata (`bytes[]`) as opposed to a single piece of metadata (`bytes extraData`) allows multiple extensions to be combined without disrupting the other metadata and requiring less coordination.

The rationale for not adding hooks is:
- In order for hooks to be useful we need to design and specify the mechanism by which the gateway and a hook communicate. While there were advances in [#1038](https://github.com/ethereum/ERCs/pull/1038) with https://github.com/ethereum/ERCs/pull/1038/commits/62ab9a38ffc288575f7bae9b6a1b62127c3f1db2, I don't think the work is done, for example although there is a `getTransientMessageContext()` getter in the gateway, as currently specified the gateway address has to be provided by the sender in the hook data and the hook must trust it.
- Structured metadata easily subsumes hooks: we can define an extension where metadata of the form `hook(address,bytes,uint256)` triggers a hook call.

### DIFF
--- a/ERCS/erc-7786.md
+++ b/ERCS/erc-7786.md
@@ -13,7 +13,7 @@ requires: 7930
 
 ## Abstract
 
-This proposal describes an interface, and the corresponding workflow, for smart contracts to send arbitrary data through cross-chain messaging protocols. The end goal of this proposal is to have all such messaging protocols accessible via this interface (natively or using "adapters") to improve their composability and interoperability. That would allow a new class of cross-chain native smart contracts to emerge while reducing vendor lock-in. This proposal is modular by design, allowing users to leverage bridge-specific features through structured metadata while providing simple "universal" access to the simple feature of "just getting a simple message through".
+This proposal describes an interface, and the corresponding workflow, for smart contracts to send arbitrary data through cross-chain messaging protocols. The end goal of this proposal is to have all such messaging protocols accessible via this interface (natively or using "adapters") to improve their composability and interoperability. That would allow a new class of cross-chain native smart contracts to emerge while reducing vendor lock-in. This proposal is modular by design, allowing users to leverage bridge-specific features through attributes (structured metadata) while providing simple "universal" access to the simple feature of "just getting a simple message through".
 
 ## Motivation
 
@@ -31,7 +31,7 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 
 ### Message Field Encoding
 
-A cross-chain message consists of a sender, recipient, payload, value (native token), and list of metadata.
+A cross-chain message consists of a sender, recipient, payload, value (native token), and list of attributes.
 
 #### Sender and Recipient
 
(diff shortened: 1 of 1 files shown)


ERCS/erc-7786.md
```diff
@@ -30,56 +31,60 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
 ### Sending Procedure
 
-An **Source Gateway** is a contract that offers a protocol to send a message to a receiver on another chain. It MUST implement `IERC7786GatewaySource`.
+An **Source Gateway** is a contract that offers a protocol to send a message to a recipient on another chain. It MUST implement `IERC7786GatewaySource`.
 
 ```solidity
 interface IERC7786GatewaySource {
-    event MessagePosted(bytes32 indexed outboxId, string sender, string receiver, bytes payload, uint256 value, bytes[] attributes);
+    event MessageSent(
+        bytes32 indexed sendId,
```
xinbenlv, line 65: Consider adding `indexed` to both `sender` and `recipient` parameters for better event filtering capabilities. This would allow efficient filtering by sender/recipient addresses in cross-chain message tracking. 

What's the rationale for not indexing these fields? Gas cost considerations or expectation that most queries will filter by `sendId` only?

xinbenlv (commented):
Thanks for the improvements to ERC-7786! I have one specific question about event indexing.

ERCS/erc-7786.md
frangio: `bytes` values cannot be indexed, or rather, when indexed only their hash is emitted.

u59149403:
@frangio , CAIP-350/ERC‑7930 is too immature for building other standards upon it. Currently it doesn't specify binary encoding for any namespace, even for EVM-compatible chains.

So, to build other standards (such as ERC-7786) on top of CAIP-350/ERC‑7930 the following conditions should be met:
- CAIP-350 and ERC‑7930 both should be merged
- Table of namespaces ( https://namespaces.chainagnostic.org ) should get own column for CAIP-350
- That column should get implementations for many namespaces, including EVM-compatible chains, i. e. EIP-155 . I. e. there should be "X" (pointing to real document) on intersection of EIP-155 row and CAIP-350 column
- And ideally, CAIP-350 should be implemented for all namespaces, for which CAIP-10 is implemented (i. e. CAIP-350 column should have at least as many Xs as CAIP-10 has)

Otherwise there is no sense in migrating from CAIP-10 to less mature standard (CAIP-350/ERC‑7930)

frangio:
> Otherwise there is no sense in migrating from CAIP-10 to less mature standard

I disagree on this part. Parsing CAIP-10 ASCII strings on chain is so bad that it makes sense to migrate to a less mature standard.

frangio:
Blocked on https://github.com/ethereum/ERCs/pull/1002.

ERCS/erc-7786.md
```diff
@@ -89,93 +94,90 @@ Initiates the sending of a message.
...
-interface IERC7786Receiver {
-    function executeMessage(
-        string calldata messageId, // gateway specific, empty or unique
-        string calldata sourceChain, // [CAIP-2] chain identifier
-        string calldata sender, // [CAIP-10] account address
+interface IERC7786Recipient {
+    function receiveMessage(
+        bytes32 receiveId,     // Unique identifier
+        bytes calldata sender, // Binary Interoperable Address
         bytes calldata payload,
```
0xValera, line 134: The comma here is redundant.

PR #1052, opened 2025-05-23
Add ERC: Permissionless CREATE2 Factory

rmeissner:


777668z0 (approved):

ERCS/erc-permissionless_create2_factory.md
```diff
@@ -0,0 +1,323 @@
+---
```
SamWilsn:
```suggestion
---
eip: 7955
```

Don't forget to rename the file as well.

ERCS/erc-7955.md
```diff
@@ -0,0 +1,324 @@
+---
+eip: 7955
+title: Permissionless CREATE2 Factory
+description: A permissionless method for the cross-chain deployment of a universal CREATE2 factory.
```
SamWilsn: Your description is mostly just restating your title in more words. I think the only new information introduced here is that it's "cross-chain". You should use your description to elaborate on the ideas introduced in your title, not just restating them.

ERCS/erc-7955.md
```diff
@@ -0,0 +1,324 @@
...
+
+However, there is a bootstrapping problem: how do you get a CREATE2 factory contract with a specific address and code?
+
+### Existing Solutions
+
+There are currently three main approaches to this problem:
+
+#### 1. Nick's Method
+
+Use Nick's method to randomly generate a signature for a transaction **without** [EIP-155](./eip-155.md) replay protection that deploys the CREATE2 factory. Nick's method ensures that there is no known private key for an account that deploys the CREATE2 factory, meaning that the resulting contract will have a deterministic address and code on all chains. This strategy is used by the _Deterministic Deployment Proxy_ (`Arachnid/deterministic-deployment-proxy`), one of the most widely used CREATE2 factory contracts.
```
SamWilsn: > _Deterministic Deployment Proxy_ (`Arachnid/deterministic-deployment-proxy`)

This is an external link and must be removed.

ERCS/erc-7955.md
```diff
@@ -0,0 +1,324 @@
...
+
+**Downsides**:
+
+- It does not work on chains that only accept EIP-155 replay-protected transactions.
+- It is sensitive to changes in gas parameters on the target chain since the gas price and limit in the deployment transaction is sealed, and a new one cannot be signed without a private key.
+- Reverts due to alternative gas schedules make the CREATE2 factory no longer deployable.
```
SamWilsn: The tense here is off, but I'm not sure how you intend it to be read. Perhaps:

```suggestion
- Reverts due to alternative gas schedules making the CREATE2 factory no longer deployable.
```

ERCS/erc-7955.md
```diff
@@ -0,0 +1,324 @@
...
+
+#### 2. Secret Private Key
+
+Keep a carefully guarded secret key and use it to sign transactions to deploy CREATE2 factory contracts. The resulting contract will have a deterministic address and code on all chains where the first transaction of the deployer account is a CREATE2 factory deployment, which can be verified post-deployment to ensure trustlessness. Additionally, this method does not have the same gas sensitivity downsides as Nick's method, as the private key can sign a creation transaction with appropriate gas parameters at the time of execution. This is the strategy used by the _Safe Singleton Factory_ (`safe-global/safe-singleton-factory`) and _CreateX_ (`pcaversaccio/createx`).
```
SamWilsn: Similar external links. Please remove them.

ERCS/erc-7955.md
```diff
@@ -0,0 +1,324 @@
...
+bytes32 salt = CREATE2_FACTORY_SALT;
+assembly ("memory-safe") {
+    create2(0, add(initCode, 32), mload(initCode), salt)
+}
+```
+
+The _bootstrap contract_ MAY implement additional features such as:
+
+- Abort early if either _CREATE2 factory contract_ is already deployed or the _deployer_ is not correctly delegated to.
+  - This can help mitigate gas griefing from the previously described front-running issue.
```
SamWilsn: Where?

ERCS/erc-7955.md
```diff
@@ -0,0 +1,324 @@
...
+
+The use of a publicly known private key enables this mechanism, as anyone can permissionlessly generate a delegation signature to **any** bootstrap contract that would cause the `DEPLOYER_ADDRESS` to execute the specified `CREATE2 (0xf5)` operation and deploy the factory contract to a completely deterministic address. Because of the use of `CREATE2 (0xf5)`, the CREATE2 factory will be deployed to `CREATE2_FACTORY_ADDRESS` if and only if it is deployed with `CREATE2_FACTORY_INIT_CODE`, thus guaranteeing a deployed code hash of `CREATE2_FACTORY_CODE_HASH`. Additionally, the semantics of `CREATE2 (0xf5)` make it so no transaction executed by `DEPLOYER_ADDRESS` can permanently block the deployment of the CREATE2 factory contract.
+
+One issue with this method is that because the `DEPLOYER_PRIVATE_KEY` is public, anyone can sign alternative delegations or transactions and front-run a legitimate CREATE2 factory deployment. We consider this to not be a serious issue however, as:
+
+1. Doing so does not prevent future deployments - meaning that an attacker can only delay the deployment of the CREATE2 factory with a sustained attack at a gas cost to the attacker.
```
SamWilsn: I don't believe the account paying for the gas of a 7702 transaction has to be the same account as the one being delegated. You never have to put funds into `DEPLOYER_ADDRESS` to execute the code contained within it.

ERCS/erc-7955.md
nlordell: Do you mean the reference to the GitHub repository? This was intended to be rendered as:

```html
<i>Deterministic Deployment Proxy</i> (<code>Arachnid/deterministic-deployment-proxy</code>)
```

And _should_ just be plain text with no link. If this is also forbidden, I can remove it (and it also appears in a few other places when referencing `safe-global/safe-singleton-factory` for example).

ERCS/erc-7955.md
nlordell: I meant "reverts" as a noun (so "the reverts ... make"), maybe just some missing punctuation:

```suggestion
- Reverts, due to alternative gas schedules, make the CREATE2 factory no longer deployable.
```

Does that make it clearer?

ERCS/erc-7955.md
nlordell: > I don't believe the account paying for the gas of a 7702 transaction has to be the same account as the one being delegated.

Correct.

> You never have to put funds into DEPLOYER_ADDRESS to execute the code contained within it.

Also correct. In fact, you should never send funds to `DEPLOYER_ADDRESS` as anyone can take them.

---

The front-running issue stems from the fact that an attacker can submit an alternate EIP-7702 transaction (from another address they control) in order to increase the `DEPLOYER_ADDRESS`'s nonce, thereby invaliding the EIP-7702 delegation signature from `DEPLOYER_PRIVATE_KEY` in a legitimate deployment transaction that is included afterwards.

ERCS/erc-7955.md
nlordell: Oops, nice catch. The issue is described below and not above. Will fix.

nlordell:
First of all, thank you for the detailed review @SamWilsn :). I _think_ I addressed all of your comments, let me know if there is still something left.

TBH, writing a good description is hard :sweat_smile: (its like naming things in software development). Anyway, open to more feedback if you have any.

ERCS/erc-7955.md
```diff
@@ -0,0 +1,330 @@
...
+0x0002: RETURNDATASIZE  # Stack: [0; 32]
+
+# First, load the salt value and compute the actual code size for the CREATE2
+# call, this is the calldata length minus 32 for the salt prefix.
+                        # Stack: [0; 32]
+0x0003: RETURNDATASIZE  # Stack: [0; 0; 32]                     | Push the calldata offset 0 of the `salt` parameter
+0x0004: CALLDATALOAD    # Stack: [salt; 0; 32]                  | Load the `salt` from calldata
+0x0005: DUP3            # Stack: [32; salt; 0; 32]              | Push 32 to the stack
+0x0006: CALLDATASIZE    # Stack: [msg.data.len; 32; salt; ...]  | Followed by the calldata length
+0x0007: SUB             # Stack: [code.len; salt; 0; 32]        | Compute `msg.data.length - 32`, which is the length of
```
frangio, line 207: Is it acceptable for `msg.data.length - 32` to potentially overflow?

nlordell: It would cause the creation transaction to revert with "out of gas" (you would be copying ~2**256 bytes into memory), so it is acceptable IMO.

I saw your implementation in https://ethereum-magicians.org/t/eip-tbd-multi-chain-deterministic-deployment-factory/24998 and think that saturating subtraction is another way to do it. Although, it may cause "silent failures" where incorrectly encoded CREATE2 deployments lead to successful transactions deployments of contracts with empty code; so it violates the "law of least surprise" for me: I would be least surprised if incorrect encoding of parameters causes a revert. There is an argument to be made to make the reverts more explicit i.e. using the `REVERT` opcode and not rely on the fact that you will run out of gas when trying to copy 1000...000 bytes of calldata into memory, especially given the later will use up all transaction gas, and the former will not.

frangio: Yeah I'll give this a think, reverting if sub 32 is probably a better idea.

777668z0 (approved):

777668z0 (approved):

Cassieminkus1 (approved):

Cassieminkus1 (approved):

Cassieminkus1 (approved):

PR #1053, opened 2025-05-23
Update ERC-165: Update erc-165.md

dancer1325:
fix boolean value returned in second call, considered as failure

 - The build passes.

### DIFF
--- a/ERCS/erc-165.md
+++ b/ERCS/erc-165.md
@@ -88,7 +88,7 @@ Implementation note, there are several logical ways to implement this function. 
 1. The source contract makes a `STATICCALL` to the destination address with input data: `0x01ffc9a701ffc9a700000000000000000000000000000000000000000000000000000000` and gas 30,000. This corresponds to `contract.supportsInterface(0x01ffc9a7)`.
 2. If the call fails or return false, the destination contract does not implement ERC-165.
 3. If the call returns true, a second call is made with input data `0x01ffc9a7ffffffff00000000000000000000000000000000000000000000000000000000`.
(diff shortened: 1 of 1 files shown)


ERCS/erc-165.md
```diff
@@ -88,7 +88,7 @@ Implementation note, there are several logical ways to implement this function.
 1. The source contract makes a `STATICCALL` to the destination address with input data: `0x01ffc9a701ffc9a700000000000000000000000000000000000000000000000000000000` and gas 30,000. This corresponds to `contract.supportsInterface(0x01ffc9a7)`.
 2. If the call fails or return false, the destination contract does not implement ERC-165.
 3. If the call returns true, a second call is made with input data `0x01ffc9a7ffffffff00000000000000000000000000000000000000000000000000000000`.
-4. If the second call fails or returns true, the destination contract does not implement ERC-165.
+4. If the second call fails or returns false, the destination contract does not implement ERC-165.
```
SamWilsn, line 91: Why would any ERC-165 supporting contract return `true` for interface id `0xffffffff`? That doesn't make sense.

dancer1325: > Why would any ERC-165 supporting contract return `true` for interface id `0xffffffff`? That doesn't make sense.

hi @SamWilsn  but that's out of the scope of this fix. or do you want to handle ALL here?

SamWilsn: The only change in this proposal is modifying the behaviour when passing an interface id `0xffffffff`. How is my question out of scope?

hilljessica8787-dot (approved):

PR #1060, opened 2025-05-29
Update ERC-7813: Move to Review

vdrg:
This EIP is already being used by MUD projects in production, and we believe it is ready to be moved to Review.


ERCS/erc-7813.md
```diff
@@ -2,9 +2,9 @@
 eip: 7813
 title: Store, Table-Based Introspectable Storage
 description: On-chain tables for automatic indexing and introspectable state
```
SamWilsn, line 4: Your description is _probably_ good, but as someone who isn't intimately familiar with the topic, I'd certainly appreciate a bit more information about what this proposal is for. Like... what is a table in the context of blockchain?

PR #1067, opened 2025-06-05
Add Rationale and Backwards Compatibility to ERC-7803

ernestognw:

 - The build passes.

### DIFF
--- a/ERCS/erc-7803.md
+++ b/ERCS/erc-7803.md
@@ -2,7 +2,7 @@
 eip: 7803
 title: EIP-712 Extensions for Account Abstraction
 description: Improvements for EIP-712 to support smart contract accounts.
-author: Francisco Giordano (@frangio)
+author: Francisco Giordano (@frangio), Ernesto García (@ernestognw)
 discussions-to: https://ethereum-magicians.org/t/erc-7803-eip-712-extensions-for-account-abstraction/21436
 status: Draft
 type: Standards Track
@@ -21,14 +21,15 @@ This ERC improves on [EIP-712] signatures to better support smart contract accou
 
 ### Signing Domains
 
-Standards like [ERC-1271] and [ERC-6492] give smart contract accounts (SCAs) the ability to produce signatures that an application can authenticate without knowledge of the abstract rules of the account. This is an important primitive for applications, as the account owner is able to authorize a third-party to act on its behalf without interacting with the chain.
+Standards like [ERC-1271], [ERC-7913], and [ERC-6492] give smart contract accounts (SCAs) the ability to produce signatures that an application can authenticate without knowledge of the abstract rules of the account. This is an important primitive for applications, as the account owner is able to authorize a third-party to act on its behalf without interacting with the chain.
 
 [ERC-1271]: ./eip-1271.md
 [ERC-6492]: ./eip-6492.md
+[ERC-7913]: ./eip-7913.md
 
 Smart contract accounts may be "owned" by cryptographic keys whose signatures are used to authorize the use of the account. There is not necessarily a one-to-one mapping between keys and accounts, because a single key may control multiple accounts, so care must be taken to prevent replay attacks across them. This is done by binding a signature to a particular account.
 
-EIP-712 introduced a scheme where signatures can be bound to a verifying domain, which corresponds to the protocol contract that will authenticate a signature. Reusing this mechanism to additionally bind a signature to the domain of the smart contract account runs into a large amount of complexity and attack surface (see [ERC-7739]), as well as yet unresolved issues with account composability (SCAs that control other SCAs). This ERC introduces *signing domains* in addition to verifying domains to natively enable wallets to generate smart contract account signatures with replay protection.
+EIP-712 introduced a scheme where signatures can be bound to a verifying domain, which corresponds to the protocol contract that will authenticate a signature. Reusing this mechanism to additionally bind a signature to the domain of the smart contract account runs into a large amount of complexity and attack surface (see [ERC-7739]), as well as yet unresolved issues with account composability (SCAs that control other SCAs). This ERC introduces _signing domains_ in addition to verifying domains to natively enable wallets to generate smart contract account signatures with replay protection.
 
 [ERC-7739]: ./eip-7739.md
 
@@ -64,7 +65,7 @@ This property is an array of smart contract account domains. Each member of the 
 
 From right to left the array lists the chain of accounts through which the signer ultimately has control over the "outermost" signing domain (i.e., that listed first).
 
-For example: 
+For example:
 
 1. A dapp requests an EIP-712 signature to a connected account via JSON-RPC. `signingDomains` is empty or undefined.
 2. The connected account is a multisig, so it requests EIP-712 signatures from its signers, prepending the domain of the multisig to `signingDomains`, which is now an array of length 1.
@@ -88,8 +89,8 @@ This property is an array of supported signature authentication methods, listed 
 Each member of the array is an object with the following keys:
 
 - `id`: An string that identifies the method. It may be one of:
-    - `ECDSA`: ECDSA signatures by Externally Owned Accounts.
-    - `ERC-{n}`: A standard type of signature specified by an ERC. `n` must not be padded with zeros.
+  - `ECDSA`: ECDSA signatures by Externally Owned Accounts.
+  - `ERC-{n}`: A standard type of signature specified by an ERC. `n` must not be padded with zeros.
 - `parameters` (optional): An array of method-specific parameters.
 
 ### JSON Schema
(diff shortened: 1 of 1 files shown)


ERCS/erc-7803.md
```diff
@@ -88,8 +90,8 @@ This property is an array of supported signature authentication methods, listed
 Each member of the array is an object with the following keys:
 
 - `id`: An string that identifies the method. It may be one of:
-    - `ECDSA`: ECDSA signatures by Externally Owned Accounts.
-    - `ERC-{n}`: A standard type of signature specified by an ERC. `n` must not be padded with zeros.
+  - `ECDSA`: ECDSA signatures by Externally Owned Accounts.
+  - `ERC-{n}`: A standard type of signature specified by an ERC. `n` must not be padded with zeros.
```
ernestognw, line 106: I believe the `id` field should be reduced to only `ERC-{n}`, so that `ECDSA` is covered by ERC-7913

ERCS/erc-7803.md
```diff
@@ -81,72 +83,121 @@ In the presence of `signingDomains` the account should encode the message to be
...
+
+The recursive encoding scheme was designed to maintain the hierarchical nature of account relationships while supporting unlimited nesting depth. This approach provides better wallet UX by preserving the context of each signing domain in the chain, making it clearer to users what they are signing for and on behalf of whom.
+
+### Array Ordering
+
+The array ordering convention (outermost to innermost) was chosen to match the natural flow of authorization, where the outermost domain represents the final destination of the signature. This ordering enables wallets to display the signing context in a logical and intuitive way, helping users understand the full chain of authorization they are participating in.
+
+### Graceful Degradation
+
+The design ensures graceful degradation when only part of the stack supports this standard. Security never falls below current [EIP-712] levels. Enhanced features simply become unavailable rather than causing failures. Users can still interact with applications, albeit with reduced context and security. This approach enables incremental adoption while maintaining backward compatibility throughout the ecosystem.
```
frangio: I don't think this is true. What does the graceful degradation look like? If the wallet doesn't support or worse if it ignores `signingDomains` it will produce a different signature than what the app expects no?

ERCS/erc-7803.md
```diff
@@ -81,72 +83,121 @@ In the presence of `signingDomains` the account should encode the message to be
...
 
 ## Backwards Compatibility
 
-<!-- TODO -->
+This specification maintains backwards compatibility with existing [EIP-712] implementations across the Ethereum ecosystem, with wallets being able to identify signatures encoded through the `\x19\x02` prefix while standard [EIP-712] signatures using `\x19\x01` continue to function normally.
+
+Smart accounts currently using [ERC-7739] can seamlessly upgrade to this specification as it represents only a semantical change (rather than a breaking change), with the [`authMethods`](#authmethods) field appropriately covering the corresponding ERC's verification behavior. Existing contracts using [ERC-1271], [ERC-6492], or [ERC-7913] require no modifications, and all [EIP-712] libraries, development tools, and frameworks continue to work without changes to their current interfaces.
```
frangio: Not sure what you mean by "semantical change rather than a breaking change".

I'd just say that accounts using ERC-7739 can upgrade or add to support for this authentication method.

ERCS/erc-7803.md
```diff
@@ -81,72 +83,121 @@ In the presence of `signingDomains` the account should encode the message to be
 
 `signingDomainSeparators` is the array of hashes of the domains included in `signingDomains`, in the same order, computed according to EIP-712's `hashStruct`.
 
+### Signature Encoding for Signing Domains
```
frangio: Can you explain this section? I don't understand what it's saying or what it refers to.

ERCS/erc-7803.md
```diff
@@ -81,72 +83,121 @@ In the presence of `signingDomains` the account should encode the message to be
...
 - `id`: An string that identifies the method. It may be one of:
-    - `ECDSA`: ECDSA signatures by Externally Owned Accounts.
-    - `ERC-{n}`: A standard type of signature specified by an ERC. `n` must not be padded with zeros.
+  - `ECDSA`: ECDSA signatures by Externally Owned Accounts.
+  - `ERC-{n}`: A standard type of signature specified by an ERC. `n` must not be padded with zeros.
 - `parameters` (optional): An array of method-specific parameters.
 
 ### JSON Schema
 
 ```javascript
```
frangio: Please either remove the formatting changes or go all the way and make this valid JSON by wrapping keys in quotes.

ernestognw: But this is javascript.

I'll rollback

ERCS/erc-7803.md
frangio: ECDSA is necessary to express that the application only supports ECDSA authentication. (Additionally ERC-7913 doesn't cover ECDSA signatures, the trick described in the ERC's Backwards Compatibility section is not part of the spec.)

ERCS/erc-7803.md
frangio: Yeah I formatted this as JavaScript ot make it easier to read but it could also be a JSON block.

ERCS/erc-7803.md
ernestognw: A smart contract receiving a signature through `isValidSignature` (or any other method, the point is on-chain validation) will need to encode the components needed to reconstruct the signature. Imo the section is pretty clear. I'll do my best to update it

ERCS/erc-7803.md
ernestognw: I agree. Needs formatting in Javascript too

ERCS/erc-7803.md
ernestognw: Yes you're right. I thought it was possible when drafting it, sorry. I noticed that wallets would require to implement detection as well. Should we add a detection mechanism ala 7739?

```solidity
function isValidSignature(bytes32 hash, bytes calldata signature) public view virtual returns (bytes4 result) {
        // For the hash `0x7803780378037803780378037803780378037803780378037803780378037803` and an empty signature,
        // we return the magic value `0x78030001` as it's assumed impossible to find a preimage for it that can be used
        // maliciously. Useful for simulation purposes and to validate whether the contract supports ERC-7803.
        return
            (_isValidSigningDomainSignature(hash, signature) || _rawSignatureValidation(hash, signature))
                ? IERC1271.isValidSignature.selector
                : (hash == 0x7803780378037803780378037803780378037803780378037803780378037803 && signature.length == 0)
                    ? bytes4(0x78030001)
                    : bytes4(0xffffffff);
    }
```

We could only suggest it or recommend so that 7803 doesn't depend on 1271

ERCS/erc-7803.md
ernestognw: Does this makes sense to you? https://github.com/ethereum/ERCs/pull/1067/commits/7d23f0c65352492c6d4f0db8e4d478c229e90d5d#diff-045b70a6945a7105810dfd486ca7fc034c8464fc3c3490420d3b19cd58318e54R88

ERCS/erc-7803.md
ernestognw: Updated. Happy to hear your thoughts on detection but non-blocking imo

ERCS/erc-7803.md
ernestognw: Yes, this makes sense. I initially thought this was only about smart contract validation logic, but I see now that `authMethods` is primarily a coordination mechanism: it tells the wallet what signature formats the application's protocol supports so the wallet can generate the correct signature type. 

The smart contract validation is separate (handled by the contract's implementation), while `authMethods` ensures the wallet and application are aligned on what authentication methods will work.

And you're right that ECDSA needs to be explicit since ERC-7913's core interface doesn't cover it, it only handles non-Ethereum keys with verifier contracts. I was thinking from the perspective of [ERC7913Utils](https://github.com/OpenZeppelin/openzeppelin-contracts-experimental/blob/891eb8f2fa5ac50ca7e81b79e97bc1c2973573a0/contracts/utils/cryptography/ERC7913Utils.sol)

ERCS/erc-7803.md
ernestognw: Rolled it back. Not sure if you're suggesting to make it a JSON block. I think this would've been clearer:

"Thanks for noticing the Javascript required formatting. Can you please make it a JSON block? I think it would be more convenient"

ERCS/erc-7803.md
ernestognw: I think I understand it now. Here's a reference [implementation PR](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5722/files#diff-9b871e4de772c0b05ce6c230e26f4c313f523cddaa30d5338002afccedea65e8R53-R71), which made me realize I was mixing terms. I updated with a different decoding and a concise rationale, appreciate your clear feedback 😄

ERCS/erc-7803.md
```diff
@@ -81,15 +83,27 @@ In the presence of `signingDomains` the account should encode the message to be
...
+To accommodate existing onchain signature validation functions like [ERC-1271]'s `isValidSignature`, it is possible to use `signature` argument (or equivalent) to ABI encode the signature and its metadata. The following format is RECOMMENDED for completeness:
+
+```solidity
+abi.encodePacked(
+    uint16 signatureLength,              // Length of the actual signature
+    bytes signature,                     // The actual cryptographic signature
+    abi.encode(bytes32[] signingDomainSeparators)  // Signing domain separators (outermost first)
+)
+```
+
```
Amxx, line 97: I don't understand why the `signingDomainSeparators` would be passed in the signature.

AFAIK, the `signingDomainSeparators` are known by the different signers that are going to process the signature. Each SCA that gets a signature request would "process" the incomming hash using its own domain. Since it knows its local hash, getting it from somewhere else (the signature) should not necessary.

ERCS/erc-7803.md
Amxx: I'm not sure I understand the opposition between ERC-{n} and ECDSA. AFAIK they are two ortogonal things.

- ERC-721 is a way to prepare the hash that is signed, its counterpart is ERC-191
- ECDSA can be used to sign (and verify) signatures, its counterpart would be P256, or ERC-1271.

What does ERC-{n} refers to here, is it 191/721, or is it 1271 ?

ERCS/erc-7803.md
```diff
@@ -81,15 +83,27 @@ In the presence of `signingDomains` the account should encode the message to be
...
 
 This property is an array of supported signature authentication methods, listed in the order that the verifying domain tries them.
 
 Each member of the array is an object with the following keys:
 
 - `id`: An string that identifies the method. It may be one of:
-    - `ECDSA`: ECDSA signatures by Externally Owned Accounts.
-    - `ERC-{n}`: A standard type of signature specified by an ERC. `n` must not be padded with zeros.
+  - `ECDSA`: ECDSA signatures by Externally Owned Accounts.
+  - `ERC-{n}`: A standard type of signature specified by an ERC. `n` must not be padded with zeros.
```
Amxx, line 106: (note, this comment is about the original ERC, not about the changes made in this PR)

I'm not sure I understand the opposition between ERC-{n} and ECDSA. AFAIK they are two ortogonal things.

- ERC-721 is a way to prepare the hash that is signed, its counterpart is ERC-191
- ECDSA can be used to sign (and verify) signatures, its counterpart would be P256, or ERC-1271.

What does ERC-{n} refers to here, is it 191/721, or is it 1271 ?

ERCS/erc-7803.md
frangio: Yeah detection is an issue, I don't think there's a way to detect capabilities, I fear that `signingDomains` can just be ignoreds by wallets. I think the only solution is to use a new RPC method.

ERCS/erc-7803.md
frangio: I think this section needs to be removed, there's no reason for this ERC to mandate an encoding. It's up to the wallet or account to define the encoding according to their needs.

ERCS/erc-7803.md
frangio: > its counterpart would be P256, or ERC-1271.

Yes exactly, IIRC this was the reason. It would be one of ERC-1271, ERC-6492, ERC-8010, etc.

PR #1072, opened 2025-06-06
Add ERC: Proof-based Broadcast in ERC-7786 Gateways

ernestognw:
Extends ERC-7786 cross-chain messaging with cryptographic proof-based verification and granular broadcasting semantics. It enables trustless message verification using blockchain consensus mechanisms rather than external validators or multisigs. 

The specification defines three ERC-7786 attributes: `route()` for multi-hop verification paths, `inclusionProof()` for cryptographic evidence of message commitment, and optional `targetBlock()` for finality validation. 

Broadcasting leverages ERC-7930 addresses to enable targeted message distribution: empty address components broadcast to all addresses on specific chains, empty chain references broadcast to all chains of a type, and fully zeroed addresses enable universal broadcasting.

### DIFF
--- a/ERCS/erc-7965.md
+++ b/ERCS/erc-7965.md
@@ -40,7 +40,7 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 
 [ERC-7786] defines broadcasting as using "omitted or zeroed" recipient addresses (containing an [ERC-7930] interoperable address with all fields set to zero). This specification extends that concept to enable granular broadcasting patterns through [ERC-7930]'s flexible address structure:
 
-[ERC-7930]: ./erc-7930.md
+[ERC-7930]: ./eip-7930.md
 
 Broadcasting semantics in this specification extend [ERC-7786] by allowing:
 


ERCS/erc-xxxx.md
```diff
@@ -0,0 +1,174 @@
...
+Gateways MAY choose to write messages to storage locations that cannot be deleted or accessed after being set, providing immutability guarantees. While not required by this standard, implementers can use [ERC-7201] to calculate namespaces for these not-mutable storage locations.
+
+### Verification Process
+
+Message verification follows these steps:
+
+1. Parse the `route` and `storageProof` attributes from the message, and optionally `targetBlock` if provided
+2. Validate all required attributes are present and well-formed
+3. For each route step, verify block hash transition using the paired proof and validate version requirements if specified (non-zero)
+4. Use the `storageProof` to verify that message data exists in the source chain's state at the target block obtained from the route verification
```
bomanaps: In here the  verification process should clarify that the storage proof must correspond to the source chain identified by the completed route verification, what I mean by this is it should use the `storageProof` to verify that message data exists in the source chain's state at the `target block`. The source chain MUST correspond to the final validated step in the route verification process.

ERCS/erc-xxxx.md
```diff
@@ -0,0 +1,174 @@
...
+
+Receivers of broadcast messages SHOULD validate the source and authenticity of messages according to their own security requirements.
+
+### Storage Proof Attributes
+
+This specification defines the following [ERC-7786] attributes for storage proof messaging. Gateways MUST return true if `supportsAttribute` is called with the selector for supported attributes.
+
+#### `route((address,bytes,uint256)[])`
+
+Specifies the verification path from destination to source chain with corresponding proofs and version requirements. Each tuple contains an address that enables verification of the next chain's state, the cryptographic proof required for that verification step, and the expected version of the verification logic (0 means any version is acceptable).
```
bomanaps: Regarding the expected version of the verification logic, the specification states that 0 means any version is acceptable. I recommend clarifying that when a non-zero version is specified, gateways MUST reject messages if the route address does not support the exact required version. Additionally, route addresses SHOULD implement version querying mechanisms to enable compatibility checking. This will ensure stronger version enforcement and improve interoperability between implementations.

ernestognw: I've [added clarifications](https://github.com/ethereum/ERCs/pull/1072/commits/5ba1441459b148756cb3f640c2e73f2da51c5574#diff-01e4f3f9f1841fcd36d9c60ac028f570623a9d08e6aa439c94c17efa9153935dR51) to the `route` attribute section. Thanks for pointing out!

ERCS/erc-xxxx.md
ernestognw: Hi! Thanks for your feedback! I've updated to clarify that "The source chain SHOULD correspond to the final validated step in the route verification process."

I'm using SHOULD rather than MUST since I'm unsure about its enforceability. Perhaps there's a use case in sending through a different chain for better prices so I don't want to be too constraining here.

ERCS/erc-xxxx.md
```diff
@@ -0,0 +1,174 @@
...
+
+// Sending a broadcast message with optional targetBlock
+bytes[] memory attributes = new bytes[](3);
+attributes[0] = abi.encodeWithSignature("route((address,bytes,uint256)[])", hops);
+attributes[1] = abi.encodeWithSignature("storageProof(bytes)", storageProofData);
+attributes[2] = abi.encodeWithSignature("targetBlock(uint256)", blockNumber); // Optional
+
+gateway.sendMessage(
+    "eip155:42161",
+    "", // broadcast
```
Amxx:
```suggestion
    abi.encodePacked(bytes8(0x0001000002a4b100)), // eip155:42161 without an address
```

ernestognw: This makes sense. I'll update this ERC as soon as 7786 is updated too, otherwise I might have to update again 😅

ERCS/erc-7965.md
```diff
@@ -0,0 +1,184 @@
...
+#### `route((bytes,bytes,uint256)[])`
+
+Specifies the verification path from destination to source chain with corresponding proofs and version requirements. Each tuple contains a bytes-encoded address that enables verification of the next chain's state or transaction history, the cryptographic proof required for that verification step, and the expected version of the verification logic (0 means any version is acceptable).
+
+When a non-zero version is specified, gateways MUST reject messages if the route address does not support the exact required version (unless `0`). Route addresses SHOULD implement version querying mechanisms to enable compatibility checking.
+
+The route MUST form a valid path where each step represents a direct relationship between chains that enables state or transaction verification. For multi-hop scenarios, the route creates a chain of trust where each step verifies the next, ultimately establishing the authenticity of the source chain's state. Gateways MUST reject messages with invalid or incomplete proof data.
+
+```solidity
+abi.encodeWithSignature("route((bytes,bytes,uint256)[])", hops, proof, version);
```
ernestognw:
```suggestion
abi.encodeWithSignature("route((bytes,bytes,uint256)[])", router, proof, version);
```

ERCS/erc-7965.md
```diff
@@ -0,0 +1,184 @@
...
+#### `route((bytes,bytes,uint256)[])`
+
+Specifies the verification path from destination to source chain with corresponding proofs and version requirements. Each tuple contains a bytes-encoded address that enables verification of the next chain's state or transaction history, the cryptographic proof required for that verification step, and the expected version of the verification logic (0 means any version is acceptable).
+
+When a non-zero version is specified, gateways MUST reject messages if the route address does not support the exact required version (unless `0`). Route addresses SHOULD implement version querying mechanisms to enable compatibility checking.
+
+The route MUST form a valid path where each step represents a direct relationship between chains that enables state or transaction verification. For multi-hop scenarios, the route creates a chain of trust where each step verifies the next, ultimately establishing the authenticity of the source chain's state. Gateways MUST reject messages with invalid or incomplete proof data.
+
+```solidity
+abi.encodeWithSignature("route((bytes,bytes,uint256)[])", router, proof, version);
```
Amxx: according to the signature, there should be a single argument that is an array of tuples. However, this mentions 3 separate arguments (are they array?), which is inconsistent.

ERCS/erc-7965.md
```diff
@@ -0,0 +1,184 @@
...
+```
+
+#### `storageProof(bytes)`
```
Amxx: inclusionProof (events) are probably as important (if not more) than storage proofs. IMO they should be support (either by this, or by another attribute)

ERCS/erc-7965.md
```diff
@@ -0,0 +1,184 @@
...
+
+### Mutability of Message Commitments
+
+Gateways MAY choose to commit messages in ways that cannot be deleted or modified after being set, providing immutability guarantees. While not required by this standard, implementers can use [ERC-7201] to calculate namespaces for immutable storage locations on EVM chains, or equivalent immutability mechanisms on other chain architectures.
+
+[ERC-7201]: ./eip-7201.md
+
+### Verification Process
+
+Message verification follows these steps:
```
Amxx, line 113: My understanding is that this verification logic happens on the receiving end (destination chain). It described how to verify something actually happened on the source chain.

What I don't understand is:

- what is the relationship between the "data" part of the ERC-7786 message and these attributes. It it the 7786 receiver the one that must match the data and the attributes it receives, and interpret the data in the context of these attributes which have been verified by the destination gateway ?

- how are these atributes set on the sending side. If the sender expected to build them before doing the `send` call on the source chain? If so, that means they have to correspond to data that was in place (part of the state) at least 1 block before the "send" is mined. The fact that setting the data in place, and sending the crosschain message must be done in separate block feels like a big downside.

ERCS/erc-7965.md
```diff
@@ -0,0 +1,184 @@
...
+- **Flexible messaging patterns**: Supports both targeted and broadcast messaging through standardized semantics, enabling new classes of applications like oracles and intent settlement systems.
+- **Composability**: Full integration with existing [ERC-7786] infrastructure and tooling
+
+## Specification
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
+
+### Broadcasting Semantics
+
+Gateways MUST interpret an empty string `""` in the receiver address of `sendMessage()` as broadcasting. This convention allows cryptographic proof gateways to support both targeted messaging (to specific receivers) and broadcast messaging (to any interested party) within the same interface.
```
Amxx: With ERC-7786 using ERC-7930 interoperable address, this translates to something like

"Gateway MUST accept ERC-7930 interoperable addresses that have an empty address component as valid recipient when sending a message."

ERCS/erc-7965.md
ernestognw: Right. I decided to keep them together with the `inclusionProof(bytes)` attribute, each gateway can validate in different ways.

ERCS/erc-7965.md
ernestognw: I addressed the first point a bit better in https://github.com/ethereum/ERCs/pull/1072/commits/cebfce80900d8aaa78ef21484398e7beb93bcd27.

For the second point, I know each protocol has a different way of generating the proof for the same block they're proposing. For example, Taiko's SignalService uses [anchor transactions](https://github.com/taikoxyz/taiko-mono/blob/main/packages/protocol/docs/how_taiko_proves_blocks.md#anchor-transactions) at the beginning of each L2 block to establish verifiable state commitments, eliminating the need for separate block commitments. I'm not sure if this standard should determine how that's done

ernestognw:
Blocked on #1050

SamWilsn (approved):

Issue #729, opened 2024-11-24
Add ERC PR - bot checks bug for HTMLProofer


frankmint2024:
### Pull Request

https://github.com/ethereum/ERCs/pull/624

### What happened?

![image](https://github.com/user-attachments/assets/687a1a1b-2213-4b54-bb42-25b62f982f65)
I have added the reference implementation contracts in the assets/ directory and link to it in the proposal.
But the HTMLProofer check shows this strange error, and i can't find the eip-7765.html. what should i do here ?

jamesavechives:
Yes, must replace ERC to EIP to pass the test. But it is wrong!

frankmint2024:
> Yes, must replace ERC to EIP to pass the test. But it is wrong!

But actually this file dose not exist at all.
./_site/EIPS/eip-7765.html:379:

Julius278:
same issue here:
https://github.com/ethereum/ERCs/pull/1081

Milisya:
💪

PR #1081, opened 2025-06-11
Update ERC-7573: Correct asset paths in documentation

Julius278:
@cfries @pekola fyi

### DIFF
--- a/ERCS/erc-7573.md
+++ b/ERCS/erc-7573.md
@@ -42,7 +42,7 @@ Here, we propose a protocol that facilitates secure delivery-versus-payment with
 
 The following methods specify the functionality of the smart contract implementing
 the locking. For further information, please also look at the interface
-documentation [`ILockingContract.sol`](../assets/eip-7573/contracts/ILockingContract.sol).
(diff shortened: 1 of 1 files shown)


cfries:
I believe the asset path has to be eip - see this comment here:
https://github.com/ethereum/ERCs/pull/871#discussion_r2027477441
I know it is wrong, but the issue is that the web page that is build will move the assets to eip. So with the "wrong" asset path, this page renders the figures: https://eips.ethereum.org/EIPS/eip-7573 - with the right one it does not.

I too don't like this, it appears to be a left over from the EIP->ERC migration.

cfries:
Similar issue here: https://github.com/ethereum/ERCs/issues/729#issuecomment-2500769982

Julius278:
> Similar issue here: [#729 (comment)](https://github.com/ethereum/ERCs/issues/729#issuecomment-2500769982)

hmm.. okay, so a bug on repository / pipeline side.. 
thanks for the information, I will close the PR and hopefully it gets fixed

I just commented in the issue to get a notification when its fixed

PR #1094, opened 2025-06-16
Add ERC: Future-block oracleless randomness

angrymouse:

 - The build passes.


ERCS/erc-future-block-randomness.md
```diff
@@ -0,0 +1,93 @@
...
+### Computation Algorithm
+
+The core of the scheme is its two-stage entropy aggregation algorithm. To compute the randomness for a given `requestId` originating from `requestBlockNumber`:
+
+1.  **Initialize**: An accumulator, `randomness_accumulator`, is initialized to zero.
+
+2.  **Aggregate**: The algorithm MUST perform the following loop:
+    *   For each block `i` in the range from `requestBlockNumber + 1` to `requestBlockNumber + BLOCK_WINDOW`:
+        a.  **Source the blockhash**: Fetch the `blockhash(i)`.
+        b.  **Bind to request**: Create a request-specific hash to prevent cross-request contamination: `intermediate_hash = keccak256(abi.encodePacked(blockhash(i), requestId))`.
```
bomanaps: Don't you think this could theoretically allow collisions under malicious input. We can tru using `abi.encode(blockhash, requestId)` or structured hashing (e.g., EIP-712 style), which is unambiguous.

angrymouse: By input, do you mean requestId? 
I think the only risk would be if the requestId is the same across different requests - making it easier for validators to manipulate them all at once. But as for other factors, since blockhashes are not defined when you get requestId, there shouldn't be a problem with collisions.

angrymouse:
CC @SamWilsn

angrymouse:
image
I could not post this on ethereum magicians for some reason. Any thoughts why?

angrymouse:
@SamWilsn

SamWilsn:
Unfortunately this proposal does not seem like a good fit for the ERC process. EIPs/ERCs exist to promote interoperability between independent implementers. Generally we require that there be multiple implementations on both sides of the standard. For example, tokens are a good fit because on one half we have many wallets and on the other half we have many different tokens.

This proposal looks like it'll have a single implementation to generate randomness, which can be picked up as a library by contracts that want it. Instead of an ERC, this proposal should get API documentation and its own git repository.

PR #1098, opened 2025-06-17
Update ERC-7715: Simplify specification to ease wallet implementation

V00D00-child:
This PR proposes simplifications to the ERC-7715 specification to support a more manageable implementation across wallets.

For additional context and rationale behind these changes, please refer to the supporting document: [Proposal to Simplify ERC-7715](https://metamask-consensys.notion.site/Proposal-to-Simplify-ERC-7715-215f86d67d688041b580dd161968b2df?pvs=143)

### DIFF
--- a/ERCS/erc-7715.md
+++ b/ERCS/erc-7715.md
@@ -2,7 +2,7 @@
 eip: 7715
 title: Grant Permissions from Wallets
 description: Adds JSON-RPC method for granting permissions from a wallet
-author: Luka Isailovic (@lukaisailovic), Derek Rein (@arein), Dan Finlay (@danfinlay), Derek Chiang (@derekchiang), Fil Makarov (@filmakarov), Pedro Gomes (@pedrouid), Conner Swenberg (@ilikesymmetry), Lukas Rosario (@lukasrosario), Idris Bowman (@V00D00-child)
+author: Luka Isailovic (@lukaisailovic), Derek Rein (@arein), Dan Finlay (@danfinlay), Derek Chiang (@derekchiang), Fil Makarov (@filmakarov), Pedro Gomes (@pedrouid), Conner Swenberg (@ilikesymmetry), Lukas Rosario (@lukasrosario), Idris Bowman (@V00D00-child), Jeff Smale (@jeffsmale90)
 discussions-to: https://ethereum-magicians.org/t/erc-7715-grant-permissions-from-wallets/20100
 status: Draft
 type: Standards Track
--- a/ERCS/erc-7715.md
+++ b/ERCS/erc-7715.md
@@ -2,7 +2,7 @@
 eip: 7715
 title: Grant Permissions from Wallets
 description: Adds JSON-RPC method for granting permissions from a wallet
-author: Luka Isailovic (@lukaisailovic), Derek Rein (@arein), Dan Finlay (@danfinlay), Derek Chiang (@derekchiang), Fil Makarov (@filmakarov), Pedro Gomes (@pedrouid), Conner Swenberg (@ilikesymmetry), Lukas Rosario (@lukasrosario)
+author: Luka Isailovic (@lukaisailovic), Derek Rein (@arein), Dan Finlay (@danfinlay), Derek Chiang (@derekchiang), Fil Makarov (@filmakarov), Pedro Gomes (@pedrouid), Conner Swenberg (@ilikesymmetry), Lukas Rosario (@lukasrosario), Idris Bowman (@V00D00-child)
 discussions-to: https://ethereum-magicians.org/t/erc-7715-grant-permissions-from-wallets/20100
 status: Draft
 type: Standards Track
@@ -13,7 +13,7 @@ requires: 4337, 5792, 7679, 7710
 
 ## Abstract
 
-We define a new JSON-RPC method `wallet_grantPermissions` for DApp to request a Wallet to grant permissions in order to execute transactions on the user’s behalf. This enables two use cases:
+We define a new JSON-RPC method `wallet_requestExecutionPermissions` for DApp to request a Wallet to grant permissions in order to execute transactions on the user’s behalf. This enables two use cases:
 
 - Executing transactions for users without a wallet connection.
 - Executing transactions for users with a wallet connection that is scoped with permissions.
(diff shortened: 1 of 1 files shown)


jeffsmale90 (approved):

ERCS/erc-7715.md
```diff
@@ -29,30 +29,170 @@ Each interaction requires the user to sign a transaction with their wallet. The
...
+  type: "native-token-stream";
+  isAdjustmentAllowed: boolean;
+  data: {
+    initialAmount?: string; // hex-encoded uint256; allowance available at startTime
+    maxAmount?: string; // hex-encoded uint256; maximum total accumulable allowance
+    timePeriod: number; // accrual period in seconds (e.g., 3600 for hourly)
+    startTime: number; // unix timestamp in seconds at which accrual begins
+    amountPerPeriod: string; // hex-encoded uint256; amount accrued every timePeriod seconds
+  };
+};
```
lukaisailovic: This looks very similar to `NativeTokenPeriodicPermission` I would suggest picking one or the other for the ERC. On the first glance it seems like stream should be enough?

ERCS/erc-7715.md
```diff
@@ -29,30 +29,170 @@ Each interaction requires the user to sign a transaction with their wallet. The
...
+  data: {
+    timestamp: number; // unix timestamp at which the permission becomes invalid
+  };
+};
+```
+
+### `wallet_requestExecutionPermissions`
 
-#### Permission schema
+We introduce a `wallet_requestExecutionPermissions` method for the DApp to request the Wallet to grant permissions.
```
lukaisailovic, line 75: Why change method name?

ERCS/erc-7715.md
```diff
@@ -61,236 +201,172 @@ type PermissionRequest = {
...
-type MultiKeySigner = {
-  type: "keys";
-  data: {
-    keys: {
-      type: KeyType;
-      publicKey: `0x${string}`;
-    }[];
-  };
-};
+The wallet SHOULD include `signerTypes` (`string[]`), `permissionTypes` (`string[]`) and `ruleTypes` (`string[]`) in the response, to specify the signer types and permission types it supports.
```
lukaisailovic: I would strongly suggesting dropping this and making whatever is in 7715 a MUST implementation for wallets. Otherwise there is too much optionality and a lot of requests where the user connects a wallet and dapp wants to request something while the wallet for example only supports streaming permission and nothing else

ERCS/erc-7715.md
```diff
@@ -61,236 +201,172 @@ type PermissionRequest = {
...
+  }
+}
 ```
 
-### Wallet-managed Sessions
+### Sending transaction to redeem permissions
+
+#### Wallet-managed Sessions
 
 If the signer is specified as `wallet`, then the wallet itself manages the session. If the wallet approves the request, it MUST accept [ERC-5792](./eip-5792.md)’s `wallet_sendCalls` with the `permissions` capability, which MAY include the session with a `permissionsContext`. For example:
```
lukaisailovic: I think wallet signer can be removed. This RPC focuses on Dapp to Wallet communication. If wallet wants, it can manage the permission itself without the RPC. Removing it means dropping the signer optionality completely as well.

ERCS/erc-7715.md
McOso: Thanks for the feedback! While they may look similar at first glance, they serve distinct use cases. Periodic permissions are best suited for subscription-style payments (e.g., $30 every month), where value unlocks in discrete intervals. Streaming permissions, on the other hand, represent a continuous linear unlock over time, more appropriate for scenarios where funds vest or accrue per second. Because these patterns enable meaningfully different behaviors, we believe both are valuable to support in the ERC.

ERCS/erc-7715.md
McOso: We already have `wallet_revokePermissions` implemented in the wallet. So we decided to change the revoke endpoint to `wallet_revokeExecutionPermission` and to keep things consistent we changed `wallet_grantPermissions` to `wallet_requestExecutionPermissions`

ERCS/erc-7715.md
McOso: We see optionality as an important feature rather than a drawback. Different wallets may prioritize different permission types depending on their product goals and user base, and we don’t want the ERC to require every wallet to implement all possible patterns from day one.

But I do think we need a discoverability mechanism that allows dapps to easily determine what a given wallet supports before making a request. In other words, discoverability mitigates the optionality problem while still allowing the ecosystem to innovate and adopt new permission types over time without forcing all wallets to upgrade simultaneously.

ERCS/erc-7715.md
McOso: The purpose of wallet signer is so the dapp doesnt have to have an account to give the permission to. Instead the wallet spins up a signer that the dapp can use to execute the permission from the user. 

We aren't thinking of implementing this right away, so perhaps we do drop this from the ERC. It will simplify it and it means we can drop the 5792 requirement as well.

Just to make sure, can you clarify what RPC you are talking about here?

ERCS/erc-7715.md
lukaisailovic: Wouldn't it be easier and better to drop the optionality completely and use ERCs for that? Like in 7715, you have x permissions and if you support it you need to support all of them. And in ERC-XXXX you add one more permission and if some signals support for ERC-XXXX everyone knows exactly what that means. 

Alternatively 7715 can be versioned and permission support can always be added as a must for different versions.

IF this is not done, then the wallet can support just one permission and the app can support just one permission and while they technically do support 7715, in most cases they will never be able to interact with one another

ERCS/erc-7715.md
lukaisailovic: No RPC, I'm just saying that the dapp doesn't need to ask for permissions if they are managed by the wallet. Wallet can just spin up a signer "locally" (no communication required) and when it accepts normal RPCs like `sendTransaction` it can execute with signer instead of the main account.

If the dapp is managing the signer, it just uses the normal flow. Thats why I think this pattern is unnecessary.

ERCS/erc-7715.md
lukaisailovic: Can't this be handled with a streaming permission?
`timePeriod = 30 days`
`amountPerPeriod = 30`

add new rule:
`capPerPeriod = true` (allowance doesn't carry over to the next period)

ERCS/erc-7715.md
V00D00-child: > Wouldn't it be easier and better to drop the optionality completely and use ERCs for that? 

 @lukaisailovic yes, it would drastically reduce the complexity of 7715 by **_removing the list of default permission types_** and **_using additional ERC-XXXX(s) to add more permissions_**.  

Optionality would then live at the ERC level, not within the core spec. Wallet chooses which additional ERCs to support to expand the set of permission offerings for dApps, while staying compatible with 7715. This enables experimentation and ecosystem growth without bloating the base standard.

Under this model, 7715’s responsibility would be limited to:
- Defining the interface for how a dApp can request/revoke a permission
- Define the base scheme type for permissions and rules to avoid collisions in downstream ERCs
- Defining the discovery interface that additional ERCs will extend to inform dApps of the new permission types

> Alternatively 7715 can be versioned and permission support can always be added as a must for different versions.

Regarding versioning: I suggest avoiding it, as it can increase complexity and fragment compatibility where certain permissions exist only in specific versions. From a dApp developer perspective, handling multiple permissions and versions quickly leads to excessive conditional logic and poor DX.

A stable core + extensible ERCs is a cleaner and more future-proof approach.

ERCS/erc-7715.md
```diff
@@ -206,17 +206,17 @@ type GetPermissionsInfoResultParams = {
...
 #### Request Specification
 
 **Request example**:
 
 ```tsx
 window.ethereum.request({
-  "method": "wallet_getSupportedPermissions",
+  "method": "wallet_getSupportedExecutionPermissions",
   "params": []
 }): Promise<GetSupportedPermissionsResult>
```
MoMannn: GetSupportedExecutionPermissionsResult

ERCS/erc-7715.md
```diff
@@ -254,17 +254,17 @@ An object keyed on all permission types supported by the Wallet expected by the
...
 #### Request Specification
 
 **Request example**:
 
 ```tsx
 window.ethereum.request({
-  "method": "wallet_getGrantedPermissions",
+  "method": "wallet_getGrantedExecutionPermissions",
   "params": []
 }): Promise<getGrantedPermissionsParams>
```
MoMannn: Probably should update response as well with getGrantedExecutionPermissionsParams

ERCS/erc-7715.md
danfinlay: The dapp would still need to ask for permissions to be granted _to its signing key_.

ERCS/erc-7715.md
danfinlay: I also think streams are basically superior to periodic permissions, but they are definitely distinct permissions. Periodic only increases the allowance at the specified interval, more like traditional subscription systems.

danfinlay (approved):
I think this is a tremendous improvement and simplification on the current spec, and all the feedback right now is about whether to declare some optional things mandatory, which I don't think is worth blocking over (since this ultimately only manifests in the interpretation of alternative implementors readings, and they're already going to decide how much of the specification to follow).

Curious: Does any organization have an implementation in progress besides MetaMask?

danfinlay:
@gakonst I was notified that Porto [has an implementation](https://porto.sh/sdk/wagmi/useGrantPermissions), so I'm letting you know about the spec change. It's simplier, but has a new method name, so it should be pretty easy to match parity with.

jxom:
Interesting that this spec is tightly coupled to ERC-7710 now (Porto doesn't use ERC-7710), and now bound to a single permission (rather than multiple). Is that intended? For the former – generally think standards should not lean into implementation detail too much.

octavio12345300 (commented):
listo

PR #1114, opened 2025-07-04
Add Gateway Attributes for Message Control

ernestognw:
Introduces standard attributes for ERC-7786 cross-chain messaging gateways, enabling consistent message lifecycle control across implementations.

**Adds 7 standard attributes:**
- `cancellable(bool)` - Message cancellation
- `timeout(uint256)` - Automatic expiration  
- `earliestExecTime(uint256)` - Delayed execution
- `retryPolicy(bytes)` - Retry mechanisms
- `revertBehavior(uint8)` - Failure handling
- `dependsOn(bytes32[])` - Message dependencies
- `minGasLimit(uint256)` - Gas requirements

**Benefits:** Ecosystem consistency, predictable APIs, advanced workflows, and reliable error handling. Extends ERC-7786 without breaking changes. Emerged from discussions with Matter Labs' InteropCenter team addressing real-world cross-chain messaging requirements.

### DIFF
--- a/ERCS/erc-xxxx.md
+++ b/ERCS/erc-xxxx.md
@@ -1,0 +1,111 @@
+---
+eip: XXXX
+title: Gateway Attributes for Message Control
+description: Gateway attributes for cancellation, timeout, retry, dependencies, and execution control in cross-chain messaging.
+author: Ernesto García (@ernestognw), Kalman Lajko (@LajkoKalman), Valera Grinenko (@0xValera)
+discussions-to: TODO
+status: Draft
+type: Standards Track
+category: ERC
+created: 2024-XX-XX
+requires: 7786
+---
+
+## Abstract
+
+This ERC defines standard attributes for ERC-7786 cross-chain messaging gateways to enable consistent cancellation, timeout, retry, dependency, and execution control mechanisms across implementations. These attributes provide applications with predictable control over message lifecycle, ordering, and execution requirements.
+
+## Motivation
+
+ERC-7786 introduces an extensible attribute system for cross-chain messaging, but leaves attribute standardization to follow-up specifications. As cross-chain applications mature, consistent patterns for message control have emerged as essential requirements:
+
+1. **Cancellation**: Applications need to cancel pending messages due to changed conditions
+2. **Timeouts**: Automatic cancellation prevents indefinite pending states
+3. **Retry Logic**: Standardized failure handling improves reliability
+4. **Revert Behavior**: Consistent error semantics across gateways
+5. **Message Dependencies**: Ensuring correct ordering when messages must execute in sequence
+6. **Gas Requirements**: Preventing execution failures due to insufficient gas
+7. **Execution Timing**: Controlling when messages can be executed for scheduling and coordination
+
+Without standardized attributes, each gateway implements these features differently, fragmenting the ecosystem and requiring application-specific integration logic.
+
+## Specification
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
+
+### Standard Attributes
+
+This specification defines standard attributes for ERC-7786 cross-chain messaging gateways.
+
+Gateways MAY implement attributes independently. Gateways MUST validate the attribute's encoding for each attribute they implement and revert the transaction if the encoding is invalid.
+
+#### `cancellable(bool)`
+
+Indicates whether a message can be cancelled after submission. This attribute uses selector `0xde986d7f`, which represents the first 4 bytes of `keccak256("cancellable(bool)")`.
+
+The attribute value is encoded as an ABI-encoded boolean, and COULD default to `false` when not specified. When set to `true`, gateways MUST provide a cancellation mechanism to allow applications to cancel pending messages due to changed conditions or requirements.
+
+#### `timeout(uint256)`
+
+Specifies a timestamp after which the message is automatically cancelled. This attribute uses selector `0x08148f7a`, derived from the first 4 bytes of `keccak256("timeout(uint256)")`.
+
+The value is encoded as an ABI-encoded Unix timestamp, and COULD default to `0` when not specified. Gateways MUST NOT execute messages after the timeout timestamp unless `0` is specified, which MUST be interpreted as no timeout.
+
+#### `earliestExecTime(uint256)`
+
+Specifies the earliest timestamp at which the message can be executed. This attribute uses selector `0x6c5875a2`, derived from the first 4 bytes of `keccak256("earliestExecTime(uint256)")`.
+
+The value is encoded as an ABI-encoded Unix timestamp, and COULD default to `0` when not specified. Gateways MUST NOT execute messages before the earliestExecTime timestamp unless `0` is specified, which MUST be interpreted as no delay. When combined with `timeout(uint256)`, this creates an execution time window.
+
+#### `retryPolicy(bytes)`
+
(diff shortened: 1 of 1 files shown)


ERCS/erc-xxxx.md
```diff
@@ -0,0 +1,111 @@
...
+
+#### `cancellable(bool)`
+
+Indicates whether a message can be cancelled after submission. This attribute uses selector `0xde986d7f`, which represents the first 4 bytes of `keccak256("cancellable(bool)")`.
+
+The attribute value is encoded as an ABI-encoded boolean, and COULD default to `false` when not specified. When set to `true`, gateways MUST provide a cancellation mechanism to allow applications to cancel pending messages due to changed conditions or requirements.
+
+#### `timeout(uint256)`
+
+Specifies a timestamp after which the message is automatically cancelled. This attribute uses selector `0x08148f7a`, derived from the first 4 bytes of `keccak256("timeout(uint256)")`.
```
kelemeno: > automatically cancelled

I would rephrase, "cannot be executed". ( with this phrasing there has to be an automatic cancellation mechanism). 

Alternatively, "implicitely cancelled".

ernestognw: Addressed in be754a5

ERCS/erc-xxxx.md
```diff
@@ -0,0 +1,111 @@
...
+
+This specification defines standard attributes for ERC-7786 cross-chain messaging gateways.
+
+Gateways MAY implement attributes independently. Gateways MUST validate the attribute's encoding for each attribute they implement and revert the transaction if the encoding is invalid.
+
+#### `cancellable(bool)`
+
+Indicates whether a message can be cancelled after submission. This attribute uses selector `0xde986d7f`, which represents the first 4 bytes of `keccak256("cancellable(bool)")`.
+
+The attribute value is encoded as an ABI-encoded boolean, and COULD default to `false` when not specified. When set to `true`, gateways MUST provide a cancellation mechanism to allow applications to cancel pending messages due to changed conditions or requirements.
```
0xValera: > COULD

Somehow this slipped past me while reading ERC-7786, but both in this draft, and in ERC-7786 the word COULD is used assumably in the same context as RFC 2119 key words, even though it (RFC 2119) doesn't define COULD.

I would suggest to change COULDs to an RFC 2119 key words, for consistency.

ERCS/erc-xxxx.md
```diff
@@ -0,0 +1,111 @@
...
+
+The format follows `abi.encodePacked(uint16(maxRetries), uint32(retryDelay), uint32(backoffMultiplier))`, where `maxRetries` specifies the maximum number of retry attempts (with 0 indicating no retries), `retryDelay` defines the initial delay between retries in seconds, and `backoffMultiplier` provides the multiplier for exponential backoff in basis points (with 10000 representing 1x multiplier).
+
+The attribute value COULD default to `0x` when not specified, equivalent to infinite retries, no delay, and no backoff (or `maxRetries = 0`, `retryDelay = 0`, and `backoffMultiplier = 0`).
+
+#### `revertBehavior(uint8)`
+
+Specifies how execution failures MUST be handled. This attribute uses selector `0x9e521a77`, representing the first 4 bytes of `keccak256("revertBehavior(uint8)")`.
+
+The value is encoded as an ABI-encoded uint8 with three possible values: `0` for silent failure (the default behavior), `1` for reverting the transaction, and `2` for emitting a failure event and continuing execution. When not specified, the attribute defaults to `0`.
```
0xValera: I would suggest to specify the expected behavior for each of possible values. Something in likes of:

- `0` – Silent Failure
  - Gateways MUST NOT revert the transaction.
  - Gateways MUST NOT emit any failure-related events.

- `1` – Revert on Failure
  - Gateways MUST revert the entire message execution when any failure occurs.

- `2` – Emit-and-Continue
  - Gateways MUST emit a `MessageFailed(bytes32 messageId, string reason)` event upon failure.  
  - Gateways MUST continue execution of subsequent messages or operations.

0xValera: Also, we did not document "a failure event" anywhere (I added it here, but not sure if it has to be present somewhere else in text as well)

ERCS/erc-xxxx.md
```diff
@@ -0,0 +1,111 @@
...
+
+#### `dependsOn(bytes32[])`
+
+Specifies message dependencies that must be executed before this message. This attribute uses selector `0xa9fed7b9`, derived from the first 4 bytes of `keccak256("dependsOn(bytes32[])")`.
+
+The value is encoded as an ABI-encoded array of message identifiers. Gateways MUST NOT execute a message until all messages specified in the `dependsOn` array have been successfully executed. When not specified or empty, the message has no dependencies. This ensures correct ordering and prevents out-of-order delivery issues.
```
0xValera: Should the way the status of the messages is tracked also be standardized?

The gateway must be able to verify if the message was executed for this attribute to work, but there are also a couple of different places where it could be useful/is required for attributes to be able to work correctly with each other.

For example, we have cancellation, which assumably can also just modify the status of the message, so the gateway knows that it shouldn't be executed/etc after cancellation.

Same argument as before goes for messages which were failed. If the message had `revertBehavior` attribute, and was processed by gateway but failed, I assume it should be executed/cancelled/etc after failing.

ERCS/erc-xxxx.md
ernestognw: Yeah this makes sense. I don't know where I got this idea at first. Updated them with `MAY`

ERCS/erc-xxxx.md
ernestognw: Your proposal makes sense. I reused the event you're proposing (`MessageFailed(bytes32 messageId, string reason)`). Maybe we can suggest encoding as in [ERC-7751](https://eips.ethereum.org/EIPS/eip-7751) without requiring it.

ERCS/erc-xxxx.md
ernestognw: Right, I suspect  ERC-7786 already covers the properties you're describing about messages. So I was thinking to reuse the definition of "delivery" from ERC-7786 rather than defining new status tracking mechanisms. ERC-7786 already establishes a clear message lifecycle: messages are sent (via `MessageSent` event) and then delivered (via successful `receiveMessage` call).

I've updated the specification to make this relationship clear; all references to "execution" have been changed to "delivery" to align with ERC-7786's terminology. I also renamed `timeout`/`earliestExecTime` to `deliverBefore`/`deliverAfter` for consistency.

ERCS/erc-7985.md
```diff
@@ -0,0 +1,130 @@
+---
+eip: 7985
+title: Gateway Attributes for Message Control
+description: Gateway attributes for cancellation, timeout, retry, dependencies, and delivery control in cross-chain messaging.
```
SamWilsn, line 4: I don't know what a "gateway attribute" is, and I'm probably representative of the average ERC reader. This is an extremely tiny nit, but if you could choose a different wording in your description that might make it more obvious, I think it would make better use of the limited character space.

ERCS/erc-7985.md
```diff
@@ -0,0 +1,130 @@
...
+
+#### `cancellable(bool)`
+
+Indicates whether a message can be cancelled after submission. This attribute uses selector `0xde986d7f`, which represents the first 4 bytes of `keccak256("cancellable(bool)")`.
+
+The attribute value is encoded as an ABI-encoded boolean, and MAY default to `false` when not specified. When set to `true`, gateways MUST provide a cancellation mechanism to allow applications to cancel pending messages due to changed conditions or requirements.
+
+#### `deliverBefore(uint256)`
+
+Specifies a timestamp after which the message cannot be delivered. This attribute uses selector `0x3e97d7ee`, derived from the first 4 bytes of `keccak256("deliverBefore(uint256)")`.
```
SamWilsn, line 52: `deliverBefore` sounds like it implies `now < value`, where the description of this method implies `now <= value`. Is that intentional?

SamWilsn (approved):

PR #1126, opened 2025-07-18
Update ERC-7682: Add section on `capabilities.auxiliaryFunds` parameter, `requiredAssets`

adonesky1:
This PR extends `ERC-7682` to specify that when a wallet supports the `auxiliaryFunds` capability, applications using `wallet_sendCalls` must include a `assetsRequiredForExecution` parameter nest under `capabilties` -> `auxiliaryFunds`. This parameter explicitly lists the assets and amounts required for successful execution, as transaction call data alone is insufficient for wallets to infer these requirements.

### DIFF
--- a/ERCS/erc-7682.md
+++ b/ERCS/erc-7682.md
@@ -146,14 +146,14 @@ This mechanism allows advanced wallets to abstract away the complexity of cross-
 
 This specification currently supports three token standards with the following field requirements:
 
-- **ERC-20**: Requires `amount` field only
-- **ERC-721**: Requires both `amount` and `tokenId` fields
-- **ERC-1155**: Requires both `amount` and `tokenId` fields
+- **[ERC-20](./eip-20.md)**: Requires `amount` field only
+- **[ERC-721](./eip-721.md)**: Requires both `amount` and `tokenId` fields
+- **[ERC-1155](./eip-1155.md)**: Requires both `amount` and `tokenId` fields
 
 The `amount` field semantics vary by token standard:
-- **ERC-20**: Represents token quantity in smallest unit (e.g., wei for 18-decimal tokens)
-- **ERC-721**: Represents NFT quantity (typically "0x01" for single NFT instances)
-- **ERC-1155**: Represents quantity of the specified token ID
+- **[ERC-20](./eip-20.md)**: Represents token quantity in smallest unit (e.g., wei for 18-decimal tokens)
+- **[ERC-721](./eip-721.md)**: Represents NFT quantity (typically "0x01" for single NFT instances)
+- **[ERC-1155](./eip-1155.md)**: Represents quantity of the specified token ID
 
 Future token standards MUST specify additional required fields as properties on the asset object root to maintain extensibility.
 
--- a/ERCS/erc-7682.md
+++ b/ERCS/erc-7682.md
@@ -106,13 +106,12 @@ type WalletSendCallsRequest = {
         };
         erc721?: {
           [asset: `0x${string}`]: {
-            tokenId: `0x${string}`; // Token ID as a hex string
+            tokenIds: `0x${string}`[]; // Array of token IDs as hex strings
           };
         };
         erc1155?: {
           [asset: `0x${string}`]: {
-            tokenId: `0x${string}`; // Token ID as a hex string
-            amount: `0x${string}`; // Amount required, as a hex string representing the integer value
+            [tokenId: `0x${string}`]: `0x${string}`; // Token ID to amount mapping, where amount is a hex string
           };
         };
       };
@@ -135,15 +134,14 @@ type ERC20Asset = {
 #### [ERC-721](./eip-721.md)
 ```typescript
 type ERC721Asset = {
-  tokenId: `0x${string}`; // Token ID as a hex string
+  tokenIds: `0x${string}`[]; // Array of token IDs as hex strings
 };
 ```
 
 #### [ERC-1155](./eip-1155.md)
 ```typescript
 type ERC1155Asset = {
-  tokenId: `0x${string}`; // Token ID as a hex string
-  amount: `0x${string}`; // Amount required, as a hex string representing the integer value
+  [tokenId: `0x${string}`]: `0x${string}`; // Token ID to amount mapping, where amount is a hex string
 };
 ```
 
--- a/ERCS/erc-7682.md
+++ b/ERCS/erc-7682.md
@@ -151,10 +151,6 @@ Apps MUST include the `assetsRequiredForExecution` parameter in the `capabilitie
 ### Gas Estimation and Native Asset Handling
 - Applications SHOULD include an estimate for the required native asset (e.g., ETH for gas) in `assetsRequiredForExecution` when possible.
 
-## Backwards Compatibility
-- Applications SHOULD only include the `assetsRequiredForExecution` parameter if the wallet advertises the `auxiliaryFunds` capability.
-- If a wallet does not recognize or support the `assetsRequiredForExecution` parameter, it SHOULD ignore the parameter or return a clear error indicating lack of support. Apps SHOULD be prepared to handle such cases gracefully, potentially falling back to legacy behavior.
-
 ## Rationale
 
 ### Alternatives
@@ -164,6 +160,11 @@ Apps MUST include the `assetsRequiredForExecution` parameter in the `capabilitie
 An alternative we considered is defining a way for apps to fetch available auxiliary balances. This could be done, for example, by providing a URL as part of the `auxiliaryFunds` capability that apps could use to fetch auxiliary balance information. However, we ultimately decided that a boolean was enough to indicate to apps that they should not block user actions on the basis of balance checks, and it is minimally burdensome for apps to implement.
 
 The shape of this capability allows for a more advanced extension if apps feel more functionality is needed.
+
+## Backwards Compatibility
+- Applications SHOULD only include the `assetsRequiredForExecution` parameter if the wallet advertises the `auxiliaryFunds` capability.
+- If a wallet does not recognize or support the `assetsRequiredForExecution` parameter, it SHOULD ignore the parameter or return a clear error indicating lack of support. Apps SHOULD be prepared to handle such cases gracefully, potentially falling back to legacy behavior.
+
 
 ## Security Considerations
 
--- a/ERCS/erc-7682.md
+++ b/ERCS/erc-7682.md
@@ -139,6 +139,9 @@ The wallet, upon receiving this request, can:
 
 This mechanism allows advanced wallets to abstract away the complexity of cross-chain or offchain funding, enabling seamless user experiences even when the user's onchain balance is insufficient on the target chain.
 
+### Auxiliary Funds and Atomic Execution
+
+Auxiliary funds provisioning (bridging, swapping, or other asset retrieval operations) is independent of the `wallet_sendCalls` execution lifecycle. The `atomicRequired` field applies only to the call bundle execution, not to auxiliary funds provisioning.
 
 ### Error Codes
 
--- a/ERCS/erc-7682.md
+++ b/ERCS/erc-7682.md
@@ -151,9 +151,8 @@ The following error codes are defined for auxiliary funds provisioning failures.
 |----------|-----------------------------------|-------------|
 | -32080   | Auxiliary funds provisioning failed | Wallet attempted to provision funds but failed (e.g., bridge/swap error). |
 | -32081   | Asset not supported             | The requested asset is not available through the wallet’s auxiliary fund system. |
-| -32082   | Auxiliary funds timeout         | The provisioning operation took too long or timed out. |
-| -32083   | Auxiliary funds capability not available | The wallet no longer supports auxiliary funds on the requested chain. |
-| -32084   | Invalid assetsRequiredForExecution | The structure of the assetsRequiredForExecution object is malformed. |
+| -32082   | Auxiliary funds capability not available | The wallet no longer supports auxiliary funds on the requested chain. |
+| -32083   | Invalid assetsRequiredForExecution | The structure of the assetsRequiredForExecution object is malformed. |
 
 Applications SHOULD use the `wallet_getCallsStatus` method to track the status of call bundles that involve auxiliary funds provisioning, as the provisioning process may take time to complete. The status response will indicate whether the auxiliary funds provisioning is in progress, completed, or failed.
 
@@ -166,9 +165,6 @@ Apps MAY include the `assetsRequiredForExecution` parameter in the `capabilities
 ### Amount Validation
 
 The wallet MUST cross-check the `amount` value in `assetsRequiredForExecution` against the `decimal()` value specified on the `asset`'s contract to ensure that the amount is interpreted correctly according to the asset's decimal precision.
-
-### Gas Estimation and Native Asset Handling
-- Applications SHOULD include an estimate for the required native asset (e.g., ETH for gas) in `assetsRequiredForExecution` when possible.
 
 ## Rationale
 
--- a/ERCS/erc-7682.md
+++ b/ERCS/erc-7682.md
@@ -2,7 +2,7 @@
 eip: 7682
 title: Auxiliary Funds Capability
 description: A capability allowing wallets to indicate that they have access to additional funds.
-author: Lukas Rosario (@lukasrosario), Wilson Cusack (@wilsoncusack)
+author: Lukas Rosario (@lukasrosario), Wilson Cusack (@wilsoncusack), Alex Donesky (@adonesky1)
 discussions-to: https://ethereum-magicians.org/t/erc-7682-auxiliary-funds-capability/19599
 status: Draft
 type: Standards Track
--- a/ERCS/erc-7682.md
+++ b/ERCS/erc-7682.md
@@ -72,9 +72,69 @@ type AuxiliaryFundsCapability = {
 }
 ```
 
-### App Implementation
+### Extended Usage: `assetsRequiredForExecution` Parameter
+
+When a wallet indicates support for the `auxiliaryFunds` capability (i.e., `supported: true`), applications that use the [`wallet_sendCalls`](./eip-5792.md#wallet_sendcalls) method MUST include an additional parameter named `assetsRequiredForExecution` in the call. 
+
+This parameter is necessary because transaction call data alone does not reliably indicate which assets and amounts are needed for successful execution. Calls may involve complex logic or multiple contracts, making it difficult for wallets to infer requirements. By specifying required assets and amounts explicitly, apps ensure wallets have the information needed to provision, bridge, or swap assets as necessary for the transaction to succeed.
+
+### `wallet_sendCalls` Extended Parameter
+
+The `assetsRequiredForExecution` parameter MUST be specified per call in the `calls` array. Each call object MAY include an `assetsRequiredForExecution` array, which details the assets and amounts required for that specific call.
+
+```typescript
+// Example call type extension
+type WalletSendCall = {
+  to: `0x${string}`;
+  data: string;
+  // ... other call fields ...
+  assetsRequiredForExecution?: Array<{
+    asset: `0x${string}`; // Asset address (use 0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE for native asset)
+    amount: string;        // Amount required, as a string representing the integer value in the asset's smallest unit
+    decimals: number;      // Number of decimals for the asset (to interpret amount correctly)
+  }>;
+};
+```
+
+#### Example Scenario: Cross-Chain Deposit to Aave
+
+Suppose a user wants to deposit DAI into Aave on Chain X, but their wallet address on Chain X has no DAI or native gas token. However, the wallet has access to funds on Chain Y (e.g., Ethereum mainnet). The app, upon detecting the `auxiliaryFunds` capability, can construct a `wallet_sendCalls` request as follows:
+
+```json
+{
+  "calls": [
+    {
+      "to": "0xAaveDAIDepositContractOnChainX",
+      "data": "0xd0e30db0", // Example encoded deposit function call
+      "assetsRequiredForExecution": [
+        {
+          "asset": "0x6B175474E89094C44Da98b954EedeAC495271d0F", // DAI address on Chain X
+          "amount": "1000000000000000000", // 1 DAI
+          "decimals": 18
+        },
+        {
+          "asset": "0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE", // Native asset for gas (e.g., ETH)
+          "amount": "10000000000000000", // 0.01 ETH
+          "decimals": 18
+        }
+      ]
+    }
+  ]
+}
+```
+
+The wallet, upon receiving this request, can:
+- Bridge or swap the required DAI from Chain Y to Chain X
+- Ensure the user has enough native asset (e.g., ETH) for gas on Chain X
+- Complete the deposit call to Aave on Chain X
+
+This mechanism allows advanced wallets to abstract away the complexity of cross-chain or offchain funding, enabling seamless user experiences even when the user's onchain balance is insufficient on the target chain.
+
+### App Implementation Guidance
 
 When an app sees that a connected wallet has access to auxiliary funds via the `auxiliaryFunds` capability in a `wallet_getCapabilities` response, the app SHOULD NOT block users from taking actions on the basis of asset balance checks.
+
+Apps SHOULD include the `assetsRequiredForExecution` parameter per call whenever they rely on the `auxiliaryFunds` capability to fulfill a user action. This ensures the wallet has the necessary information to provision assets as needed for successful execution.
 
 ## Rationale
 
@@ -86,9 +146,25 @@ An alternative we considered is defining a way for apps to fetch available auxil
 
 The shape of this capability allows for a more advanced extension if apps feel more functionality is needed.
 
+### Extensibility
+- The structure of `assetsRequiredForExecution` allows for future extension, such as specifying deadlines, slippage, or preferred bridges. Future versions of this standard MAY introduce additional fields as needed.
+
+### Decimals Source of Truth
+- Applications and wallets SHOULD verify the `decimals` value for each asset against onchain data or trusted sources to avoid errors from incorrect values.
+
+### Gas Estimation and Native Asset Handling
+- Applications SHOULD include an estimate for the required native asset (e.g., ETH for gas) in `assetsRequiredForExecution` when possible.
+
+## Backwards Compatibility
+- Applications SHOULD only include the `assetsRequiredForExecution` parameter if the wallet advertises the `auxiliaryFunds` capability.
+- If a wallet does not recognize or support the `assetsRequiredForExecution` parameter, it SHOULD ignore the parameter or return a clear error indicating lack of support. Apps SHOULD be prepared to handle such cases gracefully, potentially falling back to legacy behavior.
+
 ## Security Considerations
 
-Apps MUST NOT make any assumptions about the source of auxiliary funds. Apps' smart contracts should still, as they would today, make appropriate balance checks onchain when processing a transaction.
+- Apps MUST NOT make any assumptions about the source of auxiliary funds. Apps' smart contracts should still, as they would today, make appropriate balance checks onchain when processing a transaction.
+- Applications MUST NOT assume that the wallet will always be able to fulfill the asset requirements, and SHOULD handle failures gracefully.
+- Wallets SHOULD provide clear error messages or status codes if they are unable to fulfill the asset requirements (e.g., insufficient offchain funds, bridge/swap failure, or unsupported assets).
+- Consider whether specifying asset requirements per call could leak sensitive information about user intent or holdings. Where appropriate, apps and wallets SHOULD follow best practices to minimize privacy risks.
 
 ## Copyright
 
--- a/ERCS/erc-7682.md
+++ b/ERCS/erc-7682.md
@@ -98,10 +98,11 @@ type WalletSendCallsRequest = {
   }>;
   capabilities?: {
     auxiliaryFunds?: {
-      assetsRequiredForExecution?: Array<{
-        asset: `0x${string}`; // Asset address (use 0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE for native asset)
-        amount: `0x${string}`; // Amount required, as a hex string representing the integer value in the asset's smallest unit
-      }>;
+      assetsRequiredForExecution?: {
+        [asset: `0x${string}`]: {
+          amount: `0x${string}`; // Amount required, as a hex string representing the integer value in the asset's smallest unit
+        };
+      };
     };
   };
 };
@@ -121,12 +122,11 @@ Suppose a user wants to deposit DAI into Aave on Chain X, but their wallet addre
   ],
   "capabilities": {
     "auxiliaryFunds": {
-      "assetsRequiredForExecution": [
-        {
-          "asset": "0x6B175474E89094C44Da98b954EedeAC495271d0F", // DAI address on Chain X
+      "assetsRequiredForExecution": {
+        "0x6B175474E89094C44Da98b954EedeAC495271d0F": { // DAI address on Chain X
           "amount": "0x0de0b6b3a7640000" // 1 DAI
-        },
-      ]
+        }
+      }
     }
   }
 }
--- a/ERCS/erc-7682.md
+++ b/ERCS/erc-7682.md
@@ -221,8 +221,6 @@ The shape of this capability allows for a more advanced extension if apps feel m
 
 ## Backwards Compatibility
 - Applications SHOULD only include the `assetsRequiredForExecution` parameter if the wallet advertises the `auxiliaryFunds` capability.
-- If a wallet does not recognize or support the `assetsRequiredForExecution` parameter, it SHOULD ignore the parameter or return a clear error indicating lack of support. Apps SHOULD be prepared to handle such cases gracefully, potentially falling back to legacy behavior.
-
 
 ## Security Considerations
 
--- a/ERCS/erc-7682.md
+++ b/ERCS/erc-7682.md
@@ -43,7 +43,7 @@ In this specification, a chain's native asset (e.g. Ether on Ethereum) MUST be r
 ```typescript
 type AuxiliaryFundsCapability = {
   supported: boolean;
-  assets?: `0x${string}`[];
+  assets?: `0x${string}`[`];
 }
 ```
 
@@ -99,7 +99,7 @@ type WalletSendCallsRequest = {
   capabilities?: {
     auxiliaryFunds?: {
       assetsRequiredForExecution?: {
-        address: `0x${string}`; // Asset contract address (MUST be checksummed)
+        address: `0x${string}`;
         amount: `0x${string}`; // Amount required, as a hex string representing the integer value in the asset's smallest unit
         standard: 'erc20' | 'erc721' | 'erc1155'; // Token standard
         tokenId?: `0x${string}`; // Token ID as a hex string (required for ERC-721 and ERC-1155)
--- a/ERCS/erc-7682.md
+++ b/ERCS/erc-7682.md
@@ -125,21 +125,21 @@ type WalletSendCallsRequest = {
 
 The following type definitions specify the structure for each supported token standard:
 
-#### ERC20
+#### ERC-20
 ```typescript
 type ERC20Asset = {
   amount: `0x${string}`; // Amount required, as a hex string representing the integer value in the asset's smallest unit
 };
 ```
 
-#### ERC721
+#### ERC-721
 ```typescript
 type ERC721Asset = {
   tokenId: `0x${string}`; // Token ID as a hex string
 };
 ```
 
-#### ERC1155
+#### ERC-1155
 ```typescript
 type ERC1155Asset = {
   tokenId: `0x${string}`; // Token ID as a hex string
--- a/ERCS/erc-7682.md
+++ b/ERCS/erc-7682.md
@@ -80,7 +80,7 @@ This parameter is necessary because transaction call data alone does not reliabl
 
 ### `wallet_sendCalls` Extended Parameter
 
-The `assetsRequiredForExecution` parameter MUST be specified within the `capabilities.auxiliaryFunds` object of the `wallet_sendCalls` request. This allows the application to specify which assets and amounts are required for the successful execution of the requested calls.
+If included, the `assetsRequiredForExecution` parameter MUST be specified within the `capabilities.auxiliaryFunds` object of the `wallet_sendCalls` request.
 
 > **Note**: The `capabilities` object is specified in [EIP-5792](./eip-5792.md) where `wallet_sendCalls` is first defined: "The capabilities field is how an app can communicate with a wallet about capabilities a wallet supports. For example, this is where an app can specify a paymaster service URL from which an [ERC-4337](./eip-4337.md) wallet can request a paymasterAndData input for a user operation."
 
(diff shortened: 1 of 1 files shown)


ERCS/erc-7682.md
```diff
@@ -72,10 +72,70 @@ type AuxiliaryFundsCapability = {
...
-### App Implementation
+### Extended Usage: `assetsRequiredForExecution` Parameter
+
+When a wallet indicates support for the `auxiliaryFunds` capability (i.e., `supported: true`), applications that use the [`wallet_sendCalls`](./eip-5792.md#wallet_sendcalls) method MUST include an additional parameter named `assetsRequiredForExecution` in the call. 
+
+This parameter is necessary because transaction call data alone does not reliably indicate which assets and amounts are needed for successful execution. Calls may involve complex logic or multiple contracts, making it difficult for wallets to infer requirements. By specifying required assets and amounts explicitly, apps ensure wallets have the information needed to provision, bridge, or swap assets as necessary for the transaction to succeed.
+
+### `wallet_sendCalls` Extended Parameter
+
+The `assetsRequiredForExecution` parameter MUST be specified per call in the `calls` array. Each call object MAY include an `assetsRequiredForExecution` array, which details the assets and amounts required for that specific call.
```
jiexi: Is the first sentence required? I think you're trying to say that you can't have a `assetsRequiredForExecution` param at the root level of the wallet_sendCalls params, but I don't feel like it's necessary to say that since it doesn't look like we ever imply it is a possibility?

ERCS/erc-7682.md
```diff
@@ -72,10 +72,70 @@ type AuxiliaryFundsCapability = {
...
+```typescript
+// Example call type extension
+type WalletSendCall = {
+  to: `0x${string}`;
+  data: string;
+  // ... other call fields ...
+  assetsRequiredForExecution?: Array<{
+    asset: `0x${string}`; // Asset address (use 0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE for native asset)
+    amount: string;        // Amount required, as a string representing the integer value in the asset's smallest unit
+    decimals: number;      // Number of decimals for the asset (to interpret amount correctly)
```
jiexi: what happens if the `decimals` value provided by the dapp doesn't match what is on chain for the asset contract? would it make more sense to have the wallet look up the decimals value itself?

ERCS/erc-7682.md
```diff
@@ -72,10 +72,70 @@ type AuxiliaryFundsCapability = {
...
+        {
+          "asset": "0x6B175474E89094C44Da98b954EedeAC495271d0F", // DAI address on Chain X
+          "amount": "1000000000000000000", // 1 DAI
+          "decimals": 18
+        },
+        {
+          "asset": "0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE", // Native asset for gas (e.g., ETH)
+          "amount": "10000000000000000", // 0.01 ETH
+          "decimals": 18
+        }
```
jiexi: Hmm... just realized isn't this kind of the same thing as eth_sendTransaction's gas limit and max gas fee? Does wallet_sendCalls not have a way to specify expected gas required per call?

ERCS/erc-7682.md
```diff
@@ -86,9 +146,25 @@ An alternative we considered is defining a way for apps to fetch available auxil
 
 The shape of this capability allows for a more advanced extension if apps feel more functionality is needed.
 
+### Extensibility
+- The structure of `assetsRequiredForExecution` allows for future extension, such as specifying deadlines, slippage, or preferred bridges. Future versions of this standard MAY introduce additional fields as needed.
+
+### Decimals Source of Truth
+- Applications and wallets SHOULD verify the `decimals` value for each asset against onchain data or trusted sources to avoid errors from incorrect values.
```
jiexi: in this case, why have the dapp provide `decimals` at all? to make sure that the value they're trying to request via the `amount` field is what they intend?

ERCS/erc-7682.md
adonesky1: fair point, first sentence is redundant!

ERCS/erc-7682.md
adonesky1: Yeah I suppose this may be the right way. I suppose if we were going to add values that could otherwise be queried on the contract we'd maybe want to add the name/ticker. If we want remove it we'll just want to make sure its clear that the amount must be interpreted against the decimals that must be queried for.

adonesky1: This is making me wonder if we need to specify the standard (e.g. ERC20) for non-native assets, if only to make this extensible in the future

ERCS/erc-7682.md
```diff
@@ -72,10 +72,70 @@ type AuxiliaryFundsCapability = {
...
+- Ensure the user has enough native asset (e.g., ETH) for gas on Chain X
+- Complete the deposit call to Aave on Chain X
+
+This mechanism allows advanced wallets to abstract away the complexity of cross-chain or offchain funding, enabling seamless user experiences even when the user's onchain balance is insufficient on the target chain.
+
+### App Implementation Guidance
 
 When an app sees that a connected wallet has access to auxiliary funds via the `auxiliaryFunds` capability in a `wallet_getCapabilities` response, the app SHOULD NOT block users from taking actions on the basis of asset balance checks.
 
+Apps SHOULD include the `assetsRequiredForExecution` parameter per call whenever they rely on the `auxiliaryFunds` capability to fulfill a user action. This ensures the wallet has the necessary information to provision assets as needed for successful execution.
```
ffmcgee725:
```suggestion
Apps MUST include the `assetsRequiredForExecution` parameter per call whenever they rely on the `auxiliaryFunds` capability to fulfill a user action. This ensures the wallet has the necessary information to provision assets as needed for successful execution.
```

We previously state

> "When a wallet indicates support for the `auxiliaryFunds` capability, applications that use the [`wallet_sendCalls`] method MUST include an additional parameter named `assetsRequiredForExecution` in the call."

So should this sentence instead be

> "Apps _MUST_ include the `assetsRequiredForExecution` parameter per call whenever they rely on the `auxiliaryFunds` capability to fulfill a user action. This ensures the wallet has the necessary information to provision assets as needed for successful execution."

instead of "Apps _SHOULD_ (....)" ?

ERCS/erc-7682.md
ffmcgee725: And if we end up removing `decimals` from the required params to be provided by the dapp, we should update this to

```suggestion
- Applications and wallets MUST verify the `decimals` value for each asset against onchain data or trusted sources to avoid errors from incorrect values.
```

ERCS/erc-7682.md
adonesky1: Yeah I'm curious others thoughts on whether we should encourage a separate property where the dapp can atleast suggest what it believes the gas requirements will be

ERCS/erc-7682.md
adonesky1: Its a good question. I think there may be cases where a) fetching this data is not possible or b) slows down execution. Additionally yes it gives us some sense as to whether the amount value can be trusted. The amount should always be expressed according to what the caller believes is the number of decimals for the asset, so if the caller is incorrect about the number of decimals the asset has and then we go and try to interpret it by the number of decimals based on whats on chain then we will get that wrong. So if decimals is incorrect in our check that lets us throw an error and eject

ERCS/erc-7682.md
```diff
@@ -72,10 +72,87 @@ type AuxiliaryFundsCapability = {
 }
 ```
 
-### App Implementation
+### Extended Usage: `assetsRequiredForExecution` Parameter
+
+When a wallet indicates support for the `auxiliaryFunds` capability (i.e., `supported: true`), applications that use the [`wallet_sendCalls`](./eip-5792.md#wallet_sendcalls) method MUST include an additional parameter named `assetsRequiredForExecution` in the call.
```
matthewwalsh0: Does this not put a large burden on dApps or wallets that just want to disable additional balance checks?

I assumed this mechanism would be optional as wallets may simply want to support auxillary funds for a strict set of transactions that they can decode?

ERCS/erc-7682.md
```diff
@@ -72,10 +72,87 @@ type AuxiliaryFundsCapability = {
...
+  chainId: `0x{string}`;
+  atomicRequired: boolean;
+  calls: Array<{
+    to: `0x${string}`;
+    data: string;
+    // ... other call fields ...
+  }>;
+  capabilities?: {
+    auxiliaryFunds?: {
+      assetsRequiredForExecution?: Array<{
```
matthewwalsh0: Would an `object` keyed by the address avoid the risk of duplicate assets and be more explicit?

Since the assets should be unique?

ERCS/erc-7682.md
```diff
@@ -72,10 +72,87 @@ type AuxiliaryFundsCapability = {
...
+        asset: `0x${string}`; // Asset address (use 0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE for native asset)
+        amount: `0x${string}`; // Amount required, as a hex string representing the integer value in the asset's smallest unit
+        decimals: number;      // Number of decimals for the asset (to interpret amount correctly)
```
matthewwalsh0: Should this be optional as it wouldn't apply to all asset types, and may be derivable by the wallet?

Or should we assume that if the wallet can provision the asset, it has the means to detect metadata such as decimals?

ERCS/erc-7682.md
```diff
@@ -72,10 +72,87 @@ type AuxiliaryFundsCapability = {
...
+This mechanism allows advanced wallets to abstract away the complexity of cross-chain or offchain funding, enabling seamless user experiences even when the user's onchain balance is insufficient on the target chain.
+
+### App Implementation Guidance
 
 When an app sees that a connected wallet has access to auxiliary funds via the `auxiliaryFunds` capability in a `wallet_getCapabilities` response, the app SHOULD NOT block users from taking actions on the basis of asset balance checks.
 
+Apps MUST include the `assetsRequiredForExecution` parameter in the `capabilities.auxiliaryFunds` object whenever they rely on the `auxiliaryFunds` capability to fulfill a user action. This ensures the wallet has the necessary information to provision assets as needed for successful execution.
+
+### Decimals Source of Truth
+- Applications and wallets SHOULD verify the `decimals` value for each asset against onchain data or trusted sources to avoid errors from incorrect values.
```
matthewwalsh0: If there is still a burden on the wallet to verify this, why have the dApp provide it at all?

ERCS/erc-7682.md
```diff
@@ -86,9 +163,16 @@ An alternative we considered is defining a way for apps to fetch available auxil
...
+- Applications SHOULD only include the `assetsRequiredForExecution` parameter if the wallet advertises the `auxiliaryFunds` capability.
+- If a wallet does not recognize or support the `assetsRequiredForExecution` parameter, it SHOULD ignore the parameter or return a clear error indicating lack of support. Apps SHOULD be prepared to handle such cases gracefully, potentially falling back to legacy behavior.
+
+
 ## Security Considerations
 
-Apps MUST NOT make any assumptions about the source of auxiliary funds. Apps' smart contracts should still, as they would today, make appropriate balance checks onchain when processing a transaction.
+- Apps MUST NOT make any assumptions about the source of auxiliary funds. Apps' smart contracts should still, as they would today, make appropriate balance checks onchain when processing a transaction.
+- Applications MUST NOT assume that the wallet will always be able to fulfill the asset requirements, and SHOULD handle failures gracefully.
+- Wallets SHOULD provide clear error messages or status codes if they are unable to fulfill the asset requirements (e.g. bridge/swap failure, or unsupported assets).
```
matthewwalsh0: If we add this, are we obligated to define specific error codes?

ERCS/erc-7682.md
```diff
@@ -72,10 +72,87 @@ type AuxiliaryFundsCapability = {
...
+          "decimals": 18
+        },
+      ]
+    }
+  }
+}
+```
+
+The wallet, upon receiving this request, can:
+- Bridge or swap the required DAI from Chain Y to Chain X
```
matthewwalsh0, line 146: Should we clarify that any asset retrieval done by the wallet is not subject to the `atomicRequired` field and is separate to the "call" lifecycle from the perspective of the dApp?

ERCS/erc-7682.md
```diff
@@ -72,10 +72,87 @@ type AuxiliaryFundsCapability = {
...
+  atomicRequired: boolean;
+  calls: Array<{
+    to: `0x${string}`;
+    data: string;
+    // ... other call fields ...
+  }>;
+  capabilities?: {
+    auxiliaryFunds?: {
+      assetsRequiredForExecution?: Array<{
+        asset: `0x${string}`; // Asset address (use 0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE for native asset)
```
matthewwalsh0: Should we be explicit if we expect checksummed addresses?

ERCS/erc-7682.md
matthewwalsh0: I would argue that's entirely out of scope here as we're talking about assets that cannot be determined by the wallet, where as the gas limit and gas fee in the case of `wallet_sendCalls` is entirely calculated by the wallet (pending some future capability).

ERCS/erc-7682.md
adonesky1: Good point, dapps shouldn't/can't be forced to opt in to this capability simply because the wallet says I have it available 😅

ERCS/erc-7682.md
adonesky1: Yeah I was wondering about this as well. See [this comment above](https://github.com/ethereum/ERCs/pull/1126#discussion_r2226530584), I'm wondering what role it potentially plays in ensuring that wallet and caller are on the same page about the asset in question. I realize that's a somewhat strange pattern... and perhaps its just creating an issue we assume is not present in more basic calls like `eth_sendTransaction` where we don't double check whether the caller has their metadata correct about the asset they're prompting to send.

ERCS/erc-7682.md
adonesky1: modified [here](https://github.com/ethereum/ERCs/pull/1126/commits/6966ea9fefb9fe83e7b816d34189b5a96ca98bdd):


> When a wallet indicates support for the `auxiliaryFunds` capability (i.e., `supported: true`), applications that use the [`wallet_sendCalls`](./eip-5792.md#wallet_sendcalls) method MAY include an `assetsRequiredForExecution` parameter in the `capabilities.auxiliaryFunds` object to enable the wallet to leverage this capability.

how does that sound?

ERCS/erc-7682.md
adonesky1: Yes I suppose that makes some sense 😅

ERCS/erc-7682.md
adonesky1: added an error codes section here:
https://github.com/ethereum/ERCs/pull/1126/commits/ee2074c58c54010e540b6308c3c810047301dcf6

ERCS/erc-7682.md
adonesky1: good call out: https://github.com/ethereum/ERCs/pull/1126/commits/b0575c742552589104d49859c3674771c865419e

ERCS/erc-7682.md
adonesky1: I've decided to remove the decimals property but add a section explaining that the wallet must cross reference the asset's decimals value to ensure it is interpreting values correctly:

https://github.com/ethereum/ERCs/pull/1126/commits/7fb4cf1d19c2e92a8bfbd10e8f948a858a9bca68

>The wallet MUST cross-check the `amount` value in `assetsRequiredForExecution` against the `decimal()` value specified on the `asset`'s contract to ensure that the amount is interpreted correctly according to the asset's decimal precision.

ERCS/erc-7682.md
adonesky1: I've decided to remove the decimals property but add a section explaining that the wallet must cross reference the asset's decimals value to ensure it is interpreting values correctly:

https://github.com/ethereum/ERCs/pull/1126/commits/7fb4cf1d19c2e92a8bfbd10e8f948a858a9bca68

>The wallet MUST cross-check the `amount` value in `assetsRequiredForExecution` against the `decimal()` value specified on the `asset`'s contract to ensure that the amount is interpreted correctly according to the asset's decimal precision.

ERCS/erc-7682.md
adonesky1: I've decided to remove the decimals property but add a section explaining that the wallet must cross reference the asset's decimals value to ensure it is interpreting values correctly:

https://github.com/ethereum/ERCs/pull/1126/commits/7fb4cf1d19c2e92a8bfbd10e8f948a858a9bca68

>The wallet MUST cross-check the `amount` value in `assetsRequiredForExecution` against the `decimal()` value specified on the `asset`'s contract to ensure that the amount is interpreted correctly according to the asset's decimal precision.

ERCS/erc-7682.md
adonesky1: I've decided to remove the decimals property but add a section explaining that the wallet must cross reference the asset's decimals value to ensure it is interpreting values correctly:

https://github.com/ethereum/ERCs/pull/1126/commits/7fb4cf1d19c2e92a8bfbd10e8f948a858a9bca68

>The wallet MUST cross-check the `amount` value in `assetsRequiredForExecution` against the `decimal()` value specified on the `asset`'s contract to ensure that the amount is interpreted correctly according to the asset's decimal precision.

ERCS/erc-7682.md
adonesky1: Done here: https://github.com/ethereum/ERCs/pull/1126/commits/7860c89c879250ed0ca9e00cfd797580cb943ab2

ERCS/erc-7682.md
adonesky1: I'm not sure if we should require checksummed, we don't do that for any other APIs to my knowledge, so could create inconsistencies

ERCS/erc-7682.md
```diff
@@ -72,10 +72,104 @@ type AuxiliaryFundsCapability = {
...
+
+### Error Codes
+
+The following error codes are defined for auxiliary funds provisioning failures.
+
+| Code     | Message                           | Description |
+|----------|-----------------------------------|-------------|
+| -32080   | Auxiliary funds provisioning failed | Wallet attempted to provision funds but failed (e.g., bridge/swap error). |
+| -32081   | Asset not supported             | The requested asset is not available through the wallet’s auxiliary fund system. |
+| -32082   | Auxiliary funds timeout         | The provisioning operation took too long or timed out. |
```
jiexi: does the dapp need to differentiate between -32080 and -32082?

ERCS/erc-7682.md
```diff
@@ -72,10 +72,104 @@ type AuxiliaryFundsCapability = {
...
 When an app sees that a connected wallet has access to auxiliary funds via the `auxiliaryFunds` capability in a `wallet_getCapabilities` response, the app SHOULD NOT block users from taking actions on the basis of asset balance checks.
 
+Apps MAY include the `assetsRequiredForExecution` parameter in the `capabilities.auxiliaryFunds` object to ensure the wallet has the necessary information to provision assets as needed for successful execution.
+
+### Amount Validation
+
+The wallet MUST cross-check the `amount` value in `assetsRequiredForExecution` against the `decimal()` value specified on the `asset`'s contract to ensure that the amount is interpreted correctly according to the asset's decimal precision.
+
+### Gas Estimation and Native Asset Handling
+- Applications SHOULD include an estimate for the required native asset (e.g., ETH for gas) in `assetsRequiredForExecution` when possible.
```
jiexi: is this necessary if we've already determined the wallet capable of estimating the gas fee correctly without dapp input?

adonesky1: oh whoops I meant to remove this!

adonesky1: https://github.com/ethereum/ERCs/pull/1126/commits/dd3a05e0af065da03d7ce5a2819dce45c53d429f

ERCS/erc-7682.md
adonesky1: no probably not! Cleaned up here: https://github.com/ethereum/ERCs/pull/1126/commits/dd3a05e0af065da03d7ce5a2819dce45c53d429f

ERCS/erc-7682.md
```diff
@@ -86,9 +176,16 @@ An alternative we considered is defining a way for apps to fetch available auxil
 
 The shape of this capability allows for a more advanced extension if apps feel more functionality is needed.
 
+## Backwards Compatibility
+- Applications SHOULD only include the `assetsRequiredForExecution` parameter if the wallet advertises the `auxiliaryFunds` capability.
+- If a wallet does not recognize or support the `assetsRequiredForExecution` parameter, it SHOULD ignore the parameter or return a clear error indicating lack of support. Apps SHOULD be prepared to handle such cases gracefully, potentially falling back to legacy behavior.
```
cryptotavares: I believe this behaviour is covered in the 5792 spec. `wallet_sendCalls` `capabilities` [rpc specification](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-5792.md#wallet_sendcalls-rpc-specification) includes an optional property:
```
type Capability = {
  [key: string]: unknown;
  optional?: boolean;
}

type SendCallsParams = {
  version: string;
  id?: string;
  from?: `0x${string}`;
  chainId: `0x${string}`; // Hex chain id
  atomicRequired: boolean;
  calls: {
    to?: `0x${string}`;
    data?: `0x${string}`;
    value?: `0x${string}`; // Hex value
    capabilities?: Record<string, Capability>;
  }[];
  capabilities?: Record<string, Capability>;
};
```

And there's the error code [`5700`](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-5792.md#error-codes) for `Unsupported non-optional capability`. This should be enough, as wallets that do not support `auxiliaryFunds` capability, will either need to reply with that error, or ignore the capability if a dapp adds the `optional` property in the `wallet_sendCalls` request.

ERCS/erc-7682.md
```diff
@@ -72,10 +72,100 @@ type AuxiliaryFundsCapability = {
...
+### Error Codes
+
+The following error codes are defined for auxiliary funds provisioning failures.
+
+| Code     | Message                           | Description |
+|----------|-----------------------------------|-------------|
+| -32080   | Auxiliary funds provisioning failed | Wallet attempted to provision funds but failed (e.g., bridge/swap error). |
+| -32081   | Asset not supported             | The requested asset is not available through the wallet’s auxiliary fund system. |
+| -32082   | Auxiliary funds capability not available | The wallet no longer supports auxiliary funds on the requested chain. |
+| -32083   | Invalid assetsRequiredForExecution | The structure of the assetsRequiredForExecution object is malformed. |
```
cryptotavares: I am wondering if the error codes should follow the pattern used in the [5792 spec](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-5792.md#error-codes), instead of the json-rpc 2.0 spec

vandan: Yeah, we had something like that in [a draft MIP](https://github.com/MetaMask/metamask-improvement-proposals/pull/63/files) that corresponded to these changes. Maybe 5800 range to ensure that the errors remain distinct from the 5792 wallet_sendCalls errors.

ERCS/erc-7682.md
```diff
@@ -72,10 +72,100 @@ type AuxiliaryFundsCapability = {
...
+
+### App Implementation Guidance
 
 When an app sees that a connected wallet has access to auxiliary funds via the `auxiliaryFunds` capability in a `wallet_getCapabilities` response, the app SHOULD NOT block users from taking actions on the basis of asset balance checks.
 
+Apps MAY include the `assetsRequiredForExecution` parameter in the `capabilities.auxiliaryFunds` object to ensure the wallet has the necessary information to provision assets as needed for successful execution.
+
+### Amount Validation
+
+The wallet MUST cross-check the `amount` value in `assetsRequiredForExecution` against the `decimal()` value specified on the `asset`'s contract to ensure that the amount is interpreted correctly according to the asset's decimal precision.
```
bumblefudge: this is a hard requirement to test... or rather, it's easy to check that a wallet is "checking" but hard to check what the correct interpretation is? it might be easier just to make this a SHOULD unless you're going to get more specific about what interpretation means here.

bumblefudge: maybe instead of mandating a certain behavior where the decimal value has been misinterpreted, it would make more sense to have a general-purpose fallback recommendation, like, "where the assetsRequiredForExecution is in any way malformed, such as malformed relationship between asset's decimal() and stipulated value, wallets MAY/SHOULD..."

ERCS/erc-7682.md
adonesky1: Removed here: https://github.com/ethereum/ERCs/pull/1126/commits/7e431dc85530f5b505eb2632d3f32da5a539936e

ERCS/erc-7682.md
adonesky1: Happy to migrate them to that number range!

ERCS/erc-7682.md
adonesky1: > this is a hard requirement to test... or rather, it's easy to check that a wallet is "checking" but hard to check what the correct interpretation is? it might be easier just to make this a SHOULD unless you're going to get more specific about what interpretation means here.

Maybe the way I phrased it is confusing. What I'm trying to say here is that the value amount that is specified by the dapp can only be correctly interpreted by the wallet (for representing to the user) if the wallet reads the decimal value of the contract in question. I had previously included decimals as a property in the `assetsRequiredForExecution` object so that the wallet wouldn't even need to check the contract to know how to interpret the amount, but after [some back and forth on this](https://github.com/ethereum/ERCs/pull/1126#discussion_r2220367032), I decided [to remove it](https://github.com/ethereum/ERCs/pull/1126/commits/7fb4cf1d19c2e92a8bfbd10e8f948a858a9bca68) and add this clause. Does my explanation here make it clearer? If so how do you think I should modify this clause to make it clearer?

ERCS/erc-7682.md
adonesky1: done here: https://github.com/ethereum/ERCs/pull/1126/commits/c52d9b23375dc5557c84af09c7cebcfa2035fe1c

ERCS/erc-7682.md
```diff
@@ -99,15 +99,56 @@ type WalletSendCallsRequest = {
...
-        [asset: `0x${string}`]: {
-          amount: `0x${string}`; // Amount required, as a hex string representing the integer value in the asset's smallest unit
+        erc20?: {
+          [asset: `0x${string}`]: {
+            amount: `0x${string}`; // Amount required, as a hex string representing the integer value in the asset's smallest unit
+          };
+        };
+        erc721?: {
+          [asset: `0x${string}`]: {
+            tokenId: `0x${string}`; // Token ID as a hex string
```
matthewwalsh0: Does this need to be an array assuming it could require multiple NFTs from the same contract?

Same for ERC-1155?

ERCS/erc-7682.md
```diff
@@ -99,15 +99,56 @@ type WalletSendCallsRequest = {
...
+          };
+        };
+        erc1155?: {
```
matthewwalsh0: Just confirming that we've gone with this standard as key approach to make it explicit that `erc20` doesn't support `tokenId`?

The alternative could be an array of shared schema such as:

```
assetsRequiredForExecution?: {
    asset: `0x${string}`;
    amount: `0x${string}`;
    standard: 'erc${string}`;
    tokenId?: `0x${string}`;
}[]
```

This has the advantage of simplicity, and meaning we wouldn't necessarily have to define specific standards in this specification, meaning it is decoupled and the wallet could support future standards? Assuming we are at least specific that the standard is the ERC number in the form `ercX` etc.

One downside is the risk of duplicate assets that the wallet would have to aggregate or validate.

adonesky1: Yeah this was closer to the format I'd originally proposed. I agree its probably cleaner than building in certain specification for the 3 primary token standards standards we anticipate. The main issue I see with it is that we aren't fully avoiding the possible need for extensions in the future since the properties we're assuming here (`amount` and `tokenId`) are the superset of the required properties for these specific standards. Any new standards with additionally required properties would need to propose an extension

ERCS/erc-7682.md
adonesky1: done here: https://github.com/ethereum/ERCs/pull/1126/commits/e7e4fda0c5bee9e291e908eb267d91024890c5aa

ERCS/erc-7682.md
adonesky1: Reverted to your suggested format: https://github.com/ethereum/ERCs/pull/1126/commits/55e7241c3fa69911ccd91b4d8642f542869dd9e9

lukasrosario:
@adonesky1 dont think i have a super strong stance one way or the other currently, but just curious if you considered this as a separate capability? i actually think i agree with the overall direction here, but my first reaction was that this deserves its own capability.

ERCS/erc-7682.md
```diff
@@ -72,10 +72,118 @@ type AuxiliaryFundsCapability = {
...
+  atomicRequired: boolean;
+  calls: Array<{
+    to: `0x${string}`;
+    data: string;
+    // ... other call fields ...
+  }>;
+  capabilities?: {
+    auxiliaryFunds?: {
+      assetsRequiredForExecution?: {
+        address: `0x${string}`; // Asset contract address (MUST be checksummed)
```
lukasrosario: i dont think we should include this constraint -- this is not the case anywhere else in 5792

ERCS/erc-7682.md
```diff
@@ -72,10 +72,118 @@ type AuxiliaryFundsCapability = {
 }
 ```
 
-### App Implementation
+### Extended Usage: `assetsRequiredForExecution` Parameter
```
lukasrosario: nit on naming and maybe worth considering with the idea around a separate capability but `required` feels a little strong given what's provided could be (intentionally or unintentionally) incorrect. maybe more like `fundingSuggestion` or `fundingTip`

ERCS/erc-7682.md
lukasrosario: ^ kind of another argument for a separate capability. as a wallet, just expressing support for `auxiliaryFunds` is not expressive enough for an app to know if its `assetsRequiredForExecution` suggestion will be respected. at the very least, i think we need more than just `auxiliaryFunds: {supported: true}` in response to `wallet_getCapabilities`

ERCS/erc-7682.md
```diff
@@ -72,10 +72,118 @@ type AuxiliaryFundsCapability = {
...
+        tokenId?: `0x${string}`; // Token ID as a hex string (required for ERC-721 and ERC-1155)
+      }[];
+    };
+  };
+};
+```
+
+#### Example Scenario: Cross-Chain Deposit to Aave
+
+Suppose a user wants to deposit DAI into Aave on Chain X, but their wallet address on Chain X has no DAI or native gas token. However, the wallet has access to funds on Chain Y (e.g., Ethereum mainnet). The app, upon detecting the `auxiliaryFunds` capability, can construct a `wallet_sendCalls` request as follows:
```
arein, line 119: not following the premise here - `auxiliaryFunds` was introduced in order to signal offchain funds. In this example the funds are not just onchain but also on the same address just on another network. So the app, which needs to handle additional logic here, can already detect whether the wallet has enough funds - assuming the wallet is "connected".

> Calls may involve complex logic or multiple contracts, making it difficult for wallets to infer requirements. By specifying required assets and amounts explicitly, apps ensure wallets have the information needed to provision, bridge, or swap assets as necessary for the transaction to succeed.

Is what you state in the motivation - can you give some more specific examples about that - then it's maybe clearer why it's needed.

Also: Maybe also reference https://eips.ethereum.org/EIPS/eip-7811 as it has somewhat similar intentions.

ERCS/erc-7682.md
```diff
@@ -72,10 +72,118 @@ type AuxiliaryFundsCapability = {
...
+{
+  "calls": [
+    {
+      "to": "0xAaveDAIDepositContractOnChainX",
+      "data": "0xd0e30db0" // Example encoded deposit function call
+    }
+  ],
+  "capabilities": {
+    "auxiliaryFunds": {
+      "assetsRequiredForExecution": [
```
lukaisailovic: Why does the dapp need to provide this? Wallet can just simulate the request

adonesky1: Simulation alone is not sufficient for a wallet to reliably determine which assets are required. If the user lacks the necessary balances or approvals on the target chain, the simulated call will revert before any token transfers occur, so no asset flow data is produced. Even with successful simulation, a wallet must know which contracts and methods to monitor; without full ABI knowledge of every contract involved, it cannot consistently decode the intended transfers. Some advanced engines allow “fake balance injection,” but that requires knowing which tokens and amounts to inject in the first place — the exact information the wallet is missing. The assetsRequiredForExecution parameter solves this chicken‑and‑egg problem by letting the dapp explicitly declare the required assets, so the wallet can provision them (via bridging, swapping, or funding) before executing the call.

adonesky1:
> @adonesky1 dont think i have a super strong stance one way or the other currently, but just curious if you considered this as a separate capability? i actually think i agree with the overall direction here, but my first reaction was that this deserves its own capability.


So ATM I see `assetsRequiredForExecution` as an essential extension of `auxiliaryFunds` rather than a standalone capability. auxiliaryFunds signals what the wallet can do, while assetsRequiredForExecution provides the context it needs from the dapp in order to actually do it. Without this, the capability becomes much more constrained in practice.

For example, imagine a user wants to make a DAI deposit into Aave. The wallet has auxiliary funds it could provision, but at the moment holds no DAI on that chain (it doesn't really matter whether the assets to be provisioned are coming from off chain to be ramped just in time, or bridged or swapped or whatever). The calldata doesn’t reliably reveal which token or how much is being deposited without the wallet knowing the ABI of the contract beforehand, and simulations will revert without revealing asset data if the required assets are not there to successfully simulate. Without the dapp explicitly specifying the required assets, the wallet can’t act — even though it technically has the auxiliaryFunds capability. How do you see wallets meaningfully leveraging auxiliaryFunds in cases like this without assetsRequiredForExecution?

@lukasrosario

ERCS/erc-7682.md
adonesky1: Agreed, I meant to remove this 😅

ERCS/erc-7682.md
adonesky1: see my [longer comment](https://github.com/ethereum/ERCs/pull/1126#issuecomment-3148540794) below on this. I think that `required` is actually the right language for what I think is needed here

ERCS/erc-7682.md
adonesky1: > not following the premise here - auxiliaryFunds was introduced in order to signal offchain funds. In this example the funds are not just onchain but also on the same address just on another network. So the app, which needs to handle additional logic here, can already detect whether the wallet has enough funds - assuming the wallet is "connected".

So I think `auxIliaryFunds` ought to be considered to broadly indicate that funds can be provisioned from other sources whether they be offchain, other addresses in the users wallet, or from other chains from this same address. Why in your opinion should the `auxiliaryFunds` capability be limited to offchain funds only?

Essentially we want to provide a capability where the dapp shouldn't care about where the funds are or where they come from and they are providing the bare minimum information required by the wallet to ensure that its able to execute the transaction by gathering the assets it needs for the transaction being proposed to succeed. Normally right now dapps look at your assets on the particular chain the transaction is going to take place and don't even enable the flows for users to initiate transactions which they don't believe can be completed given the users onchain holdings for that chain. We want to say: "we can get funds from other sources, don't worry about where they're coming from, just tell us what we need for this transaction to succeed and we will provision those funds so the transaction can be executed."

ERCS/erc-7682.md
```diff
@@ -72,10 +72,118 @@ type AuxiliaryFundsCapability = {
...
+    data: string;
+    // ... other call fields ...
+  }>;
+  capabilities?: {
+    auxiliaryFunds?: {
+      assetsRequiredForExecution?: {
+        address: `0x${string}`; // Asset contract address (MUST be checksummed)
+        amount: `0x${string}`; // Amount required, as a hex string representing the integer value in the asset's smallest unit
+        standard: 'erc20' | 'erc721' | 'erc1155'; // Token standard
+        tokenId?: `0x${string}`; // Token ID as a hex string (required for ERC-721 and ERC-1155)
```
onbjerg: does `amount` make sense for erc721? what would you fill it with given you have to specify `tokenId`?

ERCS/erc-7682.md
```diff
@@ -72,10 +72,118 @@ type AuxiliaryFundsCapability = {
...
+    to: `0x${string}`;
+    data: string;
+    // ... other call fields ...
+  }>;
+  capabilities?: {
+    auxiliaryFunds?: {
+      assetsRequiredForExecution?: {
+        address: `0x${string}`; // Asset contract address (MUST be checksummed)
+        amount: `0x${string}`; // Amount required, as a hex string representing the integer value in the asset's smallest unit
+        standard: 'erc20' | 'erc721' | 'erc1155'; // Token standard
```
jiexi: i know `0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE` exists, but it doesn't feel very elegant to represent native eth this way. Thoughts on adding a `'native'` standard here as well?

ERCS/erc-7682.md
adonesky1: I think amount can probably safely stay for 721, and it should just always be 1?

lukasrosario:
@adonesky1 thanks for responses, here's what i think i'd need to feel good about getting this in:
- i have one unaddressed comment above re `getCapabilities` response:
  - as a wallet, just expressing support for `auxiliaryFunds` is not expressive enough for an app to know if its `assetsRequiredForExecution` suggestion will be respected. at the very least, i think we need more than just `auxiliaryFunds: {supported: true}` in response to wallet_getCapabilities.
  - maybe something like `auxiliaryFunds: { supported: true, executionFundingSuggestionSupported: true }`?
- i still don't feel great about the name. i understand the example you laid out, but again `required` still feels like a strong word for what is really a suggestion by the application. it doesn't even have to be malicious, it's very possible that an application miscalculates what is really "required". could we change to something like `suggestion` or `tip`?

ERCS/erc-7682.md
lukasrosario: wallets can get this from `call.value` -- i would actually add something here that explicitly states wallets should just deduce this themselves @adonesky1

ERCS/erc-7682.md
```diff
@@ -72,10 +72,118 @@ type AuxiliaryFundsCapability = {
...
+  chainId: `0x{string}`;
+  atomicRequired: boolean;
+  calls: Array<{
+    to: `0x${string}`;
+    data: string;
+    // ... other call fields ...
+  }>;
+  capabilities?: {
+    auxiliaryFunds?: {
+      assetsRequiredForExecution?: {
```
jxom: Why can't this be named something simple like `requiredAssets`? These capabilities already imply that they are intended "for execution" (it's a send calls request), so the `*ForExecution` is kind of redundant.

ERCS/erc-7682.md
adonesky1: I see what you’re getting at, but I think it’s important to clarify: `auxiliaryFunds` is inherently a best‑effort capability, not a guarantee. Even if a wallet reports support, there’s no way to ensure in advance that it can provision the assets specified — that depends on factors like liquidity, bridging availability, and user approval. That’s why the spec defines error codes for provisioning failures.

Apps can always use `assetsRequiredForExecution` to make their needs explicit, but they must be prepared for cases where the wallet can’t fulfill them. Do you think requiring wallets to declare more in `wallet_getCapabilities` would add meaningful reliability here, given that the outcome is always conditional?

Or perhaps I'm misunderstanding your feedback

adonesky1:
@lukasrosario 

>i have one unaddressed comment above re getCapabilities response:
as a wallet, just expressing support for auxiliaryFunds is not expressive enough for an app to know if its assetsRequiredForExecution suggestion will be respected. at the very least, i think we need more than just auxiliaryFunds: {supported: true} in response to wallet_getCapabilities.
maybe something like auxiliaryFunds: { supported: true, executionFundingSuggestionSupported: true }?

[replied above](https://github.com/ethereum/ERCs/pull/1126#discussion_r2255182275), but copying my response here:

I see what you’re getting at, but I think it’s important to clarify: auxiliaryFunds is inherently a best‑effort capability, not a guarantee. Even if a wallet reports support, there’s no way to ensure in advance that it can provision the assets specified — that depends on factors like liquidity, bridging availability, and user approval. That’s why the spec defines error codes for provisioning failures.

Apps can always use assetsRequiredForExecution to make their needs explicit, but they must be prepared for cases where the wallet can’t fulfill them. Do you think requiring wallets to declare more in wallet_getCapabilities would add meaningful reliability here, given that the outcome is always conditional?

So `assetsRequiredForExecution` is always conditional, so maybe you're pointing out that the wallet should indicate that it can interpret this suggestion? I suppose part of what I'm suggesting is that support for `auxiliaryFunds` should imply that you are able to interpret `assetsRequiredForExecution` since I'm not convinced the capability is broadly workable without it.

> i still don't feel great about the name. i understand the example you laid out, but again required still feels like a strong word for what is really a suggestion by the application. it doesn't even have to be malicious, it's very possible that an application miscalculates what is really "required". could we change to something like suggestion or tip?

I get the concern about calling them “required,” since the wallet can’t actually verify if the assets a dApp lists are truly necessary. On the other hand, calling them just “suggestions” feels too soft, because in the "happy path" the transaction won’t succeed without them.

**edit:** the wallet can via simulation verify whether the suggested assets will result in a successful transaction

Maybe the middle ground is: the dApp marks these as assets it says are needed, and the wallet shows that list to the user so they can decide whether to provision them. That way:
- The dApp makes clear what it thinks is required.
- The wallet isn’t pretending it can verify that claim (or perhaps in cases where it can verify it does)
- The user gets the final say after seeing what’s being asked for.

ERCS/erc-7682.md
adonesky1: I'd be fine with that

adonesky1: Done here: https://github.com/ethereum/ERCs/pull/1126/commits/13bf16bad7223c59ae7c3dedf45b593c315cb0cd

ERCS/erc-7682.md
adonesky1: Removed here: https://github.com/ethereum/ERCs/pull/1126/commits/b562bad687c0eeb975668d537134858b76e95b2e

ERCS/erc-7682.md
adonesky1: good call out @lukasrosario, adding something this effect

adonesky1: https://github.com/ethereum/ERCs/pull/1126/commits/d66b6400475ebb0d001ff7584396d7e6a57605b2

ERCS/erc-7682.md
```diff
@@ -98,10 +100,11 @@ type WalletSendCallsRequest = {
   }>;
   capabilities?: {
     auxiliaryFunds?: {
+      optional?: boolean;
```
lukasrosario, line 105: maybe redundant? this is part of the capability definition. will leave it up to you but just fyi
Screenshot 2025-08-14 at 5 50 45 PM

PR #1142, opened 2025-07-30
Update ERC-7656: Remove reference to ERC-6551

sullof:
Moving ERC-7656 to Last Call. To do so, we had to remove any direct reference to ERC-6551, since the latter has not been finalized yet. We maintain the reference to it in the reference code, linking to its Github repo.

### DIFF
--- a/ERCS/erc-7786.md
+++ b/ERCS/erc-7786.md
@@ -4,7 +4,8 @@ title: Cross-Chain Messaging Gateway
 description: An interface for contracts to send and receive cross-chain messages containing arbitrary data.
 author: Francisco Giordano (@frangio), Hadrien Croubois (@Amxx), Ernesto Garcia (@ernestognw), CJ Cobb (@cjcobb23), Sergey Gorbunov (@sergeynog), joxes (@Joxess)
 discussions-to: https://ethereum-magicians.org/t/erc-7786-cross-chain-messaging-gateway/21374
-status: Draft
+status: Last Call
+last-call-deadline: 2025-08-31
 type: Standards Track
 category: ERC
 created: 2024-10-14
--- a/ERCS/erc-7955.md
+++ b/ERCS/erc-7955.md
@@ -242,7 +242,7 @@ The `CREATE2_FACTORY_RUNTIME_CODE` corresponds to the following assembly:
 # CREATE2 reverted, propagate the revert data.
                         # Stack: [address = 0; 0; 32]
 0x0013: RETURNDATASIZE  # Stack: [r.len; 0; 0; 32]              | Push the revert data length onto the stack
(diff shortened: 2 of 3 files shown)


ERCS/erc-7656.md
```diff
@@ -17,7 +18,7 @@ This proposal defines a factory capable of deploying generic services linked to
 
 ## Motivation
 
-Existing standards like [ERC-6551](./eip-6551.md) successfully bind smart accounts to NFTs, allowing registries to deploy accounts owned by specific token IDs. However, these standards have two key limitations:
+Existing proposals like TokenBound Account (numbered as 6551) successfully bind smart accounts to NFTs, allowing registries to deploy accounts owned by specific token IDs. However, these standards have two key limitations:
```
SamWilsn: Changing the way you write the reference doesn't change the fact that you're making a reference. Either remove it entirely or help get ERC-6551 across the finish line.

SamWilsn (changes requested):

ERCS/erc-7656.md
sullof: @SamWilsn I would be happy to help the erc6551 team, but it looks like they are not working on it at this moment. I suspect that it is also because of the scope of erc6551, which tries to put two proposals in a single one, making it very hard to complete it. If think that if they had focused on a first proposal for the registry and a second one for the nft-bound account, the first one would definitely be finalized.
ERC7656 is simpler and focused only on the contract deployer. After many months and many conversations with other devs, I don't expect any change.
So, I removed any reference, referring to the TokenBound project when it makes sense.
I hope it is fine.

ERCS/erc-7656.md
```diff
@@ -1,14 +1,15 @@
 ---
 eip: 7656
 title: Generalized Contract-Linked Services
-description: Define a factory for generic services linked to any contract type including ERC-4337 accounts and NFTs
+description: Define a factory for generic services linked to any contract type, including smart accounts and NFTs
 author: Francesco Sullo (@sullof)
 discussions-to: https://ethereum-magicians.org/t/variation-to-erc6551-to-deploy-any-kind-of-contract-linked-to-an-nft/19223
-status: Review
+status: Last Call
```
xinbenlv, line 7: Status change should be a separate PR. Can you split this PR into two PRs? One with all the change, and the second one with the move to Last Call?

sullof: Sure. Thanks. I will change this one for review, and make a new one with the move to last call.

sullof:
@xinbenlv I removed the status change, as you asked, leaving only the changes requested by @SamWilsn 
To simplify the job, I am closing this PR and create a new one, since there will be no need for a reviewer

PR #1144, opened 2025-07-31
Update ERC-7683: Remove order structs, extend resolved order

frangio:
This is part of an ongoing conversation about modifying aspects of ERC-7683 that implementers consider need to be improved.

### Current Problems

- Gas inefficiencies
    - Full resolved order in `Open` event
    - Redundant `originSettler`, `originChainId` in `openFor(GaslessCrossChainOrder, ...)`
    - Parameters in calldata that may not be needed by some protocols
- Unclear criteria for standardized fields vs `orderData`
- Designed primarily for escrow
    - `openDeadline` is not needed in resource locks
- Asset recovery timelines for unfilled intents (escrow refunds, resource lock forced withdrawals) have implications for solvers but are not legible in `ResolvedCrossChainOrder`
- Non-standard chain id and address encoding outside of EVM

### This Proposal

The central change is to move everything into opaque `orderData` and lean more heavily on `resolve`. This allows protocols to encode orders as efficiently as they can.

Additionally:
- Resource locks are supported by making `open` optional.
- A settlement deadline is added, after which escrowed assets may be reclaimed by the user.
- The resolved order is extensible so protocols can encode implementation-specific important information, such as external trust/security assumptions implied by the order.
- The supported filler data format is described in resolved orders.

### Pending Issues

- Some protocols are not able to give a reasonable `maxSpent` bound on output tokens based on the order from the origin settler, or at all.
- Addresses in resolved orders should be given as ERC-7930 Interoperable Addresses.

### DIFF
--- a/ERCS/erc-7683.md
+++ b/ERCS/erc-7683.md
@@ -36,9 +36,25 @@ The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL 
 - **Raw order**: an order specific to an origin settler, in an implementation-specific encoding
 - **Resolved order**: an order in a standardized representation
 
+### Filler Procedure
+
+Once a filler receives an order for an origin settler it proceeds as follows:
+
+1. Resolves the payload with the origin settler.
+2. Resolves the destination payload with the destination settler(s).
+3. Decodes and processes all extended attributes in the resolved order.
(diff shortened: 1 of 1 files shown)


frangio:
Closing while we develop a different approach.

PR #1147, opened 2025-08-04
Update ERC-7846: Tweaks

jxom:
A few minor changes:

- Adds a `chainIds` property to the request. Rationale:
  - If an Application provides a set of chain IDs it supports, then we can make `chainId` in request capabilities optional (and also dedupe on it if there are multiple capabilities that have `chainId`) and have the wallet fill in this property with one of the root `chainIds`. 
- Makes `chainId` optional on the `signInWithEthereum` capability. Rationale: 
  - As mentioned above, the wallet will be able to fill `chainId` with one of the root `chainIds`.
  - For the case of Contract Accounts, the SIWE `chainId` must be a chain that the signing Smart Contract Account resides on, so that we can ERC-1271 verify. The Application may not know this in advance because Applications do not (and should not have to) know what chains the SCA lives on. For example, an Application may be set up with JxomChain (`69420`) and Base (`8453`), and set the SIWE `chainId` to `69420` (because the app may just supply the first or active chain). However the Wallet may not support JxomChain and therefore SIWE will fail. If a set of `chainIds` are supported on the root, then the Wallet can pick a supported chain and fill the SIWE `chainId` as Base (`8453`).

### DIFF
--- a/ERCS/erc-7846.md
+++ b/ERCS/erc-7846.md
@@ -30,7 +30,7 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 
 Requests to connect account(s) with optional capabilities.
 
-- An Application MAY provide a list of chain IDs to propose to the wallet to establish an initial connection with. 
+- An Application MAY provide a list of chain IDs (`chainIds`) to propose to the wallet to establish an initial connection with. 
 - The Wallet MUST support at least one of the provided chain IDs. If the Wallet does not support any of the provided chain IDs, then the Wallet MUST reject with an `5710 Unsupported Chain` error.
 - If the Application does not provide any chain IDs, then the Wallet MAY propose any chain ID it supports.
 - If the Application requests to add or switch chain via a `wallet_addEthereumChain` or `wallet_switchEthereumChain` request at a later time, the Wallet MAY switch to the requested chain (even if not one of the chain IDs provided at connection) or reject with an `5710 Unsupported Chain` error if not supported.
--- a/ERCS/erc-7846.md
+++ b/ERCS/erc-7846.md
@@ -2,7 +2,7 @@
 eip: 7846
 title: Wallet Connection API
 description: Adds JSON-RPC method for requesting wallet connection with modular capabilities.
-author: Conner Swenberg (@ilikesymmetry), Jake Moxey (@jxom), Lukas Rosario (@lukasrosario)
+author: Conner Swenberg (@ilikesymmetry), Jake Moxey (@jxom), Lukas Rosario (@lukasrosario), Stephan Cilliers (@stephancill)
 discussions-to: https://ethereum-magicians.org/t/erc-7846-wallet-connection-api/22245
 status: Draft
 type: Standards Track
@@ -30,16 +30,22 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 
 Requests to connect account(s) with optional capabilities.
 
+- An Application MAY provide a list of chain IDs to propose to the wallet to establish a connection with. 
+- The Wallet MUST support at least one of the provided chain IDs. If the Application does not provide any chain IDs, then the Wallet MAY propose any chain ID it supports.
+
+
 ##### Request
 
 ```ts
 type Request = {
   method: 'wallet_connect',
   params: [{
-    // JSON-RPC method version.
-    version: string;
     // Optional capabilities to request (e.g. Sign In With Ethereum).
     capabilities?: Record<string, unknown>;
+    // Chain IDs to propose to the wallet to connect to.
+    chainIds?: `0x${string}`[];
+    // JSON-RPC method version.
+    version: '1';
   }]
 }
 ```
@@ -65,13 +71,13 @@ type Response = {
 const response = await provider.request({
   method: 'wallet_connect',
   params: [{
-    version: '1',
     capabilities: {
       signInWithEthereum: {
         nonce: '12345678',
-        chainId: '0x1'
       }
-    }
+    },
+    chainIds: ['0x1', '0x2105'],
+    version: '1',
   }]
 })
 /**
@@ -86,7 +92,8 @@ const response = await provider.request({
  *         }
  *       }
  *     }
- *   ]
+ *   ],
+ *   chainIds: ['0x1', '0xa', '0x2105', '0x31b51']
  * }
  */
 ```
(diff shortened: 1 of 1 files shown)


ERCS/erc-7846.md
```diff
@@ -30,16 +30,24 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 
 Requests to connect account(s) with optional capabilities.
 
+- An Application MAY provide a list of chain IDs (`chainIds`) to propose to the wallet to establish an initial connection with. 
+- The Wallet MUST support at least one of the provided chain IDs. If the Wallet does not support any of the provided chain IDs, then the Wallet MUST reject with an `5710 Unsupported Chain` error.
```
wilsoncusack: Can we pass back which chain id(s) are not supported?

ERCS/erc-7846.md
```diff
@@ -30,16 +30,24 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 
 Requests to connect account(s) with optional capabilities.
 
+- An Application MAY provide a list of chain IDs (`chainIds`) to propose to the wallet to establish an initial connection with.
```
wilsoncusack: if future requests are made without a chain ID, what is expected behavior?

ERCS/erc-7846.md
```diff
@@ -124,28 +136,29 @@ Adds support for offchain authentication using [ERC-4361](./eip-4361.md).
 Same as ERC-4361 specification with minor modifications: 
 * The casing of multi-word fields has been adjusted to camelCase instead of kebab-case. Resources are an array field. 
 * The account address returned by `wallet_connect` MUST match the address inferred in the Sign-In with Ethereum (SIWE) message.
-* `version` is optional and defaults to an accepted version defined in ERC-4361 if not provided.
-* `domain` is optional and defaults to the domain of the requesting app if not provided.
-* `uri` is optional and defaults to the uri of the requesting app if not provided.
-* `issuedAt` is optional and defaults to the current time if not provided.
+* `chainId` is optional. If not provided, the Wallet MUST fill this field with one of the chain IDs requested on the `wallet_connect` request (or any chain ID the account supports if no chain IDs were requested).
```
wilsoncusack: can we get more explicit here? First chain id maybe?

ERCS/erc-7846.md
```diff
@@ -197,13 +209,21 @@ const result = await provider.request({
...
 ```
 
 ## Rationale
 
+### `chainIds` Request Parameter
+
+In an ever-evolving "multichain" ecosystem, it is important for Applications to be able to connect to a Wallet without being constrained to supply a chain (or set of chains) to connect to. This specification allows for this flexibility by making chain IDs optional in the `wallet_connect` request. This allows an Application to 
+more gracefully perform actions on any chain the Wallet supports.
+However, if an Application is designed to support specific chains, the Application MAY provide a `chainIds` parameter to `wallet_connect` to request connection to at least one of the provided chain IDs. This may be
+important for Applications that perform actions with contracts that reside on a chain, or perform [ERC-1271](https://eips.ethereum.org/EIPS/eip-1271) verification on Smart Contract Accounts.
```
lukasrosario: as part of the move towards the big 3 we should just do away with the idea of a "current" chain. i think we could still support this parameter so a wallet can short circuit connecting by letting the user know if it doesn't support any of the application's chains, but otherwise i think we should remove language around "initial chain", "letting the wallet choose a chain" etc as it implies a user might need to "switch" chains later (which is what i think we should get rid of as a concept, sendCalls and sign should just require chain ids).

ERCS/erc-7846.md
```diff
@@ -30,16 +30,24 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
   method: 'wallet_connect',
   params: [{
-    // JSON-RPC method version.
-    version: string;
     // Optional capabilities to request (e.g. Sign In With Ethereum).
     capabilities?: Record<string, unknown>;
+    // Chain IDs to propose to the wallet to establish an initial connection with.
+    chainIds?: `0x${string}`[];
+    // JSON-RPC method version.
+    version: '1';
```
ilikesymmetry: propose we keep this as `'0'` until we get alignment from other wallets to make an official v1

ERCS/erc-7846.md
```diff
@@ -30,16 +30,24 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
 ```ts
 type Request = {
   method: 'wallet_connect',
   params: [{
-    // JSON-RPC method version.
-    version: string;
     // Optional capabilities to request (e.g. Sign In With Ethereum).
     capabilities?: Record<string, unknown>;
+    // Chain IDs to propose to the wallet to establish an initial connection with.
+    chainIds?: `0x${string}`[];
```
ilikesymmetry: might be nice to explicitly call out hex encoded in the comment

ERCS/erc-7846.md
```diff
@@ -50,12 +58,15 @@ List of connected accounts with their associated capabilities.
...
+  // Accounts established with this connection.
   accounts: {
     // Address of the connected account.
     address: `0x${string}`;
     // Capabilities granted that is associated with this account.
     capabilities: Record<string, unknown>;
-  }[]
+  }[],
+  // Initial chain IDs established with this connection.
+  chainIds: `0x${string}`[]
```
ilikesymmetry, line 67: Curious why we have chainIds on the outside vs inside as I could see reason for both.

ERCS/erc-7846.md
ilikesymmetry: > (or any chain ID the account supports if no chain IDs were requested)

This is a sane default but also slightly concerning compatibility wise. For smart accounts, an app will need an RPC to run a stateful `isValidSignature` call. If an app doesn't declare it supports a chain, I don't have high confidence they will actually be able to validate the signature on a given chain. I feel like it may be more reliable for us to say that "`chainId` MUST be provided if `chainIds` is not".

ERCS/erc-7846.md
```diff
@@ -124,28 +136,29 @@ Adds support for offchain authentication using [ERC-4361](./eip-4361.md).
...
-    chainId: string; // EIP-155 hex-encoded
-    version?: string;
-    scheme?: string;
+    chainId?: string; // EIP-155 hex-encoded
     domain?: string;
-    uri?: string;
-    statement?: string;
-    issuedAt?: string;
     expirationTime?: string;
+    issuedAt?: string;
```
ilikesymmetry, line 155: generally agree with sorting by alphabetical but for context the previous ordering was following from the order declared in SIWE itself

ilikesymmetry (commented):
Good improvement, considered before but never got to it. Would recommend running by @jxom too

ERCS/erc-7846.md
jxom: Yes, agree.

ERCS/erc-7846.md
jxom: Kind of wanted to avoid complexity. I have never seen an app that has differing chain id scopes across accounts / why would you want to?

ERCS/erc-7846.md
jxom: Will change back.

ERCS/erc-7846.md
jxom: > If an app doesn't declare it supports a chain

The intention for an Application supplying no chain IDs would mean that it is graceful towards any chain ID. An Application could be smart enough to ERC-1271 verify against any arbitrary chain ID within the SIWE message. For example, if a `chainId` of `69420` is provided in the message, then an application could ERC-1271 verify against a `https://69420.rpc.example.com` RPC.
If the Application is not smart enough to do that, it would be recommended to pass their supported chain IDs.

ERCS/erc-7846.md
jxom: I would assume that it would use the active (or other compatible) chain of the Wallet. Analogous to optional `from` in ERC-5792, but `chainId` in this case.

ERCS/erc-7846.md
jxom: Sure

ERCS/erc-7846.md
jxom: Is this a suggestion to add to this to the rationale? or to amend the content here? (this is the sentiment I was heading towards here)

hilljessica8787-dot (approved):

PR #1154, opened 2025-08-06
Update ERC-3770: fix inconsistent author formatting in metadata

voronor:
noticed that one of the author entries didn’t follow the same format as the others - added the missing GitHub handle for consistency.


hilljessica8787-dot (approved):

SamWilsn:
If the authors haven't approved this by now, they are unlikely to.

PR #1155, opened 2025-08-07
Add ERC: Simplified Payment Verification Gateway

Hrom131:
Hey, a quick one here. Would love to hear your feedback!

### DIFF
--- a/ERCS/erc-XXXX.md
+++ b/ERCS/erc-XXXX.md
@@ -1,0 +1,304 @@
+---
+eip: XXXX
+title: Simplified Payment Verification Gateway
+description: Trustless singleton contract for on-chain verification of Bitcoin transactions through block headers
+author: Artem Chystiakov (@arvolear), Oleh Komendant (@Hrom131)
+discussions-to: 
+status: Draft
+type: Standards Track
+category: ERC
+created: 2025-07-31
+---
+
+## Abstract
+
+Introduce a singleton contract for on-chain verification of transactions that happened on Bitcoin. The contract is available at "0xTODO" <!-- TODO -->, acting as a trustless Simplified Payment Verification (SPV) gateway where anyone can submit Bitcoin block headers to. The gateway maintains the mainchain of blocks and allows the existence of Bitcoin transactions to be verified via Merkle proofs.
+
+## Motivation
+
+Ethereum's long term mission has always been to revolutionize the financial world through decentralization, trustlessness, and programmable value enabled by smart contracts. Many great use cases have been discovered so far, including the renaissance of Decentralized Finance (DeFi), emergence of Real-World Assets (RWA), and rise of privacy-preserving protocols.
+
+However, one gem has been unreachable to date -- Bitcoin. Due to its extremely constrained programmability, one can only hold and transfer bitcoins in a trustless manner. This EIP tries to expand its capabilities by laying a solid foundation for bitcoins to be also used in various EVM-based DeFi protocols, unlocking the whole new trillion dollar market.
+
+The singleton SPV gateway contract defined in this proposal acts as a trustless one-way bridge between Bitcoin and Ethereum, already enabling use cases such as using _native_ BTC as a lending collateral for stablecoin loans. Moreover, with the recent breakthroughs in the BitVM technology, the full-fledged, ownerless two-way bridge may soon become a reality, powering the permissionless and wrapless issuance of BTC on Ethereum.
+
+## Specification
+
+The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119.
+
+### General
+
+#### Bitcoin Block Header Structure
+
+In Bitcoin, each block contains a header, which has a fixed size of 80 bytes and follows the following structure:
+
+| Field          | Size     | Format             | Description                                                          |
+| :------------- | :------- | :----------------- | :------------------------------------------------------------------- |
+| Version        | 4 bytes  | little-endian      | The version number of the block.                                     |
+| Previous Block | 32 bytes | natural byte order | The block hash of a previous block this block is building on top of. |
+| Merkle Root    | 32 bytes | natural byte order | A fingerprint for all of the transactions included in the block.     |
+| Time           | 4 bytes  | little-endian      | The current time as a Unix timestamp.                                |
+| Bits           | 4 bytes  | little-endian      | A compact representation of the current target (difficulty).         |
+| Nonce          | 4 bytes  | little-endian      | A 32-bit number which miners compute to find a valid block hash.     |
+
+The fields within the block header are sequentially ordered as presented in the table above.
+
+#### Difficulty Adjustment Mechanism
+
+Bitcoin's Proof-of-Work (PoW) consensus mechanism has a probabilistic finality, thus relying on a dynamic **difficulty target** adjustments. The target's initial value is set to `0x00000000ffff0000000000000000000000000000000000000000000000000000`, which also serves as the minimum difficulty threshold.
+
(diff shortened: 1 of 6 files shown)


ERCS/erc-XXXX.md
```diff
@@ -0,0 +1,304 @@
+---
+eip: XXXX
```
abcoathup:
```suggestion
eip: 8002
```

ERCS/erc-XXXX.md
```diff
@@ -0,0 +1,304 @@
...
+title: Simplified Payment Verification Gateway
+description: Trustless singleton contract for on-chain verification of Bitcoin transactions through block headers
+author: Artem Chystiakov (@arvolear), Oleh Komendant (@Hrom131)
+discussions-to: https://ethereum-magicians.org/t/eip-xxxx-simplified-payment-verification-gateway/25038
```
abcoathup:
```suggestion
discussions-to: https://ethereum-magicians.org/t/erc-8002-simplified-payment-verification-gateway/25038
```

Hrom131: Done

ERCS/erc-8002.md
```diff
@@ -0,0 +1,304 @@
...
+     * @param merkleProof The array of hashes used to build the Merkle root
+     * @param directions The array indicating the hashing directions for the Merkle proof
+     * @return True if the txId is present in the block's Merkle tree, false otherwise
+     */
+    function checkTxInclusion(
+        bytes32 blockHash,
+        bytes32 txId,
+        bytes32[] calldata merkleProof,
+        HashDirection[] calldata directions
+    ) external view returns (bool);
```
reednaa: If you provide a transaction index, you can derive the directions. That saves you calldata.

reednaa (commented):
Hey @Hrom131,
I like the progress on Bitcoin on EVM and is personally also a big fan. We have actually been doing this for some time as Catalyst @ LI.FI with an audited and highly optimised Bitcoin SPV client implemented on EVM inspired by Bitcoin Mirror.
https://github.com/catalystsystem/bitcoinprism-evm

Additionally, we have also added it as an [oracle](https://github.com/openintentsframework/oif-contracts/blob/main/src/oracles/bitcoin/BitcoinOracle.sol) to the OIF to allow for EVM to Bitcoin swaps.

assets/erc-8002/contracts/libs/TxMerkleProof.sol
```diff
@@ -0,0 +1,74 @@
...
+
+        for (uint256 i = 0; i < proofLength; ++i) {
+            if (directions[i] == HashDirection.Left) {
+                computedHash = _doubleSHA256(computedHash, proof[i]);
+            } else if (directions[i] == HashDirection.Right) {
+                computedHash = _doubleSHA256(proof[i], computedHash);
+            } else {
+                computedHash = _doubleSHA256(computedHash, computedHash);
+            }
+        }
```
reednaa: This merkle tree algorithm is (specifically for BTC) subject to an insertion attack.
Implement something like: https://github.com/catalystsystem/bitcoinprism-evm/blob/158ff5e6361ba6813f37135c3635d6a88c6b3407/src/library/BtcProof.sol#L247 to ensure the tree does not contain inserted transactions.

Hrom131:
Hi @reednaa, thanks for the helpful links! We’ve already spotted a few areas for improvement.

I also have a few questions about `BtcPrism`:
1. On average, how much gas does it cost to add a single block and a batch of, say, 100 blocks?
2. Why was the decision made to store only the latest 2000 blocks? This way, users can’t prove a transaction existed if it’s older than about two weeks.

ERCS/erc-8002.md
Hrom131: If we pass `txIndex`, it can only encode two options: hashing to the left or hashing to the right. This means we cannot represent the third option, where a node needs to be hashed with itself. Because of this, the Merkle Proof obtained from a Bitcoin node would need additional processing to include extra siblings for the case when a node is hashed with itself. As a result, Merkle Proof verification in the contract would be entirely dependent on correct preprocessing, which is a drawback.

By passing an array of `directions`, the need for proof preprocessing is reduced, while the cost of proof verification is only slightly higher.

assets/erc-8002/contracts/libs/TxMerkleProof.sol
Hrom131: Good point! We will add necessary checks to prevent such attacks.

reednaa:
> 1. On average, how much gas does it cost to add a single block and a batch of, say, 100 blocks?

Here is a transaction on Base submitting 100 blocks to an instance of the SPV client. 451372 gas used. I believe the cost for a single block is 24000 to 40000 depending on whether the ring buffer is full.
https://basescan.org/tx/0xbc16f579e76a04ba372fdd7dfdeb0eaed6531ee41b8d0daebeca037df9f51202

> 2. Why was the decision made to store only the latest 2000 blocks? This way, users can’t prove a transaction existed if it’s older than about two weeks.

The ring buffer makes each block 18k cheaper once it is full. It makes it significantly cheaper to update. This functionality can easily be removed if needed.

ERCS/erc-8002.md
reednaa: What is the use case for hashing with itself?

Hrom131: I was talking about a case, for example, when we need to prove the last transaction in block https://learnmeabitcoin.com/explorer/block/0000000049a63b4dda3a43450c19d085d6c28bfb4cbb2e0576815d7f31919c5d.
This block has 6 transactions, and at the second level of the tree we will have 3 nodes, where the one we need will be the last and will not have a pair, so the second operation should hash it with itself.

Am I correct in understanding that in your implementation you precompute the hash of the node that needs to be hashed with itself and put it into the `MerkleProof`, so that when we reach it, we simply use the `txIndex` for the hashing direction without making any changes?

reednaa: Yes.

ERCS/erc-8002.md
```diff
@@ -0,0 +1,291 @@
+---
+eip: 8002
+title: Simplified Payment Verification Gateway
+description: Trustless singleton contract for on-chain verification of Bitcoin transactions through block headers
```
SamWilsn, line 4: Whenever there's a singleton contract, we have to ask the question: does it need an ERC?

If there's only one deployment, would it make more sense to document it on the project's page? ERCs really shine when multiple implementations need to coordinate.

That said, we do have other singletons documented here, so if your reasoning is sound, we'll include it.

ERCS/erc-8002.md
```diff
@@ -0,0 +1,291 @@
...
+
+## Security Considerations
+
+Among potential security issues, the following can be noted:
+
+The security of the `SPVGateway` is directly dependent on the security of Bitcoin's underlying PoW consensus. A successful 51% attack on the Bitcoin network would allow an attacker to submit fraudulent block headers that would be accepted by the contract, compromising its state.
+
+The block header validation rules require a Bitcoin node to check that the newly created block is not more than 2 hours ahead of the node's network-adjusted time. This check is impossible to implement on the `SPVGateway` smart contract, hence it is omitted.
+
+Unlike other blockchain systems with deterministic finality, Bitcoin's consensus is probabilistic. The `SPVGateway` contract SHOULD be designed to handle chain reorganizations of arbitrary depth, but it cannot prevent them. As a result, transactions included in a block may not be permanently final. All dApps and protocols relying on this contract MUST implement their own security policies to determine a sufficient number of block confirmations before a transaction is considered "final" for their specific use case.
```
SamWilsn, line 285: Avoid creating requirements (with UPPERCASE keywords) outside of the Specification section.

ERCS/erc-8002.md
SamWilsn: After reading, I think this does make sense as a standard. There could easily be different implementations of this interface, or similar implementations with different bootstrapping.

PR #1158, opened 2025-08-10
Add ERC: ERC-20 Pre-initialization (Sentinel Storage)

ariutokintumi:
This ERC proposes an **optional ERC-20 extension** that allows token contracts to implement a `preInitializeAddress(address)` function. This function lets users or dApps pay the high `SSTORE` gas cost of first-time balance initialization in advance, when gas is low, by writing a sentinel (magic) value to the balance slot.

The balance mapping is stored as `bytes32` internally, and the sentinel is treated as `0` for all ERC-20 reads/writes. When the user later receives tokens, the sentinel is overwritten at the cheaper 5k gas rate instead of the full 20k.

**We use bytes32** because writing zero to a slot does **not** save gas for the first real (nonzero) balance write. EVM only discounts overwrites of already-allocated slots with nonzero value.

This approach mirrors ERC-721A "gas timing" concept but in reverse: **high gas now, low gas later**, giving users more control over when they incur expensive storage writes.

**Discussion thread:** [https://ethereum-magicians.org/t/erc-tbd-erc-20-pre-initialization-extension-sentinel-storage-gas-savings-for-first-time-token-receivers/24993](https://ethereum-magicians.org/t/erc-tbd-erc-20-pre-initialization-extension-sentinel-storage-gas-savings-for-first-time-token-receivers/24993)

**Repository including contracs, testing and tooling:** [https://github.com/ariutokintumi/ERC-20-Pre-initialization](https://github.com/ariutokintumi/ERC-20-Pre-initialization)


ERCS/erc-draft-erc20-preinit-sentinel-storage.md
```diff
@@ -0,0 +1,149 @@
+---
+erc: TBD
```
abcoathup:
```suggestion
erc: 8003
```
Please also update the filename

ERCS/erc-draft-erc20-preinit-sentinel-storage.md
```diff
@@ -0,0 +1,149 @@
...
+title: ERC-20 Pre-initialization Extension (Sentinel Storage)
+author: German Maria Abal Bazzano (@ariutokintumi)
+discussions-to: https://ethereum-magicians.org/t/erc-tbd-erc-20-pre-initialization-extension-sentinel-storage-gas-savings-for-first-time-token-receivers/24993
```
abcoathup:
```suggestion
discussions-to: https://ethereum-magicians.org/t/erc-8003-erc-20-erc-20-pre-initialization-extension-sentinel-storage/24993
```
Add assigned number and update title to match ERC title

ERCS/erc-8003-erc20-preinit-sentinel-storage.md
```diff
@@ -0,0 +1,149 @@
+---
+erc: 8003
+title: ERC-8003 ERC-20 Pre-initialization Extension (Sentinel Storage)
```
abcoathup:
```suggestion
title: ERC-20 Pre-initialization Extension (Sentinel Storage)
```
Don't include the ERC number in the title

xinbenlv:
Overall look like a really interesting proposal! Just onething, i want to check if the Math still works in this new approach. Left a technical comment in EthMagic https://ethereum-magicians.org/t/erc-8003-erc-20-pre-initialization-extension-sentinel-storage/24993/6?u=xinbenlv. Please let me know when it's answered

xinbenlv (dismissed):
Also please address bot failure

ariutokintumi:
> Also please address bot failure

Done!

ERCS/erc-8003.md
```diff
@@ -0,0 +1,147 @@
+---
+eip: 8003
+title: ERC-20 Pre-initialization (Sentinel Storage)
+description: Optional extension to pre-allocate a balance slot via a sentinel bytes32 value so the first SSTORE is prepaid, reducing later transfer gas.
```
SamWilsn, line 4:
```suggestion
description: Pre-allocate a balance slot via a sentinel bytes32 value so the first SSTORE is prepaid, reducing later transfer gas.
```

Extensions are always optional because of how EIPs work.

ERCS/erc-8003.md
```diff
@@ -0,0 +1,147 @@
...
+type: Standards Track
+category: ERC
+created: 2025-07-17
+requires: 20
+---
+
+
+## Abstract
+
+This [ERC-20](../EIPS/eip-20.md) extension introduces an **optional pre-initialization function** for ERC-20 tokens, leveraging a sentinel (magic) value and a `bytes32` storage mapping for balances. This mechanism allows users to pay the high SSTORE gas cost in advance (when gas is cheap), so that the later first real token transfer to that address costs far less gas during a potential upcoming network congestion.
```
SamWilsn, line 17:
```suggestion
This [ERC-20](./eip-20.md) extension introduces an **optional pre-initialization function** for ERC-20 tokens, leveraging a sentinel (magic) value and a `bytes32` storage mapping for balances. This mechanism allows users to pay the high SSTORE gas cost in advance (when gas is cheap), so that the later first real token transfer to that address costs far less gas during a potential upcoming network congestion.
```

ERCS/erc-8003.md
```diff
@@ -0,0 +1,147 @@
...
+
+Pre-initializing an ERC-20 token balance storage slot can save significant gas for users, but a simple approach (writing 0 to a `uint256` balance) does **not** actually reduce costs for the eventual first real transfer at all. This extension proposes a safe, ERC-20-compatible method using `mapping(address => bytes32)` for balances and a unique sentinel value. The contract's ERC-20 interface remains 100% standard: all reads/writes are still performed as `uint256`. Only the internal pre-initialization function ever stores the sentinel value; all normal functions interpret it as zero, and any later transfer/mint simply overwrites it.
+
+This allows users or integrators to "prepay" storage rent and optimize gas costs before high-demand launches, presales, and trending tokens. An easy way to understand for experienced developers is that this mirrors the principle of gas-shifting found in [ERC-721](./eip-721)A, but applied in reverse for ERC-20 storage.
```
SamWilsn, line 21: What is ERC-721A? That isn't defined in ERC-721.

lucemans:
So what if the wild case occurs where a user's balance does end up matching the MAGIC value?
The contracts would start returning 0's no?

ERCS/erc-8003.md
SamWilsn: This is technically an external link, since you're referencing something other than an ERC. Since this standard is available under MIT, you could put a copy of it in your assets directory if you'd like to keep this reference.

ERCS/erc-8003.md
```diff
@@ -0,0 +1,147 @@
...
+2. **Strategic Pre-initialization for Trending Tokens**
+   - Savvy users can batch-initialize slots on tokens they may want to trade when hype spikes.
+3. **Wallet/dApp Automation**
+   - Wallets, bots, and advanced dApps can offer a "pre-initialize now" button or automatic service for users to optimize their on-chain costs.
+
+
+## Specification
+
+ERC-20 contracts implementing this extension **MUST**:
+- Store balances as `mapping(address => bytes32)`.
```
SamWilsn, line 42: This is an implementation detail. If the contract wants to implement it using some other mechanism, but the observable external effects are the same, unless you're expecting people to directly read the storage, through, for example, `eth_getStorageAt`.

ERCS/erc-8003.md
```diff
@@ -0,0 +1,147 @@
...
+
+
+## Backwards Compatibility
+
+This function and pattern are optional and fully backward compatible: all ERC-20 views, transfers, and approvals remain unchanged in interface and behavior.
+
+
+## Reference Implementation
+
+A minimal example contract is provided in the repository (github.com/ariutokintumi/ERC-20-Pre-initialization/tree/main/contracts). Don't use for production until a security audit is performed.
```
SamWilsn, line 135: Please remove the external link. If you'd like, you can include a reference implementation in your assets directory if it's under a compatible license (preferably CC0, but we also accept, for example, MIT/Apache-2.0, etc.)

ERCS/erc-8003.md
```diff
@@ -0,0 +1,147 @@
...
+## Security Considerations
+
+- No risk of funds loss: pre-initialization cannot set balances to nonzero token values or affect accounting.
+- The sentinel is never visible or accessible externally. ERC-20 APIs always interpret it as zero.
+- Minimal "storage bloat" risk; a slot with sentinel is no worse than a slot initialized via normal transfers.
+
+
+## Copyright
+
+Copyright and license: MIT
```
SamWilsn, line 147: Please use the exact license waiver from the EIP template:

https://github.com/ethereum/EIPs/blob/d0e17ebb67c6413fd6e98f538f652e8945a64e80/eip-template.md?plain=1#L128-L130

SamWilsn (changes requested):
Aside from the other comments, the last item is that this proposal deals almost entirely with implementation details: picking a particular storage layout (`mapping`) and a sentinel value. You should rewrite this so that the Specification section only describes the externally visible behaviour (so `preInitializeAddress` and "does work to ensure minting is cheaper"). You can talk about a sentinel value and mapping in your Reference Implementation section.

ariutokintumi:
> So what if the wild case occurs where a user's balance does end up matching the MAGIC value? The contracts would start returning 0's no?

Hello @lucemans, thanks for taking the time to review my proposal. 

This case is not happening because the 'MAGIC value' is not a number (and MUST not be). That means that there is no potential collision between balance and per-initialization values.

This also makes the accounting to be safe. If we use a number for the 'MAGIC value'  it will bring a lot of problems besides that one as well.

SamWilsn:
Friendly reminder to keep technical discussion on [the discussion thread](https://ethereum-magicians.org/t/erc-8003-erc-20-pre-initialization-extension-sentinel-storage/24993).

hilljessica8787-dot (approved):

PR #1163, opened 2025-08-12
Update ERC-1: Update and rename eip-1.md to erc-1.md

peersky:
Having it named EIP-1 creates a discrepancy since the EIP-1 in this repository is not in sync with https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1.md anymore


xinbenlv:
I don't think we are going to change it

PR #1178, opened 2025-08-19
Update ERC-7858: Move to Final

MASDXI:
Change status of ERC-7858 from Last Call to Final


xinbenlv (dismissed):
The specification section didn't mention whether the interface is a MUST or should.

ERCS/erc-7858.md
```diff
@@ -94,14 +95,14 @@ interface IERC7858 /**is IERC721 */ {
...
+- `isTokenExpired` is used for retrieving the status of the given `tokenId`; the function MUST return `true` if the token is expired and MUST revert if the `tokenId` does not exist. Implementations that use custom errors SHOULD revert with `ERC721NonexistentToken` following the relevant [ERC-6093](./eip-6093.md) and implementations using Solidity version below `v0.8.4` or those preferring to use revert with string error SHOULD revert with `NonexistToken`. If the `tokenId` exists and is not expired, the function MUST return `false`.
 - `startTime` and `endTime` of `tokenId`, can be `block.number` or `block.timestamp` depending on `expiryType`. The `startTime` MUST less than `endTime` and SHOULD except when both are set to `0`. A `startTime` and `endTime` of `0` indicates that the `tokenId` has no time-limited. In case `startTime` is non `0` but the `endTime` is `0` also considered that `tokenId` has no time-limited. If the `tokenId` does not exist MUST revert same as in function `isTokenExpired`.
```
SamWilsn: Should this be something like:

```suggestion
- `startTime` and `endTime` of `tokenId`, can be `block.number` or `block.timestamp` depending on `expiryType`. The `startTime` MUST be less than or equal to `endTime` and `startTime` SHOULD be strictly less than `endTime` except when both are set to `0`. A `startTime` and `endTime` of `0` indicates that the `tokenId` has no expiration. Tokens with a non-zero `startTime` and a zero `endTime` are also considered to have no expiration. If the `tokenId` does not exist, this function MUST revert in the same way as `isTokenExpired`.
```

ERCS/erc-7858.md
```diff
@@ -94,14 +95,14 @@ interface IERC7858 /**is IERC721 */ {
...
 - `supportInterface` for `IERC7858` is `0x3ebdfa31` for `IERC7858Epoch` is `0xec7ffd66`
```
SamWilsn: How about:

```suggestion
- The interface ID for [ERC-165](./eip-165.md)'s `supportsInterface` for `IERC7858` is `0x3ebdfa31`, and for `IERC7858Epoch` it is `0xec7ffd66`
```

Also, make sure to double check these before we merge this PR!

ERCS/erc-7858.md
```diff
@@ -163,6 +164,7 @@ interface IERC7858Epoch /** is IERC7858 */ {
     function validityDuration() external view returns (uint256);
 }
 ```
+### Extension Behavior Specification
 
 - `unexpiredBalanceOfAtEpoch` MUST return unexpired or usable balance of tokens held by an account at the specified `epoch`,If the specified `epoch` is expired, this function MUST return `0`. For example, if epoch `5` has expired, calling `unexpiredBalanceOfAtEpoch(5, address)` returns `0` even if there were tokens previously held in that epoch.
```
SamWilsn: The "or" here was ambiguous: are unexpired tokens distinct from usable balance, or are they both terms for the same thing?

```suggestion
- `unexpiredBalanceOfAtEpoch` MUST return unexpired (usable) balance of tokens held by an account at the specified `epoch`. If the specified `epoch` is expired, this function MUST return `0`. For example, if epoch `5` has expired, calling `unexpiredBalanceOfAtEpoch(5, address)` returns `0` even if there were tokens previously held in that epoch.
```

MASDXI: Thank you, @SamWilsn, unexpired token means usable token that still valid

SamWilsn (approved):

PR #1179, opened 2025-08-19
Add ERC: Universal Policy Engine

vpriadko:
This ERC standardizes a universal, artifact-based policy engine – composable and reusable contracts with a minimal init/exec interface – coordinated by a Policy handler structured as a DAG.

All policy parameters are passed as bytes to support generic usage both on- and off-chain.
Policies are attached as method-level hooks (pre/post) to gate calls, chain validations, and enable clean, simple reuse across contracts, domains, and deployments.


ERCS/erc-draft_composite_policy.md
```diff
@@ -0,0 +1,425 @@
+---
```
abcoathup, line 1:
```suggestion
---
eip: 8006
```

ERCS/erc-draft_composite_policy.md
```diff
@@ -0,0 +1,425 @@
...
+title: Universal Policy Engine
```
abcoathup, line 3: Would "Composite policy engine" be a more descriptive title? Universal doesn't seem to add much.

I am not an ERC editor

ERCS/erc-draft_composite_policy.md
```diff
@@ -0,0 +1,425 @@
...
+description: Composite onchain policy engine utilizing generic rule artifacts within DAG evaluator
+author: Vlad Pryadko <vpriadko@guardianlabs.org>, Vitali Grabovski <v.grabovski@guardianlabs.org>
+discussions-to: https://ethereum-magicians.org/t/new-erc-universal-policy-engine-a-composable-on-chain-compliance-policy-built-from-reusable-artifacts/25155
```
abcoathup:
```suggestion
discussions-to: https://ethereum-magicians.org/t/erc-8006-universal-policy-engine/25155
```
Added assigned number and updated title to match ERC

ERCS/erc-draft_composite_policy.md
vpriadko: Hey, thanks for the comment
"Composite" is about internal structure, "how it works". "Universal" is about integration and usage perspective, "how can I use it".
I believe, standard's name may want to reflect external properties in the first place - so, it's about "how can I use it".

ERCS/erc-draft_composite_policy.md
vgrabovski: solved

ERCS/erc-draft_composite_policy.md
vgrabovski: solved

ERCS/erc-8006.md
```diff
@@ -0,0 +1,426 @@
...
+
+## Abstract
+
+Smart contract systems face significant challenges implementing sophisticated, adaptable compliance rules that can evolve with changing regulatory requirements. This proposal presents an efficient on-chain policy engine approach. Policies are decomposed into simple rules, called artifacts, and represented as interconnected entities forming logical structures. In a recommended implementation, these entities can be organized as a directed acyclic graph (DAG), where each node is an instance of an artifact. The evaluation begins with a top-tier artifact (root node) and processes through the connected structure, with the result serving as the final policy evaluation outcome.
+
+The standard's universality stems from its preservation of artifact connectivity across different contexts, establishing a meta-layer of compliance requirements applicable to any new or existing (yet upgradable) decentralized application. It defines approaches, interfaces, and conventional traits that facilitate elegant interoperability between policies and artifacts for common usage scenarios, enabling seamless implementation and consumption of any compliance scenario.
+
+## Motivation
```
xinbenlv: The motivation section conflates multiple different problems without clearly establishing this as a distinct need from existing solutions.

The ERC describes general smart contract validation challenges but doesn't adequately differentiate from
  existing patterns like:
  - OpenZeppelin's AccessControl and modifiers
  - Proxy patterns for upgradeable validation logic
  - Existing policy/permission systems (e.g., compound governance, gnosis safe modules)
  
  Recommendation: Provide concrete examples where current approaches fail and this DAG-based policy engine provides unique
   value. Include gas cost comparisons and complexity metrics versus existing patterns.

xinbenlv (changes requested):

ERCS/erc-8006.md
```diff
@@ -0,0 +1,430 @@
+---
+eip: 8006
+title: Universal Policy Engine
+description: Composite onchain policy engine utilizing generic rule artifacts within DAG evaluator
```
bomanaps:
```suggestion
description: Composite on-chain policy engine utilizing generic rule artifacts within DAG evaluator
```

ERCS/erc-8006.md
```diff
@@ -0,0 +1,430 @@
...
+
+## Abstract
+
+Smart contract systems face significant challenges implementing sophisticated, adaptable compliance rules that can evolve with changing regulatory requirements. This proposal presents an efficient on-chain policy engine approach. Policies are decomposed into simple rules, called artifacts, and represented as interconnected entities forming logical structures. In a recommended implementation, these entities can be organized as a directed acyclic graph (DAG), where each node is an instance of an artifact. The evaluation begins with a top-tier artifact (root node) and processes through the connected structure, with the result serving as the final policy evaluation outcome.
+
+The standard's universality stems from its preservation of artifact connectivity across different contexts, establishing a meta-layer of compliance requirements applicable to any new or existing (yet upgradable) decentralized application. It defines approaches, interfaces, and conventional traits that facilitate elegant interoperability between policies and artifacts for common usage scenarios, enabling seamless implementation and consumption of any compliance scenario.
```
bomanaps: As per EIP guidelines, the abstract should only summarize what the proposal does. The second paragraph leans into rationale, but the “why” belongs in the Rationale section. Make it self-contained enough that someone can understand the core purpose without reading the full document.

ERCS/erc-8006.md
```diff
@@ -0,0 +1,430 @@
...
+1. **Component Reusability**: Quick and simple reuse of artifacts across multiple applications
+2. **Flexible Policy Evolution**: Easy addition or replacement of artifacts within policy handlers
+3. **Rapid Rule Updates**: When compliance requirements change, only targeted modifications are needed
+4. **Reduced Development Overhead**: Instead of complete development lifecycles, only new feature implementation is required
+
+When there's a need to implement additional compliance rules, developers can simply implement new artifacts (or use existing community-developed ones) and update the existing policy via the policy handler.
+
+### Outcome
+
+This approach enables sophisticated compliance systems that were previously impractical to implement, unlocking new possibilities for on-chain governance and regulatory compliance.
```
bomanaps: The Motivation section as it is currently goes beyond its intended purpose. According to the EIP guidelines, Motivation should explain why the change is needed in a concise way, highlighting the problem and the insufficiency of existing methods.

Right now, subsections such as Proposed Solution, Key Benefits, and Delegated Security Model introduce details of how the solution works and the advantages of the design. These belong in the Specification (how it works) or Rationale (why these choices were made).
I recommend restructuring Motivation to keep only:
- The problem statements,
- The challenges with existing approaches, and
- Why those approaches are insufficient.

ERCS/erc-8006.md
```diff
@@ -0,0 +1,430 @@
...
+
+## Specification
```
bomanaps, line 54: Specification should focus strictly on what the change does, not why this design is chosen. Explanatory reasoning (e.g., “most efficient architecture”, “addressing the challenge of composing complex rules”) belongs in the Rationale section. I recommend moving these there to maintain a clean separation of concerns.

Also, you introduce a definition of “policy.” can you clarify whether this definition is strictly necessary for the specification? If it is only explanatory context?

ERCS/erc-8006.md
vgrabovski: Thanks for the raised questions and ideas.

**Regarding your examples**

The examples you provided: (a) only solve specific tasks, (b) involve usage complexity, (c) locked within specific ecosystems

These are precisely the limitations that ERC-8006 aims to overcome.

Here are the limitations in detail:

**OpenZeppelin AccessControl:**
1. Only checks if the caller is authorized.
2. Authorization is based solely on roles.
3. No way to apply complex or conditional logic beyond that.

**Proxy Pattern (e.g., Diamond Standard):**
1. Requires complex and cautious upgrade procedures.
2. The `importantValidation` method used in 10-15 methods has to be a public method to upgrade it; otherwise, you have to upgrade each of these 10 methods; there is no way to upgrade a private one or an internal one (which may be essential since they often modify state (used nonces, used signatures, etc).
3. Risk of introducing bugs (e.g., typos) during upgrades.
4. Post-upgrade issues require emergency re-upgrades.
5. It’s needed to track the current state of the upgraded contract to figure out how to interact with it now.
6. This is significantly more complex than simply calling `policyHandler.upgrade(...params)`.

**Gnosis Safe Modules:**
1. Great in their own scope, but limited.
2. Only apply to Gnosis Safe wallet.
3. `ITransactionGuard` works for `execTransaction`, and `IModuleGuard` for `execTransactionFromModule`.
4. These guard interfaces only serve Gnosis Safe-like architectures.
5. Only one `IModuleGuard` can be used for all modules – dedicated guards per module are not supported; same is true for `ITransactionGuard`.
6. If you develop and contribute an `IModuleGuard` or `ITransactionGuard` for Safe, there’s no relatively easy way to contribute similar logic to other smart wallet architectures.
___
**Just to provide a bit more context on the motivation**

The _goal_ of this ERC is to enable the creation of a compliance (protection through compiance) mechanism for any on-chain **action**. This mechanism would rely on a combination of the action’s **traits**, environmental traits, and off-chain traits. 
A universal and composable compliance policy solves this problem by facilitating the creation of such a mechanism.

For example, let’s say the `action` is a token transfer. In this case, the `transfer()` contract method is the action where the policy check is attached to. Its `traits` include: amount, sender address, and receiver address. There are also sub-traits such as receiver KYC status, sender KYC status, receiver citizenship, sender verifiable credentials, etc., as well as **environmental traits** like the current block timestamp.
All of these traits can be combined and processed to ultimately evaluate whether the action is authorized (true or false).

The _second goal_ of this ERC is compatibility – with both new and existing (yet upgradeable) solutions. The idea is not to break existing codebases, but to extend them. 

Thus, there is no intent to replace OpenZeppelin, Proxy upgrade approaches (which are still useful in many cases), or Gnosis Safe modules.
The intent is to build with them or build on top of them.

**Possible Use Case:**

a. Imagine an 'DApp allowing the lending/borrowing of Game asset (EIP-721)', where a policy is attached to the `returnAssetWhenRentalFinished()` method:
```
                          [OR Artifact]
[Access-Control Artifact]               [M-of-N EIP712 Signatories Quorum Artifact]
```
In this setup, the developer could define the policy rule as follows `isAccessControl(...params) || isEIP712Quorum(...params)` to eventually validate if the return action is authorized.

b. Another example: the `ModuleGuardManager` of a Safe Wallet could implicitly use a Policy if `ModuleGuard` uses `PolicyHandler`.

The motivation in such cases is simple: the required behavior already exists within the Policy artifacts – what remains is to implement an adapter that attaches/consumes the `PolicyHandler` in the `ModuleGuard`

___
**Regarding the gas usage:**

1. The Policy-handler introduces only minimal gas overhead. It acts merely as a skeleton, scaffold, or blueprint – essentially just a container for a DAG of nodes. All “meaningful” (i.e., business-level) logic must be provided by the policy creator as policy building blocks (artifacts).
2. As such, gas consumption is not an inherent property of this standard – it depends almost entirely on the specific artifact implementations chosen by the developer or client.
3. In practice, this means that a policy like "Is the transaction sent within business hours?" can be as simple or as complex as the developer chooses – either composed of a single 'Business Hours' artifact or a combination of others, such as 'OR', 'AND', 'Current Time', and 'Restricted Days' artifacts.
4. It’s also worth noting that policy check evaluation is performed recursively over the DAG, making the call stack height the primary built-in constraint on complexity (as noted in the documentation).

___
**Final Summary**

Based on Safe case, Safe Module is like adding an extra door to the Safe Wallet.
Policy Engine is like adding advanced locks to the main door to Safe Wallet or any DApp/  smart contract.

Our goal is to create a "fit-every-need" approach for handling compliance requirements – evaluated simply as true or false.
This is achievable using Policy logic with conditions that can be fully or partially changed as needed.

vgrabovski: Thanks again for the question. I will add a shortened version of this message to the ERC to provide greater clarification.

ERCS/erc-8006.md
vpriadko: The role of this definition here is to emphasize on non-atomic policy nature, meaning it can be split. The definition is provided through rules to lead the narrative toward the concept of an "artifact". So, I'd say, the policy definition here serves to bind together everything following.

ERCS/erc-8006.md
vpriadko: Thanks, done

ERCS/erc-8006.md
vpriadko: Thanks, done

ERCS/erc-8006.md
vpriadko: Thanks, done

ERCS/erc-8006.md
vgrabovski: solved

vgrabovski:
I'm trying to figure out why the CI/CD check keeps failing

The links to internal files related to the ERC should be fine, but it still can’t pass the check:

```
For the Images check, the following failures were found:

* At ./_site/EIPS/eip-8006.html:426:

  internal image ../assets/erc-8006/simple_policy_bird_eye_view.svg does not exist

For the Links > Internal check, the following failures were found:

* At ./_site/EIPS/eip-8006.html:401:

  internally linking to ../assets/erc-8006/contracts/inheritance/DAGWithPolicyMetadata.sol, which does not exist

* At ./_site/EIPS/eip-8006.html:401:

  internally linking to ../assets/erc-8006/contracts/PolicyHandler.sol, which does not exist

* At ./_site/EIPS/eip-8006.html:507:

  internally linking to ./erc-1167.md, which does not exist

* At ./_site/EIPS/eip-8006.html:676:

  internally linking to ../assets/erc-8006/contracts/PolicyHandler.sol, which does not exist

* At ./_site/EIPS/eip-8006.html:722:

  internally linking to ./erc-1167.md, which does not exist
```

vgrabovski:
@abcoathup, could you please help with [this problem](https://github.com/ethereum/ERCs/pull/1179#issuecomment-3463421287)?

assets/erc-8006/contracts/data-structures/DAG/ContainerDAG.sol
```diff
@@ -0,0 +1,60 @@
+// SPDX-License-Identifier: Unlicensed
```
SamWilsn: We cannot accept "Unlicensed" content in the ERCs repository. If you are the copyright holder, or can get permission, please provide all files under a compatible license (preferably CC0-1.0, but MIT/Apache would be fine as well).

ERCS/erc-8006.md
```diff
@@ -0,0 +1,370 @@
...
+type: Standards Track
+category: ERC
+created: 2025-03-14
+---
+
+## Abstract
+
+Smart contract applications often need configurable and reusable compliance logic. This standard defines an on-chain policy engine that composes rule "artifacts" and evaluates the final check result based on them through a specialized rule processor called the Policy Handler.
+
+A **Directed Acyclic Graph (DAG)** is the RECOMMENDED approach for composition and evaluation in Policy Handler. Artifacts implement a minimal interface and exchange input params / output params of the main interface methods as `bytes` or `bytes[]` (see [Interfaces section](#interfaces)), allowing heterogeneous data and on/off-chain variable supply.
```
SamWilsn: Please keep requirements (indicated by UPPERCASE keywords) in the Specification section.

ERCS/erc-8006.md
```diff
@@ -0,0 +1,370 @@
+---
+eip: 8006
+title: Universal Policy Engine
+description: Composite on-chain policy engine utilizing generic rule artifacts within DAG evaluator
```
SamWilsn: It might be nice to define what a "policy engine" actually is within your description.

ERCS/erc-8006.md
```diff
@@ -0,0 +1,370 @@
...
+created: 2025-03-14
+---
+
+## Abstract
+
+Smart contract applications often need configurable and reusable compliance logic. This standard defines an on-chain policy engine that composes rule "artifacts" and evaluates the final check result based on them through a specialized rule processor called the Policy Handler.
+
+A **Directed Acyclic Graph (DAG)** is the RECOMMENDED approach for composition and evaluation in Policy Handler. Artifacts implement a minimal interface and exchange input params / output params of the main interface methods as `bytes` or `bytes[]` (see [Interfaces section](#interfaces)), allowing heterogeneous data and on/off-chain variable supply.
+
+When a DAG is used, the Policy Handler traverses the graph from a designated **Root** node and eventually returns the output of the root node (root artifact). This output is policy's final result. The standard also specifies self-descriptive methods for artifacts to support programmatic composition and external tooling.
```
SamWilsn: Your abstract does a great job of explaining _how_ your proposal works, but doesn't sufficiently explain _what_ your proposal is. What is a policy engine? What is an artifact? What types of contracts might want to use a Universal Policy Engine?

ERCS/erc-8006.md
```diff
@@ -0,0 +1,370 @@
...
+2. **Security Vulnerabilities**: Flaws in rule implementation pose serious risks to system integrity and user assets. Poor implementations can enable unauthorized access to restricted functionality, lead to direct loss of user funds through exploits, and potentially violate regulatory requirements that the rules were meant to enforce.
+
+3. **Integration Difficulties**: Adding new compliance logic to existing codebases often requires:
+   - Extensive codebase changes
+   - Separate integration work for each compliance rule implementation
+   - The need of carefully conducted contract upgrade process (which is error-prone)
+
+4. **Redundant Implementation**: Many compliance patterns appear across different projects but lack standard interfaces, forcing developers to repeatedly implement similar functionality.
+
+5. **Costly Updates**: As regulatory requirements change, contracts need updates, requiring the entire development process again - writing, testing, auditing, and deploying - increasing costs and error potential.
```
SamWilsn, line 52: This lays out the problem space quite nicely, but you should also use the Motivation to explain how your particular proposal is the best solution to that problem. Why should an implementer choose your solution over a (possibly hypothetical) alternative ERC?

ERCS/erc-8006.md
```diff
@@ -0,0 +1,370 @@
...
+    classDef box fill:white,stroke:#0066cc,rounded:true
+    classDef single fill:white,stroke:black,rounded:true
+    classDef and fill:white,stroke:black,rounded:true
+    classDef greater fill:#ccffcc,stroke:#339933,rounded:true
+    classDef citizen fill:#ffeecc,stroke:#ff9933,rounded:true
+    
+    classDef greenText color:#339933
+    classDef blackText color:black
+    classDef orangeText color:#ff9933
+```
```
SamWilsn: Mermaid is not supported by our renderer. You'll need to render it ahead of time (preferably to SVG) and link to it from your assets directory.

ERCS/erc-8006.md
```diff
@@ -0,0 +1,370 @@
...
+
+To summarize, a policy is a rule governing a specific on-chain action, regardless of its complexity.
+
+In this standard, the *rule, fundamental component* of a policy, is called an **artifact**. An artifact is atomic within the policy context. Internally, an artifact can be as simple or complex as its creator requires.
+
+For example, a policy stating "be over 21 years old and a citizen" comprises two artifacts: "be over 21 years old" and "be a citizen". While one could define a single artifact encompassing both requirements, it is generally recommended to maintain granularity that reflects natural divisions of logic. This granular approach enables policies to be constructed from simple, reusable building blocks, tackling the challenge of complex rule composition.
+
+Artifacts in this proposal are implemented as smart contracts, allowing them to represent not only rules but also arbitrary operations across any context - on-chain transactions, off-chain actions, administrative workflows, or any other decision point requiring policy enforcement. Consequently, logical operators like "and" are also implemented as artifacts.
```
SamWilsn: The definition of rule vs. policy here is a bit muddy. If a policy can be made up of other policies, but can also be treated as rules, then rules aren't really atomic?

ERCS/erc-8006.md
```diff
@@ -0,0 +1,370 @@
...
+    classDef greenText color:#339933
+    classDef blackText color:black
+    classDef orangeText color:#ff9933
+```
+
+### Interfaces
+
+#### Artifact Interfaces
+
+Artifacts may implement any logic but MUST adhere to a standard interface that allows the policy controller to integrate them correctly and ensure consistent data flow:
```
SamWilsn: Maybe:

```suggestion
Artifacts MAY implement any logic but MUST adhere to the following `IArbitracyDataArtifact` interface that allows the policy controller to integrate them correctly and ensure consistent data flow:
```

?

ERCS/erc-8006.md
```diff
@@ -0,0 +1,370 @@
...
+1. Collects the necessary variables for evaluation
+2. Initiates a recursive *traversal* of the graph
+
+For each node during the **traversal**, the handler:
+1. Calls the target artifact using the standard interface
+2. Supplies it with the appropriate variables and/or results from previously evaluated (calculated) nodes that the current node depends on
+3. Collects the output values of the artifacts for use by parent nodes
+4. Finishes the traversal once all nodes have been visited and evaluated
+
+In summary, the handler passes outputs from child nodes as inputs to parent nodes; artifacts do not invoke each other directly. After the traversal completes, the root node’s output value represents the policy’s final result. This design provides a structured mechanism for rule interaction, addressing the challenge of composing complex rules from simpler components.
```
SamWilsn: I'm assuming this is the recommended algorithm for a DAG-based implementation? It probably needs a bit of text that indicates whether it's a MUST or a RECOMMENDED or whatever.

ERCS/erc-8006.md
```diff
@@ -0,0 +1,370 @@
...
+
+![Policy bird's eye view](../assets/erc-8006/simple_policy_bird_eye_view.svg)
+
+The handler SHOULD provide a method to retrieve information about all variables used by all artifacts in the policy:
```
SamWilsn: > a method

The recommendation here is a bit ambiguous. Is it optional to include the method? Is it optional that, if such a method is included, it adheres to the following signature?

ERCS/erc-8006.md
```diff
@@ -0,0 +1,370 @@
...
+
+```solidity
+function getVariablesList() public view returns (ExecVarsMetadata[] memory list) {
+    require(isInitialized, POLICY_NOT_INITIALIZED_ERR);
+
+    list = Utils.getVarsDesriptionList(dag.getNodes());
```
SamWilsn: This is an implementation detail. ERCs should only specify the externally visible behaviour of a system, not how that system is built.

ERCS/erc-8006.md
```diff
@@ -0,0 +1,370 @@
...
+- **Acyclicity**: When DAG is used policies must be acyclic; cycles (including self- and transitive references) are disallowed.
+- **Graph size and depth limits**: Practical policy size is bounded by gas budgets and EVM call stack limits.
+- **Encoding overhead**: The all-`bytes` transport simplifies heterogeneous data flow but adds encode/decode overhead.
+- **Supply-chain risk**: Overall safety depends on the quality and audit status of third-party artifacts; policy owners should curate artifact supply chains, prefer pinned and audited implementations, and apply strict authorization and validation in handlers.
+
+### All-Bytes Mechanism & Artifact Initialization and Execution
+
+#### All-Bytes
+
+Artifacts, being contracts, require clearly typed method arguments. However, the Policy Handler cannot process all types uniformly. To avoid extensive ad hoc duplications, all input / output parameters SHOULD be encoded as `bytes` type before being supplied to and returned from artifacts.
```
SamWilsn: Avoid creating requirements (indicated with UPPERCASE keywords) outside of the specification section. An implementer should be able to make a compatible implementation if they were only handed that section.

SamWilsn (commented):
Your reference implementation looks very complex. Would it be possible to cut it down to a more manageable/minimal one?

ERCS/erc-8006.md
vgrabovski: Thanks for this question

I intentionally omitted an explanation of what a policy engine (or, alternatively, a "rules engine") is, since I consider it a general term that is frequently used in banking and finance and therefore not necessary to explain in detail here

This ERC and its content are inspired by those areas. And also by authorisation, governance approaches used in Dapps. To be honest, there are many, and they vary in their usage and design – starting from AccessControl and Gnosis Safe modules to what Magic does.. to what ERC-8006 does

My goal is to create the best way to write and assign policies for smart contracts and dapps dynamically

assets/erc-8006/contracts/data-structures/DAG/ContainerDAG.sol
vgrabovski: Agree about the license type – the MIT license is a good option for me.
–––
Regarding the codebase complexity: with this ERC, I am trying to provide a universal solution, which is why the codebase looks like this. I also follow OOP principles and clean code practices to keep the structure clear and the relationships between contracts (related structs, libraries, free functions, etc.) explicit

I would be very grateful if you could point out which parts of the codebase seem too complex to understand @samwilsn

I also noticed that some other ERC proposals include diagrams and tests

Would it help if I added diagrams and TypeScript + Hardhat tests directly to the assets directory?

ERCS/erc-8006.md
vgrabovski: Good suggestion, thanks for pointing that out.

ERCS/erc-8006.md
vgrabovski: Thanks, I will convert it to SVG format.

ERCS/erc-8006.md
vgrabovski: Good point, noted.

ERCS/erc-8006.md
vgrabovski: True, noted 👌

ERCS/erc-8006.md
vgrabovski: yes 🤝

ERCS/erc-8006.md
vgrabovski: I think, it is a very good suggestion

ERCS/erc-8006.md
vgrabovski: I re-read this one year later after it was written, and yes – it’s a bit muddy. The explanation is too complicated

The original idea was to allow any smart contract that inherits the `IArbitraryDataArtifact` interface to act as a rule (artifact). The `PolicyHandler` contract can implement `IArbitraryDataArtifact` and therefore become compatible as an artifact for (some other) policy

The point is that there’s no reason to prevent this behavior. In fact, in some cases of DeFi LEGO (policy LEGO, to be precise), it could be intentionally useful and deliberately chosen

vgrabovski: I also notice a misconception about “atomic” usage. This is a good reason to make edits.

In the context of policy, “atomic” refers to being atomic in terms of artifact being a single unit used as a building block. type and number of its input parameters and output parameters must not change, 

Imagine Artifact A. 
During `PolicyHandler` initialization, it is assigned that **artifact a** in method `exec(bytes[])` receives input  as `[abi.encode(address)]` and the output is `bytes` returned as `abi.encode(uint256)`. 

This artifact returns an address credibility score. The way the credibility rating is calculated may change over time, which could require upgrading the artifact’s codebase (and replacing the existing artifact with a new one). Which is okay.

What is not okay is if `[address]` is no longer supplied as input or if `uint256` is no longer returned as output.

ERCS/erc-8006.md
vgrabovski: I aimed to focus primarily on explaining what ERC does in the **Abstract** section.

Regarding what it is, I can adjust the **Abstract** to clarify where it may be useful (for example, in dapps that rely on dynamic rules as part of their business logic).

Regarding definitions, it is not possible to include them in the **Abstract** section direclty. EIP guidelines.
Will figure somthein out.

At the same time, the **Specification** section already contains a definition of a `policy` and its relationship to `artifacts` (`rules`). Would this be enough?

I will also add a definition of a `Policy Engine` to the **Specification** section.

ERCS/erc-8006.md
vgrabovski: I'll add more about policy engine (perhaps in the **_Specification section_**)

ERCS/erc-8006.md
vgrabovski: This is required for compatibility with the off-chain tooling we have built, but I will reconsider whether it could be optional

ERCS/erc-8006.md
vgrabovski: Sorry @SamWilsn, but it is difficult to compare this with an ERC that does not exist yet.

Instead, I can highlight in ERC content some of the features that this standard aims to provide:

1. Universality of use – it can help any contract that requires complex, large, or dynamic validation logic on top of its core business logic by providing a standardized way to separate method validation logic (policy rules) from the core business logic.
2. Simple and clear policy maintenance techniques (deploy, attach, update)
   1. including easier initial integration and rule updates without the mandatory use of an upgrade pattern
4. Reusability of components across multiple contracts, similar to how OpenZeppelin libraries are used, with support for both inheritance and composition approaches
5. Simplified and more focused testing of new or adjusted policy behavior
6. Based on all of the above, – reducing the overall complexity of the codebase and enabling a cleaner architecture.

hilljessica8787-dot (approved):

vgrabovski:
Hey @abcoathup, what else I need to do to move this forward? Should I [get a review](https://github.com/ethereum/ERCs/pull/1179#issuecomment-3200118265) from @g11tech and @jochem-brouwer?

abcoathup:
@vgrabovski-gl you need an ERC editor to review, which I am not. There is only a few of them so it does take some time. 

Make sure there are no validator issues. Get your fave AIs to review too. 

There is a regular office hours including ERCs. Check your ERC is on the agenda.

vgrabovski:
@abcoathup two of them have already left comments. I have addressed their suggestions or replied clarifying what was done and why. Some suggestions will be solved with new commits in a few days.

Do I need to get reviews from all 4 reviewers?

I will explore the agenda (I believe I should check [here](https://github.com/ethereum/pm)). Thank you!
___
Also, it is alerting a validator issue, but I don't understand what the reason is (https://github.com/ethereum/ERCs/actions/runs/19238448349/job/54994622357?pr=1179)

jochem-brouwer:
Hi @vgrabovski-gl, we have a backlog and try to review ERCs per longest-waited as seen on eipsinsight.com, this PR is for some reason not on there. 

I am closing and re-opening your PR to retrigger HTMLProofer so we can see what went wrong.

ERCS/erc-8006.md
```diff
@@ -0,0 +1,370 @@
...
+When implementing compliance conditions, developers face several important challenges:
+
+1. **Code Complexity**: Beyond simple value thresholds, compliance logic becomes increasingly complex. This complexity makes codebases significantly harder to maintain, understand, and audit effectively. The intricate nature of compliance requirements often results in convoluted implementations that obscure the underlying business logic.
+
+2. **Security Vulnerabilities**: Flaws in rule implementation pose serious risks to system integrity and user assets. Poor implementations can enable unauthorized access to restricted functionality, lead to direct loss of user funds through exploits, and potentially violate regulatory requirements that the rules were meant to enforce.
+
+3. **Integration Difficulties**: Adding new compliance logic to existing codebases often requires:
+   - Extensive codebase changes
+   - Separate integration work for each compliance rule implementation
+   - The need of carefully conducted contract upgrade process (which is error-prone)
```
jochem-brouwer: Grammar:
```suggestion
   - The need for a carefully conducted contract upgrade process (which is error-prone)
```

ERCS/erc-8006.md
```diff
@@ -0,0 +1,370 @@
...
+
+4. **Redundant Implementation**: Many compliance patterns appear across different projects but lack standard interfaces, forcing developers to repeatedly implement similar functionality.
+
+5. **Costly Updates**: As regulatory requirements change, contracts need updates, requiring the entire development process again - writing, testing, auditing, and deploying - increasing costs and error potential.
+
+## Specification
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
```
jochem-brouwer: Link the RFCs (rfc-editor.org):
```suggestion
The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC 2119](https://www.rfc-editor.org/rfc/rfc2119) and [RFC 8174](https://www.rfc-editor.org/rfc/rfc8174).
```

ERCS/erc-8006.md
```diff
@@ -0,0 +1,370 @@
...
+
+Artifacts, being contracts, require clearly typed method arguments. However, the Policy Handler cannot process all types uniformly. To avoid extensive ad hoc duplications, all input / output parameters SHOULD be encoded as `bytes` type before being supplied to and returned from artifacts.
+
+This approach allows each variable (artifact input) or result (artifact output) to be processed as bytes by both the Policy Handler and off-chain code, thereby significantly simplifying implementation. The «all-bytes» mechanism provides a unified method for data transfer regardless of the underlying type, resolving challenges related to heterogeneous data handling.
+
+#### Initialization and Execution
+
+Artifacts have two primary methods: `init` and `exec`.
+
+As declared in [Specification](#specification) section, stateful artifacts maintain a clean and independent storage state. Therefore, the `init` method must be called once for each stateful artifact during policy initialization by the Policy Handler. As result, a separate copy of the artifact must be created. This standard recommends a gas-optmized and suitable copying strategy [ERC-1167](./erc-1167.md).
```
jochem-brouwer: Repo link convention is `./eip-N.md`, and `gas-optmized` → `gas-optimized`:
```suggestion
As declared in [Specification](#specification) section, stateful artifacts maintain a clean and independent storage state. Therefore, the `init` method must be called once for each stateful artifact during policy initialization by the Policy Handler. As result, a separate copy of the artifact must be created. This standard recommends a gas-optimized and suitable copying strategy [ERC-1167](./eip-1167.md).
```

ERCS/erc-8006.md
```diff
@@ -0,0 +1,370 @@
...
+Additionally, a Policy is subject to updates, meaning the number of nodes may change after an update.
+
+Policy with the same conditions can also be implemented at different levels of granularity, resulting in a greater or smaller number of graph nodes for same policy variants. The chosen granularity level is determined by the Policy Handler developer.
+
+Therefore, the cost of invoking **evaluate** may vary from a minimum cost threshold to a maximum cost (bounded by Ethereum Virtual Machine call stack limits), depending on the current graph state. To manage this, a special utility or dry-run method could be implemented for the **evaluate** function to estimate and report the potential cost range before actual execution.
+
+## Backwards Compatibility
+
+No backward compatibility issues have been identified.policy engine in various scenarios, serving as both verification of the standard and a learning resource for implementers.
```
jochem-brouwer: This line is garbled (missing space + an orphaned fragment that reads like leftover Reference-Implementation text). Trim to the single sentence; relocate the fragment if it was meant for Reference Implementation:
```suggestion
No backward compatibility issues have been identified.
```

ERCS/erc-8006.md
```diff
@@ -0,0 +1,370 @@
...
+## Security Considerations
+
+### Delegated Security
+
+This standard introduces and encourages the use of a **Delegated Security Model**, in which security responsibilities are distributed across multiple roles. Such an approach decouples security logic from a single core codebase, simplifying the overall policy lifecycle including development, creation, upgrades.
+
+Expected roles and responsibilities:
+- **Policy Owner**: Responsible for curating, pinning, and updating approved artifact sets for off-chain or on-chain consumption during policy creation.
+- **Policy Handler Developer**: Responsible for enforcing acyclicity, authorization, and correct initialization of both the handler and its associated nodes.
+- **Artifact Developer**: Responsible for developing, conducting security audits, and maintaining their artifacts before making them publicly available. Unvetted or upgradable artifacts can introduce risks; see notes on *untrusted artifacts* and *statefull initialization problem* below.
```
jochem-brouwer: `statefull` → `stateful`:
```suggestion
- **Artifact Developer**: Responsible for developing, conducting security audits, and maintaining their artifacts before making them publicly available. Unvetted or upgradable artifacts can introduce risks; see notes on *untrusted artifacts* and *stateful initialization problem* below.
```

ERCS/erc-8006.md
```diff
@@ -0,0 +1,370 @@
...
+Such an unreliable artifact contract might contain unknown or unaudited code, code vulnerable to known attack vectors, or even modified logic (for example, introduced through an unauthorized call to an upgradable proxy, if artifact implemented as one).
+
+2. **Statefull Artifact Initialization Problem**: Stateful artifacts require clean state isolation between different policy instances. To achieve this, the Policy Handler must create new, isolated instances for each stateful artifact using approaches like minimal proxies (e.g., [ERC-1167](./erc-1167.md)). Improper implementation of this state isolation mechanism can lead to unexpected behavior – developers may encounter un-initialized artifacts or empty storage slots where data was expected or slots containing data from other policies when they should be empty. While developers may implement their own approach for state isolation, it is strongly recommended to use established patterns like ERC-1167 to avoid such corruption. The reference implementation uses minimal proxies with delegatecall functionality.
```
jochem-brouwer: `Statefull` → `Stateful`, and repo link convention is `./eip-N.md`:
```suggestion
2. **Stateful Artifact Initialization Problem**: Stateful artifacts require clean state isolation between different policy instances. To achieve this, the Policy Handler must create new, isolated instances for each stateful artifact using approaches like minimal proxies (e.g., [ERC-1167](./eip-1167.md)). Improper implementation of this state isolation mechanism can lead to unexpected behavior – developers may encounter un-initialized artifacts or empty storage slots where data was expected or slots containing data from other policies when they should be empty. While developers may implement their own approach for state isolation, it is strongly recommended to use established patterns like ERC-1167 to avoid such corruption. The reference implementation uses minimal proxies with delegatecall functionality.
```

ERCS/erc-8006.md
```diff
@@ -0,0 +1,370 @@
...
+    function description() external pure returns (string memory desc);
+}
+```
+
+
+Other interfaces are more implicit, following conventions and approaches explained in the Rationale section.
+
+### Policy Handler
+
+The Policy Handler orchestrates artifact execution. While implementations may vary, this standard RECOMMENDS a **Directed Acyclic Graph (DAG)** as the reference architecture (see [reference implementation](../assets/erc-8006/contracts/inheritance/DAGWithPolicyMetadata.sol) and [Policy Handler](../assets/erc-8006/contracts/PolicyHandler.sol)).
```
jochem-brouwer:
```suggestion
The Policy Handler orchestrates artifact execution. While implementations may vary, this standard RECOMMENDS a **Directed Acyclic Graph (DAG)** as the reference architecture (see [reference implementation](../assets/eip-8006/contracts/inheritance/DAGWithPolicyMetadata.sol) and [Policy Handler](../assets/eip-8006/contracts/PolicyHandler.sol)).
```

This fixes HTMLProofer (1/?)

ERCS/erc-8006.md
```diff
@@ -0,0 +1,370 @@
...
+
+For each node during the **traversal**, the handler:
+1. Calls the target artifact using the standard interface
+2. Supplies it with the appropriate variables and/or results from previously evaluated (calculated) nodes that the current node depends on
+3. Collects the output values of the artifacts for use by parent nodes
+4. Finishes the traversal once all nodes have been visited and evaluated
+
+In summary, the handler passes outputs from child nodes as inputs to parent nodes; artifacts do not invoke each other directly. After the traversal completes, the root node’s output value represents the policy’s final result. This design provides a structured mechanism for rule interaction, addressing the challenge of composing complex rules from simpler components.
+
+![Policy bird's eye view](../assets/erc-8006/simple_policy_bird_eye_view.svg)
```
jochem-brouwer:
```suggestion
![Policy bird's eye view](../assets/eip-8006/simple_policy_bird_eye_view.svg)
```

This fixes HTMLProofer (2/?)

ERCS/erc-8006.md
```diff
@@ -0,0 +1,370 @@
...
+
+Therefore, the cost of invoking **evaluate** may vary from a minimum cost threshold to a maximum cost (bounded by Ethereum Virtual Machine call stack limits), depending on the current graph state. To manage this, a special utility or dry-run method could be implemented for the **evaluate** function to estimate and report the potential cost range before actual execution.
+
+## Backwards Compatibility
+
+No backward compatibility issues have been identified.policy engine in various scenarios, serving as both verification of the standard and a learning resource for implementers.
+
+## Reference Implementation
+
+A reference implementation of Policy Handler is available [here](../assets/erc-8006/contracts/PolicyHandler.sol).
```
jochem-brouwer:
```suggestion
A reference implementation of Policy Handler is available [here](../assets/eip-8006/contracts/PolicyHandler.sol).
```

This fixes HTMLProofer (3/3)

ERCS/erc-8006.md
```diff
@@ -0,0 +1,370 @@
...
+}
+```
+
+This separation of concerns allows both human users and automated systems to properly interact with artifacts, supporting both manual integration and programmatic composition of policies. These descriptive capabilities make policies more transparent and easier to work with, solving discoverability and comprehension challenges.
+
+### Policy Handler Architecture Considerations
+
+The graph-based approach is optimal for artifact orchestration, offering elegant recursive traversal capabilities, memory efficiency, and support for dynamic modifications. Artifacts function naturally as node logic within this structure, with the handler defining additional node properties and relationships.
+
+Another notable architecture decision is a public **variables-list** getter method exposed on the handler. It improves off-chain orchestration and automation – for example, by enabling the precomputation of inputs and the wiring UI forms or client pipelines.
```
jochem-brouwer:
```suggestion
Another notable architecture decision is a public **variables-list** getter method exposed on the handler. It improves off-chain orchestration and automation – for example, by enabling the precomputation of inputs and the wiring User Interface (UI) forms or client pipelines.
```

Not sure if correct, but define the abbreviations before use.

ERCS/erc-8006.md
```diff
@@ -0,0 +1,370 @@
...
+
+Implementation details of Policy Handler regarding interface extensions and argument formatting remain at the discretion of the handler developer.
+
+The graph-based architecture imposes several important constraints:
+
+1. **Acyclicity requirement**: Policies MUST NOT contain cycles. Nodes cannot refer to each other cyclically in any form, including self-references and transitive references.
+
+2. **EVM call stack limitations**: Since each node traversal involves a contract call to an artifact, policy size is bounded by Ethereum Virtual Machine call stack limits.
+
+3. **Connectivity requirement**: Only nodes connected (directly as children nodes or indirectly as decendant nodes) to the root node will be traversed during evaluation. It is RECOMMENDED that graphs with disconnected nodes or subgraphs MUST be rejected during policy creation stage.
```
jochem-brouwer: This sentence contradicts itself if these are RFC 2119 keywords, something cannot be RECOMMENDED and MUST. Also, this is not the specification section, so lowercase it or move this to the specification section if it is normative.

ERCS/erc-8006.md
jochem-brouwer: Also fixes HTMLProofer

ERCS/erc-8006.md
jochem-brouwer: Also fixes HTMLProofer

jochem-brouwer (changes requested):
## Editorial review (Draft)

Editorial review only (structure / readability / EIP-1), not a technical assessment. @xinbenlv (changes requested) and @SamWilsn have already reviewed; most of their inline comments are still open, please address those.

**Blockers**
- **Asset licensing** (`@SamWilsn`): a `.sol` file was flagged as "Unlicensed". All in-repo code must be CC0-1.0 (MIT/Apache acceptable in assets) — please ensure no asset file is "Unlicensed".

HTMLProofer on CI is failing, I included suggestions for the fix.

EDIT: you need one approval from an editor, so not 4 :wink:

ERCS/erc-8006.md
```diff
@@ -0,0 +1,370 @@
+---
+eip: 8006
+title: Universal Policy Engine
+description: Composite on-chain policy engine utilizing generic rule artifacts within DAG evaluator
+author: Vlad Pryadko (@VladPryad) <vladyslav.priad@gmail.com>, Vitali Grabovski (@vitaligrabovski1) <v.grabovski@guardianlabs.org>
```
jochem-brouwer: None of the @ handles in this `author` list matches the user who opened this PR. Is this intended @vpriadko? I would assume that you are the first author? 

It is important to fix the GitHub username, as otherwise once this ERC gets merged you will not have merge rights for updates in the future.

```suggestion
author: Vlad Pryadko (@vpriadko) <vladyslav.priad@gmail.com>, Vitali Grabovski (@vitaligrabovski1) <v.grabovski@guardianlabs.org>
```

vgrabovski: This is intended. Didn't know it has to match. Will change it back.

ERCS/erc-8006.md
vgrabovski: Thank you

ERCS/erc-8006.md
vgrabovski: Thank you

assets/erc-8006/contracts/data-structures/DAG/ContainerDAG.sol
vgrabovski: License is now MIT

hilljessica8787-dot (approved):

vgrabovski:
Hi @abcoathup ,

The word spelling validation is too good. This seems to be a new behavior because it is now flagging something that has been unchanged in the erc content for over a year.

It also flags proper names and the word “orchestrator,” which is an actual word according to Merriam-Webster and many others forcing to use z insted of s.
https://github.com/ethereum/ERCs/actions/runs/30678635372/job/91310927855?pr=1179
Could you help clarify why this is happening?

abcoathup:
> Could you help clarify why this is happening

Sorry I don't know as I haven't ever looked at that. Suggest asking your fave AI tool. 

Alternatively ask in the ECH Discord EIP channel or the regular EIP/ERC office hours.

vgrabovski:
> > Could you help clarify why this is happening
> 
> Sorry I don't know as I haven't ever looked at that. Suggest asking your fave AI tool.
> 
> Alternatively ask in the ECH Discord EIP channel or the regular EIP/ERC office hours.

Thank you for the suggestion, @abcoathup. I will. The only reason I asked you is that the commits were green earlier this week, and after I added new commits, they turned red, flagging words that I hadn't touched during this green/red ping-pong
Sorry if I misunderstood your role or if this isn't something you usually deal with.

abcoathup:
> Sorry if I misunderstood your role

I'm not an ERC editor. I assign numbers to help speed up the process. 

@SamWilsn may be able to assist.

vgrabovski:
I've found the reason: spelling check enabled (https://github.com/ethereum/ERCs/commit/85e1a2921e42cfb3aa53b51b2aa65d2c05c81290)

spelling check disabled (https://github.com/ethereum/ERCs/commit/165de186e1f9340139bbef5a9e62bd085f9c4d05)

Was useful though – I do have several misspelt words, but the EIP Walidator did false positives too.

ERCS/erc-8006.md
vgrabovski: applied edtis to avoid misconceptions

assets/erc-8006/contracts/data-structures/DAG/ContainerDAG.sol
vgrabovski: Code is also much simplified and includes only that part needed to demonstrate how to declare and use a purely onchain policy.

PR #1180, opened 2025-08-19
Update ERC-7955: Fix DUP instruction

frangio:
The bytecode is correct but the assembly has an error.


nlordell:
Ah - thanks! Nice catch.

nlordell (approved):

PR #1184, opened 2025-08-21
Add ERC: Proxy Clear Signing

markvrc:
This standard introduces an alternative solution to the blind-signing problem on dedicated hardware devices, by routing transactions through a proxy contract which enforces transaction constraints. The solution is split into core/periphery contracts, with the core contract being the standard.

[Discussion](https://ethereum-magicians.org/t/erc-8009-proxy-clear-signing/25199)

### DIFF
--- a/ERCS/erc-8009.md
+++ b/ERCS/erc-8009.md
@@ -1,0 +1,200 @@
+---
+eip: 8009
+title: Proxy Clear Signing
+description: Permissionless singleton contract that proxies smart-contract calls and enforces transaction constraints specified by the end user
+author: Mark Virchenko (@borseno), Ilya Kubariev (@gymnasy55), Daniel Nagy (@nagydani)
+discussions-to: https://ethereum-magicians.org/t/erc-8009-proxy-clear-signing/25199
+status: Draft
+type: Standards Track
+category: ERC
+created: 2025-08-19
+requires: 20
+---
+
+## Abstract
+
+Introduce a singleton, stateless contract that proxies arbitrary smart‑contract calls and enforces user‑supplied outcome constraints. The proxy supports **absolute post‑balance thresholds** and **balance‑difference (delta) checks** over native ETH and [ERC-20](./eip-20.md) balances. Calls that violate the declared constraints are reverted.
+
+## Motivation
+
+Current hardware wallets struggle to present meaningful transaction information when interacting with unknown or newly deployed smart contracts. Without prior knowledge of a contract’s ABI or address, these devices can only display raw hexadecimal calldata - an unreadable and unsafe experience for most users. This practice, known as blind signing, leaves users vulnerable to unintentionally approving malicious or incorrect transactions. This vulnerability has already been exploited in high-value attacks.
+
+Instead of parsing transaction calldata, the proposed approach focuses on verifying expected balance differences after execution. This proposal introduces an alternative path to clear signing - defined here as the ability for users to verify the outcome of a transaction before signing, without requiring ABI knowledge. By routing calls through a singleton proxy contract that enforces these expectations on-chain, hardware wallets can reliably display the “balance after transaction” values to the user, taken from the smart-contract parameters. This enables a permissionless, scalable, and understandable transaction confirmation experience, even for unknown contracts.
+
+## Specification
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
+
+### Contract Overview
+
+A singleton, stateless permissionless proxy contract deployed once per network.
+
+The contract accepts transaction execution requests containing:
+
+1. **Target call data** — Encoded calldata for the intended interaction.
+2. **Approvals** — Optional [ERC-20](./eip-20.md) instructions executed **before** the target call to either approve the call target to spend proxy-held ERC-20 tokens or transfer proxy-held tokens directly to the call target.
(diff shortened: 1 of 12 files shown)


ERCS/erc-xxxx.md
```diff
@@ -0,0 +1,299 @@
+---
+eip: xxxx
```
```suggestion
eip: 8009
```

ERCS/erc-xxxx.md
```diff
@@ -0,0 +1,299 @@
...
+title: Proxy Clear Signing
+description: Permissionless singleton contract that proxies smart-contract calls and enforces transaction constraints specified by the end user
+author: Mark Virchenko (@borseno), Ilya Kubariev (@gymnasy55), Daniel Nagy (@nagydani)
+discussions-to: https://ethereum-magicians.org/t/new-erc-proxy-clear-signing/25199
```
```suggestion
discussions-to: https://ethereum-magicians.org/t/erc-8009-proxy-clear-signing/25199
```
Added assigned number

ERCS/erc-8009.md
```diff
@@ -0,0 +1,199 @@
...
+discussions-to: https://ethereum-magicians.org/t/erc-8009-proxy-clear-signing/25199
+status: Draft
+type: Standards Track
+category: ERC
+created: 2025-08-19
+---
+
+## Abstract
+
+Introduce a singleton, stateless contract that proxies arbitrary smart‑contract calls and enforces user‑supplied outcome constraints. The proxy supports **absolute post‑balance thresholds** and **balance‑difference (delta) checks** over native ETH and [ERC-20](./eip-20.md) balances. Calls that violate the declared constraints MUST revert.
```
SamWilsn:
```suggestion
Introduce a singleton, stateless contract that proxies arbitrary smart‑contract calls and enforces user‑supplied outcome constraints. The proxy supports **absolute post‑balance thresholds** and **balance‑difference (delta) checks** over native ETH and [ERC-20](./eip-20.md) balances.
```

Don't create requirements (UPPERCASE keywords) outside of the specification section.

ERCS/erc-8009.md
```diff
@@ -0,0 +1,199 @@
...
+   - **Balance‑difference rules:** minimum required deltas between pre‑ and post‑execution balances.
+
+The contract MUST NOT store persistent state. Multiple independent users MAY reuse the same instance without interference.
+
+### BalanceProxy
+
+The `BalanceProxy` MUST implement the following interface:
+
+```typescript
+// SPDX-License-Identifier: UNLICENSED
```
SamWilsn: Code within the EIP itself must be CC0-1.0. If you'd like a different license, you can put the file in your assets directory. In particular, UNLICENSED makes no sense for an ERC. You want people to use your code, right?

ERCS/erc-8009.md
```diff
@@ -0,0 +1,199 @@
...
+
+This EIP introduces a new contract standard and does not modify existing standards. No backwards-compatibility issues are expected.
+
+## Test Cases
+
+Test cases can be found [here](../assets/eip-8009/BalanceProxy.test.ts).
+
+## Reference Implementation
+
+A reference implementation of the **Core** `BalanceProxy` contract can be found [here](../assets/eip-8009/BalanceProxy.sol) and reference implementations of the **Periphery** contract `ApproveRouter` and `PermitRouter` can be found [here](../assets/eip-8009/ApproveRouter.sol) and [here](../assets/eip-8009/PermitRouter.sol) respectively.
```
SamWilsn: If this is a singleton, won't there only ever be the one implementation? Or do you intend for there to be multiple implementations that all share the same interface?

assets/erc-8009/ApproveRouter.sol
```diff
@@ -0,0 +1,191 @@
+// SPDX-License-Identifier: UNLICENSED
```
SamWilsn: We can't accept UNLICENSED code.

assets/erc-8009/BalanceProxy.sol
```diff
@@ -0,0 +1,133 @@
+// SPDX-License-Identifier: UNLICENSED
```
SamWilsn: Same

assets/erc-8009/PermitRouter.sol
```diff
@@ -0,0 +1,253 @@
+// SPDX-License-Identifier: UNLICENSED
```
SamWilsn: Same

SamWilsn (changes requested):
We generally don't like singleton contract proposals. If there's only one contract, why do you need an ERC? No one needs to coordinate, they can just use the contract itself.

markvrc:
Hey @SamWilsn, thank you for your valuable comments.
We have recently added a SafeRouter to the solution, which allows users to apply proxy clear signing from a multisig wallet. This means it now fully covers the Bybit scenario, where blind signing on a hardware wallet occurred within a multisig transaction.
On the singleton question: we wanted to push this as a standard because blind signing affects all modern hardware wallets, and each vendor will likely have their own preferences for the peripheral contract, so having a standard here would definitely help achieve a unified approach to balance-based clear-signing.

A single well-known deployment address also simplifies hardware wallet firmware integration: vendors can hardcode one address rather than making users deploy one instance per one address. To avoid fragmentation and give everyone a clear reference point, a standard feels like the right approach.

One instance per network is exactly the approach intended.

ERCS/erc-8009.md
markvrc: done

ERCS/erc-8009.md
markvrc: done

ERCS/erc-8009.md
markvrc: answered on the comment to this PR

assets/erc-8009/ApproveRouter.sol
markvrc: done

hilljessica8787-dot (approved):

PR #1189, opened 2025-08-23
Update ERC-7656: Move to Last Call

sullof:
As suggested by @xinbenlv I made a PR for all the little changes and this new one just to move the ERC to Last Call.


sullof:
> It sounds like you're trying to discourage authors from modifying existing standards, and that isn't really something we do in proposals

There a lot of ERC that require changes in the ERC-721 standard to add specific features. The big issue there is that it works only with new deployed contracts or with upgradeable contracts. Using something like ERC-7656 someone can establish a standard for a service associated to an ERC-721 that adds the required changes. For example, imagine that instead of changing ERC-721 to add data about a physical asset, you can define a standard service, deployed using ERC-7656, that adds all the data related to the physical asset. The advantage is that, even existing NFTs can be used according to the new standard. I believe that it could be a smart approach that does not penalize all the existing NFTs, but I am not suggesting that people stop creating ERCs, just that that could be a way to avoid suggesting changes to the NFT standard and instead create a standard that uses a service for the NFT. I will try to make it more clear.

sullof:
> In your security considerations section, you're using the UPPERCASE keywords. Those keywords create requirements, and should only be used in the Specification section. Just switch them to lowercase.

Done. Thanks.

sullof:
> This is a bit of a philosophical goal

I changed it to a more technical consideration.

PR #1192, opened 2025-08-27
 Update ERC-6909: Move to Last Call

z0r0z:
Reasoning: The discussion in the official discussion forum has settled, only window-dressing niceties remain, and there are existing deployments using ERC-6909 interface (Uniswap, 0xSplits, zAMM). At the risk of a production interface being deemed stagnant because of procedure unrelated to application layer, this ERC should moved to Last Call.

https://ethereum-magicians.org/t/erc-6909-multi-token-standard/13891/77?u=z0r0z


xinbenlv (approved):

xinbenlv:
congratulations

Zodomo:
Yay!

PR #67, opened 2023-10-30
Add ERC: Dynamic Compliant Interop Security Token

rajatwasan:
Dynamic Compliant Interop Security Token(DyCIST) is a proposed token standard that extends ERC-1155 and combines its functionality with other existing token standards, introducing unique features for interoperability, cross-chain operation, and token wrapping.

PR #1204, opened 2025-09-09
Update ERC 7518: Move to Review

rajatwasan:
Reference implementation: https://github.com/zoniqx/erc7518_reference
Discussion: https://ethereum-magicians.org/t/erc-7518-dynamic-compliant-interop-security-token-dycist
Supersedes: https://github.com/ethereum/ERCs/pull/67


rajatwasan:
> * Line 21: Somewhat awkward to start a sentence with "Build upon of". Perhaps "This standard builds upon ERC-1155 by introducing partitions..."
Updated to “This standard builds upon ERC-1155 by introducing partitions…”.

> * Line 181: "ie current ..." -> "(i.e. current ...)"
Corrected

> * Line 212: "MUST check if ..." doesn't actually have an observable effect. Does it revert if it wasn't previously restricted?
transfer restriction checks revert if conditions are not met.

> * Line 305: I don't think a contract can "consider implications". Is this an instruction to implementers or users, and if so, it should be in Security Considerations.
Moved language on “consider implications” to Security Considerations

> * Line 351+: ERC1155 -> ERC-1155
Standardized to “ERC-1155”.

> * Lines 461-466: This content belongs in Motivation, because it is providing justification for the standard as a whole. The specification section should exclusively describe the observable behaviour of the implementation.
Moved justification to Motivation.

> * Line 470: We prefer an SVG if you can provide one, but a PNG is fine if you cannot.
Replaced
> * Line 486: Similarly, this looks like Motivation to me.
Restructured
> * Lines 496-540: Definitely Motivation. While examples can exist in the Specification section to demonstrate how an implementation should behave, these examples are too high level to be useful to implementers in that way.
Restructured

ERCS/erc-7518.md
```diff
@@ -18,12 +18,67 @@ This proposal is a security token standard that extends [ERC-1155](./eip-1155.md
 
 The growing demand for tokenized real-world assets necessitates a token standard that can accommodate the unique requirements of security tokens. Existing standards, while powerful, do not fully address the need for flexible partitioning and comprehensive compliance management.
 
-Build upon of [ERC-1155](./eip-1155.md) to introduce partitions, allowing for the creation of semi-fungible tokens representing fractional ownership, different share classes, or other distinct units within a single token contract. This flexibility is crucial for tokenizing complex real-world assets like real estate or funds.
+This standard builds upon [ERC-1155](./eip-1155.md) by introducing partitions, allowing for the creation of semi-fungible tokens representing fractional ownership, different share classes, or other distinct units within a single token contract. This flexibility is crucial for tokenizing complex real-world assets like real estate or funds.
```
xinbenlv, line 21: The functionality of "fractionalization" is not very obvious from the description of this ERC.

Also for maximal interoperabibility please research to find existing ERCs that touch the fractionalization, and discuss within this ERC to 
1. adopt that ERC
2. or, intentionally choose to disregard that ERC and use a new interface/approach, and *rationale* of such decision

ERCS/erc-7518.md
```diff
@@ -4,7 +4,7 @@ title: Dynamic Compliant Interop Security Token
 description: Security token framework with semi-fungible partitions for dynamic regulatory compliance management and cross-chain interoperability
 author: Abhinav (@abhinav-d3v) <abhinav@zoniqx.com>, Prithvish Baidya (@d4mr) <pbaidya@zoniqx.com>, Rajat Kumar (@rajatwasan) <rwasan@zoniqx.com>, Prasanth Kalangi <pkalangi@zoniqx.com>
 discussions-to: https://ethereum-magicians.org/t/eip-7518-dynamic-compliant-interop-security-token-dycist/15822
-status: Draft
+status: Review
```
xinbenlv, line 7: Another feedback is to provide reference implementations. For example, it's not very clear how the partition works  based on the specification which only describes
```

/**
* @notice Wraps a specified amount of tokens from a given partition into the main balance.
* @dev This function allows users to convert tokens from a specific partition back to the main balance,making them fungible with tokens from other partitions.
* @param partitionId The unique identifier of the partition from which tokens will be wrapped.
* @param id The unique identifier of the token.
* @param amount The amount of tokens to be wrapped from the specified partition.
* @param data Additional data that may be used to handle the wrap process (optional).
* @return success A boolean indicating whether the wrapping operation was successful or not.
*/
```

And it's not very easy to find from the ref-impl you shared https://github.com/zoniqx/erc7518_reference

rajatwasan: Added in the reference: https://github.com/zoniqx/erc7518_reference/pull/3/files

ERCS/erc-7518.md
rajatwasan: I’ve updated the file to clarify the “fractionalization” functionality within the specification and motivation sections.

ERCS/erc-7518.md
```diff
@@ -18,12 +18,67 @@ This proposal is a security token standard that extends [ERC-1155](./eip-1155.md
...
+
+4. Transfer of Ownership: The ownership of each floor's NFT partition can be transferred seamlessly to token holders using the safeTransferFrom function. This enables the seamless transfer of ownership rights for specific floors.
+
+5. Compliance Management: Different compliance rules and transfer restrictions can be applied to each partition (floor) based on regulatory requirements or issuer-defined rules. The canTransfer function can be used to enforce these rules before allowing transfers.
+
+6. Payouts: The payout and batchPayout functions can be used to distribute rental income, dividends, or other payouts to token holders of specific floor partitions efficiently.
+
+By leveraging proposal, this use case demonstrates the ability to tokenize complex real estate assets while maintaining granular control over ownership, pricing, compliance, and payouts for individual units within the property.
```
SamWilsn, line 56:
```suggestion
By leveraging this proposal, this use case demonstrates the ability to tokenize complex real estate assets while maintaining granular control over ownership, pricing, compliance, and payouts for individual units within the property.
```

ERCS/erc-7518.md
```diff
@@ -18,12 +18,67 @@ This proposal is a security token standard that extends [ERC-1155](./eip-1155.md
...
+
+#### Use Case 2: Tokenization of Securities with Reg S and Reg D Partitions
+
+In this use case, a company is tokenizing its securities and wants to comply with different regulations for U.S. accredited investors (Reg D) and non-U.S. investors (Reg S).
+
+1. Initial Partitions: The company deploys an proposed standard and creates two partitions: one for Reg D investors (accredited U.S. investors) and another for Reg S investors (non-U.S. investors).
+
+2. Dynamic Allocation: As the offering progresses, the company can dynamically mint tokens into the appropriate partition based on investor eligibility. For example, if a U.S. accredited investor wants to participate, tokens can be minted in the Reg D partition, while tokens for non-U.S. investors are minted in the Reg S partition.
+
+3. Compliance Management: Each partition can have its own set of compliance rules and transfer restrictions. The canTransfer function can be integrated with off-chain compliance services to verify the eligibility of a transfer based on the specific rules for each partition.
```
SamWilsn, line 66:
```suggestion
3. Compliance Management: Each partition can have its own set of compliance rules and transfer restrictions. The `canTransfer` function can be integrated with off-chain compliance services to verify the eligibility of a transfer based on the specific rules for each partition.
```

ERCS/erc-7518.md
```diff
@@ -18,12 +18,67 @@ This proposal is a security token standard that extends [ERC-1155](./eip-1155.md
...
+
+4. Temporary Non-Fungibility: During the initial offering period, tokens in the Reg D and Reg S partitions may need to be treated as non-fungible due to different regulatory requirements. However, after the holding period, the company can create a new joint partition and allow token holders to deposit their old partitioned tokens to receive the new joint partition tokens, merging the two classes.
+
+5. Payouts: The payout and batchPayout functions can be used to distribute dividends, interest payments, or other payouts to token holders in each partition based on their respective rights and privileges.
```
SamWilsn, line 70:
```suggestion
5. Payouts: The `payout` and `batchPayout` functions can be used to distribute dividends, interest payments, or other payouts to token holders in each partition based on their respective rights and privileges.
```

Use backticks (\`) when using inline code.

SamWilsn (approved):
- Not a requirement, but you might want to define what a "security token" is somewhere early on in your abstract.
- Line 225/280: requiring that the function call `canTransfer` specifically is an implementation detail. A compliant contract should be able to implement the checks however it wants, as long as the observable behaviour is correct. In this case, you could say something like "MUST revert if `canTransfer` would return false, and MUST NOT revert otherwise".

PR #1208, opened 2025-09-11
Update ERC-6909: Move to Final

z0r0z:


ERCS/erc-6909.md
```diff
@@ -4,8 +4,7 @@ title: Minimal Multi-Token Interface
 description: A minimal specification for managing multiple tokens by their id in a single contract.
```
SamWilsn, line 4: Ideally we'd omit "minimal specification" from the description, because it's subjective and doesn't really add any meaningful information, but because we're already at last call, I'll let it slide.

PR #1213, opened 2025-09-16
Make ERC-7913 final

Amxx:

 - The build passes.


Amxx:
@SamWilsn 

When using a ERC-7913 verifier contract, the key must not be empty. However, one could use the "identity representation" that is `address ++ key` with an empty key to represent "old" ECDSA and ERC-1271. This is only possible because of this specific construction that ERC-7913 verifiers don't support empty keys, and so there is no ambiguity between "this identity should be interpreted as ERC-1271" versus "this identity should be interpreted as ERC-7913 with an empty key"

SamWilsn (approved):

PR #1235, opened 2025-10-02
Update ERC-7836: Remove array request / response

lukasrosario:
- remove arrays from requests / responses. these endpoints would not expect multiple entries in these fields.

### DIFF
--- a/ERCS/erc-7291.md
+++ b/ERCS/erc-7291.md
@@ -4,8 +4,7 @@ title: Purpose bound money
 description: An extension to ERC-1155 implementing a hybrid form of programmable payment and money.
 author: Orchid-Dev (@proj-orchid-straitsx), Victor Liew (@alcedo), Wong Tse Jian (@wongtsejian), Jacob Shan (@Jacobshan429), Chin Sin Ong (@chinsinong), Praveen Kumar (@veenkumarr)
 discussions-to: https://ethereum-magicians.org/t/eip-7291-purpose-bound-money/14973
-status: Last Call
-last-call-deadline: 2025-06-10
+status: Final
 type: Standards Track
 category: ERC
 created: 2023-06-24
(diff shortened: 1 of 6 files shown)


ERCS/erc-7836.md
```diff
@@ -36,7 +36,7 @@ Instructs a Wallet to prepare a call bundle according to the Account's implement
 ```typescript
 type Request = {
   method: 'wallet_prepareCalls',
-  params: [{
+  params: {
```
azf20: I think to be consistent with 5792 this should still be an object in an array, but not an array of objects?

ERCS/erc-7836.md
```diff
@@ -149,7 +149,7 @@ Instructs a Wallet to execute a prepared call bundle (response of `wallet_prepar
 ```typescript
 type Request = {
   method: 'wallet_sendPreparedCalls',
-  params: [{
+  params: {
```
azf20: I think this should still be an array - see this example https://github.com/ethereum/EIPs/blob/master/EIPS/eip-5792.md#wallet_sendcalls-example-parameters

ERCS/erc-7836.md
```diff
@@ -182,15 +182,15 @@ type Request = {
 type Response = {
   id: string,
   capabilities: Record<string, any>
-}[]
```
azf20, line 185: as well as this change, I think we need to change the callout above which says "except that it is a list"?

ERCS/erc-7836.md
```diff
@@ -99,7 +99,7 @@ type Response = {
 ```typescript
 const response = await provider.request({
   method: 'wallet_prepareCalls',
-  params: [{
+  params: {
```
azf20: Same comment (should still be an array)

ERCS/erc-7836.md
```diff
@@ -182,15 +182,15 @@ type Request = {
...
+}
 ```
 
 #### Example
 
 ```typescript
 const { digest, ...request } = await provider.request({
   method: 'wallet_prepareCalls',
-  params: [{
+  params: {
```
azf20: should still be an array

ERCS/erc-7836.md
```diff
@@ -207,24 +207,22 @@ const { digest, ...request } = await provider.request({
...
-  }]
+  }
 })
 
 const signature = p256.sign(digest, privateKey)
 
 const response = await provider.request({
   method: 'wallet_sendPreparedCalls',
-  params: [{
+  params: {
```
azf20: should still be an array

PR #1241, opened 2025-10-06
Update ERC-3009: Fix typos, add missing example code, and move to Final

petejkim:
Fixed typos, added missing example code, and moved ERC status to Final.

This ERC has been live in production within the USDC smart contract since 2020, and serves as a critical component of the [x402](https://www.x402.org/) standard.


petejkim:
Thanks for the comments, will make updates

petejkim:
@SamWilsn thanks, i checked 7758, it looks like the only difference is fixing of some typos (which this PR also does), will contact the author to get that closed.

PR #1260, opened 2025-10-15
Add ERC 8049: Contract-Level Onchain Metadata

nxt3d:
### Summary

Introduces a standard for storing contract-level metadata onchain using ERC-8042's Diamond Storage pattern. Extends ERC-7572's concept with onchain storage for cross-chain compatibility and upgradable contracts.

### Key Features

- **Diamond Storage**: Uses ERC-8042 for predictable storage locations
- **Cross-Chain Compatible**: Storage slots remain consistent across deployments  
- **ERC-7572 Compatible**: Maintains compatibility with existing contract metadata standards
- **Upgradable Support**: Metadata persists through contract upgrades

### Technical Details

- Diamond Storage pattern for predictable storage locations
- Support for name, description, image, etc.

### Example Use Case

**Contract ENS Naming**: Any contract can store its ENS name using this standard:
- `name`: "MyToken"
- `description`: "A decentralized exchange for trading ERC-20 tokens"
- `ens_name`: "mycontract.eth"

This enables contracts to self-identify with ENS names while maintaining consistent metadata across chains and through upgrades.


ERCS/erc-draft_contract-metadata.md
```diff
@@ -0,0 +1,136 @@
+---
+eip: TBD
```
abcoathup:
```suggestion
eip: 8049
```

ERCS/erc-draft_contract-metadata.md
```diff
@@ -0,0 +1,136 @@
...
+title: Contract-Level Onchain Metadata
+description: A standard for storing contract-level metadata onchain using Diamond Storage pattern for predictable storage locations.
+author: Prem Makeig (@nxt3d)
+discussions-to: https://ethereum-magicians.org/t/add-erc-contract-level-metadata-with-diamond-storage/25819
```
abcoathup:
```suggestion
discussions-to: https://ethereum-magicians.org/t/erc-8049-contract-level-metadata-with-diamond-storage/25819
```
Updated with assigned number

ERCS/erc-8049.md
```diff
@@ -0,0 +1,187 @@
+---
+eip: 8049
+title: Contract-Level Onchain Metadata
+description: A standard for storing contract-level metadata onchain with optional Diamond Storage pattern for predictable storage locations.
```
SamWilsn:
```suggestion
description: Onchain Contract-level metadata with optional Diamond Storage pattern for predictable storage locations.
```

SamWilsn: It might be beneficial to avoid restating "contract-level metadata" directly here, and instead define that term in the description. Space is at a bit of a premium, however, so organize the description however you feel is most valuable.

ERCS/erc-8049.md
```diff
@@ -0,0 +1,187 @@
...
+author: Prem Makeig (@nxt3d)
+discussions-to: https://ethereum-magicians.org/t/erc-8049-contract-level-metadata-with-diamond-storage/25819
+status: Draft
+type: Standards Track
+category: ERC
+created: 2025-10-10
+---
```
SamWilsn, line 11:
```suggestion
requires: 7572, 8042
---
```

An "extends" relationship generally means the base proposal is required reading, and since you mention ERC-8042 right in your abstract, I assume it is as well.

ERCS/erc-8049.md
```diff
@@ -0,0 +1,187 @@
...
+
+## Abstract
+
+This ERC defines a standard for storing contract-level metadata onchain. It extends [ERC-7572](./eip-7572.md)'s contract-level metadata concept by providing onchain storage. Contracts MAY optionally use [ERC-8042](./eip-8042.md)'s Diamond Storage pattern for predictable storage locations, enabling cross-chain compatibility and supporting upgradable contracts.
+
+## Motivation
+
+ERC-7572 provides a standard for contract-level metadata via `contractURI()`, but it primarily focuses on offchain metadata storage. This ERC extends that concept by providing onchain storage. Contracts that need predictable storage locations for cross-chain compatibility or upgradable contracts can optionally use the Diamond Storage pattern from ERC-8042.
```
SamWilsn: Your motivation is mostly just restating your abstract. Instead, this section should explain why the ecosystem needs this new standard. What use cases does it enable? Why is this particular proposal better than other (possibly hypothetical) competing proposals?

ERCS/erc-8049.md
```diff
@@ -0,0 +1,187 @@
...
+
+## Specification
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
+
+### Scope
+
+This ERC is an optional extension that MAY be implemented by any smart contract that wishes to store contract-level metadata onchain.
```
SamWilsn: All ERCs are optional, at least in the general sense.

ERCS/erc-8049.md
```diff
@@ -0,0 +1,187 @@
...
+    function getContractMetadata(string calldata key) external view returns (bytes memory);
+    
+    /// @notice Emitted when contract metadata is updated.
+    event ContractMetadataUpdated(string indexed indexedKey, string key, bytes value);
+}
+```
+
+- `getContractMetadata(key)`: Returns the contract metadata value for the given key as bytes
+
+Contracts implementing this ERC MAY also expose a `setContractMetadata(string calldata key, bytes calldata value)` function to allow metadata updates, with write policy determined by the contract.
```
SamWilsn, line 41: I'd recommend putting `setContractMetadata` into its own interface. That makes, for example, ERC-165 implementation easier.

ERCS/erc-8049.md
```diff
@@ -0,0 +1,187 @@
...
+```solidity
+event ContractMetadataUpdated(string indexed indexedKey, string key, bytes value);
+```
+### Key/Value Pairs
+
+This ERC specifies that the key is a string type and the value is bytes type. This provides flexibility for storing any type of data while maintaining an intuitive string-based key interface.
+
+### Optional Key Parameters
+
+Keys MAY include parameters to represent variations or instances of a metadata type. When keys include parameters, they SHOULD use TOON format (a minimal agent-optimized format similar to YAML, developed by Johann Schopplich) with a colon and space separator.
```
SamWilsn:
```suggestion
Keys MAY include parameters to represent variations or instances of a metadata type. When keys include parameters, they SHOULD use Token-Oriented Object Notation (TOON) format.
```

SamWilsn: Mentioning TOON here is kinda an external reference. Generally we let this slide for JSON and other well-known formats, but I don't believe TOON qualifies as well-known yet.

We have two options here. Either we allow the TOON specification as an external origin (see [EIP-5757](https://eips.ethereum.org/EIPS/eip-5757)), or you just copy the spec (it's available under MIT) into your assets directory and link to it there. How would you like to proceed?

ERCS/erc-8049.md
```diff
@@ -0,0 +1,187 @@
...
+If no standard is specified for a metadata value, clients MAY assume the value is a UTF-8 encoded string (bytes(string)) unless otherwise specified by the implementing contract or protocol.
+
+### Examples
+
+#### Example: Basic Contract Information
+
+A contract can store basic information about itself:
+
+- Key: `"name"` → Value: `bytes("MyToken")`
+- Key: `"description"` → Value: `bytes("A decentralized exchange for trading ERC-20 tokens")`
```
SamWilsn: If this is just an example, it'll be easier to remove the `ERC-20` from the string than to update the linting tool.

That said, this is technically a bug in the tool, so if you'd like to keep this, you may and I'll fix the tool.

ERCS/erc-8049.md
```diff
@@ -0,0 +1,187 @@
...
+
+- Key: `"ens_name"` → Value: `bytes("mycontract.eth")`
+
+This allows clients to discover the contract's ENS name and resolve it to get additional information about the contract.
+
+### Optional Metadata Hooks
+
+Contracts implementing this ERC MAY use metadata hooks to redirect record resolution to a different contract for secure resolution from known contracts, such as singleton registries with verifiable security properties.
+
+For the full specification of metadata hooks, see [ERC-XXXX](./erc-hooks.md) (Metadata Hooks). That standard defines the hook function signature `ContractMetadataHook(string,address)` with selector `0xcae9bb7e` for use with this ERC.
```
SamWilsn:
```suggestion
For the full specification of metadata hooks, see (Metadata Hooks) <!-- TODO: insert link -->. That standard defines the hook function signature `ContractMetadataHook(string,address)` with selector `0xcae9bb7e` for use with this ERC.
```

ERCS/erc-8049.md
```diff
@@ -0,0 +1,187 @@
...
+- Non-supporting clients can ignore the scheme.
+
+## Reference Implementation
+
+The interface is defined in the Required Metadata Function and Event section above. Here are reference implementations:
+
+### Basic Implementation
+
+```solidity
+// SPDX-License-Identifier: MIT
```
SamWilsn: Code within the proposal itself must be CC0-1.0 licensed, but if you'd like to use MIT, you can put the file in your assets directory and link to it from here.

ERCS/erc-8049.md
```diff
@@ -0,0 +1,187 @@
...
+        _metadata[key] = value;
+        emit ContractMetadataUpdated(key, key, value);
+    }
+}
+```
+
+### Diamond Storage Implementation
+
+```solidity
+// SPDX-License-Identifier: MIT
```
SamWilsn: Same license comment.

SamWilsn (commented):
Looking pretty good so far! Just a few small changes; nothing major.

ERCS/erc-8049.md
nxt3d: ok nice!

ERCS/erc-8049.md
nxt3d: There are many ways that metadata can be set. I am unsure about this change.

ERCS/erc-8049.md
nxt3d: deleted the SPDX-License-Identifier

SamWilsn:
Still a few outstanding comments, plus the broken links.

SamWilsn (approved):
Fine for a draft, but do update your requires preamble field when you get the chance.

PR #1265, opened 2025-10-16
Update ERC-7930: Move to Review

clowestab:
Updates to ERC-7930 to make it more clear and accessible based on community feedback and discussions within the Interop working group.

 - The build passes.

### DIFF
--- a/ERCS/erc-7930.md
+++ b/ERCS/erc-7930.md
@@ -115,7 +115,7 @@ The 4-byte `<checksum>` is the first 4 bytes of the hexadecimal ([RFC 4648]) str
 
 A 4-byte checksum MUST be computed and included when sharing an _Interoperable Name_. 
 
-If a user-provided _Interoperable Name_ includes a checksum, clients MUST derive the underlying _Interoperable Address_, recalculate the checksum, and compare it to the provided value. In case of a mismatch, clients MUST warn the user.
+If a user-provided _Interoperable Name_ includes a checksum, clients MUST derive the underlying _Interoperable Address_, recalculate the checksum, and compare it to the provided value. In case of a mismatch, clients MUST warn the user and require explicit user input to continue with the operation.
 
 Clients MAY include the checksum when displaying an _Interoperable Name_ within their interface.
 Clients MAY accept _Interoperable Name_ inputs without a checksum.


ERCS/erc-7930.md
```diff
@@ -2,144 +2,127 @@
...
+- Binds chain identification to the raw address.
+- Is compact for usage with cross-chain message passing and intent declaration.
+- Includes checksums to protect users.
+- Extends beyond EVM blockchains.
 
-### Guarantees this standard aims to provide
-- Any future Interoperable Address will be trivially convertible to Interoperable Address v1 (specified below), which, together with the point below, makes them a canonical representation for users who need to treat them opaquely.
-- Two Interoperable Addresses, converted to this canonical version, will be bitwise-equal if and only if they represent the same _target address_.
-- Checksums are short enough to be compared by humans, and allow for easy differentiation of _target addresses_ independently of any extra data the Interoperable Address may contain.
+## Comparisons with other standards
```
0xrcinus:
```suggestion
### Comparisons with other standards
```

ERCS/erc-7930.md
```diff
@@ -2,144 +2,127 @@
...
 
-### Comparisons with other standards
+### CAIP-10
```
0xrcinus:
```suggestion
#### CAIP-10
```

ERCS/erc-7930.md
```diff
@@ -205,25 +210,38 @@ Interoperable Address:
...
+It is therefore advised that implementers requiring canonicity of addresses (if for example you are using them as keys in smart contract mappings), to thoroughly review the [CAIP-350] profile of a chain namespace for the possibility of a lack of canonicity. These considerations should be noted in the profile's 'Extra Considerations' section.
+
+## Copyright
+Copyright and related rights waived via [CC0](../LICENSE.md).
 
 [ERC-55]: ./eip-55.md
-[ERC-7785]: ./eip-7785.md
 [CAIP-2]: https://github.com/ChainAgnostic/CAIPs/blob/2a7d42aebaffa42d1017c702974395ff5c1b3636/CAIPs/caip-2.md
 [CAIP-10]: https://github.com/ChainAgnostic/CAIPs/blob/2a7d42aebaffa42d1017c702974395ff5c1b3636/CAIPs/caip-10.md
 [CAIP-50]: https://github.com/ChainAgnostic/CAIPs/blob/2a7d42aebaffa42d1017c702974395ff5c1b3636/CAIPs/caip-50.md
```
0xrcinus, line 266: Should we add CAIP-350 (referenced in the text)

ERCS/erc-7930.md
```diff
@@ -2,144 +2,136 @@
...
-Interoperable Addresses build on insights from <!-- TODO: [ERC-7828] -->, [CAIP-10] and [CAIP-50], offering a unified format which combines:
-- Binding chain specificity (via explicit chain identifiers) to the raw address.
-- Compact & canonical binary format for use on cross-chain message passing and intent declaration.
-- Checksums for name collision mitigation & user error prevention.
+This proposal builds on insights from [CAIP-10] and [CAIP-50]. It offers a format definition which:
 
-Furthermore, this proposal adds a few features to make addresses interoperable with infrastructure at the edges of and beyond the ethereum ecosystem, by supporting the representation of addresses of non-evm chains as well, for the benefit of cross-chain liquidity movements and cross-chain message passing.
+- Binds chain identification to the raw address.
+- Is compact for usage with cross-chain message passing and intent declaration.
+- Includes checksums to protect users.
```
0xteddybear: this paragraph loses a great deal of specificity when rewritten. While I get that the previous iteration could be needlessly verbose, this no longer differentiates between binary and text formats (which like half of the value proposition of this ERC) and doesn't say what or how it protects users from

ERCS/erc-7930.md
```diff
@@ -2,144 +2,136 @@
...
+**Interoperable Name**
+: A string representation of an _Interoperable Address_.
 
-Target address
-: The (address, chain ID) pair a particular Interoperable Address points to.
+### Binary Format Definition
 
-### Interoperable Address V1 binary format definition
-In binary format, addresses MUST have the following encoding:
+An _Interoperable Address_ as defined by this standard MUST have the following binary format:
```
0xteddybear, line 63: Good fix on using the name whenever applicable

ERCS/erc-7930.md
```diff
@@ -2,144 +2,136 @@
...
-Interoperable Address
-: A binary payload which unambiguously identifies a target address and allows conversion to a human-readable name.
+### Terminology
+**Interoperable Address**
+: A binary payload which unambiguously identifies a target address on a target chain.
 
-Interoperable Name
-: A string representation of an interoperable address, meant to be used by humans.
+**Interoperable Name**
+: A string representation of an _Interoperable Address_.
```
0xteddybear: the human-readability of it is a crucial clarification

ERCS/erc-7930.md
```diff
@@ -2,144 +2,136 @@
...
 
-Version
-: A 2-byte version identifier. For version 1, this must be `0x0001` (big-endian). Future versions SHOULD be standardized in separate ERCs.
+**Version**
+: A 2-byte version identifier. For version 1 (this specification), this must be `0x0001` (big-endian). Future versions SHOULD be standardized in separate ERCs.
 
-ChainType
-: A 2-byte value as defined in [CAIP-350], corresponding to a [CAIP-2] _namespace_, which allows users to know how to interpret & display the other two fields.
+**ChainType**
+: A 2-byte value as defined in [CAIP-350], corresponding to a [CAIP-2] namespace, which allows users to know how to interpret and display the `ChainReference` and the `Address`.
```
0xteddybear: it's standard practice to have new concepts highlighted in italics

ERCS/erc-7930.md
```diff
@@ -2,144 +2,136 @@
...
 
-AddressLength
-: 1-byte integer encoding the length of Address in bytes. Note that it MAY be zero, in which the Interoperable Address will not include an address. It MUST NOT be zero if `ChainReferenceLength` is also zero.
+**AddressLength**
+: 1-byte integer encoding the length of Address in bytes. Note that it MAY be zero, in which case the _Interoperable Address_ MUST NOT include an address. It MUST NOT be zero if the `ChainReferenceLength` is also zero.
 
-Address
-: Variable length field containing the binary format of the address component. For serialization details of different types of addresses see the [CAIP-350] profile for the chain type.
+**Address**
+: Variable length field containing the binary encoding of the address. To discern how to serialize an address for a specific chain type, implementors MUST reference the respective [CAIP-350] profile.
```
0xteddybear: nit: I think it makes sense to normative-MUST the algorithm complying with the CAIP profile, and not the implementor referencing the spec

ERCS/erc-7930.md
```diff
@@ -2,144 +2,136 @@
...
+
+A 4-byte checksum MUST be computed and included when sharing an _Interoperable Name_. 
+
+If a user-provided _Interoperable Name_ includes a checksum, clients MUST discern the underlying _Interoperable Address_ recompute the checksum and compare it to the provided value. On mismatch, clients MUST alert the user.
 
-## Test Cases
+Clients MAY include the checksum when displaying an _Interoperable Name_ within their interface.
+Clients MAY accept _Interoperable Name_ inputs without a checksum.
+
+## Examples
```
0xteddybear: iirc EIP-1 requires this is called test cases

ERCS/erc-7930.md
```diff
@@ -205,25 +219,38 @@ Interoperable Address:
...
 ```
-keccak256 input for checksum: `0x00022045296998a6f8e2a784db5d9f95e18fc23f70441a1039446801089879b08c7ef000`.
-Note the version field is removed before hashing.
 
-## Security Considerations
-While this standard aims to be a foundation to be able to canonically refer to addresses on different chains, that guarantee is going to be a leaky abstraction in the real world, given that e.g. a particular chain namespace might define a serialization scheme that can't guarantee canonicity of addresses, or a given network might have two valid [CAIP-2] ids referring to it.
+## Rationale
 
-It is therefore advised for implementers requiring canonicity of addresses (e.g by using them as keys in smart contract mappings or other key-value stores), to thoroughly review the [CAIP-350] profile of a chain namespace for the possibility of a lack of canonicity of addresses (which should be noted in the profile's 'Extra Considerations' section) as well as collisions with other already-supported namespaces.
+- The `@` symbol is used as a separator as it is clear, easy for software to parse, and avoids confusion with the colon (`:`) symbol utilized in [CAIP-2] identifiers.
```
0xteddybear: I feel the rewrite of this section is more vague

0xteddybear (changes requested):
thanks for iterating on this!

Some of the edits do make the document easier to read, however some sections lost a great deal of normative power (e.g. the definition of the interoperable names) and I'd like for this to be as clear-cut of a spec as possible. The better-flowing prose is very well suited for an explainer-type document, though.

Also, some editorial concerns:
- the reordering breaks automated checks on the document structure (e.g. no extraneous sections, required sections, etc)
- the status should be moved to Review, and then we should ask an editor to move it to Last Call

ERCS/erc-7930.md
clowestab: EIP 1 states _"Test cases for an implementation are mandatory for EIPs that are affecting consensus changes."_ 

Because this ERC isn't related to consensus, and 'Test cases' is an optional heading I opted to call it 'Examples' because it's semantically clearer IMO..

ERCS/erc-7930.md
clowestab: 👍 CAIP-350 is linked.

ERCS/erc-7930.md
clowestab: Fixed.

ERCS/erc-7930.md
clowestab: Fixed.

ERCS/erc-7930.md
clowestab: Updated for clarity.

ERCS/erc-7930.md
clowestab: Noted. Updated.

ERCS/erc-7930.md
clowestab: Specifically as it pertains to canonicity as a security consideration? Happy to flesh out - is there an obvious example of a collision with another namespace?

ERCS/erc-7930.md
clowestab: Ha. Valid point. Updated.

ERCS/erc-7930.md
clowestab: I've iterated on this further to address these concerns.

clowestab:
> thanks for iterating on this!
> 
> Some of the edits do make the document easier to read, however some sections lost a great deal of normative power (e.g. the definition of the interoperable names) and I'd like for this to be as clear-cut of a spec as possible. The better-flowing prose is very well suited for an explainer-type document, though.
> 
> Also, some editorial concerns:
> 
> * the reordering breaks automated checks on the document structure (e.g. no extraneous sections, required sections, etc)
> * the status should be moved to Review, and then we should ask an editor to move it to Last Call

I've updated, and commented as appropriate.

I've moved it to 'Review' and fixed the validation checks.

I think my fundamental issue was that on first read I was unable to follow exactly what was going on. The separation of the _Interoperable Address_ and _Interoperable Name_ definitions in my opinion makes this significantly more accessible. 

I do agree that the spec should be as clear-cut as possible, so if I've inadvertently removed meaning I apologise, and am happy to iterate to make sure it is all in there.

This is my first foray into contributing to specs, and my perception is that a spec should be easy to understand independent of any external resources or explainers so as to incentivize people to actually read and contribute.

ERCS/erc-7930.md
```diff
@@ -2,144 +2,136 @@
...
+This proposal builds on insights from [CAIP-10] and [CAIP-50]. It offers both a binary canonical _Interoperable Address_ format and a human readable _Interoperable Name_ format which:
 
-Furthermore, this proposal adds a few features to make addresses interoperable with infrastructure at the edges of and beyond the ethereum ecosystem, by supporting the representation of addresses of non-evm chains as well, for the benefit of cross-chain liquidity movements and cross-chain message passing.
-
-### Guarantees this standard aims to provide
-- Any future Interoperable Address will be trivially convertible to Interoperable Address v1 (specified below), which, together with the point below, makes them a canonical representation for users who need to treat them opaquely.
-- Two Interoperable Addresses, converted to this canonical version, will be bitwise-equal if and only if they represent the same _target address_.
-- Checksums are short enough to be compared by humans, and allow for easy differentiation of _target addresses_ independently of any extra data the Interoperable Address may contain.
+- Bind together chain identification and the raw address.
```
0xteddybear: nit: should be 'binds' ?

ERCS/erc-7930.md
```diff
@@ -2,144 +2,136 @@
...
-Interoperable Addresses is a binary format to describe an address specific to one chain with a companion 'Interoperable Names' specification for human-readable display.
+This proposal introduces a **binary format** to describe a chain specific address. Additionally, it defines a human-readable version of that identifier for improved user experience for user facing interactions.
 
 This is achieved through a versioned, length-prefixed binary envelope that supports arbitrary-length data. The interpretation and serialization rules for the data within this envelope are defined by companion standards, which provide profiles for each chain type.
 
 ## Motivation
-Current Ethereum addresses ([ERC-55]) lack chain specificity and have optional checksums, creating several challenges:
+The address format utilized on Ethereum mainnet is shared by a large number of other blockchains. The format does not include details of the chain on which an interaction should occur. This introduces risk if, for example, a transaction is mistakenly executed on a chain where the address is inaccesible. This risk is particularly pronounced for addresses that represent smart contracts or smart accounts.
```
0xteddybear, line 19: this introduction is clearly better, thanks

ERCS/erc-7930.md
```diff
@@ -2,144 +2,136 @@
...
-
-### Guarantees this standard aims to provide
-- Any future Interoperable Address will be trivially convertible to Interoperable Address v1 (specified below), which, together with the point below, makes them a canonical representation for users who need to treat them opaquely.
-- Two Interoperable Addresses, converted to this canonical version, will be bitwise-equal if and only if they represent the same _target address_.
-- Checksums are short enough to be compared by humans, and allow for easy differentiation of _target addresses_ independently of any extra data the Interoperable Address may contain.
+- Bind together chain identification and the raw address.
+- Is compact for usage with cross-chain message passing and intent declaration.
+- Includes short, easily verififiable checksums to protect users.
+- Extends beyond EVM blockchains.
+- Is flexible and extensible
```
0xteddybear: I believe this is a bit easier to follow on _what_ this standard provides, but not why/how. I wrote it like this to address actual questions from people in the ecosystem (with @frangio in particular being the guinea pig on how to sell a smart contract implementor on this). My impression is that every standard sells itself as flexible and extensible, and it's crucial to explain what it means in this context

ERCS/erc-7930.md
```diff
@@ -2,144 +2,136 @@
...
+[CAIP-10] proposes a standard text format to represent an address on a specific chain (referenced by its [CAIP-2] identifier).
 
-Interoperable Names are convertible to [CAIP-10] without even going through the binary representation, so backwards-compatibility with actors expecting [CAIP-10] identifiers should not be an issue.
+The standard **does not** concern itself with the serialization/deserialization of the chain specific address format.
 
-The interop roadmap is better served by having a standardized binary format for addresses first, which allows the message passing and intents verticals to move forward on a common interface, with a good-enough text representation which is familiar to users and useful for developers, and as a next step develop a chain & address name resolving standard on top of it, leveraging its uniformity and extensibility.
+It is trivial to add support for chains to [CAIP-10], but the standard has its drawbacks:
 
+- There is no canonical identifier for a specific address on a specific chain - there could be multiple valid [CAIP-10] addresses pointing to the same target address.
+- The format is not optimized for usage within smart contracts.
```
0xteddybear: imo not saying _how_ it is unoptimized is a loss for readability and convince-people-to-use-this

ERCS/erc-7930.md
```diff
@@ -2,144 +2,136 @@
...
+### Terminology
+**Interoperable Address**
+: A binary payload which unambiguously identifies a target address on a target chain.
 
-Interoperable Name
-: A string representation of an interoperable address, meant to be used by humans.
+**Interoperable Name**
+: A string representation of an _Interoperable Address_, meant to be used by humans.
 
-Target address
```
0xteddybear, line 62: this is still used throughout the document, it's worth it to define it

ERCS/erc-7930.md
```diff
@@ -2,144 +2,136 @@
...
 
-ChainReferenceLength
-: A 1-byte integer encoding the length of ChainReference in bytes. Note that it MAY be zero, in which the Interoperable Address will not include a chain reference.
+**ChainReferenceLength**
+: A 1-byte integer encoding the length of _ChainReference_ in bytes. Note that it MAY be zero, in which case the _Interoperable Address_ MUST NOT include a chain reference.
 
-ChainReference
-: Variable length, binary representation of [CAIP-2] chain namespace & reference serialized as explained in the [CAIP-350] profile for the chain type, encoding the chain ID.
+**ChainReference**
+: Variable length, binary representation of the [CAIP-2] chain ID. To discern how to serialize the _ChainReference_ for a specific chain type, implementors MUST reference the respective [CAIP-350] profile. Chain profiles are maintained by the Chain-Agnostic Standards Alliance (CASA).
```
0xteddybear, line 83: my previous comment on 'implementors MUST reference' still applies to this

ERCS/erc-7930.md
```diff
@@ -2,144 +2,136 @@
...
 
-AddressLength
-: 1-byte integer encoding the length of Address in bytes. Note that it MAY be zero, in which the Interoperable Address will not include an address. It MUST NOT be zero if `ChainReferenceLength` is also zero.
+**AddressLength**
+: 1-byte integer encoding the length of Address in bytes. Note that it MAY be zero, in which case the _Interoperable Address_ MUST NOT include an address. It MUST NOT be zero if the _ChainReferenceLength_ is also zero.
```
0xteddybear, line 86: iiuc the interoperable address not including an address in that case is not a requirement of the spec, but a consequence of there not being any place where to put it, so the 2119 MUST doesn't really apply?

ERCS/erc-7930.md
```diff
@@ -205,25 +219,40 @@ Interoperable Address:
...
+- The address field includes `%` as a valid character to allow for URL-encoded address formats.
+- We chose to support zero-length addresses and chain IDs to make this standard flexible and to allow developers to use a single, uniform standard for many different jobs.
+- We chose **not** to utilize alternate encoding formats (`base58` or `base64` for example) to reduce the size of the _Interoperable Address_ and/or to abstract away namespace specific address serialization standards. We opted not to noting that readability and familiarity are important user experience considerations.
 
-<!-- TODO:
 
-[ERC-7828]: ./eip-7828.md
+## Security Considerations
+This standard aims to be a foundational standard to allow for canonical representations of chain specific addresses. In the real world it can not be guaranteed that there will only be one valid representation of a chain specific address. For example, a particular chain namespace might define a serialization scheme that can't guarantee canonicity of addresses, or a given network might have two valid [CAIP-2] IDs referring to it.
```
0xteddybear: s/aims to be a foundational standard / aims to be a foundation

ERCS/erc-7930.md
```diff
@@ -205,25 +219,40 @@ Interoperable Address:
...
 
-## Security Considerations
-While this standard aims to be a foundation to be able to canonically refer to addresses on different chains, that guarantee is going to be a leaky abstraction in the real world, given that e.g. a particular chain namespace might define a serialization scheme that can't guarantee canonicity of addresses, or a given network might have two valid [CAIP-2] ids referring to it.
+### Versioning
+
+These rules ensure that future standards that build upon this one maintain backwards compatibility.
+
+Future versions..
+
+- MUST set the most significant bit of the version field to 1 if they are unable to parse the _Interoperable Address_ format outlined herein
```
0xteddybear: the interoperable addresses are parsed themselves, they do not parse other things. The idea is to set the bit to 1 if they cannot be parsed with the same algorithm as v1

ERCS/erc-7930.md
```diff
@@ -205,25 +219,40 @@ Interoperable Address:
...
+
+These rules ensure that future standards that build upon this one maintain backwards compatibility.
+
+Future versions..
+
+- MUST set the most significant bit of the version field to 1 if they are unable to parse the _Interoperable Address_ format outlined herein 
+- MUST support defining an address, a chain, or both
+- MUST allow the client to discern the data necessary to calculate the checksum as outlined in this standard
+- MAY only be able to represent a subset of the CAIP namespaces
+- MAY expand upon the _Interoperable Name_ definition
```
0xteddybear: 'expand' is much less specific than 'assign extra syntactic restrictions', what do you mean precisely by 'expand'

ERCS/erc-7930.md
```diff
@@ -205,25 +219,40 @@ Interoperable Address:
...
+- We chose **not** to utilize alternate encoding formats (`base58` or `base64` for example) to reduce the size of the _Interoperable Address_ and/or to abstract away namespace specific address serialization standards. We opted not to noting that readability and familiarity are important user experience considerations.
 
-<!-- TODO:
 
-[ERC-7828]: ./eip-7828.md
+## Security Considerations
+This standard aims to be a foundational standard to allow for canonical representations of chain specific addresses. In the real world it can not be guaranteed that there will only be one valid representation of a chain specific address. For example, a particular chain namespace might define a serialization scheme that can't guarantee canonicity of addresses, or a given network might have two valid [CAIP-2] IDs referring to it.
 
--->
+It is therefore advised that implementers requiring canonicity of addresses (if for example you are using them as keys in smart contract mappings), to thoroughly review the [CAIP-350] profile of a chain namespace for the possibility of a lack of canonicity. These considerations should be noted in the profile's 'Extra Considerations' section.
```
0xteddybear, line 258: this is repeated with the 'rationale' section

ERCS/erc-7930.md
```diff
@@ -205,25 +219,40 @@ Interoperable Address:
...
+- MAY expand upon the _Interoperable Name_ definition
+- MAY define extra fields
+
+## Rationale
 
 It is therefore advised for implementers requiring canonicity of addresses (e.g by using them as keys in smart contract mappings or other key-value stores), to thoroughly review the [CAIP-350] profile of a chain namespace for the possibility of a lack of canonicity of addresses (which should be noted in the profile's 'Extra Considerations' section) as well as collisions with other already-supported namespaces.
 
-## Copyright
-Copyright and related rights waived via [CC0](../LICENSE.md).
+- The `@` symbol is used as a separator as it is clear, easy for software to parse, and avoids confusion with the colon (`:`) symbol utilized in [CAIP-2] identifiers.
```
0xteddybear: software parsing is not particularly hard in the case of using another separator, but might be less obvious for humans

ERCS/erc-7930.md
```diff
@@ -205,25 +219,40 @@ Interoperable Address:
...
+- MAY define extra fields
+
+## Rationale
 
 It is therefore advised for implementers requiring canonicity of addresses (e.g by using them as keys in smart contract mappings or other key-value stores), to thoroughly review the [CAIP-350] profile of a chain namespace for the possibility of a lack of canonicity of addresses (which should be noted in the profile's 'Extra Considerations' section) as well as collisions with other already-supported namespaces.
 
-## Copyright
-Copyright and related rights waived via [CC0](../LICENSE.md).
+- The `@` symbol is used as a separator as it is clear, easy for software to parse, and avoids confusion with the colon (`:`) symbol utilized in [CAIP-2] identifiers. 
+- No length restriction is placed on addresses, allowing for chains with longer address formats to be represented. This is flexible and future proof.
```
0xteddybear: not sold on the 'This is flexible and future proof', feels too chatgpt-y while being less specific than the previous wording

ERCS/erc-7930.md
```diff
@@ -205,25 +219,40 @@ Interoperable Address:
...
+- The address field includes `%` as a valid character to allow for URL-encoded address formats.
+- We chose to support zero-length addresses and chain IDs to make this standard flexible and to allow developers to use a single, uniform standard for many different jobs.
+- We chose **not** to utilize alternate encoding formats (`base58` or `base64` for example) to reduce the size of the _Interoperable Address_ and/or to abstract away namespace specific address serialization standards. We opted not to noting that readability and familiarity are important user experience considerations.
```
0xteddybear: 'we opted not to noting that readability...' not sure I follow this sentence

0xteddybear (changes requested):
kudos on complying with the editorial requirements, but there is still a significant loss of specificity

ERCS/erc-7930.md
```diff
@@ -149,17 +141,24 @@ Interoperable Address:
...
-Address: `MJKqp326RZCHnAAbew9MDdui3iCKWco7fsK9sVuZTX2`
+#### Example 2: Solana mainnet address
+
+| Key | Value |
+| :--- | :--- |
+| **Chain Name** | Solana Mainnet |
+| **CAIP-2 Namespace** | `solana` |
+| **CAIP-2 Chain ID** | `5eykt4UsFv8P8NJdTREpY1vzqKqZKvdpKuc147dw2N9d` |
+| **Address** | `MJKqp326RZCHnAAbew9MDdui3iCKWco7fsK9sVuZTX2` |
+| **Checksum Input** | `0x00000100x00022045296998a6f8e2a784db5d9f95e18fc23f70441a1039446801089879b08c7ef02005333498d5aea4ae009585c43f7b8c30df8e70187d4a713d134f977fc8dfe0b5114D8DA6BF26964AF9D7EED9E03E53415D37AA96045` |
```
0xteddybear: this input is incorrect, it includes the address from the example above. But the interoperable names's checksum hasn't changed

ERCS/erc-7930.md
clowestab: 'assign extra syntactic restrictions' feels constraining in that 7828 for example actually loosens restrictions - it allows you to use ENS names. My usage of 'expand' was I guess intentionally less specific as we don't know _how_ people may build upon this in the future.

I've gone for '_MAY modify the syntactic restrictions of the _Interoperable Name_ definition_' for now..

ERCS/erc-7930.md
clowestab: I was trying to explain our justification - that the space savings are not worth making it less familiar or readable for users. I have clarified further.

ERCS/erc-7930.md
clowestab: Removed. I think the flexibility is implied anyway

ERCS/erc-7930.md
clowestab: Updated. Still left in '_easy for software to parse_' because it's still true, even if other separators are just as easy.

ERCS/erc-7930.md
clowestab: I have rephrased this

ERCS/erc-7930.md
clowestab: Fixed

ERCS/erc-7930.md
clowestab: Fixed

ERCS/erc-7930.md
clowestab: I'm not sure whats happened here. These were from the original. Possibly miscopied or something by me. Regardless, fixed.

ERCS/erc-7930.md
clowestab: Not sure what happened there. Removed from 'Rationale' as more appropriate in 'Security Considerations'.

ERCS/erc-7930.md
clowestab: Defined, how it is used within this updated spec.

ERCS/erc-7930.md
clowestab: Fair. Added.

ERCS/erc-7930.md
clowestab: Fair. I have referenced the original and modified this section accordingly to add some more 'why'. I think that the 'how' is pushed in the Specification.

ERCS/erc-7930.md
```diff
@@ -2,144 +2,142 @@
...
-status: Draft
+status: Review
 type: Standards Track
 category: ERC
 created: 2025-02-02
 ---
 
 ## Abstract
-Interoperable Addresses is a binary format to describe an address specific to one chain with a companion 'Interoperable Names' specification for human-readable display.
+This proposal introduces a **binary format** to describe a chain specific address. Additionally, it defines a human-readable version of that identifier for improved user experience for user facing interactions.
```
0xMonoAx, line 14: Since ERC-7930 is quite dependent on CAIP-350, do you think it would make sense to clarify this in the Abstract?

I was thinking of something like:

_"This ERC defines a generic, versioned binary container format for representing an address bound to a specific blockchain. The interpretation and serialization rules for the chain reference and address components are delegated to complementary standards, namely, CAIP-350 profiles."_

What do you think?

ERCS/erc-7930.md
```diff
@@ -2,144 +2,142 @@
...
-: Variable length field containing the binary format of the address component. For serialization details of different types of addresses see the [CAIP-350] profile for the chain type.
+**AddressLength**
+: 1-byte integer encoding the length of Address in bytes. Note that it MAY be zero, in which case the _Interoperable Address_ MUST NOT include an address. It MUST NOT be zero if the _ChainReferenceLength_ is also zero.
 
-#### Version Compatibility
+**Address**
+: Variable length field containing the binary encoding of the address. The serialization for a specific _ChainType_ MUST follow the rules of its corresponding [CAIP-350] profile.
 
-These rules ensure that future versions of Interoperable Addresses maintain backwards compatibility and consistent behavior across implementations:
+### Interoperable Name Definition
```
0xMonoAx, line 91: I think it would be helpful to clarify the relationship between this standard and [ERC-7828].
We could add something like the following:

_"Resolution of Interoperable Names MAY be delegated to name service standards such as ENS (see [ERC-7828]) or other equivalent systems. Implementations MUST ensure that any binary data derived from such name resolution strictly conforms to the Interoperable Address encoding defined in this specification."_

ERCS/erc-7930.md
```diff
@@ -2,144 +2,142 @@
...
-    - It makes the Interoperable Address <-> raw address relationship evident to the user, which will make the cases where the former are not (yet) supported by a dapp or hardware wallet easier to reason with, as users will evidently see the address is the same, but it's the chain&checksum that is being trimmed.
-    - Its extra human-readability will make it more useful as a a stopgap solution until a standard focused on chain and address names is finalized.
-- We chose to support zero-length addresses and chainids to be able to use just one standard to represent both potentially foreign _target addresses_ (e.g. the recipient of a cross-chain message) and also plain addresses or chainids (e.g. to specify the origin-chain hook address of a cross-chain message), for greater uniformity across implementations.
+`<checksum>` is a 4-byte checksum calculated by taking the first 4-bytes of the hexadecimal (as defined in [RFC 4648]) representation of the keccak256 hash of the concatenated _ChainType_, _ChainReferenceLength_, _ChainReference_, _AddressLength_ and _Address_ fields of the binary _Interoperable Address_. The `Version` field is **not** included.
```
0xMonoAx: Why is it typed like this? What do you think about it?

_The checksum is a 4-byte value calculated by taking the first 4 bytes of the hexadecimal  (as defined in [RFC 4648] string representing the `keccak256` hash.
 The hash is computed over the concatenation of the following binary fields of the Interoperable Address: `ChainType`, `ChainReferenceLength`, `ChainReference`, `AddressLength`, and `Address`.
The `Version` field MUST NOT be included in the hashed data._

ERCS/erc-7930.md
```diff
@@ -2,144 +2,142 @@
...
+
+#### Checksums
+
+A 4-byte checksum MUST be computed and included when sharing an _Interoperable Name_. 
+
+If a user-provided _Interoperable Name_ includes a checksum, clients MUST discern the underlying _Interoperable Address_ recompute the checksum and compare it to the provided value. On mismatch, clients MUST alert the user.
```
0xMonoAx: What do you think about adding this section? I think it would improve clarity regarding the relationship with CAIP-350 (if I add this, I would remove the clarification from the Abstract)

_**Parsing and Serialization:**_

_When parsing an Interoperable Name to generate its binary Interoperable Address, clients MUST follow the normalization and serialization rules defined in the relevant CAIP-350 profile for the given `<chain>` and `<address>.` This ensures that different valid text representations (e.g., case variations in an address) resolve to a single, canonical binary form, which is essential for consistent checksum calculation and data integrity._

ERCS/erc-7930.md
```diff
@@ -2,144 +2,142 @@
...
 
 ## Specification
 The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
 
-### Concepts
-Interoperable Address
-: A binary payload which unambiguously identifies a target address and allows conversion to a human-readable name.
+### Terminology
+**Target Address**
+: The target address being represented. Serialized subject to the rules outlined in the [CAIP-350] profile for the namespace of the chain being targetted.
```
0xMonoAx: Typo “targetted” -> “targeted”

ERCS/erc-7930.md
```diff
@@ -205,25 +225,38 @@ Interoperable Address:
...
-## Security Considerations
-While this standard aims to be a foundation to be able to canonically refer to addresses on different chains, that guarantee is going to be a leaky abstraction in the real world, given that e.g. a particular chain namespace might define a serialization scheme that can't guarantee canonicity of addresses, or a given network might have two valid [CAIP-2] ids referring to it.
+### Versioning
 
-It is therefore advised for implementers requiring canonicity of addresses (e.g by using them as keys in smart contract mappings or other key-value stores), to thoroughly review the [CAIP-350] profile of a chain namespace for the possibility of a lack of canonicity of addresses (which should be noted in the profile's 'Extra Considerations' section) as well as collisions with other already-supported namespaces.
+These rules ensure that future standards that build upon this one maintain backwards compatibility.
 
-## Copyright
-Copyright and related rights waived via [CC0](../LICENSE.md).
+Future versions..
```
0xMonoAx: Typo "Future versions.." -> "Future versions:"

ERCS/erc-7930.md
0xMonoAx: > clients MUST discern the underlying _Interoperable Address_ recompute the checksum and compare it to the provided value. On mismatch, clients MUST alert the user.

I think “dicern” is a bit ambiguous. What do you think of this phrase:

_clients MUST derive the underlying Interoperable Address, recalculate the checksum, and compare it to the provided value. In case of a mismatch, clients MUST warn the user._

clowestab: Updated.

ERCS/erc-7930.md
clowestab: Fixed

ERCS/erc-7930.md
clowestab: Fixed. (You added a typo in your comment XD)

ERCS/erc-7930.md
clowestab: Yes. I like this a lot. Added verbatim.

ERCS/erc-7930.md
clowestab: Reworded based on your suggestion.

ERCS/erc-7930.md
clowestab: Rather than add this here, I have added the final statement to the 'Future versions' sections.
My view is that highlighting 7828 and name services is too concrete, as specs can build upon this one in a multitude of ways as long as they follow the rules outlined.

ERCS/erc-7930.md
clowestab: I've added the section you outlined below. I also added explicit reference to CAIP-350 here.

ERCS/erc-7930.md
0xteddybear: > 'assign extra syntactic restrictions' feels constraining in that 7828 for example actually loosens restrictions - it allows you to use ENS names.

This is not correct. A grammar is defined as a set of strings (some grammars can also be _described_ by things such as regular expressions of BNF notation). The grammar defined in 7930 is less restrictive than 7828's. As an example, `0x00@...` would be a syntactically valid 7930 name (albeit semantically meaningless), but invalid 7828, since AFAICT the ENS grammar specifies you got to have labels in between dots.

> My usage of 'expand' was I guess intentionally less specific as we don't know how people may build upon this in the future. I've gone for 'MAY modify the syntactic restrictions of the Interoperable Name definition' for now..

I'm against this. While we should not make standards maximally _restrictive_ (as in, disallowing future usecases), we should make them maximally _specific_ (as in, being easy to figure out if something fits the standard or doesn't). People will write parsers for this; they'll want to be able to figure out if a string they receive is a 7828 or just 7930, and the standard should ensure that decision is computable.

Another way to think about the specificity requirements for this document is that it should be possible to develop two independent implementations of libraries that are compatible with one another, even if the teams developing them have an adversarial relationship between them

ERCS/erc-7930.md
```diff
@@ -2,144 +2,146 @@
 eip: 7930
 title: Interoperable Addresses
 description: An extensible binary format to refer to an address specific to one chain.
-author: Teddy (@0xteddybear), Joxes (@0xJoxess), Nick Johnson (@Arachnid), Francisco Giordano (@frangio), Skeletor Spaceman (@skeletor-spaceman), Racu (@0xRacoon), TiTi (@0xtiti), Gori (@0xGorilla), Ardy (@0xArdy), Onizuka (@onizuka-wl), Sam Kaufman (@SampkaML), Marco Stronati (@paracetamolo), Yuliya Alexiev (@yuliyaalexiev), Jeff Lau (@jefflau), Sam Wilson (@samwilsn), Vitalik Buterin (@vbuterin)
+author: Teddy (@0xteddybear), Joxes (@0xJoxess), Nick Johnson (@Arachnid), Francisco Giordano (@frangio), Skeletor Spaceman (@skeletor-spaceman), Racu (@0xRacoon), TiTi (@0xtiti), Gori (@0xGorilla), Ardy (@0xArdy), Onizuka (@onizuka-wl), Sam Kaufman (@SampkaML), Marco Stronati (@paracetamolo), Yuliya Alexiev (@yuliyaalexiev), Jeff Lau (@jefflau), Sam Wilson (@samwilsn), Vitalik Buterin (@vbuterin), Thomas Clowes (@clowestab), Mono (0xMonoAx)
```
0xteddybear, line 5: mono's username is missing the `@`

ERCS/erc-7930.md
```diff
@@ -2,144 +2,146 @@
...
 
 ## Specification
 The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
 
-### Concepts
-Interoperable Address
-: A binary payload which unambiguously identifies a target address and allows conversion to a human-readable name.
+### Terminology
+**Target Address**
+: The target address being represented. Serialized subject to the rules outlined in the [CAIP-350] profile for the namespace of the chain being targeted.
```
0xteddybear, line 53: so far we have used 'target address' to mean an address already scoped to a given chain, aka the object of this proposal, and we haven't used the 'target chain' concept. Also, this definition is recursive :grin:

ERCS/erc-7930.md
```diff
@@ -2,144 +2,146 @@
...
-- Interoperable Address versions MAY only be able to represent a subset of the CAIP namespaces.
-- Interoperable Address versions MAY assign extra syntactic restrictions on the associated Interoperable Name, coupled to novel semantic meaning (e.g. use of ENS or other naming registries)
-- Interoperable Address versions MAY define extra fields for purposes such as storing information on how to display the addresses.
+Recall that an _Interoperable Name_ is a human readable representation of an underlying _Interoperable Address_.
 
-### Interoperable Name format definition
+It's format is `<address> @ <chain> # <checksum>` where the components match the following regular expressions:
 
 #### Syntax
 ```bnf
```
0xteddybear, line 98: what's the rationale behind removing the element we're actually trying to define from the grammar definition? keep in mind this is a kinda standard format: https://en.wikipedia.org/wiki/Backus%E2%80%93Naur_form

ERCS/erc-7930.md
```diff
@@ -2,144 +2,146 @@
...
-Where:
+These components have the following meanings:
 
-Chain
-: String representation of [CAIP-2] blockchain identifier, recovered from the binary representation described in the [CAIP-350] profile for the chain type. In the case where `ChainReferenceLength` is zero, it should be the CAIP namespace alone, and no trailing colon. Note that it's not possible to specify a reference without a namespace.
+`<address>` is the text representation of an address for the chain namespace in question. 
 
-Address
-: Chain namespace specific text representation of the address from the binary representation. Mapping between the two described in the [CAIP-350] profile for the chain type. In the case where `AddressLength` is zero, it should be the empty string.
+`<chain>` is the string representation of a specific blockchain as defined in [CAIP-2]. For example `eip155:1` for Ethereum Mainnet.
```
0xteddybear, line 108: this is way less specific than the current wording, and even explicitly discards the handling of an edge case :sob:

0xteddybear: also, it should not be CAIP-2 but CAIP-350. They can be different when e.g. the chain reference is truncated in CAIP-2 but used in full in CAIP-350

ERCS/erc-7930.md
```diff
@@ -2,144 +2,146 @@
...
-    - It makes the Interoperable Address <-> raw address relationship evident to the user, which will make the cases where the former are not (yet) supported by a dapp or hardware wallet easier to reason with, as users will evidently see the address is the same, but it's the chain&checksum that is being trimmed.
-    - Its extra human-readability will make it more useful as a a stopgap solution until a standard focused on chain and address names is finalized.
-- We chose to support zero-length addresses and chainids to be able to use just one standard to represent both potentially foreign _target addresses_ (e.g. the recipient of a cross-chain message) and also plain addresses or chainids (e.g. to specify the origin-chain hook address of a cross-chain message), for greater uniformity across implementations.
+The 4-byte `<checksum>` is the first 4 bytes of the hexadecimal ([RFC 4648]) string representation of a keccak256 hash. This hash is computed over the concatenation of the _Interoperable Address_'s binary fields: `ChainType`, `ChainReferenceLength`, `ChainReference`, `AddressLength`, and `Address`. The `Version` field MUST NOT be included in the hashed data.
```
0xteddybear, line 112: nit: 'the 4byte checksum is the first 4 bytes...' feels like could be improved writing-wise, but can't think of an actually better alternative. It does not compromise clarity though, so not a blocker

ERCS/erc-7930.md
```diff
@@ -205,25 +229,38 @@ Interoperable Address:
...
+- MAY modify the syntactic restrictions of the _Interoperable Name_ definition but MUST ensure that any binary data derived from such name resolution strictly conforms to the _Interoperable Address_ encoding defined in this specification.
+- MAY define extra fields
 
-[ERC-7828]: ./eip-7828.md
+## Rationale
 
--->
+- The `@` symbol is used as a separator as it provides visual clarity to humans, is easy for software to parse, and avoids confusion with the colon (`:`) symbol utilized in [CAIP-2] identifiers. 
+- No length restriction is placed on addresses, allowing for chains with longer address formats to be represented.
+- The address field includes `%` as a valid character to allow for URL-encoded address formats.
```
0xteddybear, line 250: the previous expression was clearer than this one: 

>  Similarly, the address field includes `%` as a valid character to allow for url-encoding of any other characters.

it clearly states that the purpose of using `%` is to url-encode characters not natively supported by the standard. On the other hand, it's not clear what 'URL-encoded' address formats' means, precisely. Are there address formats out there which are in some way URLs?

ERCS/erc-7930.md
```diff
@@ -205,25 +229,38 @@ Interoperable Address:
...
+- MUST allow the client to discern the data necessary to calculate the checksum as outlined in this standard
+- MAY only be able to represent a subset of the CAIP namespaces
+- MAY modify the syntactic restrictions of the _Interoperable Name_ definition but MUST ensure that any binary data derived from such name resolution strictly conforms to the _Interoperable Address_ encoding defined in this specification.
+- MAY define extra fields
 
-[ERC-7828]: ./eip-7828.md
+## Rationale
 
--->
+- The `@` symbol is used as a separator as it provides visual clarity to humans, is easy for software to parse, and avoids confusion with the colon (`:`) symbol utilized in [CAIP-2] identifiers.
```
0xteddybear, line 248: the previous wording is a more comprehensive explaination of _why_ we chose to use the `@` character. In particular:
- it makes it trivial to distinguish CAIP-10 ids from Interoperable Names
- it also makes sure we can fit into this format anything that can fit in a CAIP-10

ERCS/erc-7930.md
```diff
@@ -205,25 +229,38 @@ Interoperable Address:
...
 
-[ERC-7828]: ./eip-7828.md
+## Rationale
 
--->
+- The `@` symbol is used as a separator as it provides visual clarity to humans, is easy for software to parse, and avoids confusion with the colon (`:`) symbol utilized in [CAIP-2] identifiers. 
+- No length restriction is placed on addresses, allowing for chains with longer address formats to be represented.
+- The address field includes `%` as a valid character to allow for URL-encoded address formats.
+- We chose to support zero-length addresses and chain IDs to make this standard flexible and to allow developers to use a single, uniform standard for many different jobs.
+- We chose **not** to utilize alternate encoding formats (`base58` or `base64` for example) to reduce the size of the _Interoperable Address_ and/or to abstract away namespace specific address serialization standards. We decided that using the chain specific address formats that users are used to provides a better user experience.
```
0xteddybear, line 252: This rewrite includes much less on _why_ we went for this approach, which is the purpose of the rationale section

clowestab:
Closing as changes merged here: https://github.com/ethereum/ERCs/pull/1395

PR #1267, opened 2025-10-16
Update ERC-7828: Updates to ERC-7828 to make it more clear and accessible.

clowestab:
Updates to ERC-7828 to make it more clear and accessible based on community feedback and discussions within the Interop working group.

 - The build passes.

### DIFF
--- a/ERCS/erc-7828.md
+++ b/ERCS/erc-7828.md
@@ -82,7 +82,11 @@ It is not guaranteed that an ENS name defined within the `<address>` component w
 
 In some cases resolvers return a new address on each resolution. In this scenario the validation of a previously generated checksum will always fail. 
 
-When checksum validation fails clients MUST alert the user and require explicit user input to continue with the operation.
+When checksum validation fails clients MUST warn the user and require explicit user input to continue with the operation.
+
+### Parsing and Serialization
+
+Once a `<name-service-address>` has been resolved to a `<raw-address>`, and the target chain has been determined, the binary _Interoperable Address_ can be generated. Clients MUST follow the normalization and serialization rules defined in the relevant [CAIP-350] profile for the chain in question. This ensures that different valid text representations (e.g., case variations in an address) resolve to a single, canonical binary form, which is essential for consistent checksum calculation and data integrity.
 
 ### Examples
 
@@ -127,7 +131,7 @@ example.eth@bitcoin#CCCCCCCC
 
 If the `<chain>` component matches the regular expression `[a-z-]+` it is to be interpreted as a label under a reserved ENS second-level domain, `<namespace>.eth`. It should be resolved subject to the ENS resolution process.
 
-If the `<chain>` is `ethereum`, you would resolve the `chain-id` data record for `ethereum.<namespace>.eth` subject to ENSIP-24: Arbitrary Data Resolution.
+If the `<chain>` component is `ethereum`, implementations MUST resolve the `chain-id` data record for `ethereum.<namespace>.eth` according to ENSIP-24: Arbitrary Data Resolution.
 
 Behind the scenes an all-in-one Registry-Resolver smart contract acts as a single source of truth for mapping these human-readable chain names to their corresponding chain identifiers encoded in the [ERC-7930] binary format.
 
@@ -177,7 +181,7 @@ The example below describes a possible flow for a user facing wallet. Lets assum
 3. The underlying label for 'Optimism' is 'optimism'. Append `<namespace.eth>` to that label => `optimism.<namespace>.eth`
 
 #### Step 3: Resolve the [ERC-7930] _Interoperable Address_
-4. Discern the resolver for `optimism.<namespace>.eth` and resolve `data(bytes32 namehash, string key)` (ENSIP-24: Arbitrary Data Resolution) passing in the `namehash` (see ENSIP-1) of `optimism.<namespace>.eth` and the key, `chain-id`. 
+4. Determine the resolver for `optimism.<namespace>.eth` and resolve `data(bytes32 namehash, string key)` (ENSIP-24: Arbitrary Data Resolution) passing in the `namehash` (see ENSIP-1) of `optimism.<namespace>.eth` and the key, `chain-id`. 
 
 #### Step 4: Resolve alice.eth **for** the specified chain
 5. Extract the CAIP-2 `ChainReference` from the [ERC-7930] identifier returned in the previous step.


ERCS/erc-7828.md
```diff
@@ -14,227 +14,208 @@ requires: 155, 7930
...
+
+#### Step 2: Build the _Interoperable Name_
+3. The underlying label for 'Optimism' is 'optimism'. Append `<namespace.eth>` to that label => `optimism.<namespace>.eth`
+
+#### Step 3: Resolve the [ERC-7930] _Interoperable Address_
+4. Discern the resolver for `optimism.<namespace>.eth` and resolve `data(bytes32 namehash, string key)` (ENSIP-24: Arbitrary Data Resolution) passing in the `namehash` (see ENSIP-1) of `optimism.<namespace>.eth` and the key, `chain-id`. 
+
+#### Step 4: Resolve alice.eth **for** the specified chain
+5. Extract the CAIP-2 `ChainReference` from the [ERC-7930] identifier returned in the previous step.
+6. Convert the `ChainReference` into an ENS `coinType` (see ENSIP-11).
```
ndeto: For non-EVM chains, ENSIP-11 does not apply.

We could define how both EVM and non-EVM chains can expose their coin type/SLIP-44 constant through the chain registry for consistency. 

Quick idea would be specifying a `coin-type` text key that returns the value?

0xrcinus: This would be equivalent to slip-44 in https://chainid.network/chains.json? It doesn't seem to be there for all chains in that list, so there would need to be some kind of fallback?

ERCS/erc-7828.md
```diff
@@ -14,227 +14,208 @@ requires: 155, 7930
 
 ## Abstract
 
-This proposal extends [ERC-7930](./eip-7930.md) (Interoperable Addresses & Names) by standardizing a human-readable format for chain-specific addresses in the form `<address>@<chain>#<checksum>`. It introduces:
+This proposal extends the definition of an _Interoperable Name_ outlined in [ERC-7930](./eip-7930.md). 
+
+The `<chain>` component definition is extended to allow for a human-readable chain name to be used. This identifier is resolved to an [ERC-7930] _Interoperable Address_ using the Ethereum Name Service (ENS). This moves chain metadata discovery on-chain.
```
0xteddybear, line 19: let's keep at least a mention of 7930's `<address>@<chain>#<checksum>` format

ERCS/erc-7828.md
```diff
@@ -14,227 +14,208 @@ requires: 155, 7930
...
 
-In the same spirit, the address could be a human-readable name as well, which is already a use case for ENS. By coupling the TLD to the resolving method used, this standard could leverage current and future features of ENS as well as other naming registries.
+In the same spirit, we propose that the `<address>` component of the definition allow for a human-readable name to be utilized. By coupling the TLD to the resolution method used, this standard could leverage current and future features of ENS as well as other naming systems.
 
-Moreover, the above can be leveraged to allow for a name to represent the same entity corresponding to different addresses on different chains, mitigating the risk of sending funds to an address the intended recipient doesn't actually control.
+The flexibility of this specification and the underlying naming systems can be leveraged to allow for different addresses on different chains to be represented by the same human-readable string. This mitigates the risk of sending funds to an address the intended recipient doesn't actually control.
+
+The usage of human-readable names provides additional clarity and confidence to users.
```
0xteddybear: this is vague and redundant

ERCS/erc-7828.md
```diff
@@ -14,227 +14,208 @@ requires: 155, 7930
...
-<chain-name>           := (<label> ".")* "<namespace>.eth"  
-<ens-name>:            := (<label> .)+ <label>
+Recall that [ERC-7930] defines an _Interoperable Name_ as a human readable representation of an underlying _Interoperable Address_.
+
+It's format is `<address> @ <chain> # <checksum>`
+
+These components now have the following extended definitions:
+
+```bnf
+<address>     := <raw-address> | <name-service-domain>
```
0xteddybear, line 61: this is significantly more vague and relies on prose instead of standard BNF notation

ERCS/erc-7828.md
```diff
@@ -14,227 +14,208 @@ requires: 155, 7930
...
-<chain>:               := <raw-address> | <chain-name>
-<checksum>:            := [0-9A-F]{8}
-
-<raw-address>:         := [-:_%a-zA-Z0-9]+
-<raw-chain>:           := [-:_a-zA-Z0-9]+
-<chain-name>           := (<label> ".")* "<namespace>.eth"  
-<ens-name>:            := (<label> .)+ <label>
+Recall that [ERC-7930] defines an _Interoperable Name_ as a human readable representation of an underlying _Interoperable Address_.
+
+It's format is `<address> @ <chain> # <checksum>`
```
0xteddybear: I believe it should be 'its'
also, having this in the prose is good but let's please also keep it in the formal definition as well

ERCS/erc-7828.md
```diff
@@ -14,227 +14,208 @@ requires: 155, 7930
...
-- `<label>`: Any valid string label per UTS46, as defined in ENSIP-1.
+`<raw-address>` is the namespace specific text representation of an address. For example chains within the `eip155` namespace utilize a 20-byte hex-encoded address format, whilst chains within the `bip122` namespace utilize a base58/bech32 encoded address format.
 
-This allows for Interoperable Addresses' text representation to mix and match 'resolved' and 'unresolved' usages in both the chain and address parts.
+`<name-service-domain>` is a domain name from a known, and supported naming service. For example an Ethereum Name Service (ENS) name like `example.eth`. These names should be resolved to a `<raw-address>` subject to the resolution specifications of the name service in question.
 
-A few examples below:
+`<caip-2-chain-id>` is the string representation of the [CAIP-2] blockchain identifier. For example, `eip155:1` or `bip122:000000000019d6689c085ae165831e93`. An ERC-7828 _Interoperable Name_ MUST specify both the namespace and reference elements of the CAIP-2 identifier as ENS resolves addresses on a chain specific basis.
```
0xteddybear, line 71: this is contradictory with 7930, why are you defining it like this?

0xrcinus: I think this only applies in the case where the address is an ENS name (i.e. `vitalik.eth@eip155` is underspecified because `vitalik.eth` could resolve to different addresses on different chains). `0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045@eip155` is valid but that isn't a 7828 name (it's readable 7930)

ERCS/erc-7828.md
ndeto: ENSIP-11 can derive the coin type for EVM chains using their chain ID. For non-EVM chains, we use the SLIP-44 constant

The idea is to abstract that complexity away and surface the appropriate value for each chain using a `coin-type` text record. i.e resolving the text record `coin-type` for `optimism.<namespace>.eth` returns the ENSIP-11 coin type / SLIP-44 value

clowestab: The original spec had the following: "_For addresses on chains that are both supported by BIP-44 registered coin type identifiers and the special scheme for the `eip155` namespace defined in ENSIP-11, the latter should be used._".

I will readd this for clarity, and add additional commentary to outline what @ndeto  has mentioned, namely:

_"The idea is to abstract that complexity away and surface the appropriate value for each chain using a coin-type text record. i.e resolving the text record coin-type for optimism.<namespace>.eth returns the ENSIP-11 coin type / SLIP-44 value"_

ERCS/erc-7828.md
clowestab: Yes, as outlined by @0xrcinus , this was discussed with @skeletor-spaceman - the problem is that ENS name resolution is chain specific NOT namespace specific. I wouldn't say this contradicts, but rather modifies the syntactic restrictions outlined in ERC-7930

ERCS/erc-7828.md
clowestab: Fair. Removed.

ERCS/erc-7828.md
clowestab: Correct. Fixed.

ERCS/erc-7828.md
clowestab: Are you requesting the definitions of `<raw-address>` and `<name-service-domain>` be in the bnf block too?

ERCS/erc-7828.md
clowestab: Done

ERCS/erc-7828.md
clowestab: OK. Re-added.

ERCS/erc-7828.md
```diff
@@ -14,227 +14,206 @@ requires: 155, 7930
...
+The address `0x123..789` on Ethereum Mainnet could be represented as either of the following:
 
-If a user-provided address includes a checksum, wallets MUST recompute it and compare it to the provided value. On mismatch, wallets MUST alert the user and MUST NOT allow sending funds without explicit override.
+```
+0x123..789@eip155:1#FFFFFFFF
+0x123..789@ethereum#FFFFFFFF
+```
 
-Wallets MAY display the checksum when formatting addresses, and MAY accept inputs without a checksum by computing it internally.
+The checksum is optional and is calculated as outlined in [ERC-7930].
```
0xMonoAx: To maintain consistency with ERC-7930, I would keep the previous sentence: _"Clients MAY accept Interoperable Name inputs without a checksum."_

ERCS/erc-7828.md
```diff
@@ -14,227 +14,206 @@ requires: 155, 7930
...
+This specification proposes that the `<chain>` component of the _Interoperable Name_ definition outlined in [ERC-7930] be extended to allow for a human-readable chain name resolved on-chain via ENS. This registry would provide a single source of truth for mapping a chain name to an [ERC-7930] _Interoperable Address_ and associated chain metadata.
 
-In the same spirit, the address could be a human-readable name as well, which is already a use case for ENS. By coupling the TLD to the resolving method used, this standard could leverage current and future features of ENS as well as other naming registries.
+In the same spirit, we propose that the `<address>` component of the definition allow for a human-readable name to be utilized. By coupling the TLD to the resolution method used, this standard could leverage current and future features of ENS as well as other naming systems.
```
0xMonoAx: > In the same spirit, we propose...

Change to:

_“This specification proposes…”_

(EIPs prefer an impersonal tone)

ERCS/erc-7828.md
```diff
@@ -14,227 +14,206 @@ requires: 155, 7930
...
+It is not guaranteed that an ENS name defined within the `<address>` component will always resolve to the same address. This is exactly why checksums are useful. 
 
-Assuming the on-chain registry list adds a few other CAIP namespaces
-- alice.eth@btc#CCCCCCCC
-```
+In some cases resolvers return a new address on each resolution. In this scenario the validation of a previously generated checksum will always fail. 
 
-> [!NOTE]
-> This standard explicitly defines the use of an agreed <namespace>.eth second-level domain for chain names. Beyond this, hierarchical name structures continue to be inherited from ENS, as exemplified in the ENS-in-rollup example.
+When checksum validation fails clients MUST alert the user and require explicit user input to continue with the operation.
```
0xMonoAx: > ..and require explicit user input to continue with the operation.

This is not specified in ERC-7930. Should we add it?

0xMonoAx: Since ERC- 7930 delegates parsing to CAIP-350, it would be ideal to repeat a line here that reinforces that requirement.

Similar: https://github.com/ethereum/ERCs/pull/1265#discussion_r2471469739

ERCS/erc-7828.md
```diff
@@ -14,227 +14,206 @@ requires: 155, 7930
...
+If the `<chain>` component matches the regular expression `[a-z-]+` it is to be interpreted as a label under a reserved ENS second-level domain, `<namespace>.eth`. It should be resolved subject to the ENS resolution process.
 
-1.  Apply UTS-46 lower-case normalisation to the chain name.
-2.  Append the reserved suffix: `full = "arbitrum" + ".l2.eth"`.
-3.  Compute `node = namehash(full)` as per ENSIP-1.
-4.  Build resolver calldata: `callData = abi.encodeWithSelector(IL2Resolver.chainId.selector, node)`.
-5.  Query the Universal Resolver: `result = UniversalResolver.resolve(dnsEncode(full), callData)`.
-6.  ABI-decode `result` as `(bytes chainBytes)`.
-    • If `chainBytes` is empty → no mapping exists; treat the label as unknown.
+If the `<chain>` is `ethereum`, you would resolve the `chain-id` data record for `ethereum.<namespace>.eth` subject to ENSIP-24: Arbitrary Data Resolution.
```
0xMonoAx: > If the `<chain>` is `ethereum`, you would..

“You” is not used in standards. I propose:

_If the `<chain>` component is `ethereum`, implementations MUST resolve the `chain-id` data record for ethereum.`<namespace>.eth` according to [ENSIP-24: Arbitrary Data Resolution]._

ERCS/erc-7828.md
```diff
@@ -14,227 +14,206 @@ requires: 155, 7930
...
 
+Behind the scenes an all-in-one Registry-Resolver smart contract acts as a single source of truth for mapping these human-readable chain names to their corresponding chain identifiers encoded in the [ERC-7930] binary format.
 
-#### Supported CASA namespaces
 
-While currently only `eip155` chains are supported, the on-chain registry could start listing other chains in order to make human-readable names for non-evm chains possible.
+### Reverse resolution
 
-### Resolving address names
+In the context of this specification 'Reverse Resolution' refers to the process of discerning the `<chain-label>` of a chain from its _Interoperable Address_.
```
0xMonoAx, line 141: > 'Reverse Resolution' 

change to:

_Reverse Resolution_

ERCS/erc-7828.md
```diff
@@ -14,227 +14,206 @@ requires: 155, 7930
...
+
+#### Step 1: User Input
+1. Let the user input an ENS name. For examples, `alice.eth`.
+2. Let the user select a chain. In this case, the user would select 'Optimism'.
+
+#### Step 2: Build the _Interoperable Name_
+3. The underlying label for 'Optimism' is 'optimism'. Append `<namespace.eth>` to that label => `optimism.<namespace>.eth`
+
+#### Step 3: Resolve the [ERC-7930] _Interoperable Address_
+4. Discern the resolver for `optimism.<namespace>.eth` and resolve `data(bytes32 namehash, string key)` (ENSIP-24: Arbitrary Data Resolution) passing in the `namehash` (see ENSIP-1) of `optimism.<namespace>.eth` and the key, `chain-id`.
```
0xMonoAx: “Dicern” is ambiguous; I recommend using “Determine.”

ERCS/erc-7828.md
```diff
@@ -14,227 +14,206 @@ requires: 155, 7930
...
-- Wallet developers should be aware of possible unicode glyph collisions in resolved names and warn users about them in order to keep checksums effective, since an attacker could, in order to impersonate `alice.eth@chain.<<namespace>.eth#00112233`:
-    - Mine an address on `chain.<namespace>.eth` such that the checksum of the Interoperable Address is `00112233`.
-    - Register `аlice.eth` (using the russian vowel `а` instead of the latin `a`), and point it to the address above.
-- ENSIP-19 introduces trust assumptions when resolving reverse records offchain:
-    - ZK-rollups may offer trust-minimized resolution via on-chain proofs.
-    - Sidechains may rely on centralized CCIP-read gateways or signature authorities.
+Consideration should be given to ENSIP-15: Name Normalization so as to avoid homograph attacks. 
 
-    Wallets MUST surface whether reverse resolution was trust-minimized or trusted.
+Wherever possible clients should utilize the optional checksum outlined in [ERC-7930]. It should be verified when a raw _Interoperable Name_ is pasted, and it should be appended when an _Interoperable Name_ is being generated based on user input.
```
0xMonoAx, line 214: Should you define what happens when the resolver returns invalid data?

clowestab: Which aspect are you referring to here? Are you suggesting documenting that if (for example) a non existent chain name is resolved `0x` will be returned?

ERCS/erc-7828.md
clowestab: Updated.

ERCS/erc-7828.md
clowestab: Updated

ERCS/erc-7828.md
clowestab: I've added a block about this in the context of 7828.

ERCS/erc-7828.md
clowestab: Updated. I havn't linked to ENSIP-24 as ENSIPs are not in the allowed external links..

ERCS/erc-7828.md
clowestab: Fixed

ERCS/erc-7828.md
clowestab: Fixed

ERCS/erc-7828.md
clowestab: A similar statement is included in 7930. I've added to it for completeness: _"If a user-provided _Interoperable Name_ includes a checksum, clients MUST derive the underlying _Interoperable Address_, recalculate the checksum, and compare it to the provided value. In case of a mismatch, clients MUST warn the user and require explicit user input to continue with the operation."_

PR #1289, opened 2025-10-22
Update ERC-7730: Remove embedded ABI; keys-as-schema

llbartekll:
## Summary
This PR simplifies ERC-7730 by removing embedded ABIs and making display.formats the single source of truth for routing, decoding, and labeling.

Functions only: scope limited to calldata clear-signing; events/errors out of scope.

Keys = schema & router: each display.formats key is a Human-Readable ABI function with parameter names (e.g., transfer(address to,uint256 value)).

Selector matching: wallets compute keccak256(<type-only signature>)[:4] from the key to match calldata.

Decoding from key: wallets parse the key’s canonical types to decode; names from the key populate paths.

Why: reduces duplication and drift, shrinks descriptors, and aligns with existing wallet UX while keeping overloads and complex types (tuples/arrays) precise via canonical types.

### DIFF
--- a/assets/erc-7730/erc7730-v1.schema.json
+++ b/assets/erc-7730/erc7730-v1.schema.json
@@ -352,7 +352,7 @@
 
                 "formats": {
                     "title": "List of field formats",
-                    "description": "The list includes formatting info for each field of a structure. This list is indexed by the full function signature with parameter names.",
+                    "description": "The list includes formatting info for each field of a structure. For contract bindings, entries are keyed by the full function signature with parameter names; for EIP712 bindings, entries are keyed by the message primary type.",
                     "type": "object",
                     "propertyNames": {
                         "pattern": "^[A-Za-z_][A-Za-z0-9_]*\\(\\s*(?:(?:(?:tuple(?:\\s*\\((?:[^()]|\\([^()]*\\))*\\))?|[A-Za-z0-9_]+)(?:\\[[0-9]*\\])*(?:\\s+(?:memory|calldata|storage))?\\s+[A-Za-z_][A-Za-z0-9_]*)(?:\\s*,\\s*(?:(?:tuple(?:\\s*\\((?:[^()]|\\([^()]*\\))*\\))?|[A-Za-z0-9_]+)(?:\\[[0-9]*\\])*(?:\\s+(?:memory|calldata|storage))?\\s+[A-Za-z_][A-Za-z0-9_]*))*\\s*)?\\)$"


kuzdogan:
I'm actually thinking if a full ABI JSON in this spec is an overkill. In a way this information is redundant because we are already providing it in the format spec of each function:

```json
        "formats": {
            "myFunction(address _address,uint256 _amount,MyParamType _param)": {
```

The key here contains all the information needed to identify and ABI-decode the parameters. This is the preferred format in [ethers.js](https://docs.ethers.org/v5/api/utils/abi/formats/#abi-formats--human-readable-abi) and is fully translatable to other formats. 

We can even fully remove the "abi" field and require proper formatting of `display.formats` keys.

Right now the spec allows 3 possibilities on this key: 
> For contracts, the key names MUST be one of those three forms:
> -    A 4-bytes selector of the function being described, in hex string notation: the corresponding function MUST be in the abi specified in context.contract.abi.
> - The function signature used to compute the selector, i.e. without parameter names and without spaces (refer to Solidity documentation): the corresponding function MUST be in the abi specified in context.contract.abi.
> - A full solidity signature, including parameter names: the corresponding function MUST be in the abi specified in context.contract.abi and the parameters names MUST match those in the ABI.

With that the spec would only allow the last one, "full Solidity signature", or maybe better defined as the [ether.js Human-readable ABI](https://docs.ethers.org/v5/api/utils/abi/formats/#abi-formats--human-readable-abi)

The ABI itself requires a verification/validation step anyway so I don't see the benefit of including the full ABI.

ERCS/erc-7730.md
```diff
@@ -89,26 +77,21 @@ The following is an example of how to clear sign a `transfer` function call on a
...
-                        "params": {
-                            "tokenPath": "@.to"
-                        }
+                        "path": "to",
+                        "label": "To",
+                        "format": "addressName"
                     }
-                ],
-                "required": ["_to", "_value"],
-                "excluded": []
```
kuzdogan, line 100: Maybe keep these in the example? On lines 118-119 (in new changes) we also describe what these fields do.

ERCS/erc-7730.md
```diff
@@ -161,7 +144,7 @@ Displaying structured data is often done by wallets to review its content before
 
 Current specification covers EVM smart contract calldata:
 * Defined in Solidity
-* The schema is the ABI (expected in json format when linked to)
+* The schema is derived from the matched `display.formats` function fragment
```
kuzdogan: The schema of the "EVM smart contract calldata" in this spec is still the ABI. Here it's referring to the binary "Application Binary Interface" format of EVM. And for each contract function we can infer this format in the "human-readable ABI fragment" in this spec.

```suggestion
* The schema is the ABI (derived from the matched `display.formats` function fragment)
```

ERCS/erc-7730.md
```diff
@@ -200,10 +183,10 @@ For paths referring to structured data fields, if a field has a variable length
...
 
 References to values in the serialized structured data
-* `#.params.amountIn` or `params.amountIn` refers to parameter `amountIn` in top-level structure `params` as defined in the ABI
-* `#.data.path.[0].path.[-1].to` or `data.path.[0].path.[-1].to` refers to the field `to` taken from last member of `path` array, itself taken from first member of enclosing `path` array, itself part of top level `data` structure.
-* `#.params.path.[:20]` or `#.params.path.[0:20]` refers to the first 20 bytes of the `path` byte array
-* `#.params.path.[-20:]` refers to the last 20 bytes of the `path` byte array
+* `#.` or relative `value` refers to the decoded argument named `value` from the matched function fragment
+* `#.order.price` refers to the `price` field inside the `order` tuple argument
+* `#.recipients.[0]` refers to the first entry of the `recipients` dynamic array
+* `#.order.path.[0:20]` refers to the first 20 bytes of a `path` byte array nested inside the `order` tuple
```
kuzdogan: I see we are dropping `params` `data` etc. Can we just drop them like this?

ERCS/erc-7730.md
```diff
@@ -200,10 +183,10 @@ For paths referring to structured data fields, if a field has a variable length
 *Examples*
 
 References to values in the serialized structured data
-* `#.params.amountIn` or `params.amountIn` refers to parameter `amountIn` in top-level structure `params` as defined in the ABI
-* `#.data.path.[0].path.[-1].to` or `data.path.[0].path.[-1].to` refers to the field `to` taken from last member of `path` array, itself taken from first member of enclosing `path` array, itself part of top level `data` structure.
-* `#.params.path.[:20]` or `#.params.path.[0:20]` refers to the first 20 bytes of the `path` byte array
-* `#.params.path.[-20:]` refers to the last 20 bytes of the `path` byte array
+* `#.` or relative `value` refers to the decoded argument named `value` from the matched function fragment
```
kuzdogan: Do we mean `#.value` there instead of `#.`?

ERCS/erc-7730.md
```diff
@@ -852,8 +899,8 @@ Interpolated paths MUST reference fields that have corresponding format specific
     "interpolatedIntent": "Approve {spender} to spend up to {amount} on your behalf until {deadline}",
     "fields": [
         {"path": "spender", "label": "Spender", "format": "addressName"},
-        {"path": "amount", "label": "Amount", "format": "tokenAmount", "params": {"tokenPath": "@.to"}},
```
lcastillo-ledger, line 855: params should not be dropped here and in next line

assets/erc-7730/erc7730-v1.schema.json
```diff
@@ -362,7 +349,7 @@
 
                 "formats": {
                     "title": "List of field formats",
-                    "description": "The list includes formatting info for each field of a structure. This list is indexed by a key identifying uniquely the message's type in the abi. For smartcontracts, it is the selector of the function or its signature; and for EIP712 messages it is the primaryType of the message.",
+                    "description": "The list includes formatting info for each field of a structure. This list is indexed by the full function signature with parameter names.",
```
lcastillo-ledger: Missing the index type for EIP-712 (since they do not have an equivalent of human readable ABI)

assets/erc-7730/erc7730-v1.schema.json
```diff
@@ -66,19 +66,6 @@
                     "description": "The contract binding context is a set constraints that are used to bind the ERC7730 file to a specific smart contract.",
 
                     "properties": {
-                        "abi": {
```
lcastillo-ledger, line 69: Can we leave the ABI key and mark it as deprecated? To avoid breaking a lot of existing files?

ERCS/erc-7730.md
lcastillo-ledger: +1 to keeping them, "required" is actually required in this example

assets/erc-7730/erc7730-v1.schema.json
```diff
@@ -362,7 +349,7 @@
 
                 "formats": {
                     "title": "List of field formats",
-                    "description": "The list includes formatting info for each field of a structure. This list is indexed by a key identifying uniquely the message's type in the abi. For smartcontracts, it is the selector of the function or its signature; and for EIP712 messages it is the primaryType of the message.",
+                    "description": "The list includes formatting info for each field of a structure. This list is indexed by the full function signature with parameter names.",
                     "type": "object",
```
lcastillo-ledger, line 356: Since we're making the key itself a reference of the schema, might be useful to validate the format itself using the schema. Something like this should work:
```
"propertyNames": {
    "pattern": "^[A-Za-z_][A-Za-z0-9_]*\\([A-Za-z0-9_\\[\\]()]+ [A-Za-z_][A-Za-z0-9_]*(?:,[A-Za-z0-9_\\[\\]()]+ [A-Za-z_][A-Za-z0-9_]*)*\\)$"
  }
```

ERCS/erc-7730.md
llbartekll: added it back

assets/erc-7730/erc7730-v1.schema.json
llbartekll: Thanks for surfacing that. I’ve reintroduced a stricter propertyNames guard so schema validation enforces the “full function signature with parameter names” contract. The pattern now allows:

  - Named parameters with optional array suffixes and memory|calldata|storage qualifiers.
  - Tuple/struct parameters, including nested tuples (tuple((uint256,address) info)).
  - Optional whitespace and the empty-parameter case (pause()).

  Here’s what landed:
```
  "propertyNames": {
    "pattern": "^[A-Za-z_][A-Za-z0-9_]*\\(\\s*(?:(?:(?:tuple(?:\\s*\\((?:[^()]|\\([^()]*\\))*\\))?|[A-Za-z0-9_]+)(?:\\[[0-9]*\\])*(?:\\s+(?:memory|calldata|
  storage))?\\s+[A-Za-z_][A-Za-z0-9_]*)(?:\\s*,\\s*(?:(?:tuple(?:\\s*\\((?:[^()]|\\([^()]*\\))*\\))?|[A-Za-z0-9_]+)(?:\\[[0-9]*\\])*(?:\\s+(?:memory|
  calldata|storage))?\\s+[A-Za-z_][A-Za-z0-9_]*))*\\s*)?\\)$"
  }
```

assets/erc-7730/erc7730-v1.schema.json
llbartekll: added a short mention

assets/erc-7730/erc7730-v1.schema.json
Frank122480: Thanx for all the help

lcastillo-ledger (approved):

lightclient (approved):

PR #1319, opened 2025-10-28
Add ERC: Groups - Multi-Member Containers

jamesavechives:
This proposal specifies a general-purpose "Group" primitive as a first-class, onchain object. Each group is deployed as its own contract (identified by its contract address) and has an owner, an extensible set of members, and optional metadata. The standard defines a minimal interface for direct member management and membership introspection. Unlike token standards (e.g., [ERC-20](./eip-20.md)/[ERC-721](./eip-721.md)) that model units of transferable ownership, Groups model multi-party membership and coordination. The goal is to enable interoperable social, organizational, and collaborative primitives.

### DIFF
--- a/ERCS/erc-7984.md
+++ b/ERCS/erc-7984.md
@@ -17,7 +17,7 @@ The following standard describes confidential fungible tokens via pointers. All 
 
 ## Motivation
 
-Confidential tokens enable private value transfer which is vital for many usecases such as payroll, confidential DeFi, institutional settlement, and more. 
+Confidential tokens enable private value transfer which is vital for many usecases such as payroll, confidential DeFi, institutional settlement, and more.
 A standard interface allows pointer based confidential tokens on Ethereum to be reused by other applications: from privacy-focused wallets to decentralized exchanges, while keeping transaction amounts private from public view.
 
 ## Specification
@@ -26,7 +26,7 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 
 ### Nomenclature
 
-- All amounts in this ERC are pointer based amounts represented by `bytes32` pointers unless otherwise specified. The resolution and manipulation of these pointers is implementation specific.
+All amounts in this ERC are pointer based amounts represented by `bytes32` pointers unless otherwise specified. The resolution and manipulation of these pointers is implementation specific.
 
 ### Token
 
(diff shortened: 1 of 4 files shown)


ERCS/erc-0000.md
```diff
@@ -0,0 +1,131 @@
+---
+eip: 0000
```
abcoathup:
```suggestion
eip: 8063
```

ERCS/erc-8063.md
```diff
@@ -0,0 +1,131 @@
+---
+eip: 0000
+title: Groups — Multi-member onchain containers
+description: Onchain groups with membership, invitations, and shared resource metadata.
+author: James Savechives (@jamesavechives)
+discussions-to: https://ethereum-magicians.org/t/groups-multi-member-onchain-containers-for-shared-resources/25999
```
abcoathup:
```suggestion
discussions-to: https://ethereum-magicians.org/t/groups-multi-member-onchain-containers-for-shared-resources/25999
```
Updated with assigned number

sullof:
Interesting proposal. From my perspective, the same functionality could be obtained by modifying ERC-721 or extending it with ERC-7656–style patterns. What is the motivation for exposing an ERC-20 interface that appears potentially misleading given its limitations?

ERCS/erc-8063.md
```diff
@@ -0,0 +1,147 @@
...
+
+    /// @notice Transfer membership to another address (caller loses membership)
+    /// @param to Address to receive membership (must not already be a member)
+    function transferMembership(address to) external;
+}
+```
+
+### Semantics
+
+- Group creation: The contract constructor MUST set the deployer as the group owner and initial member. Implementations SHOULD accept a name and optional metadataURI as constructor parameters.
```
xinbenlv: Mandating constructor behavior restricts the use of proxy patterns (e.g., ERC-1167, ERC-1967) where initialization happens via a separate function. Consider defining the requirement in terms of state upon initialization rather than strictly the constructor.

```suggestion
- Group creation: Upon creation or initialization, the contract MUST set the deployer (or initializer) as the group owner and initial member. Implementations SHOULD accept a name and optional metadataURI.
```

ERCS/erc-8063.md
```diff
@@ -0,0 +1,147 @@
...
+- Adding members: Only the group owner MAY call `addMember`. The account MUST NOT already be a member. The implementation MUST add them as a member and emit `MemberAdded`.
+- Leaving: Any member (except the owner) MAY call `leaveGroup` to voluntarily exit. The implementation MUST remove their membership and emit `MemberLeft`. The owner MUST NOT be allowed to leave.
```
xinbenlv: The restriction "The owner MUST NOT be allowed to leave" ensures stability but creates ambiguity if ownership transfer is implemented (as an extension). If ownership is transferred, does the *new* owner become a member automatically? Can the *old* owner then leave? Consider clarifying that this restriction applies to the *current* owner address.

ERCS/erc-8063.md
```diff
@@ -0,0 +1,147 @@
...
+### Deployment model
+
+Following the [ERC-20](./eip-20.md) pattern, each group is its own contract deployment:
+
+- **Group identity**: A group is uniquely identified by its contract address, just as an ERC-20 token is identified by its contract address.
+- **Deployment**: To create a group, deploy a new contract implementing `IERC8063`. The deployer becomes the owner and initial member.
+- **Discovery**: Applications can discover groups through events, registries, or direct address references, similar to token discovery patterns.
+- **Naming**: Implementations SHOULD accept a human-readable name in the constructor and expose it via `name()`. Empty names are permitted.
+
+### Optional extensions (non-normative)
```
xinbenlv, line 91: Since `owner()` is part of the core interface, the lack of a standardized `transferOwnership` function limits interoperability for management tools (e.g., a dashboard allowing a user to hand off a group). Consider adding `function transferOwnership(address newOwner) external;` to the interface or explicitly addressing why it is excluded in the Rationale.

ERCS/erc-8063.md
```diff
@@ -0,0 +1,147 @@
...
+
+- `decimals()` MUST return `0`.
+- `balanceOf(account)` MUST be either `0` (non-member) or `1` (member).
+- `totalSupply()` MUST equal the current member count.
+- `transfer(to, amount)` MUST require `amount == 1` and MUST transfer membership of the caller to `to`.
+- `approve(spender, amount)` and `allowance(owner, spender)` MAY be implemented with values clamped to `{0,1}`; `transferFrom(from, to, amount)` MUST require `amount == 1` and, if permitted, MUST transfer membership from `from` to `to`.
+- All operations MUST preserve the invariant that no account's balance exceeds `1` and MUST NOT allow the owner account to be transferred.
+
+Reference: see the ERC-20 compatibility implementation in the assets directory.
+
```
xinbenlv, line 131: The rationale for the ERC-20 compatibility mode is missing. Please explain *why* an implementer would want to expose a Group as an ERC-20 token (e.g., for compatibility with existing governance tools or wallet visualization). This addresses the feedback raised in the community discussion.

assets/erc-8063/README.md
```diff
@@ -0,0 +1,8 @@
+Groups reference implementation assets
+
+- After creating the Ethereum Magicians discussion, update `discussions-to` in `ERCS/EIP-0000.md` with the live URL.
+- Contracts:
+  - `IERCGroup.sol`: interface
```
xinbenlv: The filenames listed here do not match the actual files provided in the PR (`IERC8063.sol` and `ERC8063.sol`).

```suggestion
  - `IERC8063.sol`: interface
  - `ERC8063.sol`: minimal implementation
```

jamesavechives:
> Interesting proposal. From my perspective, the same functionality could be obtained by modifying ERC-721 or extending it with ERC-7656–style patterns. What is the motivation for exposing an ERC-20 interface that appears potentially misleading given its limitations?

Thanks for the thoughtful review.

### Why not extend ERC‑721 (or ERC‑7656‑style patterns)?
ERC‑8063 is intentionally **not token‑based**. ERC‑721/1155 extensions still frame the primitive as “ownership of a token,” which brings along token semantics (token IDs, approvals, enumerable assumptions, marketplace expectations, etc.). A group, in our view, is a **first‑class on‑chain container/identity** identified by its **contract address**, with a minimal membership surface (`addMember`, `leaveGroup`, `transferMembership`, `isMember`). That keeps the primitive:
- token‑agnostic (no token IDs, no scarcity assumptions),
- lightweight for integrators (one membership check),
- composable with existing standards (resources can be associated externally like with any address).

You *can* model membership via ERC‑721/1155, but it’s a different abstraction: **“you own a token”** vs **“you are a member of a group.”** ERC‑8063 aims to standardize the latter directly.

### Motivation for the ERC‑20 interface (and why it’s optional)
The ERC‑20 interface is **explicitly optional** and meant as an **interoperability shim**, not as “this is a fungible token.” Many existing tools/protocols already accept ERC‑20 inputs (portfolios, accounting, distribution contracts, governance modules, indexers). Exposing an ERC‑20 view with `decimals = 0` and strict invariants (balances in {0,1}, transfer amount must be 1, supply == member count) allows those systems to integrate membership with minimal custom code.

To avoid misleading semantics, the spec constrains behavior so it’s clear this is a **membership tokenization adapter**:
- balances are binary (member/non‑member),
- transfers represent membership handoff,
- optional and implementations may omit it entirely.

So the motivation is **compatibility with existing ERC‑20‑centric infrastructure**, while keeping the *core* primitive membership‑centric and minimal.

ERCS/erc-8063.md
jamesavechives: Good catch — agreed this needs to be explicit.

In the base spec, **owner is always a member** (owner set at deployment/initialization, and membership is binary), so the intent of “owner MUST NOT be allowed to leave” is **to prevent the group from ending up with no owner-member**.

To remove ambiguity when an ownership-transfer extension is implemented, we can clarify:

- The restriction applies to the **current** `owner()` address at call time: **`leaveGroup()` MUST revert if `msg.sender == owner()`**.
- If an ownership transfer mechanism is added, the implementation **MUST ensure the new owner is a member** (i.e., `isMember(newOwner) == true` after transfer). This can be done by requiring the new owner is already a member, or by automatically adding them during the transfer.
- After ownership transfer, the old owner is no longer `owner()`, so they **may leave** (or transfer) like any other member, assuming they remain a member.

I’ll update the wording in the semantics section to say “the **current owner address** MUST NOT be allowed to leave,” and add a note in the ownership-transfer extension about the new owner’s membership requirement.

ERCS/erc-8063.md
jamesavechives: Thanks — I agree this is an important interoperability consideration.

We left `transferOwnership` out of the **core** interface intentionally to keep ERC‑8063 minimal and avoid forcing a specific governance / admin model. Ownership handoff can be contentious (e.g., should it require the new owner to already be a member? should it be 2‑step accept? should it be multisig/vote‑based?), and different applications will want different policies.

That said, management tooling is a valid motivation for standardizing it. Two ways forward:

1) **Keep core minimal**, but add a **standard optional extension** (recommended) such as `IERC8063Ownership`:
- `event OwnershipTransferred(address indexed previousOwner, address indexed newOwner);`
- `function transferOwnership(address newOwner) external;`
with semantics like:
- only current `owner()` can call,
- new owner MUST be (or become) a member,
- after transfer, the new owner is the “current owner” for the `leaveGroup()` restriction.

2) If the group expects “owner handoff” to be a universal need, we can move it into the core interface.

My preference is (1): define it as an **optional extension** so dashboards can interoperate where supported, without constraining all implementations. I’ll update the spec to either add this optional extension explicitly or explain in Rationale why it’s not in core.

ERCS/erc-8063.md
jamesavechives: Agreed — that’s a fair request, and we should make the motivation explicit.

The ERC‑20 compatibility mode is **optional** and intended as an **interoperability shim** for existing infrastructure that already understands ERC‑20 but doesn’t have a notion of “membership contracts.” Concretely, exposing a constrained ERC‑20 view (decimals=0, balances ∈ {0,1}, transfer amount=1) lets implementers plug groups into:

- **Wallet / portfolio UIs** that only display token holdings (a member sees balance=1, non‑member sees 0).
- **Governance and permission tooling** that accepts an ERC‑20 address as the “membership/weight” source (even if weights are binary rather than proportional).
- **Distribution / accounting / gating utilities** that operate over ERC‑20 interfaces and can treat membership as a boolean token.

The spec also constrains behavior to avoid misleading semantics: this is not a fungible asset; it’s a binary “membership tokenization” adapter.

I’ll add a short rationale paragraph under the ERC‑20 compatibility section (or in Rationale) explaining these motivations and explicitly calling it an optional adapter for compatibility with existing tools.

jamesavechives:
> Thank you for this proposal! The concept of a 'Group' primitive that separates membership from asset ownership is a valuable abstraction for onchain coordination. The draft is well-structured and the distinction from existing token standards is clear.
> 
> The proposal is in good shape, but requires a response to community feedback regarding the ERC-20 compatibility motivation. Additionally, please address the flexibility issue regarding constructors/proxies and fix the filenames in the assets README.

 Thanks for the review!

 - **ERC‑20 compatibility motivation**: added a non‑normative rationale explaining this OPTIONAL mode is an interoperability shim for ERC‑20‑centric tooling (wallet/portfolio display, accounting/distribution utilities, governance/gating systems), with strict constraints to avoid implying fungible-value semantics. (Commit `774074dc`)
 - **Constructor/proxy flexibility**: updated semantics to cover creation **or initialization** (deployer/initializer). (Commit `774074dc`)
 - **Assets README filenames**: fixed stale filenames in the ERC‑8063 assets README (removed `GroupContainer.sol`, listed current contracts including the optional ERC‑20 compatibility implementation). (Commit `e3aba30c`)

ERCS/erc-8063.md
```diff
@@ -0,0 +1,167 @@
+---
+eip: 8063
+title: Groups - Multi-Member Containers
+description: Onchain groups with membership and direct member management.
```
SamWilsn, line 4: So, there's nothing wrong really with your description, but... I feel like it's missing... something. Like I can imagine what an "on-chain group" might be, but I don't know if my imagining matches what this proposal is about. Maybe include "groups of what" or the purpose behind the groups?

In any case, this isn't a hard blocker for merging, just some musing on my part.

ERCS/erc-8063.md
```diff
@@ -0,0 +1,167 @@
...
+    /// @notice Transfer membership to another address (caller loses membership)
+    /// @param to Address to receive membership (must not already be a member)
+    function transferMembership(address to) external;
+}
+```
+
+### Semantics
+
+- Group creation: Upon creation or initialization, the contract MUST set the deployer (or initializer) as the group owner and initial member. Implementations SHOULD accept a name and optional metadataURI.
+- Adding members: Only the group owner MAY call `addMember`. The account MUST NOT already be a member. The implementation MUST add them as a member and emit `MemberAdded`.
```
SamWilsn, line 77: > The account [...]

Ambiguous: do you mean the group owner's account, or the account being added.

SamWilsn, line 77: > Only the group owner MAY call [...]

This is oddly worded. You're creating a requirement on humans here, instead of on the contract's observable behaviour.

Something like:

```markdown
When called by a non-owner account, `addMember` MUST revert.
```

might be more understandable?

ERCS/erc-8063.md
```diff
@@ -0,0 +1,167 @@
...
+- Leaving: Any member MAY call `leaveGroup` to voluntarily exit. The implementation MUST remove their membership and emit `MemberLeft`. The **current owner address** (i.e., `owner()` at the time of the call) MUST NOT be allowed to leave.
```
SamWilsn, line 78: Maybe a little overly restrictive? Is it possible an implementation might assign a new owner if the current owner leaves instead, or is there a solid reason to prevent that behaviour?

ERCS/erc-8063.md
```diff
@@ -0,0 +1,167 @@
...
+    function transferOwnership(address newOwner) external;
+}
+```
+
+If implemented:
+
+- Only the current `owner()` MUST be allowed to call `transferOwnership`.
+- The `newOwner` MUST be a member after the transfer completes. Implementations MAY require `isMember(newOwner) == true` beforehand, or MAY add `newOwner` as a member during the transfer.
+- The `leaveGroup()` restriction applies to the **current** `owner()` address (i.e., after ownership is transferred, the previous owner is no longer restricted by that rule).
+
```
SamWilsn, line 118: Might want a note about ERC-165 here as well.

ERCS/erc-8063.md
```diff
@@ -0,0 +1,167 @@
...
+## Backwards Compatibility
+
+No known backward compatibility issues. This is a new interface with [ERC-165](./eip-165.md) detection.
+
+## Reference Implementation
+
+Reference contracts are provided in the `assets/` directory:
+
+- `IERC8063.sol` — Interface definition
+- `ERC8063.sol` — Minimal reference implementation
```
SamWilsn, line 154: You can link to them from here.

ERCS/erc-8063.md
```diff
@@ -0,0 +1,167 @@
...
+
+## Security Considerations
+
+- Membership griefing: Implementations SHOULD bound per-tx iterations and avoid unbounded member enumeration onchain.
+- Owner privileges: The owner has unilateral power to add members. Applications relying on group membership SHOULD consider this trust assumption.
+- Permanent membership: Members cannot be forcibly removed by the owner, only voluntarily leave or transfer. This design choice prioritizes decentralization but means malicious members cannot be expelled without governance extensions.
+- Access control: Clearly document who can add members; the baseline implementation grants this power exclusively to the owner.
+
```
SamWilsn, line 162: Avoid creating requirements (with UPPERCASE keywords) outside of the specification section. If the requirement is necessary to make a secure implementation, include it in the specification section and discuss its implications here.

ERCS/erc-8063.md
jamesavechives: Thanks — that’s a great point. The intent is “groups of addresses/accounts” (EOAs and contracts) as a lightweight onchain membership primitive for coordination/access control, not a token or asset container. I’ll adjust the Abstract/description to explicitly say “a group of Ethereum accounts” and add 1–2 concrete examples (teams/DAOs/guilds/cohorts) so readers don’t have to infer what’s being grouped.

ERCS/erc-8063.md
jamesavechives: Good point, thanks. I updated the Abstract/description to explicitly state these are groups of Ethereum accounts (EOAs and contracts) and added a short note on intended coordination/access-control use cases with examples.

jamesavechives: Good catch — I meant the account being added. I’ve updated the semantics to remove the ambiguity by stating the observable behavior: addMember MUST revert if called by a non-owner, and when called by owner() it MUST revert if the target account is already a member; otherwise it adds that account and emits MemberAdded(account, msg.sender).

ERCS/erc-8063.md
jamesavechives: Good point — that restriction was too strong for the base spec. I updated the semantics so that if the current owner() calls leaveGroup, implementations MAY either revert or reassign ownership to another existing member and then allow the owner to leave. This keeps “owner can’t leave” as a valid stability choice while permitting implementations that support automatic owner succession.

ERCS/erc-8063.md
jamesavechives: ok, done!

PR #1321, opened 2025-10-28
Update ERC-7730: Added common keywords section to ERC-7730

arikg:


llbartekll (approved):

ERCS/erc-7730.md
```diff
@@ -1380,6 +1380,9 @@ In the future, this specification could be extended to structured data like Meta
...
 
+### Common Keywords
+The DeFi ecosystem has matured and has some repeating usecases with well defined terms for different actions. While this ERC does not define specific standards for DEX or Lending protocol intents, for the sake of easier user readibility, teams with similar products should align on similar terminology to describe actions and actors. We encourage the creation of another layer of standartization on top of ERC-7730 to be recommeneded and even enforced in clear signing registries for improved user security.
```
kuzdogan, line 1384: A real-world example would be helpful here IMO.

arikg: just generally adding a mention of dex swap interface or a full definition of what it would look like as a "standard". That would seem excessive to me in this ERC, as it's just a consideration

kuzdogan: I didn't want to mean a detailed example but more like: 

> ...well defined terms for different actions e.g. using "Swap" for all similar actions

but just a small suggestion. Feel free to include or not :)

PR #1355, opened 2025-11-18
Update ERC-7780: Add module type 8,9,10

leekt:

 - The build passes.


ERCS/erc-7780.md
```diff
@@ -34,6 +34,9 @@ This standard introduces three new module types on top of the existing modules i
 - Policy (type id: 5)
 - Signer (type id: 6)
 - Stateless Validator (type id: 7)
+- PreValidationHookERC1271 (type id: 8)
```
SamWilsn: Hm... This is a bit weird because this isn't exactly "code" in the strictest sense but putting the policy names in backticks (\`) would be the easiest way to get the linter to stop complaining about "ERC1271".

My recommendation here would be to change the name to something like "Pre-Validation Hook (ERC-1271)", unless you need it to be exactly as written currently. If it's the latter case, I'd go with the backticks route (\`PreValidationHookERC1271\`).

PR #1358, opened 2025-11-18
Update ERC-7866: Move to Review

anistark:


SamWilsn (commented):
Should probably replace \`ERC721\` on line 247 with `[ERC-721](./eip-721.md)`, but that's a minor thing.

In the "Comparison with Related Work" section, you're essentially creating links to these external services. ENS has a few ERCs you can link to directly, and the DID standards likely have something you can link to with a [format from EIP-1](https://eips.ethereum.org/EIPS/eip-1#linking-to-external-resources), but Gravatar is a product and we'd prefer if you didn't call them out specifically.

On line 265, put ISoulProfile in backticks (\`). Same with line 310 and SoulProfile.

On 392, replace \`EIP137\` with `[ERC-137](./eip-137.md)`

anistark:
@SamWilsn Thanks for the review. Have made the changes. Please have another look when you can. 🙌🏼

SamWilsn (approved):

PR #1390, opened 2025-12-02
Add ERC-8098: Non-increasing supply ERC-20

ten-io-meta:
This proposal introduces ERC-8098, an ERC-20 behavioral profile with a strictly non-increasing total supply.

A conformant implementation:

- cannot increase `totalSupply()` after deployment,
- permanently removes burned units,
- exposes a single canonical ERC-20 balance and supply state,
- and cannot use upgrades, privileged roles, or internal alternate accounting mechanisms to reintroduce supply-increasing behavior.

ERC-8098 does not introduce a new token interface. It standardizes observable supply invariants on top of ERC-20 compatibility so that auditors, integrators, and applications can reason consistently about irreversible supply semantics.

External wrappers, third-party representations, and application-layer abstractions are explicitly out of scope.

Discussion thread (Ethereum Magicians):
https://ethereum-magicians.org/t/erc-8098-zw-token-irreversible-supply-erc-20/26685


ten-io-meta:
Thanks everyone for the previous guidance — especially @abcoathup.

As suggested, the earlier PR was closed and the proposal has now been correctly submitted to the ERCs repository.

This PR (#1390) contains the cleaned-up ERC draft, aligned with the required format and category, and it now links to the updated discussion thread on Ethereum Magicians:

🔗 https://ethereum-magicians.org/t/erc-9999-zw-token-irreversible-supply-erc-20/26685

I’ll continue refining the draft based on any feedback from reviewers.
Thanks for your time and support.

abcoathup:
@ten-io-meta please note, ERCs are not for marketing or promotional purposes.
Recommend removing any brand specific names and only use generic terms in the ERC. (Note: I am not an editor).

erc-9999.md
```diff
@@ -0,0 +1,105 @@
+---
+erc: 9999
```
abcoathup:
```suggestion
erc: 8098
```

erc-9999.md
```diff
@@ -0,0 +1,105 @@
...
+title: ZW-Token — Irreversible ERC-20
```
abcoathup:
```suggestion
title: Non-increasing supply ERC-20
```
Suggest using a more descriptive title.  ZW token isn't defined.  Note: I am not an editor.

erc-9999.md
```diff
@@ -0,0 +1,105 @@
...
+description: A fungible token with strictly non-increasing supply and a privacy layer that only obfuscates transfers, never holding value.
+author: TEN (@ten-io-meta) <ten.io.metaverse@gmail.com>
+discussions-to: https://ethereum-magicians.org/t/erc-xxxx-zw-token-single-state-irreversible-supply-erc-20/21540
```
abcoathup:
```suggestion
discussions-to: https://ethereum-magicians.org/t/erc-8098-zw-token-irreversible-supply-erc-20/26685
```
Updated with assigned number

ten-io-meta:
Thanks @abcoathup — all suggested changes have now been applied.

The ERC number, filename and discussion link are updated, and the draft follows the required neutral format.

Let me know if anything else is needed. I appreciate the guidance.

ten-io-meta:
Hi all,

Just a quick note to mention that the draft has been updated to address the previous feedback and is now aligned with the required neutral ERC format.

There are no open questions on my side at this point.  
Happy to make any further adjustments if reviewers have additional comments.

Thanks for your time.

erc-8098.md
```diff
@@ -0,0 +1,105 @@
...
+created: 2025-11-23
+requires: ERC-20
+---
+
+## Abstract
+
+This ERC introduces **ZW-Token**, a fungible token standard with **strictly irreversible supply**, **no re-minting path**, a **single canonical state**, no synthetic mirrors, and no alternate supply layers.
+
+The standard enforces a **single canonical supply state**.  
+Optional privacy layers may be added via obfuscation, **not** by creating parallel supply representations.
```
SamWilsn: I'm not sure these terms are widely known enough to use without an explanation. What is a re-minting path? What is a reversible supply or synthetic mirror?

Additionally your abstract should include a high-level (but still technical) overview of _how_ your proposal accomplishes its goals, not just stating those goals.

erc-8098.md
```diff
@@ -0,0 +1,105 @@
...
+   Burned units are permanently removed.
+
+2. **Single-State Supply Representation**  
+   No minting paths after burn events and no synthetic replicas.
+
+3. **No Alternate Mirrors nor Wrapped Layers**  
+   Only one canonical supply state can exist on-chain.
+
+4. **Optional Privacy Without Supply Duplication**  
+   Privacy is achieved only by obfuscation—not by creating hidden or mirrored balances.
```
SamWilsn: Good motivation so far! I think you should include a bit of text explaining why someone might want a token with these properties. For example, what's bad about "reclaimable burns"?

erc-8098.md
```diff
@@ -0,0 +1,105 @@
...
+All balances MUST reference a single canonical ERC-20 state.  
+The token MUST NOT be wrapped or mirrored into any alternate ERC-20 representation that affects supply semantics.
+
+### 4. Optional Privacy Layer
+
+Privacy extensions may be included **only** if they:
+
+- do not alter total supply,  
+- do not create alternate supply trees,  
+- and do not maintain synthetic or private mirrors.
```
SamWilsn: This isn't really a technical standard. It isn't necessary to adhere to these rules for an implementation to be compatible.

For example, a wallet compatible with an ERC-20 token is automatically fully compatible with an ERC-8098 token. There's nothing here that requires changes to the interface.

If you want to make a particular token with these properties, you can, and it'll work with every other part of the DeFi infrastructure perfectly. No ERC required.

erc-8098.md
```diff
@@ -0,0 +1,105 @@
...
+Implementations **must not** provide any function capable of:
+
+- increasing total supply after deployment,
+- restoring burned units,
+- or minting synthetic equivalents.
+
+### 3. Single-State Value Representation
+
+All balances MUST reference a single canonical ERC-20 state.  
+The token MUST NOT be wrapped or mirrored into any alternate ERC-20 representation that affects supply semantics.
```
SamWilsn: This is an infeasible technical requirement (unless you have a solution ready to demonstrate). How would an implementation enforce this?

erc-8098.md
```diff
@@ -0,0 +1,105 @@
...
+## Rationale
+
+Fragmented supply representations introduce discrepancies and auditing failures.  
+A strictly irreversible supply eliminates:
+
+- double representation of burned amounts,  
+- inconsistencies across wrapped layers,  
+- and synthetic minting vulnerabilities.
+
+This design is based on the live on-chain experiment **TEN.IO — Fragment 0 ("El Umbral")**, which demonstrates the feasibility of irreversible supply and unfragmented value representation.
```
SamWilsn: This is an external link (whether or not it's a hyperlink is irrelevant). Please remove it.

erc-8098.md
```diff
@@ -0,0 +1,105 @@
...
+- and do not maintain synthetic or private mirrors.
+
+## Rationale
+
+Fragmented supply representations introduce discrepancies and auditing failures.  
+A strictly irreversible supply eliminates:
+
+- double representation of burned amounts,  
+- inconsistencies across wrapped layers,  
+- and synthetic minting vulnerabilities.
```
SamWilsn: This is justifying (I think) the proposal as a whole. The Rationale section should be used to explain technical choices made within the proposal itself.

erc-8098.md
```diff
@@ -0,0 +1,105 @@
...
+
+- burn logic,  
+- prevention of hidden mint paths,  
+- prevention of synthetic re-issuance via wrappers or mirrors.
+
+Implementers must guarantee that **no upgrade path** or proxy mechanism can introduce new minting functionality.
+
+## Copyright
+
+Copyright and related rights waived via CC0.
```
SamWilsn: Please use the exact waiver from the [eip template](https://github.com/ethereum/EIPs/blob/master/eip-template.md?plain=1).

SamWilsn (changes requested):

SamWilsn:
Woops, sorry, forgot to rename the file for you.

erc-8098.md
ten-io-meta: Thanks for the clarification. I’ve updated the Copyright section to use the exact CC0 waiver text from the EIP template.

erc-8098.md
ten-io-meta: Understood. The external reference has been removed from the document.

erc-8098.md
ten-io-meta: Agreed. I’ve restructured the Rationale section to focus specifically on the technical design choices made in this proposal rather than general motivation.

erc-8098.md
ten-io-meta: Good point. To clarify, this requirement is not intended to mandate enforcement against external wrappers or third-party contracts.

The invariant is defined at the level of the ERC-8098 implementation itself: the token contract MUST expose only a single canonical ERC-20 balance and supply state, and MUST NOT include any internal wrapping, mirroring, or alternate accounting mechanisms.

External systems may of course build on top of the token, but such constructions are explicitly out of scope for this standard.

erc-8098.md
ten-io-meta: The intent of ERC-8098 is not to introduce new interfaces, but to standardize a set of strict invariants around ERC-20 supply behavior.

While individual tokens can of course implement these properties ad hoc, a standard provides a common, reviewable specification that wallets, auditors, and integrators can reference when reasoning about irreversible supply semantics.

This proposal is therefore invariant-focused rather than interface-focused, while remaining fully ERC-20 compatible.

erc-8098.md
ten-io-meta: Reclaimable burns undermine the reliability of total supply as an invariant.

If burned units can later be restored, then supply becomes dependent on off-chain governance decisions, privileged roles, or future contract upgrades. This complicates auditing, weakens economic guarantees, and makes it impossible to reason about supply monotonicity from on-chain state alone.

ERC-8098 targets use cases where supply finality itself is a core property.

erc-8098.md
ten-io-meta: Thanks for the feedback. I’ve updated the Abstract to briefly define the terms “re-minting path” and “synthetic mirror”, and to include a high-level technical overview of how the proposal achieves its goals.

Specifically, the Abstract now explains that irreversibility is enforced by permanently reducing totalSupply via burns, by omitting any minting functionality from the contract interface and logic, and by constraining the implementation to a single canonical ERC-20 balance and supply state.

ten-io-meta:
Hi @SamWilsn,

Thanks again for your earlier feedback.

All requested changes have now been addressed:
- clarified terminology in the abstract
- added high-level technical explanation
- refined motivation and rationale sections
- clarified scope of internal vs external representations
- aligned wording with EIP editorial standards

Could you please take another look when convenient?

Thank you for your time.

ten-io-meta:
Hi @SamWilsn,

Thank you again for the previous review and feedback.

After revisiting the proposal, I believe the central question is not whether ERC-8098 introduces a new ERC-20 interface.

It does not.

The question is whether irreversibility itself is a property that external systems may safely rely upon.

ERC-8098 attempts to standardize a behavioral invariant:

> Burned supply can never be restored.

The value of this invariant is not primarily for wallets or end users. It is for third-party systems that must determine whether supply reduction is reversible or irreversible.

Without a standard, every integration must independently analyze token logic, governance controls, upgrade paths, and mint authorities to determine whether a burn is truly final.

With a standard, external systems can treat irreversibility as a declared and verifiable property rather than a token-by-token audit exercise.

In that sense, the proposal is less about introducing new ERC-20 methods and more about standardizing a property that external systems may choose to rely upon.

The key question becomes:

> Is irreversibility a sufficiently useful interoperable property to justify standardization?

I believe that is the core claim ERC-8098 is attempting to make, and I would appreciate your thoughts on that specific point.

Thank you for your time.

hilljessica8787-dot (approved):

ten-io-meta:
Hi @SamWilsn,

I revisited ERC-8098 again with your earlier review and, in particular, your concern about whether a proposal that introduces no new interface belongs as an ERC.

The current draft has now been narrowed further:

the prescribed burn function signature has been removed,
burn requirements are defined purely as behavioral invariants,
the optional privacy layer has been removed,
the motivation now focuses on supply guarantees that ERC-20 itself does not specify,
external wrappers and third-party representations are explicitly out of scope,
and upgradeability is constrained so that an implementation capable of introducing future supply increases is not conformant.

All of the specific points from your previous review have been addressed in the current draft.

When convenient, could you please take another look? I would especially appreciate your view on whether this narrower framing — a standardized, externally reviewable ERC-20 supply invariant rather than a new interface — addresses your earlier concern.

Thank you again for your time and feedback.

PR #1403, opened 2025-12-14
Update ERC-7943: erc-7943.md - nitpicking edit of RFC2119 terms

chaals:
- change CAN (not an RFC2119 term) to MAY (which is)
- edit to avoid using "MAY NOT" which isn't an RFC2119 term (because its ordinary english meaning doesn't obviously match what it would mean in a spec)
- trim to the most common and easily interpreted terms
- edit text to remove any usage that isn't intended to be a conformance statement, removing the possibility of confusions that RFC 8174 addresses


ERCS/erc-7943.md
```diff
@@ -297,15 +297,15 @@ Implementations MUST ensure their transfer methods exhibit the following behavio
...
-- **Burning** in permissionless contexts (e.g., public `burn` functions) MUST respect `canTransfer` check, MUST respect the `canTransact` check on the token holder and MUST NOT allow burning more assets than the unfrozen amount. In permissioned contexts (e.g., authorized burning by privileged roles), burning MAY NOT be restricted by `canTransfer` or `canTransact` checks on the token holder and MAY burn more assets than the unfrozen amount, in which case the contract MUST update the frozen status accordingly and emit a `Frozen` event before the underlying base token transfer event.
+- **Burning** in permissionless contexts (e.g., public `burn` functions) MUST respect `canTransfer` check, MUST respect the `canTransact` check on the token holder and MUST NOT allow burning more assets than the unfrozen amount. In permissioned contexts (e.g., authorized burning by privileged roles), burning MAY succeed for accounts where `canTransact` on the recipient would return `false`, and MAY burn more assets than the unfrozen amount, in which case the contract MUST update the frozen status accordingly and emit a `Frozen` event before the underlying base token transfer event.
```
xaler5: I guess you wanted to say "token holder" instead of "recipient" here ?

chaals: yep
```suggestion
- **Burning** in permissionless contexts (e.g., public `burn` functions) MUST respect `canTransfer` check, MUST respect the `canTransact` check on the token holder and MUST NOT allow burning more assets than the unfrozen amount. In permissioned contexts (e.g., authorized burning by privileged roles), burning MAY succeed for accounts where `canTransact` on the token holder would return `false`, and MAY burn more assets than the unfrozen amount, in which case the contract MUST update the frozen status accordingly and emit a `Frozen` event before the underlying base token transfer event.
```

PR #1423, opened 2025-12-23
Update ERC-7943: Move to Last Call

xaler5:

 - The build passes.


xinbenlv (commented):
Thank you for your work on this proposal. The uRWA interface provides a valuable standardization for compliance-heavy assets. As this PR moves the EIP to **Last Call**, I have reviewed it for readiness, clarity, and completeness. There are a few clarifications needed regarding the `forcedTransfer` behavior and some minor editorial fixes before it enters the final review period.

The proposal is in good shape. Please address the clarification regarding `forcedTransfer` unfreezing logic and the return value behavior. Additionally, please confirm on the discussion thread that open suggestions have been considered. Once these points are resolved, the PR is ready for Last Call. **Question for Last Call:** Do you have multiple independent adopters or implementations planned?

---

**ERCS/erc-7943.md:79**

Please clarify the return behavior for `forcedTransfer` and `setFrozenTokens` (and their counterparts in other interfaces). Do they return `false` on failure, or do they `revert`?

Standard practice is to revert on failure (especially for access control or logic errors). If they revert, the boolean return is typically always `true` on success.

```suggestion
    /// @return result True if the transfer executed correctly. MUST revert on failure.
    function forcedTransfer(address from, address to, uint256 amount) external returns(bool result);
```

(https://github.com/xaler5/ERCs/blob/5b7f10abde80352a96badf0172e7ebdd9dfc8e57/ERCS/erc-7943.md?plain=1#L79)

---

**ERCS/erc-7943.md:263**

Typo correction.

```suggestion
    ///     - MUST return false in general, if any permissioned rule would prevent a given transfer from succeeding. A transfer refers to any operation that emits the token’s canonical transfer event. A permissioned check can be a pausing mechanism, a call to `canTransact` or anything else that requires privileged actors.
```

(https://github.com/xaler5/ERCs/blob/5b7f10abde80352a96badf0172e7ebdd9dfc8e57/ERCS/erc-7943.md?plain=1#L263)

---

**ERCS/erc-7943.md:275**

The requirement to "unfreeze the assets first" is slightly ambiguous if the assets weren't frozen to begin with (e.g., if `canTransfer` failed due to a pause or blocklist, not a freeze).

If the intention is to ensure the frozen balance is updated *if* frozen tokens are moved, consider clarifying:

> If this happens, and the transfer involves tokens that are currently counted as frozen, it MUST update the frozen status (emitting a `Frozen` event) to reflect the reduction in frozen balance before the transfer.

If the intention is to *always* emit a `Frozen` event even if the frozen amount was 0, that seems inefficient.

(https://github.com/xaler5/ERCs/blob/5b7f10abde80352a96badf0172e7ebdd9dfc8e57/ERCS/erc-7943.md?plain=1#L275)

---

**ERCS/erc-7943.md:290**

Please verify that these interface IDs are calculated correctly based on the final function signatures in this specification. Any changes to function definitions (e.g. return types) will change these IDs.

(https://github.com/xaler5/ERCs/blob/5b7f10abde80352a96badf0172e7ebdd9dfc8e57/ERCS/erc-7943.md?plain=1#L290)

---

**ERCS/erc-7943.md:295**

Typo correction.

```suggestion
Implementations of these interfaces MUST implement the necessary functions of their chosen base standard (e.g., [ERC-20](./eip-20.md) for the fungible interface, [ERC-721](./eip-721.md) for the non-fungible interface and [ERC-1155](./eip-1155.md) for the multi token interface) and MUST also restrict access to sensitive functions like `forcedTransfer` and `setFrozenTokens` using an appropriate access control mechanism (e.g., `onlyOwner`, Role-Based Access Control). The specific mechanism is NOT mandated by this interface standard.
```

(https://github.com/xaler5/ERCs/blob/5b7f10abde80352a96badf0172e7ebdd9dfc8e57/ERCS/erc-7943.md?plain=1#L295)

xaler5:
Thank you for this review @xinbenlv !

I've tried to address everything in my last comment. Let me know if those are good to you :) 

> Question for Last Call: Do you have multiple independent adopters or implementations planned?

Yes, we are a group of several companies trying to push for standardization, some public names are
Brickken, Bit2Me, DigiShares, Dekalabs, Hacken, Compellio, RealEstate.Exchange, Stobox, FullyTokenized, Forte Protocol, Zoth as you can easily google from latest PR.

Current public implementation is on OpenZeppelin community contracts (which needs to be updated with latest state of the EIP): https://github.com/OpenZeppelin/openzeppelin-community-contracts/blob/feature/erc-7943/contracts/token/ERC20/extensions/ERC20uRWA.sol

WIP implementations are scheduled for many of the contributing companies and are expected to be around during this year. I'm personally involved in one of them.

This EIP was also originally and spontaneously taken at an hackaton https://ethglobal.com/showcase/rwalink-4nyfv

---

On the return value of `setFrozenTokens` / `forcedTransfer` I made it clear that it should return true or revert. I wish to maintain the boolean return value so that integrators can decide to explicitly check for a true return instead of relying on a non-reverting call.

ERCS/erc-7943.md
```diff
@@ -2,9 +2,10 @@
 eip: 7943
 title: uRWA - Universal Real World Asset Interface
 description: Interfaces for common base tokens defining compliance checks, transfer controls, and enforcement actions for Real World Assets (RWAs).
-author: Dario Lo Buglio (@xaler5)
+author: Dario Lo Buglio (@xaler5), Tino Martinez Molina (@tinom9), Mihai Colceriu (@mihaic195)
 discussions-to: https://ethereum-magicians.org/t/erc-universal-rwa-interface/23972
-status: Review
+status: Last Call
```
xinbenlv, line 7: Can you keep updating Status into a separate PR.

.gitmodules
```diff
@@ -0,0 +1,6 @@
+[submodule "assets/erc-7943/lib/openzeppelin-contracts"]
```
xinbenlv: Defer to EIP Editors together to determine whether we will accept git submodules.

xaler5: They just slipped through. Was not intentional to add. Sorry about that 😊

Issue #1420, opened 2025-12-23
ERC Standard can't reference itself without linking to itself


mudgen:
### Pull Request

https://github.com/ethereum/ERCs/pull/1417/changes#diff-48702441bc6823f0cc6eb82e6c7097e7574e8bca77a651704351ef42c4272a1fR351-R352

### What happened?

My standard mentions it self, but does not link to itself, so gets this error:

I don't think an ERC should have to link to itself in order to mention itself.

<img width="840" height="276" alt="Image" src="https://github.com/user-attachments/assets/be82dcf0-bc3f-44b5-a40d-24ffdac462da" />

### Relevant log output

```shell

```

vamshi567bh:
@mudgen can you please assign this to me

PR #1442, opened 2026-01-01
Update ERC-8109: Fix self-references to ERC-8109 with canonical links

codebyankita:
This PR updates ERC-8109 to link self-references to the canonical specification file, resolving repository linting errors without changing the specification semantics.

Related to #1420.

### DIFF
--- a/ERCS/erc-8109.md
+++ b/ERCS/erc-8109.md
@@ -268,62 +268,46 @@ The `upgradeDiamond` function MUST adhere to the following implementation requir
 > The complete definitions of the events and custom errors referenced below are given earlier in this document.
 
 1. **Function Inputs**
-   
    - `_addFunctions` maps new selectors to their facet implementations.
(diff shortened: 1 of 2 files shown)


mudgen:
@codebyankita
Thanks for wanting to improve the standard.

> This PR updates ERC-8109 to link self-references to the canonical specification file, resolving repository linting errors without changing the specification semantics.

The standard doesn't have any linting errors to fix. I have no need to make the changes in this PR.
If you want to fix something then I suggest that you submit a PR that removes the requirement that a standard must link to itself in order to reference itself.

codebyankita:
Thanks mudgen, that clarifies it.

I understand that the issue is with the linting requirement itself rather than ERC-8109. I’ll close this PR and instead look into proposing a change to the lint rule so standards can reference themselves without requiring a canonical self-link.

PR #1446, opened 2026-01-02
Update ERC-7828: Move to Draft

clowestab:
This serves as an initial iterative pull request for updates to ERC-7828 in relation to Separation of Concerns (SoC) with ERC-7930.

This PR only updates the Abstract/Motivation sections.

Based on extensive discussions with the team at Wonderland, Josh Rudolf (EF) and the broader interop community.

### DIFF
--- a/ERCS/erc-7828.md
+++ b/ERCS/erc-7828.md
@@ -234,6 +234,11 @@ Copyright and related rights waived via [CC0](../LICENSE.md).
 
 [ERC-7930]: ./eip-7930.md
 [CAIP 2]: https://github.com/ChainAgnostic/CAIPs/blob/36ef9ee906da090366cb679634a0850e91785db6/CAIPs/caip-2.md
+[UTS #46]: https://www.unicode.org/reports/tr46/tr46-35.html
+[ENSIP-1]: https://docs.ens.domains/ensip/1/
+[ENSIP-9]: https://docs.ens.domains/ensip/9/
+[ENSIP-11]: https://docs.ens.domains/ensip/11/
+[ENSIP-19]: https://docs.ens.domains/ensip/19/
 [BIP 44]: https://github.com/bitcoin/bips/blob/1d371a58978fd2f313c9d162762de04d3b697bf5/bip-0044.mediawiki
 [EIP-155]: ./eip-155.md
 [ERC-7930]: ./eip-7930.md


ERCS/erc-7828.md
```diff
@@ -1,45 +1,36 @@
...
+description: Defines a human readable chain specific address format that allows for chain discovery using ENS.
+author: Sam Kaufman (@SampkaML), Marco Stronati (@paracetamolo), Yuliya Alexiev (@yuliyaalexiev), Jeff Lau (@jefflau), Sam Wilson (@samwilsn), Vitalik Buterin (@vbuterin), Teddy (@0xteddybear), Joxes (@Joxess), Racu (@0xRacoon), Skeletor Spaceman (@0xskeletor-spaceman), TiTi (@0xtiti), Gori (@0xGorilla), Ardy (@0xArdy), Onizuka (@onizuka-wl), Lumi (@oxlumi), Moebius (@0xmoebius), Thomas Clowes (@clowestab), Prem Makeig (@nxt3d)
 discussions-to: https://ethereum-magicians.org/t/erc-7828-chain-specific-addresses-using-ens/21930
-status: Review
+status: Draft
 type: Standards Track
 category: ERC
 created: 2024-11-27
-requires: 155, 7930
+requires: 155
```
0xrcinus: I think 7930 is still required as the checksum approach still depends on it?

ERCS/erc-7828.md
```diff
@@ -1,45 +1,36 @@
...
-- The use of human-readable chain names, with resolution to chain identifiers via ENS.
-- The use of human-readable account names, with resolution to addresses via ENS.
-- An on-chain registry mapping chain names to identifiers, enabling decentralized resolution of chain metadata.
-- The ENS domain suffix used for chain names SHALL be abstracted from users for readability, ensuring a simpler display format.
+The target chain component of an _Interoperable Name_ may be expressed as a human-readable label. This specification defines how such labels can be resolved, via the Ethereum Name Service (ENS), to chain-specific metadata.
 
 ## Motivation
 
-The current Ethereum address landscape is leading to an ecosystem that will have hundreds and eventually thousands of L2s that use the same address format as Ethereum mainnet. This means an address by itself is not enough information to know which chain the address is related to. This can be problematic if funds are sent to an unreachable address on the incorrect chain. From the user account it should be possible to obtain the right chain identifier (chainID) to include in a transaction.
```
0xrcinus, line 26: We seem to be losing some of the motivation for chain-specific addresses, which I think is worth keeping (i.e. specifying an address and a chain in one copy-pastable thing, as well as the risks if the wrong chain is used)

ERCS/erc-7828.md
```diff
@@ -1,45 +1,36 @@
...
-- It does not scale with the growing number of L2s.
-- The list maintainer is a trusted centralized entity.
-- It does not (currently) support non-EVM chains, even when naming systems (such as ENS, since [ENSIP-9]) do.
+ENS is a well-established blockchain identity primitive that is significantly more consumable than the alphanumeric address formats used by most blockchains.
 
-Instead of using non-human-readable numeric chain identifiers, this specification SHALL require a human-readable chain name resolved on-chain via ENS wildcard resolver. The centralized chain list maintained in `ethereum-lists/chains` SHALL be superseded by an on-chain registry. This registry provides a single source of truth for mapping chain names to chain identifiers and enables decentralized, extensible chain metadata resolution.
+An on-chain registry provides a single source of truth for mapping a chain label to metadata associated with that chain. This specification defines an architecture that enables chain operators to take ownership of their chain-specific data, thereby reducing reliance on a single centralized entity and ensuring data integrity.
```
0xrcinus: This also misses out on why a registry of any kind is required:
- Chains are currently identified by `chainId`, which are not end-user friendly
- Users want to refer to chains by their readable name (e.g. `base`, not `8543`)
- We need a way to map readable names to chain IDs

ERCS/erc-7828.md
```diff
@@ -1,45 +1,36 @@
...
 
-The current Ethereum address landscape is leading to an ecosystem that will have hundreds and eventually thousands of L2s that use the same address format as Ethereum mainnet. This means an address by itself is not enough information to know which chain the address is related to. This can be problematic if funds are sent to an unreachable address on the incorrect chain. From the user account it should be possible to obtain the right chain identifier (chainID) to include in a transaction.
+The core motivation for introducing the _Interoperable Name_ standard is to provide maximally readable, _chain-specific addresses_ for user-facing interactions.
 
-The mapping from chain names to identifiers has, since [EIP-155], been maintained off chain using a centralized list. This solution has a few shortcomings:
-- It does not scale with the growing number of L2s.
-- The list maintainer is a trusted centralized entity.
-- It does not (currently) support non-EVM chains, even when naming systems (such as ENS, since [ENSIP-9]) do.
+ENS is a well-established blockchain identity primitive that is significantly more consumable than the alphanumeric address formats used by most blockchains.
```
0xrcinus: It's unclear whether this is referring to ENS as the address resolver, or ENS as the foundation for the chain registry (maybe both?)

ERCS/erc-7828.md
```diff
@@ -1,45 +1,36 @@
...
-- An on-chain registry mapping chain names to identifiers, enabling decentralized resolution of chain metadata.
-- The ENS domain suffix used for chain names SHALL be abstracted from users for readability, ensuring a simpler display format.
+The target chain component of an _Interoperable Name_ may be expressed as a human-readable label. This specification defines how such labels can be resolved, via the Ethereum Name Service (ENS), to chain-specific metadata.
 
 ## Motivation
 
-The current Ethereum address landscape is leading to an ecosystem that will have hundreds and eventually thousands of L2s that use the same address format as Ethereum mainnet. This means an address by itself is not enough information to know which chain the address is related to. This can be problematic if funds are sent to an unreachable address on the incorrect chain. From the user account it should be possible to obtain the right chain identifier (chainID) to include in a transaction.
+The core motivation for introducing the _Interoperable Name_ standard is to provide maximally readable, _chain-specific addresses_ for user-facing interactions.
```
0xrcinus, line 24: I think this is missing the additional problem that we are trying to solve, i.e. that even if you have a chain-specific address in the `0x123..@eip155:8543` format, that is still not very readable. So we want to define that foundational format, with maximally human-readable options for both addresses (ENS) and chains (chain registry). The foundational format is necessary, so that there isn't a dependency on ENS or a chain registry, but the readable options are necessary, for general useability and adoption

0xrcinus (changes requested):
Thanks! I think we can keep a bit more detail in the motivation

ERCS/erc-7828.md
clowestab: My logic is that EIP-1 states "_A requires dependency is created when the current EIP cannot be understood or implemented without a concept or technical element from another EIP. Merely mentioning another EIP does not necessarily create such a dependency._". Because of the SoC (see this comment: https://github.com/ethereum/ERCs/pull/1447#discussion_r2664479504), this spec doesn't really _require_ 7930, its just another piece of data you can get from the registry-resolver.

ERCS/erc-7828.md
clowestab: Good point. Updated.

ERCS/erc-7828.md
clowestab: I have removed this for now, and noted it to add back in the 'Rationale' section instead (with additional clarity).

ERCS/erc-7828.md
clowestab: Added further clarity.

ERCS/erc-7828.md
clowestab: Updated with additional clarity.

I think the "The core motivation for introducing the _Interoperable Name_ standard is to provide maximally readable, _chain-specific addresses_ for user-facing interactions." bit highlights the core concern using the previously defined terminology?

ERCS/erc-7828.md
```diff
@@ -13,33 +13,25 @@ requires: 155, 7930
 
 ## Abstract
 
-This proposal extends [ERC-7930] \(Interoperable Addresses & Names\) by standardizing a human-readable format for chain-specific addresses in the form `<address>@<chain>#<checksum>`. It introduces:
+This proposal extends the definition of an _Interoperable Name_ outlined in [ERC-7930](./eip-7930.md), and defined as `<address>@<chain>#<checksum>`.
```
0xMonoAx: Since Interoperable Names were removed from the scope of ERC-7930, this section should clarify that the proposal extends ERC-7930 (Interoperable Addresses) by introducing a human-readable format for chain-specific addresses. This format, referred to as an Interoperable Name, represents canonical chain addresses in the following form: `<address>@<chain>#<checksum>`

ERCS/erc-7828.md
clowestab: Reverted this back to complete the split of 7930 and 7828 maintaining the content. Once fully executed we can discuss further.

ERCS/erc-7828.md
```diff
@@ -1,10 +1,10 @@
 ---
 eip: 7828
-title: Readable Interoperable Addresses using ENS
-description: An iteration of Interoperable Address' format that allows resolution to shorter, hierarchical strings using naming registries
-author: Sam Kaufman (@SampkaML), Marco Stronati (@paracetamolo), Yuliya Alexiev (@yuliyaalexiev), Jeff Lau (@jefflau), Sam Wilson (@samwilsn), Vitalik Buterin (@vbuterin), Teddy (@0xteddybear), Joxes (@Joxess), Racu (@0xRacoon), Skeletor Spaceman (@0xskeletor-spaceman), TiTi (@0xtiti), Gori (@0xGorilla), Ardy (@0xArdy), Onizuka (@onizuka-wl), Lumi (@oxlumi), Moebius (@0xmoebius)
+title: Interoperable Names
+description: Defines a human readable chain specific address format that allows for chain discovery using ENS.
```
0xrcinus: nit: "chain-specific"

ERCS/erc-7828.md
```diff
@@ -13,33 +13,28 @@ requires: 155, 7930
 
 ## Abstract
 
-This proposal extends [ERC-7930] \(Interoperable Addresses & Names\) by standardizing a human-readable format for chain-specific addresses in the form `<address>@<chain>#<checksum>`. It introduces:
+This proposal defines an _Interoperable Name_, a human-readable _chain-specific address_ format.
 
-- A unified format for accounts that specifies, together with the address, the chain where the address lives.
-- The use of human-readable chain names, with resolution to chain identifiers via ENS.
-- The use of human-readable account names, with resolution to addresses via ENS.
```
0xrcinus, line 20: Is there a reason we don't mention usage of ENS for the account as well as the chain?

ERCS/erc-7828.md
```diff
@@ -65,8 +60,8 @@ Where:
 
 - `<raw-chain>`, `<raw-address>` and `<checksum>` are defined to maintain backwards compatibility with [ERC-7930], and their semantics remain the same.
 - `<chain-name>`: A human-readable chain label within the reserved `<namespace>.eth` domain. This standard uses `<namespace>.eth` to provide a consistent abstraction for chain identity, mapping these labels to their corresponding canonical chain identifiers.
-- `<ens-name>`: An ENS name, which should be resolved following the appropiate ENSIPs (which this standard does not aim to overwrite). It's grammar definition is slightly different than that of [ENSIP-1], as this standard does not support names consisting of only a TLD, since they'd be of limited usefulness.
-- `<label>`: Any valid string label per [UTS #46], as defined in [ENSIP-1].
+- `<ens-name>`: An ENS name, which should be resolved following the appropiate ENSIPs (which this standard does not aim to overwrite). It's grammar definition is slightly different than that of ENSIP-1, as this standard does not support names consisting of only a TLD, since they'd be of limited usefulness.
```
0xrcinus: nit: "Its grammar definition"

ERCS/erc-7828.md
```diff
@@ -176,11 +171,11 @@ This section is non-normative, as future ENSIPs can modify the way in which ENS
 
 1. Let the user input an address name. Assume it's `alice.eth`.
 2. Let the user select a chain. Assume it's `optimism`. The `l2.eth` suffix is implicit and hidden from end users.
-3. Append `l2.eth` and compute the ENS namehash as per [ENSIP-1]: `namehash("optimism.l2.eth")`.
+3. Append `l2.eth` and compute the ENS namehash as per ENSIP-1: `namehash("optimism.l2.eth")`.
```
0xrcinus, line 176: This section seems to just be removing the [ENSIP] links correct? (i.e. it is expected that the surrounding content is still out-of-date)

ERCS/erc-7828.md
```diff
@@ -13,33 +13,28 @@ requires: 155, 7930
...
 
-Moreover, the above can be leveraged to allow for a name to represent the same entity corresponding to different addresses on different chains, mitigating the risk of sending funds to an address the intended recipient doesn't actually control.
+This specification defines an architecture that enables chain operators to take ownership of their chain-specific data, thereby reducing reliance on a single centralized entity and ensuring data integrity.
 
-Desired properties:
-- Backwards compatibility with [ERC-7930].
-- The address portion can be either the appropriate type of address for the chain, or a domain name.
-- The address portion and the chain portion should be resolved separately.
+Historically, the mapping from chain names to identifiers has, since [EIP-155], been maintained off-chain using a centralized list.
```
0xrcinus, line 32: I might move this above the "This specification defines an architecture..." so it is more Motivation->Spec

ERCS/erc-7828.md
```diff
@@ -13,33 +13,28 @@ requires: 155, 7930
 
 ## Abstract
 
-This proposal extends [ERC-7930] \(Interoperable Addresses & Names\) by standardizing a human-readable format for chain-specific addresses in the form `<address>@<chain>#<checksum>`. It introduces:
+This proposal defines an _Interoperable Name_, a human-readable _chain-specific address_ format.
```
0xrcinus: Is there a reason not to have some `<address>@<chain>` example to make it immediately tangible? from EIP-1 "This should be a very terse and human-readable version of the specification section. Someone should be able to read only the abstract to get the gist of what this specification does."

ERCS/erc-7828.md
```diff
@@ -1,10 +1,10 @@
 ---
 eip: 7828
-title: Readable Interoperable Addresses using ENS
-description: An iteration of Interoperable Address' format that allows resolution to shorter, hierarchical strings using naming registries
-author: Sam Kaufman (@SampkaML), Marco Stronati (@paracetamolo), Yuliya Alexiev (@yuliyaalexiev), Jeff Lau (@jefflau), Sam Wilson (@samwilsn), Vitalik Buterin (@vbuterin), Teddy (@0xteddybear), Joxes (@Joxess), Racu (@0xRacoon), Skeletor Spaceman (@0xskeletor-spaceman), TiTi (@0xtiti), Gori (@0xGorilla), Ardy (@0xArdy), Onizuka (@onizuka-wl), Lumi (@oxlumi), Moebius (@0xmoebius)
+title: Interoperable Names
```
0xMonoAx, line 3: Didn’t we agree that it would be Interoperable Names using ENS, leaving the door open for other Interoperable Names using XYZ in the future?

clowestab: I'm not super precious about the name, but noting 7930 being 'Interoperable Addresses' and that you can still have `0x123@eip:1` I.E. no ENS, it felt best if ENS wasn't explicitly in the title..

ERCS/erc-7828.md
clowestab: Fixed

ERCS/erc-7828.md
clowestab: Added in response to previous comment

ERCS/erc-7828.md
clowestab: :) Thanks. Fixed.

ERCS/erc-7828.md
clowestab: Yes, sorry. I think i noted this somewhere. This is just so the auto validators for merging pass and then we will come back to this stuff in the PRs for those sections.

ERCS/erc-7828.md
clowestab: Done

ERCS/erc-7828.md
clowestab: Fair point. Added.

PR #1447, opened 2026-01-02
Update ERC-7828: Updated Specification section

clowestab:
This serves as the **second** iterative pull request for updates to ERC-7828 in relation to Separation of Concerns (SoC) with ERC-7930.

This PR only updates the Specification section. A further PR will clarify the resolution of chain identifiers using ENS.

Based on extensive discussions with the team at Wonderland, Josh Rudolf (EF) and the broader interop community.

### DIFF
--- a/ERCS/erc-7828.md
+++ b/ERCS/erc-7828.md
@@ -1,10 +1,10 @@
 ---
 eip: 7828
-title: Readable Interoperable Addresses using ENS
-description: An iteration of Interoperable Address' format that allows resolution to shorter, hierarchical strings using naming registries
-author: Sam Kaufman (@SampkaML), Marco Stronati (@paracetamolo), Yuliya Alexiev (@yuliyaalexiev), Jeff Lau (@jefflau), Sam Wilson (@samwilsn), Vitalik Buterin (@vbuterin), Teddy (@0xteddybear), Joxes (@Joxess), Racu (@0xRacoon), Skeletor Spaceman (@0xskeletor-spaceman), TiTi (@0xtiti), Gori (@0xGorilla), Ardy (@0xArdy), Onizuka (@onizuka-wl), Lumi (@oxlumi), Moebius (@0xmoebius)
+title: Interoperable Names
+description: Defines a human readable chain-specific address format that allows for chain discovery using ENS.
+author: Sam Kaufman (@SampkaML), Marco Stronati (@paracetamolo), Yuliya Alexiev (@yuliyaalexiev), Jeff Lau (@jefflau), Sam Wilson (@samwilsn), Vitalik Buterin (@vbuterin), Teddy (@0xteddybear), Joxes (@Joxess), Racu (@0xRacoon), Skeletor Spaceman (@0xskeletor-spaceman), TiTi (@0xtiti), Gori (@0xGorilla), Ardy (@0xArdy), Onizuka (@onizuka-wl), Lumi (@oxlumi), Moebius (@0xmoebius), Thomas Clowes (@clowestab), Prem Makeig (@nxt3d), Mono (@0xMonoAx), Orca (@0xrcinus)
 discussions-to: https://ethereum-magicians.org/t/erc-7828-chain-specific-addresses-using-ens/21930
-status: Review
+status: Draft
 type: Standards Track
 category: ERC
 created: 2024-11-27
@@ -13,33 +13,30 @@ requires: 155, 7930
 
 ## Abstract
 
-This proposal extends [ERC-7930] \(Interoperable Addresses & Names\) by standardizing a human-readable format for chain-specific addresses in the form `<address>@<chain>#<checksum>`. It introduces:
+This proposal defines an _Interoperable Name_, a human-readable _chain-specific address_ format that MAY represent the _target address_ as an ENS name.
 
-- A unified format for accounts that specifies, together with the address, the chain where the address lives.
-- The use of human-readable chain names, with resolution to chain identifiers via ENS.
-- The use of human-readable account names, with resolution to addresses via ENS.
-- An on-chain registry mapping chain names to identifiers, enabling decentralized resolution of chain metadata.
-- The ENS domain suffix used for chain names SHALL be abstracted from users for readability, ensuring a simpler display format.
+Additionally, the target chain component of an _Interoperable Name_ may be expressed as a human-readable label. This specification defines how such labels can be resolved, via the Ethereum Name Service (ENS), to chain-specific metadata.
+
+The result is _chain-specific addresses_ such as `wallet.ensdao.eth@eip155:1`, `0xFe89cc7aBB2C4183683ab71653C4cdc9B02D44b7@eip155:1` or the maximally readable `example.eth@base`.
 
 ## Motivation
 
-The current Ethereum address landscape is leading to an ecosystem that will have hundreds and eventually thousands of L2s that use the same address format as Ethereum mainnet. This means an address by itself is not enough information to know which chain the address is related to. This can be problematic if funds are sent to an unreachable address on the incorrect chain. From the user account it should be possible to obtain the right chain identifier (chainID) to include in a transaction.
+The core motivation for introducing the _Interoperable Name_ standard is to provide maximally readable, _chain-specific addresses_ for user-facing interactions.
 
-The mapping from chain names to identifiers has, since [EIP-155], been maintained off chain using a centralized list. This solution has a few shortcomings:
-- It does not scale with the growing number of L2s.
-- The list maintainer is a trusted centralized entity.
-- It does not (currently) support non-EVM chains, even when naming systems (such as ENS, since ENSIP-9) do.
+[ERC-7930] introduced a binary format for representing a target address on a specific blockchain. While this binary data is well suited for low-level usage (e.g. in smart contracts), its meaning is semantically opaque to human users.
 
-Instead of using non-human-readable numeric chain identifiers, this specification SHALL require a human-readable chain name resolved on-chain via ENS wildcard resolver. The centralized chain list maintained in `ethereum-lists/chains` SHALL be superseded by an on-chain registry. This registry provides a single source of truth for mapping chain names to chain identifiers and enables decentralized, extensible chain metadata resolution.
+A foundational text representation (e.g. `0xFe89cc7aBB2C4183683ab71653C4cdc9B02D44b7@eip155:1`) enables a significantly more readable _chain-specific address_ format. An expanded form that leverages the name resolution capabilities of the Ethereum Name Service (ENS) further improves readability, allowing addresses to be expressed in a maximally human-friendly form (e.g. `wallet.ensdao.eth@ethereum`).
 
-In the same spirit, the address could be a human-readable name as well, which is already a use case for ENS. By coupling the TLD to the resolving method used, this standard could leverage current and future features of ENS as well as other naming registries.
+An on-chain registry provides a canonical source of truth for mapping human-readable chain labels to the metadata associated with each chain. Today, chains are identified using a variety of specifications (e.g. [CAIP 2] and [EIP-155] for EVM-compatible networks), whose formats are not necessarily semantically clear to human users. By introducing an on-chain registry, applications can refer to a chain using a readable identifier such as `base`, rather than an opaque identifier like `eip155:8543` or `8543`.
 
-Moreover, the above can be leveraged to allow for a name to represent the same entity corresponding to different addresses on different chains, mitigating the risk of sending funds to an address the intended recipient doesn't actually control.
+Historically, the mapping from chain names to identifiers has, since [EIP-155], been maintained off-chain using a centralized list.
 
-Desired properties:
-- Backwards compatibility with [ERC-7930].
-- The address portion can be either the appropriate type of address for the chain, or a domain name.
-- The address portion and the chain portion should be resolved separately.
+This approach has several shortcomings:
+- It does not scale with the growing number of blockchains
+- It relies on a trusted centralized maintainer
+- It does not support non-EVM chains
+
+This specification defines an architecture that enables chain operators to take ownership of their chain-specific data, thereby reducing reliance on a single centralized entity and ensuring data integrity.
 
 ## Specification
 
@@ -264,7 +261,7 @@ Starting from the Interoperable Address serialized above: `0x00010000010A14AAAAA
 - Wallet developers should be aware of possible unicode glyph collisions in resolved names and warn users about them in order to keep checksums effective, since an attacker could, in order to impersonate `alice.eth@chain.<<namespace>.eth#00112233`:
     - Mine an address on `chain.<namespace>.eth` such that the checksum of the Interoperable Address is `00112233`.
     - Register `аlice.eth` (using the russian vowel `а` instead of the latin `a`), and point it to the address above.
-- [ENSIP-19] introduces trust assumptions when resolving reverse records offchain:
+- ENSIP-19 introduces trust assumptions when resolving reverse records offchain:
     - ZK-rollups may offer trust-minimized resolution via on-chain proofs.
     - Sidechains may rely on centralized CCIP-read gateways or signature authorities.
 
--- a/ERCS/erc-7841.md
+++ b/ERCS/erc-7841.md
@@ -1,0 +1,550 @@
+---
+eip: 7841
+title: Cross-chain Message Format and Mailbox
+description: A mailbox API and message format to provide a unified cross-chain developer experience
+author: Ellie Davidson (@elliedavidson), Alex Xiong (@alxiong), Philippe Camacho (@philippecamacho), and Ben Fisch (@benafisch)
+discussions-to: https://ethereum-magicians.org/t/erc-7841-cross-chain-message-format-and-mailbox/22185
+status: Draft
+type: Standards Track
+category: ERC
+created: 2024-12-12
+---
+
+## Abstract
+
+This ERC proposes a **mailbox API** and **message format** for sending and receiving data between L2s. This standard makes it easier for developers to build cross-chain applications that work over a variety of VMs, chain settlement mechanisms (e.g., ZK or optimistic settlement), and messaging protocols (e.g., synchronous or asynchronous protocols). This ERC accomplishes this by 1.) defining a mailbox interface through which cross-chain messages can be sent and received independent of their payload; 2.) defining a VM-agnostic message format and address type to make the interface compatible with many VMs;  3.) keeping the mailbox interface minimal to make it compatible with many kinds of cross-chain communication.  Example applications include an intent-based bridge, a liquidity-unifying DEX operating across multiple chains, or a cross-chain lending application.
+
+## Motivation
+
+L2s have scaled Ethereum and unlocked new avenues for innovation, but left the ecosystem *fragmented*. To address this, there are a variety of cross-chain communication protocols designed to make L2s composable with each other, each implements its own message format that is incompatible with others. This ERC proposes a neutral, standard format for sending and receiving cross-chain messages. By standardizing the interface chains for messaging, we achieve:
+
+- **Unified developer experience:** This standard abstracts away the low-level details of message passing from applications. This allows application developers to achieve the following, even among chains with different VMs, coordination protocols, or settlement logic:
+  - send/receive messages to/from many chains using the **same interface**.
+  - deploy their application across multiple chains with **little-to-no** code changes.
+  - focus on their application’s design instead of cross-chain infrastructure.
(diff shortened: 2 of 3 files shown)


ERCS/erc-7828.md
```diff
@@ -45,71 +45,102 @@ Desired properties:
...
+If the ENS name `example.eth` resolved to `0x123..789` subject to the ENS resolution process outlined in ENSIP-9, then the same information could be represented as:
+
+```
+example.eth@eip155:1#FFFFFFFF
+example.eth@ethereum#FFFFFFFF
 ```
 
-> [!NOTE]
-> This standard explicitly defines the use of an agreed <namespace>.eth second-level domain for chain names. Beyond this, hierarchical name structures continue to be inherited from ENS, as exemplified in the ENS-in-rollup example.
+**Note:** `alice.eth@eip155#FFFFFFFF` is *not a valid* _Interoperable Name_. It's `<chain>` component must include both a [CAIP-2] namespace and reference as the address to which an ENS name resolves is chain-specific.
```
0xrcinus: In practice I think lots of ENS names don't have chain-specific resolvers, do we expect client libraries to assume mainnet if just an ENS is provided without a chain?

small nit: "It's" is a typo

ERCS/erc-7828.md
```diff
@@ -45,71 +45,102 @@ Desired properties:
...
-Assuming the on-chain registry list adds a few other CAIP namespaces
-- alice.eth@btc#CCCCCCCC
+### Examples
+
+#### Ethereum Mainnet
+
+The address `0x123..789` on Ethereum Mainnet could be represented as either of the following:
+
+```
+0x123..789@eip155:1#FFFFFFFF
```
0xrcinus: Is it worth having fully worked examples (with checksums calculated etc)

ERCS/erc-7828.md
```diff
@@ -236,7 +267,6 @@ Copyright and related rights waived via [CC0](../LICENSE.md).
 
 
 [ERC-7930]: ./eip-7930.md
-[CAIP 2]: https://github.com/ChainAgnostic/CAIPs/blob/36ef9ee906da090366cb679634a0850e91785db6/CAIPs/caip-2.md
 [UTS #46]: https://www.unicode.org/reports/tr46/tr46-35.html
 [ENSIP-1]: https://docs.ens.domains/ensip/1/
```
0xrcinus: We have some non-relative link warnings, do we need to link to specific Github commits?

ERCS/erc-7828.md
```diff
@@ -45,71 +45,102 @@ Desired properties:
...
 
-- `<raw-chain>`, `<raw-address>` and `<checksum>` are defined to maintain backwards compatibility with [ERC-7930], and their semantics remain the same.
-- `<chain-name>`: A human-readable chain label within the reserved `<namespace>.eth` domain. This standard uses `<namespace>.eth` to provide a consistent abstraction for chain identity, mapping these labels to their corresponding canonical chain identifiers.
-- `<ens-name>`: An ENS name, which should be resolved following the appropiate ENSIPs (which this standard does not aim to overwrite). It's grammar definition is slightly different than that of [ENSIP-1], as this standard does not support names consisting of only a TLD, since they'd be of limited usefulness.
-- `<label>`: Any valid string label per [UTS #46], as defined in [ENSIP-1].
+- `<address>` can either be:
 
-This allows for Interoperable Addresses' text representation to mix and match 'resolved' and 'unresolved' usages in both the chain and address parts.
+    - A _target address_ as defined in [ERC-7930].
+    - An empty string if you are targeting all addresses within a CASA namespace.
```
0xrcinus: Should we have an empty string example in the Examples below?

ERCS/erc-7828.md
```diff
@@ -45,71 +45,102 @@ Desired properties:
...
+#### Ethereum Mainnet
+
+The address `0x123..789` on Ethereum Mainnet could be represented as either of the following:
+
+```
+0x123..789@eip155:1#FFFFFFFF
+0x123..789@ethereum#FFFFFFFF
+```
+
+If the ENS name `example.eth` resolved to `0x123..789` subject to the ENS resolution process outlined in ENSIP-9, then the same information could be represented as:
```
0xrcinus: I think it's worth having separate examples for the ENS names, rather than embedded in the same example?

ERCS/erc-7828.md
```diff
@@ -45,71 +45,102 @@ Desired properties:
...
-- alice.eth@btc#CCCCCCCC
+### Examples
+
+#### Ethereum Mainnet
+
+The address `0x123..789` on Ethereum Mainnet could be represented as either of the following:
+
+```
+0x123..789@eip155:1#FFFFFFFF
+0x123..789@ethereum#FFFFFFFF
```
0xrcinus: We should clarify that this is dependent on the `ethereum` label in the registry resolving to the chain identifier for `eip155:1`

ERCS/erc-7828.md
clowestab: Assuming mainnet would be a bad choice IMO and antithetical to the approaches ENS have taken in other areas.

Will fix typo.

ERCS/erc-7828.md
clowestab: Dependent on checksum conversation outcome, but if we keep them then yes.

ERCS/erc-7828.md
clowestab: I'm not averse. The reason I did it like that was because both examples in each case will have the same checksum. As you suggest above, if we actually calculate and do full work examples with correct checksums, then this is clearer. That said if we have separate examples for the ENS names we can just add a note to indicate that.

ERCS/erc-7828.md
clowestab: I've opted to remove this as I don't think it adds value anymore noting the separation. We might identify 'all addresses within a chain namespace' with 7930 but I cant envision the starting point being an Interoperable Name.

ERCS/erc-7828.md
clowestab: In EIP-1 they use full paths to specific commits like this one for CAIP-2. It's a requirement: "_Permitted Chain Agnostic URLs must anchor to a specific commit_".

I assume it's because in principle these are 'living documents' but I don't think i finalized spec should link to a specific version of something that can change otherwise it defeats the point of allowing it to change, but it is what it is..

ENSIPs are throwing errors, and I think we simply have to **not link** to them as its not allowed in EIP-1. Updated.

ERCS/erc-7828.md
clowestab: Done.

clowestab:
The main outstanding discussion is the value of checksums within this spec.

At the moment address poisoning is a big issue with 0x addresses. ENS mitigates that issue. 

A checksum is valuable if it protects the user in some way.

The argument with Interoperable Names with ENS addresses as the `<address>` component is that ENS is an abstraction in the same way that domain names abstract IP addresses. Ultimately very few end users validate the underlying IP - they trust the DNS protocol and assume security at the registrar level to avoid DNS hijacking. I think similar tradeoffs i.e. no checksums when using ENS names are valid here. Most users don't want to have to see checksums (its confusing - what are these wierd characters), and its an additional implementation hurdle for wallets etc. That said it would still be pragmatic for the spec to RECOMMEND clients resolve the ENS name and display the underlying address as an additional validation for those who want to..

When using a _target address_ (`0x123`) in the `<address>` component we _could_ have a checksum of some sort. @nxt3d outlined an argument that we should not calculate the checksum from components of the underlying ERC-7930 Interoperable Address but rather checksum the Interoperable Name string e.g. `keccak(0x123@base)`. This validates the input integrity. Yes, many different Interoperable Names could point to the same underlying _chain specific address_ but that is not an issue - their needs to be a canonical 7930 Interoperable Address but there is not a need for a canonical Interoperable Name. My inclination would be that when sharing an Interoperable Name the checksum could be appended e.g. when clicking a copy button in an interface, but when displayed in an interface or entered manually, it would not be included.

Writing this out has provided clarity to me. I think what I've outlined above is the best balance of spec simplicity/functionality. Have I missed anything and does anyone have any other thoughts? @nxt3d @0xMonoAx @0xrcinus

ERCS/erc-7828.md
```diff
@@ -45,71 +45,111 @@ Desired properties:
...
+
+- `<checksum>` is defined as the first 4 bytes of the Keccak-256 hash, represented as a lowercase hexadecimal string. The hash is computed over the concatenation of the following binary fields from the canonical [ERC-7930] Interoperable Address:
+
+    - `ChainType`   
+    - `ChainReferenceLength`
+    - `ChainReference`
+    - `AddressLength`
+    - `Address`
+
+**Note:** The Version field MUST NOT be included in the hashed data.
```
0xMonoAx, line 84: Since ERC-7828 is directly based on the semantics of ERC-7930, it may be worthwhile to add a versioning section that inherits the guarantees of ERC-7930, to clearly define the extension limits and preserve exact bit convertibility

ERCS/erc-7828.md
```diff
@@ -28,7 +28,7 @@ The current Ethereum address landscape is leading to an ecosystem that will have
 The mapping from chain names to identifiers has, since [EIP-155], been maintained off chain using a centralized list. This solution has a few shortcomings:
```
0xMonoAx: > ## Abstract
> This proposal extends [ERC-7930] \(Interoperable Addresses & Names\) by standardizing a human-readable format for chain-specific addresses in the form `<address>@<chain>#<checksum>`. It introduces:

Since the definition of _Interoperable Names_ has been removed from ERC-7930, I suggest updating the Abstract and adding some clarifying notes.                                                                                                                                                                                

_This proposal extends [ERC-7930] (Interoperable Addresses) by standardizing a human-readable format for chain-specific addresses, referred to as an Interoperable Name, in the form `<address>@<chain>#<checksum>`._

_The definition of the `<chain>` component is extended to allow the use of human-readable chain names. These identifiers are resolved to an [ERC-7930] Interoperable Address using the Ethereum Name Service (ENS), moving chain metadata discovery and resolution on-chain._

_The definition of the `<address>` component is extended to allow the use of ENS names, as well as names from other compatible naming services._

_This specification introduces:..._

clowestab: I'm not sure I follow? Updates to the Abstract/Motivation are in this PR: https://github.com/ethereum/ERCs/pull/1446. Other PRs will follow once these initial two are merged.

ERCS/erc-7828.md
clowestab: This relates to the checksum conversation, so I guess the action depends on that outcome.

That said, I think with this new separation of concerns ERC-7828 is actually quite abstracted from 7930. It will be even more so if we decide not to use checksums based on it (which i think is a good thing). 7828 then becomes a more abstract human readable naming spec that uses an on-chain registry-resolver.

ERCS/erc-7828.md
0xMonoAx: I hadn’t seen this. I’ll move my comments to that PR

ERCS/erc-7828.md
0xMonoAx: I agree that it depends on what is decided with the checksum

0xrcinus:
> The main outstanding discussion is the value of checksums within this spec.

Just to confirm that I understand what you are suggesting @clowestab...

1. You are proposing that if an ENS name is used as the `address`, checksums should not be part of an Interoperable Name (as in that case the user is delegating responsibility to ENS)
2. You are proposing that the checksumming takes place on the string, rather than the resolved Interoperable Address (as that is simpler for implementers)

Is that correct?

clowestab:
@0xrcinus Yes.

Checksums are good to avoid address poisoning, and we can do that based on a hash of the _Interoperable Name_ **text**. We should make it optional for all** _Interoperable Names_.

There is no added benefit to checksumming the 7930 address - it creates implementation overhead and ties 7828 to 7930 unnecessarily. As proposed 7828 would provide a generic registry-resolver than can get any/all chain specific metadata.

0xMonoAx:
> The main outstanding discussion is the value of checksums within this spec.
> 
> At the moment address poisoning is a big issue with 0x addresses. ENS mitigates that issue.
> 
> A checksum is valuable if it protects the user in some way.
> 
> The argument with Interoperable Names with ENS addresses as the `<address>` component is that ENS is an abstraction in the same way that domain names abstract IP addresses. Ultimately very few end users validate the underlying IP - they trust the DNS protocol and assume security at the registrar level to avoid DNS hijacking. I think similar tradeoffs i.e. no checksums when using ENS names are valid here. Most users don't want to have to see checksums (its confusing - what are these wierd characters), and its an additional implementation hurdle for wallets etc. That said it would still be pragmatic for the spec to RECOMMEND clients resolve the ENS name and display the underlying address as an additional validation for those who want to..
> 
> When using a _target address_ (`0x123`) in the `<address>` component we _could_ have a checksum of some sort. @nxt3d outlined an argument that we should not calculate the checksum from components of the underlying ERC-7930 Interoperable Address but rather checksum the Interoperable Name string e.g. `keccak(0x123@base)`. This validates the input integrity. Yes, many different Interoperable Names could point to the same underlying _chain specific address_ but that is not an issue - their needs to be a canonical 7930 Interoperable Address but there is not a need for a canonical Interoperable Name. My inclination would be that when sharing an Interoperable Name the checksum could be appended e.g. when clicking a copy button in an interface, but when displayed in an interface or entered manually, it would not be included.
> 
> Writing this out has provided clarity to me. I think what I've outlined above is the best balance of spec simplicity/functionality. Have I missed anything and does anyone have any other thoughts? @nxt3d @0xMonoAx @0xrcinus

Hi, we wanted to clarify a few things and make our stance on this discussion clear

It's true that ENS mitigates address poisoning attacks. However, homoglyph-based attacks and UI-level ambiguities can still occur even when using ENS names. Additionally, it is not always clear how consistently applications handle these [edge cases](https://dl.acm.org/doi/epdf/10.1145/3696410.3714675).

That said, we believe:

- Checksums must be computed from the Interoperable Address (not from a string name)
- We recommend displaying the checksum to the user
- We recommend using the checksum when sharing names

Why?

- It gives non-ENS addresses first-class security against address poisoning
- It adds an extra layer of defense against Unicode homoglyph attacks on ENS
- Hardware wallets especially benefit from this, as they can't resolve ENS names on their own, while the dApp requesting the signature can (and may not display the resulting address and chain)

All of these representations should resolve to the same Interoperable Address and therefore the same checksum:

```
vitalik.eth@eip155:1
vitalik.eth@ethereum
0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045@ethereum
0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045@eth
0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045@eip155:1
```

The checksum should be applied to the resolved name, precisely for canonicity reasons. What the user typed resolves to something, and it's important that both source and destination actually point to the same thing. If the checksum is based only on the text, it just means both parties can compute a hash of that text. However, if the checksum of the resolved value matches, it means both are interpreting the input as the same address on the same chain.

Since checksums are recommended not required, we don't think it makes sense to introduce special cases or additional recommendations specifically for ENS names. The mechanism should remain uniform regardless of the input format.

clowestab:
> It's true that ENS mitigates address poisoning attacks. However, homoglyph-based attacks and UI-level ambiguities can still occur even when using ENS names. Additionally, it is not always clear how consistently applications handle these [edge cases](https://dl.acm.org/doi/epdf/10.1145/3696410.3714675).

This is true. Neither proposed checksum approach meaningfully improves homoglyph safety - that is down to accurate implementation of ENSIP 15 (Name normalization) and appropriate warning UI in clients.

> It gives non-ENS addresses first-class security against address poisoning
Text based checksums do too.

> It adds an extra layer of defense against Unicode homoglyph attacks on ENS
I don’t see how this provides additional protection. With either checksumming approach if the name is not appropriately normalized a checksum will be generated for the attackers address.

> Hardware wallets especially benefit from this, as they can't resolve ENS names on their own, while the dApp requesting the signature can (and may not display the resulting address and chain)
I don't quite follow? Hardware wallets are capable of `keccak256` hashing a singular input, but they can't easily derive constituent components of the 7930 Interoperable Address (the client would have to provide them). 

> All of these representations **should** resolve

*emphasis mine

This really depends on what you are trying to achieve which I think is essentially what we disagree on. I don't believe that the benefits of 7930 component derived checksums that you've outlined hold in practice. This approach means that different addresses that resolve to the same underlying location have the same checksum **BUT** this has complexity and implementation costs. Even if libraries are produced to support the implementation, the paper linked above demonstrates that clients etc don't always use the libraries and don't always do things correctly. Case in point is that whilst updating this spec to add the checksums I struggled to calculate the checksum for the bitcoin example simply because of the complexity of its derivation..

My argument is that the main pitch for the 7930 derived checksum is that there is a shared consistent checksum for different text representations that **currently** resolve to the same underlying address. There are however significant trade offs to achieving that.

ERCS/erc-7828.md
clowestab: So based on discussion of splitting 7930 and 7828 but maintaining content for now, I have added calculated checksums.

ERCS/erc-7828.md
```diff
@@ -45,71 +45,111 @@ Desired properties:
...
-```
-Mainnet
-- 0x12345...6789@eip155:1#FFFFFFFF
-- 0x12345...6789@ethereum#FFFFFFFF
-- alice.eth@eth#FFFFFFFF
+    - The string representation of a specific blockchain as defined in [CAIP-350]. For example `eip155:1` for Ethereum Mainnet, or `solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdpKuc147dw2N9d` for Solana Mainnet.
+    - A human-readable label identifier for a specific chain, defined and registered as a subdomain of the `on.eth` ENS name. An on-chain registry-resolver maps these labels to their corresponding canonical [ERC-7930] chain identifiers, and exposes additional chain metadata.
+    - When a _target address_ is used directly within the `<address>` component, a CASA namespace, with no trailing colon can be used - for example, `eip155` or `solana`. When an ENS name is used you MUST specify both the namespace and reference elements of the [CAIP-2] identifier as ENS resolves addresses on a chain-specific basis.
+
+- `<checksum>` is defined as the first 4 bytes of the Keccak-256 hash, represented as a lowercase hexadecimal string. The hash is computed over the concatenation of the following binary fields from the canonical [ERC-7930] Interoperable Address:
```
0xrcinus: I think checksum is currently uppercase per https://www.rfc-editor.org/rfc/rfc4648 and in the examples below (similarly we can discuss in the checksum PR)

0xrcinus (changes requested):
One small note on the checksum casing, otherwise this looks good to me

ERCS/erc-7828.md
clowestab: Updated to maintain consistency pre-split. I do think that in light of the update to 7930 to RECOMMEND lower case, this should also match. It would also be consistent with Ethereum conventions. We can come back to this..

ERCS/erc-7828.md
```diff
@@ -200,20 +240,20 @@ Starting from the Interoperable Address serialized above: `0x00010000010A14AAAAA
...
     2. Call `resolver(namehash(8000000a.reverse))` on the mainnet ENS registry. It'll return `0x00000beef055f7934784d6d81b6bc86665630dba`, the address of the `L2ReverseRegistrar` *on the network with chainid 10*.
 8.  Verify the return of the `coinType` method of the `L2ReverseRegistrar` matches the coinType as serialized in step 7.
 9.  Call `nameForAddr(bytes address)` with the address as serialized in step 8. This will return the human-readable name `alice.eth`.
-10.  Check forward resolution of the name as described in [ENSIP-11] and repeated in the section above.
+10.  Check forward resolution of the name as described in ENSIP-11 and repeated in the section above.
     - If it resolves to the same address, then proceed normally.
     - If forward resolution of the ENS name returns an address different from the original, or returns an empty byte array, the wallet MUST display the raw, human-readable [ERC-7930] serialized address, indicating the address is unresolved. The wallet SHOULD also warn the user of the mismatch to prevent potential spoofing or confusion.
 11.  Compute the checksum as described in [ERC-7930]: `0xC69BEB13`.
```
0xMonoAx, line 256: We have to modify this because the checksum isn't part of the ERC-7930

ERCS/erc-7828.md
```diff
@@ -45,71 +45,111 @@ Desired properties:
...
+
+    - `ChainType`   
+    - `ChainReferenceLength`
+    - `ChainReference`
+    - `AddressLength`
+    - `Address`
+
+**Note:** The Version field MUST NOT be included in the hashed data.
+
+#### Checksums
```
0xMonoAx, line 86: We should include an example that clearly shows how to calculate the checksum

PR #1448, opened 2026-01-04
Update ERC-7930: Added clarifications and new sections

0xMonoAx:
This PR extends ERC-7930 with clarifications and additions:
- Motivation: Clarifies that protocols currently define address/chain combinations using ad hoc, multi-field conventions. This ERC standardizes that representation.
- Text representation: Adds a recommended user-facing text representation derived from the canonical binary Interoperable Address via CAIP-350 profiles
- ~~Extensions: Documents an extension model for building higher-level or human-readable formats on top of the base binary encoding~~

### DIFF
--- a/ERCS/erc-7246.md
+++ b/ERCS/erc-7246.md
@@ -63,7 +63,7 @@ interface IERC7246{
      * Grants to `taker` a guaranteed right to transfer `amount` from the caller's balance by using `transferFrom`.
      *
      * MUST revert if caller does not have `amount` tokens available
-     * (e.g. if `balanceOf(caller) - encumbrances(caller) < amount`).
+     * (e.g. if `balanceOf(caller) - encumberedBalanceOf(caller) < amount`).
      *
      * Emits an {Encumber} event.
      */
@@ -76,7 +76,7 @@ interface IERC7246{
      * The function SHOULD revert unless the owner account has deliberately authorized the sender of the message via some mechanism.
      *
      * MUST revert if `owner` does not have `amount` tokens available
-     * (e.g. if `balanceOf(owner) - encumbrances(owner) < amount`).
+     * (e.g. if `balanceOf(owner) - encumberedBalanceOf(owner) < amount`).
      *
      * Emits an {Encumber} event.
      */
@@ -190,7 +190,7 @@ contract EncumberableERC20 is ERC20, IERC7246 {
         if (exceedsEncumbrance)  {
             uint excessAmount = amount - encumberedToTaker;
 
-            // check that enough enencumbered tokens exist to spend from allowance
+            // check that enough unencumbered tokens exist to spend from allowance
            require(availableBalanceOf(src) >= excessAmount, "insufficient balance");
 
            // Exceeds Encumbrance , so spend all of it
@@ -211,6 +211,8 @@ contract EncumberableERC20 is ERC20, IERC7246 {
         uint newEncumbrance = currentEncumbrance - amount;
         encumbrances[owner][taker] = newEncumbrance;
         encumberedBalanceOf[owner] -= amount;
+
+        emit Release(owner, taker, amount);
     }
 }
 ```
--- a/ERCS/erc-7866.md
+++ b/ERCS/erc-7866.md
@@ -4,7 +4,7 @@ title: Decentralised User Profiles
 description: Decentralised Interoperable User Identity with segmented Avatars per DApp Profile
 author: Kumar Anirudha (@anistark)
 discussions-to: https://ethereum-magicians.org/t/add-erc-decentralised-profile-standard/22610
-status: Draft
+status: Review
 type: Standards Track
 category: ERC
 created: 2025-01-22
@@ -244,12 +244,12 @@ The design of the Decentralised Profile Standard was guided by the need for a un
    - A centralised system would simplify implementation but contradict the core principles of decentralisation and user sovereignty.
 
 4. **Non-SBT-Based Implementation**
-   - Using standard `ERC721` tokens for profiles was considered but rejected since transferability is unsuitable for identity management. SBTs enforce the immutability of identity ownership.
+   - Using standard [ERC-721](./eip-721.md) tokens for profiles was considered but rejected since transferability is unsuitable for identity management. SBTs enforce the immutability of identity ownership.
 
 ### Comparison with Related Work
 
-- **ENS**: Provides a robust naming service but lacks dApp-specific metadata and cross-chain functionality.
-- **Gravatar**: Centralised and cannot leverage blockchain-specific advantages like immutability, decentralisation, and SBT integration.
+- **[ERC-137](./eip-137.md) (ENS)**: Provides a robust naming service but lacks dApp-specific metadata and cross-chain functionality.
+- **Centralised identity services**: Cannot leverage blockchain-specific advantages like immutability, decentralisation, and SBT integration.
 - **DID Standards**: Profile must aligns with DID specifications, ensuring it fits into the broader decentralised identity ecosystem while offering features tailored to blockchain-based applications.
 
 ### Scalability and Extensibility
@@ -262,7 +262,7 @@ The design balances simplicity, extensibility, and user control, making it well-
 
 The standard includes the following interface:
 
-#### ISoulProfile
+#### `ISoulProfile`
 
 ```solidity
 // SPDX-License-Identifier: CC0-1.0
@@ -307,7 +307,7 @@ interface ISoulProfile {
 
 ### Reference Implementation
 
-#### SoulProfile
+#### `SoulProfile`
 
 ```solidity
 // SPDX-License-Identifier: CC0-1.0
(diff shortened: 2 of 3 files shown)


ERCS/erc-7930.md
```diff
@@ -215,13 +217,40 @@ Future versions:
...
+2. Converting each component to its text representation according to the rules specified in the namespace's [CAIP-350] profile
+
+This approach ensures that users see familiar, chain-native address formats while maintaining the canonical binary representation for on-chain and protocol-level usage.
+
+**Example: text representation**
+
+| Key | Value |
+| :--- | :--- |
+| **Interoperable Address: Ethereum mainnet** | `0x0000010114d8da6bf26964af9d7eed9e03e53415d37aa96045` |
+| **Text representation** | `0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045@eip155:1` |
```
0xrcinus: I think rather than the `<address>@<chain` format (which is kind of jumping to the Interoperable Name definition in 7828), let's separately list the Address (`0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045`) and the ChainReference (`eip155:1`), and let's link to the [eip155 caip-350 profile](https://github.com/ChainAgnostic/namespaces/blob/main/eip155/caip350.md) for reference

ERCS/erc-7930.md
```diff
@@ -215,13 +217,40 @@ Future versions:
...
+
+For an example of an extension that provides human-readable text representations of _Interoperable Addresses_, see [ERC-7828].
+
 ## Rationale
 
 The interoperability roadmap benefits significantly from first having a standardized binary format for addresses, which allows the message passing and intents verticals to move forward on a consistent common interface.
 
 The rationale for some of the low level specification decisions are outlined below:
 
-- We maintain a strict Separation of Concerns (SoC) between the underlying binary format for _chain-specific addresses_ and human-readable abstractions defined in standards like [ERC-7828].
```
0xrcinus, line 224: I think it's reasonable to have a note about focusing on the binary format here, but note that standard encodings for other contexts (such as human-readable addresses, or a format for programmatic usage) will be defined in further standards such as 7828 (I am not sure we need to talk about a Separation of Concerns)

0xrcinus (changes requested):
Thanks! Two suggested adjustments

ERCS/erc-7930.md
clowestab: I think this is important, including the reference to SoC as part of the Rationale. If anyone comes to this spec and asks the question as to why this spec has such tight scope they immediately know and get to avoid the debates we all had XD

ERCS/erc-7930.md
```diff
@@ -215,13 +217,40 @@ Future versions:
 - MAY add fields but MUST NOT alter or omit any data required to reconstruct the Version 1 _Interoperable Address_ exactly, bit for bit
 - MAY only be able to represent a subset of the CAIP namespaces
 
+### Text representation in user-facing contexts
```
clowestab: This is just adding back in what we removed, no? I think this adds confusion and complexity.

"_SHOULD NOT represent the binary _Interoperable Address_ directly as `base16`, `base58`, or `base64`_"

This is non-exhaustive. Why are we specifying these specific encoding formats with SHOULD NOT rather than having simple a RECOMMENDED (which we do).

ERCS/erc-7930.md
```diff
@@ -215,13 +217,40 @@ Future versions:
...
+When displaying _Interoperable Addresses_ to users, implementers SHOULD NOT represent the binary _Interoperable Address_ directly as `base16`, `base58`, or `base64` encoded text. Instead, implementers MAY display the `ChainReference` and `Address`components of an _Interoperable Address_ using the text representations defined in the corresponding [CAIP-350] profile for the namespace.
+
+The text representation of the _ChainReference_ and _Address_ components should be obtained by:
```
clowestab: I think this is also unnecessary. We don't need to hand hold and tell people how to get the constituent parts out of an Interoperable Address. Otherwise the spec gets longer and more complex and its also non-exhaustive and opinionated - what things are we going to hand hold, and what are we not..

ERCS/erc-7930.md
```diff
@@ -215,13 +217,40 @@ Future versions:
...
+This approach ensures that users see familiar, chain-native address formats while maintaining the canonical binary representation for on-chain and protocol-level usage.
+
+**Example: text representation**
+
+| Key | Value |
+| :--- | :--- |
+| **Interoperable Address: Ethereum mainnet** | `0x0000010114d8da6bf26964af9d7eed9e03e53415d37aa96045` |
+| **Text representation** | `0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045@eip155:1` |
+
+### Extensions
```
clowestab: I also noted mention of an explicit 'Extensions' section in @0xrcinus 's Google doc. I opted to keep it succinct with the following in 'Rationale' - "_We maintain a strict Separation of Concerns (SoC) between the underlying binary format for _chain-specific addresses_ and human-readable abstractions defined in standards like [ERC-7828]._"

I think this is _fuzzy_ - an abstraction that is additive doesn't have compatibility considerations. I don't think what we are doing with 7828 is an extension but just a completely separate spec that happens to use 7930 identifiers. I can perhaps see the merit if people are going to essentially write a spec that updates 7930 but even then is this just spec bloat? If they want to extend the spec and don't make sensible design decisions with their extension then no-one will use it. I also think there is merit in **not** pushing the idea of extending 7930 - if people are going to go to the effort of implementing it, it needs to be good and worthwhile to do so.. if we dont think the spec is good and think it's going to need to be updated we shouldn't be pushing it IMO.

ERCS/erc-7930.md
0xrcinus: I am specifically objecting to "Separation of concerns" as jargon, I agree that we need a line in the rationale

ERCS/erc-7930.md
0xrcinus: Here we are saying - don't directly display the Interoperable Address. If you want to display to users, use text representations of the address + chin as defined in the relevant CAIP-350 profile.

It is not otherwise obvious to me that CAIP-350 has text representations that can be used in that way (and otherwise we are requiring that people go read that spec).

I see the RECOMMENDED now - I had actually missed that change originally, which I would say is "information". FWIW I find the hex only useful in a debugging / validation context (i.e. I struggle to recommend it)

ERCS/erc-7930.md
0xrcinus: I am open to removing this section, as long as its goals are maintained:
- conveying that 7930 is a low level chain specific address, other formats are desired
- pointing directly to 7828 as a human readable version

Maybe we believe we can do that in the one line rationale, but that is the reasoning here

ERCS/erc-7930.md
0xrcinus: I think without this section, the reader does not know that CAIP-350 profiles include text representations, and I think it is reasonable to provide this pointer

0xMonoAx: It doesn't add additional complexity, unlike Interoperable Name. The text representation is derived directly from CAIP-350 profiles, as clarified in this section

ERCS/erc-7930.md
clowestab: This is a technical spec though..

I'm not tied to the specific jargon, bit its a commonly understood term amongst developers who are ultimate the target of this spec.

ERCS/erc-7930.md
clowestab: But the Interoperable Address is binary data. At @nxt3d 's request as noted I added "_If you choose to display an Interoperable Address it is RECOMMENDED that you display it as a lower case hexadecimal string._" to the definition, but really how these things are displayed is highly opinionated (as you highlight) which is why I don't think it should be a consideration in this spec beyond that simple suggestion.

CAIP-350 is it's own thing, and probably needs updating in light of of our separation. Lets discuss further.

ERCS/erc-7930.md
0xrcinus: Sure, but I think we don't need to use the jargon.

- This ERC defines a compact binary format for a chain-specific address, primarily for use in smart contracts. Other contexts (such as display to end-users or programmatic use in APIs) may prefer a different representation of a chain-specific address. For the sake of simplicity, definition of these representations is left to other ERCs, such as [ERC-7828], which defines a human-readable string representation of a chain-specific address.

ERCS/erc-7930.md
clowestab: Update: Don't show, if must show hex, additionally refer to companion standard CAIP-350.

ERCS/erc-7930.md
clowestab: Covered by above

ERCS/erc-7930.md
clowestab: Update: remove

ERCS/erc-7930.md
clowestab: This text looks good.

ERCS/erc-7930.md
clowestab: Update: happy to remove.

PR #1468, opened 2026-01-12
Update ERC-7930: Final grammar, and authors.

clowestab:
Cleaned up some grammar, removed some duplication, and added outstanding contributors.


ERCS/erc-7930.md
```diff
@@ -218,14 +218,12 @@ Future versions:
...
-
-[ERC-7828] follows this approach and augments it to provide a more human-readable text representation of an _Interoperable Address_ for end-users.
+The companion standard, CAIP-350, specifies on a per-namespace basis, the rules for converting the `ChainReference` and `Address` components of the _Interoperable Address_ into human readable representations. These MAY be displayed in user-facing contexts.
 
 ## Rationale
-This ERC defines a compact binary format for a chain-specific address, primarily for use in smart contracts. Other contexts (such as display to end-users or programmatic use in APIs) may prefer a different representation of a chain-specific address. For the sake of simplicity, definition of these representations is left to other ERCs, such as [ERC-7828], which defines a human-readable string representation of a chain-specific address.
+This ERC defines a compact binary format for a _chain-specific address_, primarily for use in smart contracts. The scope is focussed for simplicity.
```
0xMonoAx: >The scope is focussed for simplicity.

This doesn’t add much clarity, since without the context of everything we’ve been discussing, there’s no point of reference to understand what a much more complex ERC-7930 could be. The previous text is clearer

ERCS/erc-7930.md
```diff
@@ -218,14 +218,12 @@ Future versions:
...
 
 ### Text representation
-This specification defines a canonical binary representation that is not intended to be used directly in user-facing contexts. If an implementer chooses to display an _Interoperable Address_ directly, it is RECOMMENDED that it be represented in lowercase hexadecimal string.
+This specification defines a canonical binary representation that **is not** intended to be used directly in user-facing contexts. If an implementer chooses to display an _Interoperable Address_ directly, it is RECOMMENDED that it be represented as a lowercase hexadecimal string. [ERC-7828] is one example of a human-readable _chain-specific address_ definition that can be displayed in user-facing contexts.
 
-The conversion of an _Interoperable Address_ to and from text representations isn’t defined by this ERC, but by the companion standards established in CAIP-350. CAIP-350 specifies, on a per-_namespace profile_ basis, the rules for converting the binary components (`ChainReference` and `Address`) into their text representations, which MAY be used for the presentation of chain-specific addresses in user-facing contexts.
```
0xMonoAx, line 223: I think this wording is much clearer, and I would keep this version. As we discussed, we were using the human-readable definition only for Interoperable Names, right?

clowestab: My approach here was just avoiding duplication, so this section says 'this spec isn't intended for user-facing contexts' and then the 'Rationale' section says why.

My confusion with this bit was that the `ChainReference` and `Address` as independent units are not a _chain specific address_ as per our definition.

ERCS/erc-7930.md
clowestab: Fair. I've reverted this but rephrased slightly. Is this OK?

ERCS/erc-7930.md
0xMonoAx: > Chain-specific Address : An address representation that includes both the target address and the chain being targeted

Got it, but here we’re clarifying that both components can be used to obtain a chain specific address, which is exactly what this text explains. . I think it’s good for developers to know that they can extract it from here

clowestab: With the EIP-155 profile for example (https://github.com/ChainAgnostic/namespaces/blob/main/eip155/caip350.md) there are 'Chain References' > 'Text Representation' and 'Addresses' > 'Text Representation' sections.

OK.. typing out my thoughts here. I think your point is that the profiles explain how you can represent (as text) the `ChainReference` **and** the `Address`. I think the confusion is that I am being pedantic on semantics - a _chain specific address_ is one thing that has both, as opposed to two separate things.

If my understanding is correct, does the modification "_The companion standard, CAIP-350, specifies on a per-namespace basis, the rules for converting the `ChainReference` and `Address` components of the _Interoperable Address_ into human readable representations. These MAY be displayed in user-facing contexts._" not cover this?

0xMonoAx: I think it does cover it, but it would be better to be more explicit

What I mean is that developers need to clearly understand what we mean by “human-readable representations”, since what we actually want to refer to (under the hood) is a chain specific address

> **Chain-specific Address**
> : An address representation that includes both the _target address_ **and** the chain being targeted. The following are examples of chain-specific addresses:
> 
> - The _Interoperable Address_ definition outlined in this specification
> - The addressing format outlined in [ERC-3770], e.g. `arb:0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045`
> - The _Interoperable Name_ definition outlined in [ERC-7828]

Now that I think about it, we also don’t explicitly clarify the component order (whether it’s `address + chain `or `chain + address`), although I think it’s implied that we prefer `address + chain`

clowestab: Using 'Example 1' from the spec, from the Interoperable Address `0x00010000010114d8da6bf26964af9d7eed9e03e53415d37aa96045`  I can get the `ChainReference` `1`, and the `Address` `0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045`. CAIP-350 outlines that that they should be displayed as `eip155:1` and `0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045` respectively BUT displaying those individual components is **not** a _chain-specific address_, for it to be a _chain-specific address_ it has to be the target address and the chain **in the same _thing_**.

If under the hood a wallet is fetching an Interoperable Address, they may parse out these components but in their user interface they would like show the address and then an avatar representing the chain for example. Tangentially that data can all be fetched from the on-chain registry (from 7828) to bypass this manual labor.

axlvaz (commented):
I’m withdrawing my previous approval. It was a mistake.

axlvaz (approved):

PR #1481, opened 2026-01-21
Update ERC-7828: Update resolution flow with high level specifications

clowestab:
This PR removes the bloated and inaccurate outline of full end-to-end resolution flow for the on-chain registry-resolver, replacing it with a high level overview of the way that _Interoperable Name_'s are parsed and processed. We point to the appropriate ENS specifications where appropriate.

This PR is to get 7828 to a position of high level of completeness whilst discussions are ongoing relating to checksums and the breadth/focus of the standard.

Additionally fixed a few small grammar/parsing errors.

### DIFF
--- a/ERCS/erc-5516.md
+++ b/ERCS/erc-5516.md
@@ -4,41 +4,43 @@ title: Soulbound Multi-owner Tokens
 description: An interface for non-transferable, Multi-owner NFTs binding to Ethereum accounts
 author: Lucas Martín Grasso Ramos (@LucasGrasso), Matias Arazi (@MatiArazi)
 discussions-to: https://ethereum-magicians.org/t/EIP-5516-soulbound-multi-token-standard/10485
-status: Stagnant
+status: Draft
 type: Standards Track
 category: ERC
 created: 2022-08-19
-requires: 165, 1155
+requires: 165
 ---
 
 ## Abstract
-This EIP proposes a standard interface for non-fungible double signature Soulbound multi-tokens. Previous account-bound token standards face the issue of users losing their account keys or having them rotated, thereby losing their tokens in the process. This EIP provides a solution to this issue that allows for the recycling of SBTs.
+
+This EIP proposes a standard interface for non-transferable, multi-owner Soulbound tokens.
+Previous account-bound token standards face the issue of users losing their account keys or having them rotated, thereby losing their tokens in the process. This EIP provides a solution to this issue that allows for the recycling of SBTs.
 
 ## Motivation
-This EIP was inspired by the main characteristics of the [EIP-1155](./eip-1155.md) token and by articles in which benefits and potential use cases of Soulbound/Accountbound Tokens (SBTs) were presented. 
+
+This EIP was inspired by the main characteristics of the [ERC-1155](./eip-1155.md) token standard and by articles in which benefits and potential use cases of Soulbound/Accountbound Tokens (SBTs) were presented.
 This design also allows for batch token transfers, saving on transaction costs. Trading of multiple tokens can be built on top of this standard and it removes the need to approve individual token contracts separately. It is also easy to describe and mix multiple fungible or non-fungible token types in a single contract.
 
 ### Characteristics
+
 - The NFT will be non-transferable after the initial transfer
-- Partially compatible with [EIP-1155](./eip-1155.md)
-- Double Signature
 - Multi-Token
 - Multi-Owner
 - Semi-Fungible
 
 ### Applications
+
 - Academic Degrees
 - Code audits
 - POAPs (Proof of Attendance Protocol NFTs)
 
 ## Specification
+
 The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL NOT”, “SHOULD”, “SHOULD NOT”, “RECOMMENDED”, “MAY”, and “OPTIONAL” in this document are to be interpreted as described in RFC 2119.
 
-**Smart contracts implementing this EIP MUST implement all of the functions in the `EIP-5516` interface.**
+**Smart contracts implementing this EIP MUST implement all of the functions in the [ERC-5516](./eip-5516.md) interface.**
 
-**Smart contracts implementing this EIP MUST implement the [EIP-165](./eip-165.md) `supportsInterface` function and MUST return the constant value `true` if `0x8314f22b` is passed through the `interfaceID` argument. They also MUST implement the [EIP-1155](./eip-1155.md) Interface and MUST return the constant value `true` if `0xd9b67a26` is passed through the `interfaceID` argument. Furthermore, they MUST implement the [EIP-1155](./eip-1155.md) Metadata interface, and MUST return the constant value `true` if `0x0e89341c` is passed through the `interfaceID` argument.**
-
-_See [EIP-1155](./eip-1155.md#specification)_
+**Smart contracts implementing this EIP MUST implement the [EIP-165](./eip-165.md) `supportsInterface` function and MUST return the constant value `true` if `0x45b253ba` is passed through the `interfaceID` argument.**
 
 ```solidity
 // SPDX-License-Identifier: CC0-1.0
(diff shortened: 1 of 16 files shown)


ERCS/erc-7828.md
```diff
@@ -17,17 +17,17 @@ This proposal defines an _Interoperable Name_, a human-readable _chain-specific
 
 Additionally, the target chain component of an _Interoperable Name_ may be expressed as a human-readable label. This specification defines how such labels can be resolved, via the Ethereum Name Service (ENS), to chain-specific metadata.
 
-The result is _chain-specific addresses_ such as `wallet.ensdao.eth@eip155:1`, `0xFe89cc7aBB2C4183683ab71653C4cdc9B02D44b7@eip155:1` or the maximally readable `example.eth@base`.
+The result is _chain-specific addresses_ such as `wallet.ensdao.eth@eip155:1`, `0xFe89cc7aBB2C4183683ab71653C4cdc9B02D44b7@eip155:1` or the maximally readable `wallet.ensdao.eth@ethereum`.
```
0xrcinus, line 20: Should we update the other examples below (which reference `example.eth`) to use `wallet.ensdao.eth`

ERCS/erc-7828.md
```diff
@@ -148,136 +138,50 @@ example.eth@bip122:000000000019d6689c085ae165831e93#597D21A1
...
+If the `<chain>` component of an _Interoperable Address_ does not contain a colon (`:`) it is interpreted as being label under the `on.eth` ENS namespace.
 
-1.  Encode the [CAIP 2] identifier in [ERC-7930] v1 binary form (`AddressLength = 0`); name the result `chainBytes`.
-2.  Build resolver calldata: `callData = abi.encodeWithSelector(IL2Resolver.chainName.selector, chainBytes)`.
-3.  Query the Universal Resolver: `result = UniversalResolver.resolve(dnsEncode("l2.eth"), calldata)`.
-4.  ABI-decode `result` as `(string fullName)`.
-5.  Verify `fullName` ends with `.l2.eth`.
-6.  Strip the suffix and apply UTS #46 lower-case normalisation.
-    • Use the remaining label (e.g. **“arbitrum”** instead of **“arbitrum.l2.eth”**).
+The resolver for `on.eth` MUST implement ENSIP-24 - Arbitrary Data Resolution, and allow for the resolution of the [ERC-7930] _Interoperable Address_ using the key, `interoperable-address`. For example if you resolve the `interoperable-address` data record for `ethereum.on.eth` the _Interoperable Address_ `0x000100000101` SHOULD be returned.
```
0xrcinus: Maybe it is because I am less familiar with the internals of ENS resolution, but I think it would be good for the example to include the actual function call to get the data record

ERCS/erc-7828.md
```diff
@@ -148,136 +138,50 @@ example.eth@bip122:000000000019d6689c085ae165831e93#597D21A1
...
 
-To obtain the **chain identifier** corresponding to a human-readable chain name, clients SHALL:
+The resolver for `on.eth` MUST implement ENSIP-5 - Text Records, and allow for the resolution of text records for the `reverse.on.eth` namespace of the form `chain-label:` followed by the [ERC-7930] _Interoperable Address_ you are trying to discern the label for. For example, the _Interoperable Address_ representing Ethereum Mainnet is `0x000100000101`. Resolution of the text record `chain-label:0x000100000101` for the `reverse.on.eth` namespace SHOULD return `ethereum`.
```
0xrcinus: As above, I think a slightly more detailed example would be helpful here

ERCS/erc-7828.md
0xrcinus:
```suggestion
The resolver for `on.eth` MUST implement ENSIP-24 - Arbitrary Data Resolution, and allow for the resolution of the [ERC-7930] _Interoperable Address_ for the chain using the key, `interoperable-address`. For example if you resolve the `interoperable-address` data record for `ethereum.on.eth` the _Interoperable Address_ `0x000100000101` SHOULD be returned.
```

ERCS/erc-7828.md
```diff
@@ -148,136 +138,50 @@ example.eth@bip122:000000000019d6689c085ae165831e93#597D21A1
...
 
-1.  Apply UTS #46 lower-case normalisation to the chain name.
-2.  Append the reserved suffix: `full = "arbitrum" + ".l2.eth"`.
-3.  Compute `node = namehash(full)` as per ENSIP-1.
-4.  Build resolver calldata: `callData = abi.encodeWithSelector(IL2Resolver.chainId.selector, node)`.
-5.  Query the Universal Resolver: `result = UniversalResolver.resolve(dnsEncode(full), callData)`.
-6.  ABI-decode `result` as `(bytes chainBytes)`.
-    • If `chainBytes` is empty → no mapping exists; treat the label as unknown.
+Additional implementation details about the registry-resolver contract implementation can be discerned by referencing the ENS documentation, and viewing the verified source code of the contract set as the resolver for `on.eth`.
```
0xrcinus: I think it is worth mentioning aliasing, i.e. that multiple labels may resolve to the same Interoperable Address (but an Interoperable Address will only resolve to one label)

ERCS/erc-7828.md
```diff
@@ -148,136 +138,50 @@ example.eth@bip122:000000000019d6689c085ae165831e93#597D21A1
...
 
 
-#### Supported CASA namespaces
+### Resolving the `<address>` component
 
-While currently only `eip155` chains are supported, the on-chain registry could start listing other chains in order to make human-readable names for non-evm chains possible.
+If an ENS address is used within the `<address>` component, it MUST be resolved subject to the ENS resolution specifications, giving consideration to the target chain specified in the `<chain>` component.
 
-### Resolving address names
+The specifications of note are ENSIP-9: Multichain address resolution, ENSIP-10: Wildcard resolution, and ENSIP-11: EVM compatible Chain Address Resolution. These specification outline how one resolves an ENS name to discern the address **for a specific chain**.Consideration MUST be given to both current and future ENSIPs pertaining to address resolution.
```
0xrcinus:
```suggestion
The specifications of note are ENSIP-9: Multichain address resolution, ENSIP-10: Wildcard resolution, and ENSIP-11: EVM compatible Chain Address Resolution. These specification outline how one resolves an ENS name to discern the address **for a specific chain**. Consideration MUST be given to both current and future ENSIPs pertaining to address resolution.
```

ERCS/erc-7828.md
```diff
@@ -148,136 +138,50 @@ example.eth@bip122:000000000019d6689c085ae165831e93#597D21A1
...
-6.  ABI-decode `result` as `(bytes chainBytes)`.
-    • If `chainBytes` is empty → no mapping exists; treat the label as unknown.
+Additional implementation details about the registry-resolver contract implementation can be discerned by referencing the ENS documentation, and viewing the verified source code of the contract set as the resolver for `on.eth`.
 
 
-#### Supported CASA namespaces
+### Resolving the `<address>` component
 
-While currently only `eip155` chains are supported, the on-chain registry could start listing other chains in order to make human-readable names for non-evm chains possible.
+If an ENS address is used within the `<address>` component, it MUST be resolved subject to the ENS resolution specifications, giving consideration to the target chain specified in the `<chain>` component.
```
0xrcinus, line 219: Is it worth being specific about how to identify whether it's an ENS address (i.e. is it assuming that anything with a `.` is?)

ERCS/erc-7828.md
0xrcinus: We also don't mention anything about the "chains taking ownership of their metadata" thing which is in the rationale.. I agree that this isn't necessary for the naming standard, but I think if it isn't covered we shouldn't mention it in the motivation - but it may be worth mentioning at least the expected mechanic?

clowestab: I've added a section 'The Canonical Resolver` that addresses this.

ERCS/erc-7828.md
clowestab: Yes. Added.

ERCS/erc-7828.md
clowestab: Fixed.

ERCS/erc-7828.md
clowestab: Good point. I've added a sentence to each of the forward resolution and reverse resolution sections to mention this.

ERCS/erc-7828.md
clowestab: Updated

ERCS/erc-7828.md
clowestab: Added

ERCS/erc-7828.md
clowestab: Fair enough. Added.

ERCS/erc-7828.md
clowestab: I have done this. `wallet.ensdao.eth` DOES resolve to `0xFe89cc7aBB2C4183683ab71653C4cdc9B02D44b7` on Ethereum mainnet (at least for now). The only confusion is with the bitcoin example - `wallet.ensdao.eth` doesn't actually resolve to anything on the Bitcoin network. I'm unsure if this is unclear/confusing. Thoughts?

ERCS/erc-7828.md
```diff
@@ -17,17 +17,17 @@ This proposal defines an _Interoperable Name_, a human-readable _chain-specific
 
 Additionally, the target chain component of an _Interoperable Name_ may be expressed as a human-readable label. This specification defines how such labels can be resolved, via the Ethereum Name Service (ENS), to chain-specific metadata.
```
0xMonoAx, line 18: I think the abstract does not clearly convey what a developer will encounter. I believe it would be more beneficial for the ERC to be more direct, so I propose the following new abstract:

> This proposal defines an Interoperable Name, a human-readable, chain-specific address format of the form `<address>@<chain>#<checksum>`, where both the `<address>` and `<chain>` components MAY be expressed as human-readable names and resolved via ENS.
> An on-chain registry enables decentralized resolution of chain labels to their canonical identifiers, while address resolution leverages the existing ENS infrastructure.
> The result is chain-specific addresses such as:
> • `0xFe89cc7aBB2C4183683ab71653C4cdc9B02D44b7@eip155:1#80B12379`
> • `alice.eth@eip155:1#80B12379`
> • or the most human-readable version, `alice.eth@base#80B12379`

The checksum is optional, and this is explained in the Checksum section

ERCS/erc-7828.md
```diff
@@ -17,17 +17,17 @@ This proposal defines an _Interoperable Name_, a human-readable _chain-specific
...
 
-The result is _chain-specific addresses_ such as `wallet.ensdao.eth@eip155:1`, `0xFe89cc7aBB2C4183683ab71653C4cdc9B02D44b7@eip155:1` or the maximally readable `example.eth@base`.
+The result is _chain-specific addresses_ such as `wallet.ensdao.eth@eip155:1`, `0xFe89cc7aBB2C4183683ab71653C4cdc9B02D44b7@eip155:1` or the maximally readable `wallet.ensdao.eth@ethereum`.
 
 ## Motivation
 
 The core motivation for introducing the _Interoperable Name_ standard is to provide maximally readable, _chain-specific addresses_ for user-facing interactions.
```
0xMonoAx, line 24: I think this section is missing important context about the underlying problem: addresses alone are no longer sufficient when the same format is reused across hundreds of L2s. Adding the following paragraph would make the section more complete:

> The current Ethereum address landscape is leading to an ecosystem that will have hundreds and eventually thousands of L2s that use the same address format as Ethereum mainnet. This means an address by itself is not enough information to know which chain the address is related to. This can be problematic if funds are sent to an unreachable address on the incorrect chain. From the user account it should be possible to obtain the right chain identifier (chainID) to include in a transaction.

ERCS/erc-7828.md
```diff
@@ -124,160 +114,144 @@ bc1qwz2lhc40s8ty3l5jg3plpve3y3l82x9l42q7fk@bip122:000000000019d6689c085ae165831e
...
+Assuming that `wallet.ensdao.eth` resolves to `bc1qwz2lhc40s8ty3l5jg3plpve3y3l82x9l42q7fk` **for Bitcoin mainnet** either of the following could be used to represent the same _target address_:
 
 ```
-example.eth@bip122:000000000019d6689c085ae165831e93#597D21A1
-example.eth@bitcoin#597D21A1
+wallet.ensdao.eth@bip122:000000000019d6689c085ae165831e93#597D21A1
+wallet.ensdao.eth@bitcoin#597D21A1
+```
+
+### The Canonical Resolver
```
0xMonoAx: This paragraph mentions some governance/operational details (like the "5/7 multisig", "reputable industry participants", future transfers, etc.) that might not fit well in the Specification section since they could change over time. Per EIP-1, specs should focus on technical details that allow interoperable implementations. Maybe these details would work better in Rationale or Security Considerations?

ERCS/erc-7828.md
```diff
@@ -124,160 +114,144 @@ bc1qwz2lhc40s8ty3l5jg3plpve3y3l82x9l42q7fk@bip122:000000000019d6689c085ae165831e
...
+
+While the specific internal implementation of the resolver remains extensible to allow for future upgrades, this specification defines the core functions that the resolver MUST support.
+
+### Resolving the `<chain>` component
+
+If the `<chain>` component of an _Interoperable Address_ does not contain a colon (`:`) it is interpreted as being label under the `on.eth` ENS namespace.
+
+The resolver for `on.eth` MUST implement ENSIP-24 - Arbitrary Data Resolution, and allow for the resolution of the [ERC-7930] _Interoperable Address_ for the specified chain label using the key, `interoperable-address`. For example if you resolve the `interoperable-address` data record for `ethereum.on.eth` the _Interoperable Address_ `0x000100000101` SHOULD be returned.
```
0xMonoAx: I feel that we are misusing not only the term but also the implementation of the Interoperable Address.

ERC-7930 defines it as follows:

> Interoperable Address : A binary payload which unambiguously identifies a target address on a target chain. e.g. 0x00010000010114d8da6bf26964af9d7eed9e03e53415d37aa96045

I think we should explain in more detail how to extract the component, not just in this section but also in the next one

ERCS/erc-7828.md
```diff
@@ -124,160 +114,144 @@ bc1qwz2lhc40s8ty3l5jg3plpve3y3l82x9l42q7fk@bip122:000000000019d6689c085ae165831e
...
+If the `<address>` component contains a period (`.`), it is assumed to be an ENS name. Because ENS fully integrates with the Domain Name System (DNS), any traditional domain name can function as an ENS name, as can any name using the blockchain-native `.eth` extension.
 
-#### Supported CASA namespaces
+If an ENS address is used within the `<address>` component, it MUST be resolved subject to the ENS resolution specifications, giving consideration to the target chain specified in the `<chain>` component.
 
-Forward resolution (Interoperable Address -> Interoperable Name): Supported via the multichain address resolution standard defined in [ERC-2304], which uses [BIP 44] registered coin type identifiers to resolve names for various chains, including the `eip155` namespace.
+The specifications of note are ENSIP-9: Multichain address resolution, ENSIP-10: Wildcard resolution, and ENSIP-11: EVM compatible Chain Address Resolution. These specification outline how one resolves an ENS name to discern the address **for a specific chain**. Consideration MUST be given to both current and future ENSIPs pertaining to address resolution.
```
0xMonoAx, line 221: > Consideration MUST be given to both current and future ENSIPs pertaining to address resolution

This should be in the Security Considerations section

ERCS/erc-7828.md
```diff
@@ -124,160 +114,144 @@ bc1qwz2lhc40s8ty3l5jg3plpve3y3l82x9l42q7fk@bip122:000000000019d6689c085ae165831e
...
 
-- Wallet developers should be aware of possible unicode glyph collisions in resolved names and warn users about them in order to keep checksums effective, since an attacker could, in order to impersonate `alice.eth@chain.<<namespace>.eth#00112233`:
-    - Mine an address on `chain.<namespace>.eth` such that the checksum of the Interoperable Address is `00112233`.
-    - Register `аlice.eth` (using the russian vowel `а` instead of the latin `a`), and point it to the address above.
-- ENSIP-19 introduces trust assumptions when resolving reverse records offchain:
-    - ZK-rollups may offer trust-minimized resolution via on-chain proofs.
-    - Sidechains may rely on centralized CCIP-read gateways or signature authorities.
-
-    Wallets MUST surface whether reverse resolution was trust-minimized or trusted.
+- Implementers MUST give consideration to the name normalization specifications of the Ethereum Name Service so as to avoid homoglyph or spoofing attacks. These are outlined in ENSIP-1 and ENSIP-15.
```
0xMonoAx, line 232: I think we should be clearer here: while ENSIP-15 provides normalization against the most common homoglyph attacks, it's not a complete solution. Applications displaying ENS names should implement additional protection layers (e.g., checksum verification).

ERCS/erc-7828.md
```diff
@@ -124,160 +114,144 @@ bc1qwz2lhc40s8ty3l5jg3plpve3y3l82x9l42q7fk@bip122:000000000019d6689c085ae165831e
...
+- Users should stay vigilant of address poisoning attacks when using a raw _target address_ in the `<address>` component.
```
0xMonoAx, line 233: I would also clarify that they can use the checksum

ERCS/erc-7828.md
0xrcinus: I think this is one slightly unintuitive thing about Interoperable Addresses, in that they can refer to just a chain (as they do in this case) - that definition in 7930 is maybe misleading

ERCS/erc-7828.md
clowestab: As suggested by @0xrcinus externally, this should probably be a separate PR if you want to come back to these sections.

ERCS/erc-7828.md
clowestab: As suggested by @0xrcinus externally, this should probably be a separate PR if you want to come back to these sections.

ERCS/erc-7828.md
clowestab: Fair. Fixed. Will add longer general comment.

ERCS/erc-7828.md
clowestab: Interestingly, I got the 'chain only' Interoperable Addresses wrong, and we all missed it :| Oops. Fixed. But yes, these take advantage of the flexibility of 7930 to represent a chain only.

ERCS/erc-7828.md
clowestab: I put it here because whilst it is a security consideration it is also necessarily part of the spec because if ENS changes how resolution works, and thats not noted here the logical conclusion would be that this spec is essentially hard-coding old resolution standards..

ERCS/erc-7828.md
clowestab: Thats referenced explicitly in the Checksum section which maintaining the previous state of affairs requires checksum verification if provided, and requires them when sharing. I don't agree that this is the right approach but to get to a base from where to have these conversations we were just keeping things as is for now.

clowestab:
So, I reflected on this over the weekend having spoken to Josh and Barnabe.

I agree, governance should not be included in this spec. I have removed it.

Noting that the on-chain registry-resolver is an incredible powerful standalone product I have opted to build out documentation for it on the ENS documentation site. Governance stuff can be housed there, and that is an appropriate living location if we ever choose to upgrade the implementation down the line. A pull request can be found here: https://github.com/ensdomains/docs/pull/508

PR #1484, opened 2026-01-22
Update ERC-8019: Move to Review

Ivshti:
title says it all - sufficient time has passed, this ERC is now implemented live in prod by fileverse and ambire, so it is time to move it to Last Call and then to Final

### DIFF
--- a/ERCS/erc-8019.md
+++ b/ERCS/erc-8019.md
@@ -140,4 +140,4 @@ function shouldAutoSign(M, P) {
 
 ## Copyright
 
-Copyright and related rights waived via CC0.
+Copyright and related rights waived via [CC0](../LICENSE.md).


Ivshti:
hey @jochem-brouwer, done!

ERCS/erc-8019.md
SamWilsn, line 3: > Minimal

This word gets thrown around _a lot_ in ERCs. It is allowed, but you could probably find a more unique or useful adjective to describe your proposal (if you like).

ERCS/erc-8019.md
```diff
@@ -4,7 +4,7 @@ title: Minimal Wallet-Managed Auto-Login for SIWE
 description: Defines a wallet-local allowlist for automatic signing of ERC-4361 messages when simple match rules succeed.
```
SamWilsn, line 4: Great description though! Gives a solid idea of what the proposal is about.

ERCS/erc-8019.md
SamWilsn, line 16: Not a hard requirement, but adding a brief example of what a match rule might look like could help the reader understand what to expect.

SamWilsn:
```suggestion
Users repeatedly sign identical Sign-In With Ethereum (SIWE) messages for trusted apps. A small, explicit match policy enables zero-prompt login without involving apps.
```

SIWE is _probably_ well known enough to omit the expanded form, but if you want to, here's a suggestion for it.

SamWilsn:
````suggestion
```js
````

If you're going to use comments in your code fence, they'll get highlighted as an error if you use `json`.

Comment (SamWilsn) [inline ERCS/erc-8019.md]: Please use the exact copyright waiver from the [template](https://github.com/ethereum/EIPs/blob/master/eip-template.md?plain=1).

Comment (Ivshti) [reply ERCS/erc-8019.md:3]: I'd rather not - can't come up with anything more meaningful right now. Perhaps "rules-based" but the point here is that it is literarily a minimal addition to a SIWE implementation

Comment (Ivshti) [reply ERCS/erc-8019.md:16]: there is an example in the specification (policy = rule), I think that's enough?

File: ERCS/erc-8019.md [hunk 2]
```diff
@@ -31,7 +31,7 @@ The term “wallet” refers to wallet user interfaces, regardless of whether mo
 
 Each allow-policy is defined as:
 
-```json
+```js
```
jochem-brouwer, line 34:
```suggestion
```json
```

This is still json, right? I will approve nevertheless

jochem-brouwer (approved):
LGTM

PR #1494, opened 2026-01-28
Update ERC-7730: [ERC 7730] Improve maps handling

lcastillo-ledger:


lcastillo-ledger:
Included in combined PR #1499

PR #1521, opened 2026-02-06
Add ERC: Content-Addressable Logic Modules (CALM)

radeksvarz:

 - The build passes.


ERCS/erc-8122.md
```diff
@@ -0,0 +1,200 @@
...
+discussions-to: https://ethereum-magicians.org/t/content-addressable-logic-modules-calm-deterministic-pure-runtime-bytecode-deployment/23070
+status: Draft
+type: Standards Track
+category: ERC
+created: 2026-01-18
+---
+
+## Abstract
+
+This standard defines Content-Addressable Logic Modules (CALM) - minimal, deterministic blocks of code designed for execution via delegatecall within Diamond ([ERC-2535](./eip-2535.md), [ERC-8109]((./eip-8109.md))), UUPS ([ERC-1822](./eip-1822.md)) and other proxy architectures.
```
SamWilsn:
```suggestion
This standard defines Content-Addressable Logic Modules (CALM) - minimal, deterministic blocks of code designed for execution via delegatecall within Diamond ([ERC-2535](./eip-2535.md), [ERC-8109](./eip-8109.md)), UUPS ([ERC-1822](./eip-1822.md)) and other proxy architectures.
```

ERCS/erc-8152.md
```diff
@@ -0,0 +1,200 @@
...
+
+    fallback() external payable {
+        assembly {
+            // Store the value 18 (0x12) in memory at offset 0x00.
+            // mstore(offset, value) stores a 32-byte word.
+            mstore(0x00, 0x12)
+
+            // Return 32 bytes of data from memory starting at offset 0x00.
+            // This returns the uint256 representation of 18, which ABI-decoding will interpret as uint8.
+            return(0x00, 0x20)
```
wjmelements, line 141: use `msize()` instead of `0x20`

wjmelements:
```suggestion
            return(0x00, msize())
```

wjmelements: ah, doesn't work in yul:
```
Error (6553): The msize instruction cannot be used when the Yul optimizer is activated because it can change its semantics. Either disable the Yul optimizer or do not use the instruction.
```

wjmelements: Would you consider just writing the optimal bytecode (`0x60125f52595ff3`) instead of prescribing it in solidity? Seems solidity can't do it.

You can use [my assembly dialect](https://github.com/wjmelements/evm) if you want to show source:
```
// eighteen.evm
MSTORE(0, 18)
RETURN(0, MSIZE)
```
```sh
$ evm eighteen.evm
60125f52595ff3
```

radeksvarz: I see I might be misleading with the Reference implementation chapter.
That Solidity code is provided as the example of CALM like contract.
0x60125f52595ff3 is literally written in the subchapter 3 below the code.

But I think it is useful to write there as well the EVM assembly version below. Thanks for hint!

ERCS/erc-8122.md
```diff
@@ -0,0 +1,200 @@
+---
+eip: 8122
```
abcoathup:
```suggestion
eip: 8152
```

ERCS/erc-8122.md
```diff
@@ -0,0 +1,200 @@
...
+title: Content-Addressable Logic Modules (CALM)
+description: Content-addressable logic modules for proxy architectures. Multichain consistent and permissionlessly replicable, leading to gas-efficiency.
+author: Radek Svarz (@radeksvarz), Nick Mudge (@mudgen)
+discussions-to: https://ethereum-magicians.org/t/content-addressable-logic-modules-calm-deterministic-pure-runtime-bytecode-deployment/23070
```
abcoathup:
```suggestion
discussions-to: https://ethereum-magicians.org/t/erc-8152-content-addressable-logic-modules-calm/23070
```
Updated to use assigned number, and ERC title.

ERCS/erc-8152.md
```diff
@@ -0,0 +1,200 @@
...
+
+While such constants are used uniformly with all related CALM contracts on any chain.
+
+*Note: CALMs may be authored in any language (e.g. Solidity, Vyper, Huff, Yul) as long as the resulting bytecode adheres to the runtime constraints. The standard focuses on the bytecode output, not the source language.*
+
+### Runtime Constraints
+
+ - No Self-Destruct: The contract MUST NOT contain the SELFDESTRUCT (0xFF) opcode. This ensures permanent availability for the proxies relying on the logic on those chains that are not compliant with [EIP-6049](./eip-6049.md), [EIP-6780](./eip-6780.md) and other SELFDESTRUCT corresponding changes.
+
+ - Stateless Execution for itself: The contract MUST NOT directly access its own storage.
```
mudgen, line 62: I don't think it is necessary to have "Stateless Execution for itself: The contract MUST NOT directly access its own storage." as a runtime constraint.  They shouldn't do that, but I don't think it is necessary to require that.  Or is there some reason they must not do this?

If a CALM's external function is called using a regular external function call (not delegatecall) then it will access its own state, if there is any code accessing contract storage, breaking this rule.

ERCS/erc-8152.md
```diff
@@ -0,0 +1,200 @@
...
+
+### Logic Dispatching Models
+CALM supports two distinct models for execution:
+
+
+| Model |Description | Primary Use Case |
+|--|--|--|
+| **Multi-Method (Standard)** | Contains an internal dispatcher (e.g. Solidity) that routes calls based on the first 4 bytes in calldata ( standard `msg.sig` in Solidity ). | Complex diamond facets containing multiple related functions. Example: Transaction, approval and metadata logic of [ERC-20](./eip-20.md) contract  |
```
mudgen: Just because a facet has more than one function does not mean it is a complex facet. I suggest removing the word "Complex" here.

ERCS/erc-8152.md
```diff
@@ -0,0 +1,200 @@
...
+ - Simple Logic: For trivial CALMs (like Const18 above), stripping metadata is preferred. The bytecode is short enough to be verified via decompilation.
+
+ - Complex Logic: For sophisticated modules - such as a highly optimized unconditional `ERC20.transfer` utilizing ERC-8042 (Diamond Storage) - preserving full CBOR metadata is recommended. This ensures that comments, security warnings, and the exact audit environment are cryptographically bound to the CALM's identity.
+
+
+## Security Considerations
+
+Since a CALM is "stateless" but "state-manipulating," it must be carefully audited to ensure it only touches the storage namespaces it is authorized for.
+
+Proxies such as Diamonds using CALMs SHOULD audit and test their overall configuration in order to ensure CALMs are not misaligned in the storage handling. For instance, security vulnerabilities can arise if one CALM utilizes the [ERC-7201](./eip-7201.md) storage pattern while another CALM interacting with the same logical state utilizes the [ERC-8042](./eip-8042.md) pattern, potentially leading to storage collisions or desynchronized data states.
```
mudgen: This sentence is inaccruate:

> For instance, security vulnerabilities can arise if one CALM utilizes the ERC-7201 storage pattern while another CALM interacting with the same logical state utilizes the ERC-8042 pattern, potentially leading to storage collisions or desynchronized data states.

If ERC-7201 code and ERC-8042 code both try to write data to storage at the same location, for example at "erc20.token", there is no chance of storage collision because they write to completely different, random locations. 

I suggest changing the sentence to this:

> For example, mixing ERC-7201 and ERC-8042 storage patterns across CALMs will cause bugs because they reference different storage locations for the same logical state.

mudgen:
I think this is a great, well written, ERC. Good job!

ERCS/erc-8152.md
radeksvarz: The point is that once the self storage access is allowed it will effectively lead to mistakes with governed facet.

And that could impact several diamond proxies using such facet.

Example:

I store directly condition = true  at the facet storage.

Now within the facet once delgated from diamond I can do the call to itself to read that storage like <given_facet_address>.value and behave accordingly. So I fear this capability might be rather misused. 

Ofc the facet could read such value from any other contract and thus having the same behaviour, the point is, that influencing its own storage looks like obfuscation.

ERCS/erc-8152.md
radeksvarz: thx, reflected

ERCS/erc-8152.md
radeksvarz: thx, reflected

ERCS/erc-8152.md
mudgen: Yes, I agree.  I just think that this EIP is about content addressable logic modules.  Restricting what it can do with its contract storage state goes a bit outside the scope,  which to me is content-logic addressability across blockchains. Doing so may limit some things (I don't know what, maybe it doesn't matter) that are valid that we haven't thought of yet.  But if you want this, I say go ahead and keep it in the standard.

Even in the EIP-2535 Diamonds, or any diamond standard I write, I do not limit what people can do with state.  People can use `selfdestruct` in facets or diamonds. Such a thing does not affect compliance with the standard. I put such dangerous things in the Security Considerations section as warnings, not the Specification section which determines compliance with the standard.

Forbidding constructors and immutable variables in this standard in the Specification section makes sense,  because the standard doesn't work without that -- those things affect content addressability.

mudgen:
@xinbenlv This pull request is ready to be merged as a draft.  Please approve.

xinbenlv (approved):

PR #1532, opened 2026-02-10
Add ERC: Agent Onchain Metadata

nxt3d:
This PR introduces an onchain metadata extension for **ERC-8004 agent registries**. It defines a set of standard snake_case keys (name, description, image, agent_account, x402_support, active, supported_trust, and endpoint/<name> / endpoint/<name>/version) stored via ERC-8004’s existing getMetadata / setMetadata functions. Each key maps directly to a specific field in the ERC-8004 registration JSON, so the full JSON payload can be mirrored onchain when desired.
**The extension is fully compatible with ERC-8004**: existing agents that don’t use onchain metadata continue to work unchanged via agentURI, and registries don’t need new contracts or storage layouts. When both onchain values and an agentURI JSON exist, clients following this ERC are required to treat the per-key onchain values as authoritative overrides of the corresponding JSON fields, **while agentURI maintains compatibility with ERC-8004**.


ERCS/_metdata_extension.md
```diff
@@ -0,0 +1,149 @@
+---
+eip: XXXX
```
abcoathup:
```suggestion
eip: 8156
```

hilljessica8787-dot (approved):

PR #1534, opened 2026-02-10
Add ERC: Accept pre-approved tip amounts

nxt3d:
This PR adds a minimal tipping interface for payable contracts that want to support optional tips without opening themselves up to unbounded overpayment. Contract operators can pre-approve specific tip amounts, and when a user overpays by exactly one of these amounts, the extra is treated as a tip and surfaced via a standard TipReceived event. Overpayments that do not match a pre-approved amount can either be rejected or explicitly refunded, depending on the implementation.
The ERC is intentionally small: it standardizes setTip, tipAmounts, and the TipUpdated / TipReceived events, plus a simple overpayment-handling pattern that wallets and frontends can integrate with. The Security Considerations section calls out overpayment as a risk, notes that refunding is perfectly acceptable (with reentrancy protection), and discusses mitigation options for operator front‑running (e.g., blocking newly added tip amounts for a short time or constraining acceptable tip ranges).

### DIFF
--- a/ERCS/erc-minimal-tipping.md
+++ b/ERCS/erc-minimal-tipping.md
@@ -1,8 +1,8 @@
 ---
 eip: TBD
 title: Minimal Tipping Interface
-description: "A standard interface for contracts to accept pre-approved tip amounts."
-author: @nxt3d
+description: A standard interface for contracts to accept pre-approved tip amounts.
+author: Prem Makeig (@nxt3d)
 discussions-to: TBD
 status: Draft
 type: Standards Track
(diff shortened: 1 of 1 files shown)


ERCS/erc-minimal-tipping.md
```diff
@@ -0,0 +1,101 @@
+---
+eip: TBD
+title: Minimal Tipping Interface
```
abcoathup:
```suggestion
title: Accept pre-approved tip amounts
```
Note: I am not an editor.
Could you have a more descriptive title?

ERCS/erc-minimal-tipping.md
```diff
@@ -0,0 +1,101 @@
+---
+eip: TBD
```
abcoathup:
```suggestion
eip: 8157
```

hilljessica8787-dot (approved):

PR #1536, opened 2026-02-12
Add ERC: Primary Agent Registry

nxt3d:
## Summary

Adds a new draft ERC: **Primary Agent Registry**, a singleton registry that maps each Ethereum address to exactly one agent identity (registry contract + token ID). It provides a single, chain-wide source of truth for resolving “this address → which agent?” and works with existing agent registries (e.g. ERC-8004, ERC-8122).

## Motivation

Agents act on-chain via Ethereum addresses, but an address alone doesn’t identify the agent. There’s no standard way to go from address → agent. This ERC defines a canonical registry (one per chain, deterministic address) so wallets, protocols, and indexers can resolve any address to the agent it represents and complete a verification loop with the agent registry.

## What’s in the spec

- **Interface**: `register`, `resolveAgentId`, `agentData`, `isRegistered` (and event).
- **Storage**: Packed `bytes` (registry address + token ID), single-slot for common cases.
- **Access control**: Admin and registrar roles (implementation-agnostic; OZ optional).
- **Registrar types**: Self, ERC-1271, ownable, and role-check registrars for different verification methods.
- **Verification loop**: Forward/reverse resolution and how to verify with ERC-8004 `getAgentWallet` and ERC-8122 `agent_account`.
- **Singleton deployment**: Deterministic address per chain (e.g. CREATE2).

## Status

- **Status**: Draft  
- **Requires**: ERC-1271  
- **Discussion**: https://ethereum-magicians.org/t/erc-xxxx-primary-agent-registry  

No EIP number assigned yet (placeholder XXXX in front matter).


ERCS/_primary_agent_registry.md
```diff
@@ -0,0 +1,232 @@
+---
+eip: XXXX
```
abcoathup:
```suggestion
eip: 8160
```

PR #1566, opened 2026-02-25
Update ERC-7401: Fix duplicate words in ERC-7401, ERC-7813

Princetimix69:
ERC-7401 contains "just before before the sale" and ERC-7813 contains "without without breaking" — both are clear word duplications that make the text harder to read. This removes the extra word in each case, leaving the intended meaning intact.


steven2308 (approved):

SamWilsn (approved):
Looks good to me, but you'll need author approval to merge unless you split this into two PRs.

alvrs (approved):

hilljessica8787-dot (approved):

PR #1576, opened 2026-02-26
Add ERC: Descriptor Attestations for ERC-7730

llbartekll:
## Summary

- Defines **ERC-8176**, an EAS-based attestation format for ERC-7730 clear signing descriptor files
- Canonical EAS schema (`bytes32 descriptorHash`) registered on Ethereum mainnet with UID `0xe023eef113c1670774801c34b377fdf612dd8a4d2fa92fe382e15bd91fafb5c2` ([easscan](https://easscan.org/schema/view/0xe023eef113c1670774801c34b377fdf612dd8a4d2fa92fe382e15bd91fafb5c2))
- Attestations may be created onchain (via EAS `attest()` / `attestByDelegation()`) or offchain (EIP-712 signed JSON, distributed alongside descriptors)
- Verification follows standard EAS rules plus matching `descriptorHash` against the descriptor being verified
- Revocation uses EAS's native primitives (`revoke()` for onchain, `revokeOffchain()` for offchain) — only the original attester can meaningfully revoke
- Wallets fetch attestations from sources they trust and maintain their own trust model for which attesters to accept


ERCS/erc-9999.md
```diff
@@ -0,0 +1,245 @@
+---
+eip: 9999
```
```suggestion
eip: 8176
```

ERCS/erc-8176.md
```diff
@@ -0,0 +1,245 @@
+---
+eip: 9999
+title: Integrity Verification for ERC-7730
+description: Signer attestation and integrity verification for ERC-7730 clear signing descriptor files.
+author: Bartosz Rozwarski (@llbartekll)
+discussions-to: TBD
```
abcoathup: Please create a discussions topic in Eth Magicians with a link to this PR

llbartekll: I think I may not have permissions to do it

abcoathup: What is your username and I can bump your permissions

llbartekll: `llbartekll`

ERCS/erc-8176.md
```diff
@@ -0,0 +1,245 @@
...
+
+## Motivation
+
+[ERC-7730](./eip-7730.md) defines a JSON format that enables wallets to display human-readable transaction details before a user signs. Wallets fetch these descriptor files from registries and rely on them to format transaction data correctly. If an attacker can contribute a misleading descriptor to a registry, the wallet may display deceptive information, undermining the security guarantees that clear signing is designed to provide.
+
+The primary threat motivating this proposal is **registry or repository poisoning.** ERC-7730 descriptors may be contributed to shared registries by multiple parties. A malicious contributor could submit a subtly incorrect descriptor — for example, one that hides a token approval amount — which passes cursory review but misleads end users. Signatures from trusted parties (wallet vendors, auditors) allow consumers to distinguish vetted descriptors from unvetted ones.
```
kuzdogan: Maybe also mention a man-in-the-middle attack or supply chain attacks. Ie. the registry is not compromised but the descriptor is altered on the transport, or you're just unknowingly fetching from a malicious registry.

ERCS/erc-8176.md
```diff
@@ -0,0 +1,245 @@
...
+```json
+{
+  "integrity": {
+    "<CAIP-10 account ID>": {
+      "signerType": "EOA" | "contract",
+      "signature": "0x<hex-encoded signature>"
+    }
+  }
+}
+```
```
kuzdogan: Here I understand it needs to be signed by an Ethereum account? Does it need to be? This also means we are tied to a single encryption type. 

Can we have arbitrary signer keys with a specific signature scheme? I was expecting to see:
- signer public key
- signature
- signature scheme 

like 

```json
  "integrity": [
    {
      "cryptoSuite": "eddsa-rdfc-2022",
      "signerPubKey": "0xaabc..df32",
      "signature": "0x21b3...ad28f"
    },
    ...
    {
      "cryptoSuite": "eddsa-rdfc-2022",
      "signerPubKey": "0x4aec..1e21",
      "signature": "0xab3c...cf212"
    }
  ]
```

ERCS/erc-8176.md
```diff
@@ -0,0 +1,245 @@
...
+
+If the referenced chain is unavailable (e.g., no RPC endpoint configured), the verification result for that entry is "inconclusive".
+
+### Signing Algorithm
+
+To sign a descriptor (producer-side):
+
+1. Compute the integrity hash as defined in [Hash Computation](#hash-computation).
+2. Construct the message string as defined in [Message Format](#message-format).
+3. Sign the message using [EIP-191](./eip-191.md) personal message signing with the signer's private key.
```
kuzdogan: Isn't this only for EOA? How does a contract sign?

kuzdogan (commented):
Looks great overall, thank you! And sorry for the late review

My main doubt is if we should base the signatures on Ethereum/EVM accounts. I lean towards having non-chain-centric signatures with just the public key and the signature scheme. 

I proposed an alternative schema for the integrity field. Let's discuss if that's a good alternative

ERCS/erc-8176.md
llbartekll: I think that direction is valid, but I’d prefer to keep this as simple as possible unless we have a concrete reason to support non-EVM signer types.

ERCS/erc-8176.md
llbartekll: Do you think this ERC should avoid contract verification entirely and rely only on EOA signatures, or keep contract verification and tighten the spec to explain the contract-side flow more explicitly?

kuzdogan: No I just didn't understand how the contract signs, since contracts don't have corresponding private keys

llbartekll: I meant the owner would sign

ERCS/erc-8176.md
llbartekll: it's covered now

manuelwedler (commented):
Looks good to me. Matches what we discussed in the last meeting. Using just the descriptor hash and not the full descriptor in the EAS schema is the right way to go.

Some aspects are too verbose and too AI-sounding for my taste (for example explaining why we do keccak256), but that's nothing blocking IMO.

kuzdogan (commented):
lgtm overall too, thank you

Two things top of my head:
- How does the offchain revokation work? Is that even possible? Didn't we say revokations must be onchain?
- It might be out of scope of this PR but we need to address when and how to revoke Attesters in the case when a key is compromised. Like another attestation contract with a council of attesters that would mark an attester as unauthorative and this needs to be checked before checking a descriptor attestation.

llbartekll:
> * How does the offchain revokation work? Is that even possible? Didn't we say revokations must be onchain?

it's onchain revocation for offchain attestation, see an example:
https://easscan.org/offchain/url/#attestation=eNolkU2KXDEMhO%2Fy1k3Qv%2BRlJj19iZCFLckHCAnk%2BON%2BwRiMXfpcUv284BvZ9bjwbPj3HQn8xR8fk0g%2BKUdy2lPny%2FgJvVCDfrSG32IeVC6jQI2nWOpmle0VW3jEIZXzLHofGmv6QNGRtIV2TDS6IacojIjX2mgB3rRG76Oq6IwDzXdtjrWVt0KjTLZauASEjPp6kL85HxuO7iUqlQ0QxZ%2B09lO8tDHs5SjupHx%2FusKTy%2Bc6ztXOaqscp8HdOKFqTOhRqrPc2%2Bi8D3YcojAW9ZQb0kDcvRE50RzcJQCTZbH7rm1IVTGlaM9BuzmoUVcN3HMvzbv964HuHkOZ4QH3xZ%2Fff%2FuNj2ikNIEzGdlBZxaHtaQsec7UgRwOyycWC%2BLqwlRrhlh9bNZ1ePQ%2FplTFUKc59IQBW8DZJUOOzTMVhTjt2QmzMDCHrabBUfmOI%2Bv69QU2j499

https://etherscan.io/tx/0x164cc763a714e570b7f1d00a31c98ae6078456a4d734b57815ad52ed84db660e

> * It might be out of scope of this PR but we need to address when and how to revoke Attesters in the case when a key is compromised. Like another attestation contract with a council of attesters that would mark an attester as unauthorative and this needs to be checked before checking a descriptor attestation.

yeah, I mentioned that it should be communicated out of bound and it's out of scope for this ERC. But I see a few options here: we could have some governance around EAS based "trusted" attesters.  Signed by "governance multi-sig" or something of that sort. But I think something to discuss out of here.

kuzdogan:
@llbartekll Ah ok, thanks for the example. In that case we can make this a bit more explicit that the revocations are always onchain, on the attestation contract.

SamWilsn:
Friendly reminder to keep technical discussion on the Ethereum Magicians thread.

ERCS/erc-8176.md
```diff
@@ -0,0 +1,254 @@
+---
+eip: 8176
+title: Descriptor Attestations for ERC-7730
+description: EAS-based attestation format for ERC-7730 clear signing descriptor files.
```
SamWilsn, line 4: It'd be ideal if there's some way to expand EAS in the description. You do restate quite a bit of the title in your description, so there might be space within the character limit.

ERCS/erc-8176.md
SamWilsn, line 1: The Ethereum Attestation Service is not defined through the ERC process. Without that foundation, I believe there are too many holes in this specification for someone to be able to implement it without having to look elsewhere for information.

ERCS/erc-8176.md
```diff
@@ -0,0 +1,254 @@
...
+The canonical EAS schema for attestations conforming to this ERC is registered on Ethereum mainnet:
+
+| Field | Value |
+|---|---|
+| Schema | `bytes32 descriptorHash` |
+| Schema UID | `0xe023eef113c1670774801c34b377fdf612dd8a4d2fa92fe382e15bd91fafb5c2` |
+| Resolver | `0x0000000000000000000000000000000000000000` |
+| Revocable | `true` |
+| EAS Contract | `0xA1207F3BBa224E2c9c3c6D5aF63D0eb1582Ce587` |
+| Chain | Ethereum mainnet (`chainId = 1`) |
```
SamWilsn, line 54: This feel like a pile of external references, where the content either lives on mainnet or in the EAS documentation. I'm willing to be convinced otherwise, however.

ERCS/erc-8176.md
```diff
@@ -0,0 +1,254 @@
...
+
+The schema's resolver address is `0x0000...0000`, meaning the schema is permissionless — anyone can attest. Trust filtering happens at the wallet layer (see [Multi-Attester Semantics](#multi-attester-semantics)).
+
+### Descriptor Hash Computation
+
+To compute the `descriptorHash`:
+
+1. Let `D` be the parsed JSON descriptor object.
+2. Resolve any `includes` references in `D` per [ERC-7730](./eip-7730.md) merge rules, recursively until no `includes` key remains. Let `D'` be the resulting fully-resolved descriptor.
+3. Serialize `D'` to a byte string using RFC 8785 (JCS — JSON Canonicalization Scheme).
```
SamWilsn:
```suggestion
3. Serialize `D'` to a byte string using [RFC 8785](https://www.rfc-editor.org/rfc/rfc8785) (JCS — JSON Canonicalization Scheme).
```

ERCS/erc-8176.md
```diff
@@ -0,0 +1,254 @@
...
+4. Compute the Keccak-256 hash of the resulting byte string.
+5. Encode the hash as a `0x`-prefixed lowercase hexadecimal string (66 characters total).
+
+Implementations MUST use Keccak-256 as specified in the Ethereum Yellow Paper (i.e., the pre-standardization variant used throughout the Ethereum ecosystem), NOT the NIST SHA3-256 standard.
```
SamWilsn, line 70: Please link to a specific git hash of the Yellow Paper, as described in EIP-1

ERCS/erc-8176.md
```diff
@@ -0,0 +1,254 @@
...
+eip: 8176
+title: Descriptor Attestations for ERC-7730
+description: EAS-based attestation format for ERC-7730 clear signing descriptor files.
+author: Bartosz Rozwarski (@llbartekll)
+discussions-to: https://ethereum-magicians.org/t/erc-8176-integrity-verification-for-erc-7730/27911
+status: Draft
+type: Standards Track
+category: ERC
+created: 2026-02-26
+requires: 7730
```
SamWilsn:
```suggestion
requires: 712, 1271, 7730
```

ERCS/erc-8176.md
```diff
@@ -0,0 +1,254 @@
...
+| Supply-chain substitution | Attacker compromises a distribution node and serves forged attestations. | Wallets reject attestations from untrusted attesters. |
+| Downgrade (no attestation fetched) | Attacker delivers the descriptor without any attestation, hoping the wallet trusts it unattested. | Wallets that maintain a trusted attester list can detect the absence of expected attestations and act accordingly. Trust policy is a wallet concern. |
+| Stale attestation | Attester has lost operational capability but old attestations continue to circulate. | `expirationTime` bounds exposure. Attesters SHOULD set a bounded TTL. |
+| Replay across descriptors | Attacker copies an attestation signed over one descriptor and presents it alongside a different descriptor. | `descriptorHash` is encoded in the signed `data`. A mismatch between the attestation's `descriptorHash` and the descriptor being verified causes rejection. |
+| Hidden revocation | A distribution source omits notice that the attester has revoked. | Wallets SHOULD query the canonical EAS contract directly for revocation status. |
```
SamWilsn, line 221: Keep requirements (defined with UPPERCASE keywords) in the specification section. They're just too easy to miss if they're elsewhere. Instead, include the requirements in Specification, but discuss their implications here.

PR #1591, opened 2026-03-10
Update ERC-3156: Make ERC-3156 Flash Mint Reference Implementation compliant with its own spec

red-swan:
Add check in `maxFlashLoan` for unsupported tokens

According to ERC-3156:

> The `maxFlashLoan` function MUST return the maximum loan possible for `token`. If a `token` is not currently supported `maxFlashLoan` MUST return 0, instead of reverting.

However in the **Flash Mint Reference Implementation**, the `maxFlashLoan` function returns it's own token's maximum loan amount regardless of the `token` parameter that is provided when called. This PR simply adds a check like in `flashFee` to see in the user's `token` parameter is `this` and to return correctly based on that check.


hilljessica8787-dot (approved):

fubuloubu (approved):
Approved Informational change

SamWilsn (approved):
Seems to make sense. See `flashFee` for an example where the contract restricts supported tokens to just `this`.

PR #1617, opened 2026-03-22
Add ERC: Token Payments

jacopo-eth:
This ERC standardizes payment signaling for token transfers. It introduces a `Payment` event for individual payments and an optional `Order` event for transaction-level context, and conventions for typed metadata payloads.

The standard is designed to work across both native token integrations and settler-based flows, giving applications and indexers a common way to recognize payment intent and associated metadata across various architectures.

### DIFF
--- a/ERCS/erc-8204.md
+++ b/ERCS/erc-8204.md
@@ -1,8 +1,9 @@
 ---
+eip: 8204
 title: Token Payments
 description: Payment signaling for token transfers in commerce applications.
 author: Jacopo Ranalli (@jacopo-eth) <jacopo@slice.so>, Domenico Macellaro (@zerohex-eth) <dom@slice.so>
-discussions-to: 
+discussions-to: https://ethereum-magicians.org/t/erc-8204-token-payments/28046
 status: Draft
 type: Standards Track
 category: ERC
@@ -32,7 +33,8 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 ### Definitions
 
 - **native transfer event:** the canonical token movement event defined by the underlying token standard.
-- **native transfer arguments:** the ordered argument list of the native transfer event, including argument names, types, order, and indexedness.
+- **native transfer arguments:** the ordered argument list of the native transfer event, including argument names, types, and order.
+- **indexed-argument limit:** the maximum number of indexed event arguments permitted for an event. For standard EVM events, this limit is `3`.
 - **buyer:** the intended beneficiary of the goods or services, as asserted by the caller or settler contract.
 - **payer:** the address funding the payment. When the corresponding transfer semantics identify a source address, `payer` refers to that source address.
 - **recipient:** the logical payee represented by a payment integration. It is not necessarily the same as the native transfer `to` address.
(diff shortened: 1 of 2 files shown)


ERCS/erc-8204.md
```diff
@@ -0,0 +1,392 @@
+---
```
abcoathup, line 1:
```suggestion
---
eip: 8204
```

ERCS/erc-xxxx.md
```diff
@@ -0,0 +1,392 @@
+---
+title: Token Payments
+description: Payment signaling for token transfers in commerce applications.
+author: Jacopo Ranalli (@jacopo-eth) <jacopo@slice.so>, Domenico Macellaro (@zerohex-eth) <dom@slice.so>
+discussions-to:
```
```suggestion
discussions-to: https://ethereum-magicians.org/t/erc-8204-token-payments/28046
```

ElcaNels (approved):

ERCS/erc-8204.md
```diff
@@ -0,0 +1,398 @@
...
+eip: 8204
+title: Token Payments
+description: Payment signaling for token transfers in commerce applications.
+author: Jacopo Ranalli (@jacopo-eth) <jacopo@slice.so>, Domenico Macellaro (@zerohex-eth) <dom@slice.so>
+discussions-to: https://ethereum-magicians.org/t/erc-8204-token-payments/28046
+status: Draft
+type: Standards Track
+category: ERC
+created: 2026-03-19
+---
```
rafael-abuawad, line 11: Should add

`requires: 165, 20`

ERCS/erc-8204.md
```diff
@@ -0,0 +1,398 @@
...
+
+```solidity
+event Order(address indexed buyer, address indexed payer, bytes[] data);
+```
+
+`Order` MAY be emitted for metadata that applies to multiple payments in a transaction.
+
+`Payment.data` and `Order.data` MUST use the Metadata Payload Encoding defined below.
+
+### Metadata Payload Encoding
```
rafael-abuawad, line 76: I dont understand the need for this section? Could you elaborate, maybe im missing something

rafael-abuawad (changes requested):
As an initial draft, it looks really good! The definitions are a little too ambiguous at the moment, at least, that was my first impression. This could be really interesting to implement, especially for something more commerce-oriented

ERCS/erc-8204.md
jacopo-eth: ERC-165 should probably be required and mentioned in the spec.

ERC-8204 doesn't depend on ERC-20 though. ERC-20 is just mentioned as a reference implementation, but the spec is defined to be applicable to any token standard.

ERCS/erc-8204.md
jacopo-eth: What is unclear? The spec defines a generic `bytes[] data` field, and that section specifies canonical fields (Memo, MerchantReference, PurchaseReference, URI) along with guidelines for encoding and decoding

ERCS/erc-8204.md
rafael-abuawad: Understood, yeah thinking about it, it does make sense

Tq101510 (approved):

Frank122480 (commented):
Awesome

PR #1623, opened 2026-03-26
Update ERC-7730: Fix inconsistencies in ERC-7730 and its examples

manuelwedler:
While implementing the clear signing library, I came across a few inconsistencies in the standard, mostly in its examples. This PR addresses these inconsistencies.

- Alligns naming of map keywords across examples
- The EIP-712 examples had a whitespace between types. This doesn't follow the `encodeType` function of EIP-712. Also, made the EIP-712 key regex much stricter in regards to whitespaces. 
- Fixes schema and the examples using wrong keywords for the visibility parameters

### DIFF
--- a/ERCS/erc-7730.md
+++ b/ERCS/erc-7730.md
@@ -587,7 +587,7 @@ Sometimes constants are dependant on some part of the context of the transaction
 Each key under the `maps` object is a *map* name. 
 
 A map is a json object with two keys:
-* The `keyType` key contains an non normative indication of the expected type of the map key passed when referencing the map.
+* The `$keyType` key contains an non normative indication of the expected type of the map key passed when referencing the map.
 * The `values` key contains a list of key / value pairs, each key being used to match the data pointed to in the reference `keyPath` key. The corresponding value is used as the context-dependent constant when referenced in the `display` section.
 
 Maps references can be used anywhere a parameter with constant value would be used. A map reference is a json object with two keys:
@@ -605,7 +605,7 @@ An example can be found in [example-maps.json](../assets/eip-7730/example-maps.j
     "metadata": {
         "maps": {
             "underlyingToken": {
-                "keyPath": "@.chainId",
+                "$keyType": "Chain ID",
                 "values" : {
                     "1": "0xaabbccddeeff...",
                     "17000": "0x112233445566..."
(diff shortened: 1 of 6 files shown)


kuzdogan (approved):
(duplicate comment removed)

kuzdogan (approved):
I'm not deeply familiar with this spec and it's fairly complex, so I deferred the review to my AI agent (Claude, claude.ai) under my supervision.

This PR was reviewed by Claude under human supervision.

All 8 changes are correct. Here's what was verified:

- **`keyType` → `$keyType`:** Correct — aligns with the spec's own key naming convention where `$`-prefixed keys are non-normative/internal. The description of `$keyType` as "a non normative indication" matches this.

- **Map inline example restructuring:** The original example incorrectly placed `keyPath` inside the map *definition* object. Per the spec, `keyPath` belongs in the map *reference* (the call site). The fix correctly separates the two: `$keyType` in the definition, `keyPath` in the reference.

- **`deposit(amount)` → `deposit(uint256 amount)`:** The type name was missing from the function signature. The spec requires full human-readable ABI fragments.

- **`$.display.constants` → `$.metadata.constants`:** `display.constants` is not defined in the spec. Constants live under `metadata.constants`. The old path referenced a non-existent location.

- **EIP-712 space removal:** EIP-712's `encodeType` appends referenced types directly with no separator. The space between type definitions was wrong per the EIP-712 spec.

- **EIP-712 regex tightening:** The new regex correctly removes leading whitespace allowance and spaces between type definitions. Tested against known valid type strings — all pass.

- **`mustBe` → `mustMatch` and constraint fix:** This also fixes a silent pre-existing bug. In the original schema, the `ifNotIn` property was already correctly named `ifNotIn`, but the mutual-exclusion constraint referenced "notIn" (a non-existent property), making it vacuously true — the constraint never actually enforced mutual exclusion. The PR fixes the constraint to reference [`ifNotIn`, `mustMatch`] correctly.

- **`key` → `keyPath` in example files:** `key` was used in the example JSON files but `keyPath` is the canonical name defined in the spec.

One minor suggestion: the maps prose now says "A map is a json object with **two keys**: `$keyType` and `values`.". Since `$keyType` is non-normative and presumably optional, "two keys" is slightly misleading — consider clarifying that `values` is required and `$keyType` is optional.

PR #1632, opened 2026-03-28
Add ERC: Agent Assurance

wangbin9953:
This PR introduces a new Draft ERC: Agent Assurance Protocol (AAP).

AAP defines a programmable fulfillment assurance primitive for Agent commerce, building on ERC-8183. It introduces AssuranceAccount, JobAssurance, and Claim primitives with a self-funded collateral model.

Discussion: https://ethereum-magicians.org/t/draft-erc-agent-assurance-protocol/28097

### DIFF
--- a/ERCS/erc-1632.md
+++ b/ERCS/erc-1632.md
@@ -19,7 +19,7 @@ An `AssuranceAccount` is a per-Agent, account-level reserve of self-funded colla
 
 AAP is not a pooled underwriting model. It is a self-funded collateral model in which each Assured Agent locks its own capital as a fulfillment guarantee. Payouts are sourced from the Agent's own AssuranceAccount rather than from third-party liquidity pools.
 
-This specification assumes that each AssuranceAccount uses a single ERC-20 settlement asset and defines only the minimum surface required for interoperability. Collateral pricing, AML screening, and advanced resolution mechanisms are defined as optional extensions.
+This specification assumes that each AssuranceAccount uses a single [ERC-20](./eip-20.md) settlement asset and defines only the minimum surface required for interoperability. Collateral pricing, AML screening, and advanced resolution mechanisms are defined as optional extensions.
 
 ## Motivation
 
@@ -125,9 +125,9 @@ Protection is based on concrete transaction and Job states, making Claims easier
 
 ### Settlement Asset
 
-Implementations MUST specify the ERC-20 settlement asset used for collateral deposits and claim payouts. This specification assumes a single settlement asset per AssuranceAccount. Cross-chain assurance is outside the scope of this version and is reserved for future extensions.
+Implementations MUST specify the [ERC-20](./eip-20.md) settlement asset used for collateral deposits and claim payouts. This specification assumes a single settlement asset per AssuranceAccount. Cross-chain assurance is outside the scope of this version and is reserved for future extensions.
 
-Implementations MUST NOT use ERC-20 tokens with fee-on-transfer or rebasing mechanisms as the settlement asset. Such tokens cause the actual amount received by the contract to differ from the `amount` parameter, breaking the `totalFunded == availableAmount + lockedAmount + paidOutAmount` invariant.
+Implementations MUST NOT use [ERC-20](./eip-20.md) tokens with fee-on-transfer or rebasing mechanisms as the settlement asset. Such tokens cause the actual amount received by the contract to differ from the `amount` parameter, breaking the `totalFunded == availableAmount + lockedAmount + paidOutAmount` invariant.
 
 Implementations SHOULD use integer arithmetic throughout internal fund accounting to avoid rounding errors introduced by division. In cross-chain or Layer 2 environments where the settlement asset's `decimals` may differ from the mainnet, implementations MUST normalize incoming funds to the contract's internal accounting precision upon inflow to prevent dust accumulation from breaking the invariant.
 
@@ -668,7 +668,7 @@ The core specification intentionally does not introduce partial payout or timeou
 
 **Withdrawal race.** An Agent may attempt to withdraw `availableAmount` while Claims are pending to avoid payout obligations. The implementation addresses this by crediting `committedAmount` to `lockedAmount`, ensuring sufficient payout funds. `withdrawAvailableAssurance` MUST NOT allow withdrawal of the `lockedAmount` portion.
 
-**Reentrancy.** `payout()` and `withdrawAvailableAssurance()` transfer ERC-20 tokens to external addresses. Implementations MUST follow the checks-effects-interactions pattern: update all internal accounting state (deducting from `lockedAmount` or `availableAmount`, crediting `paidOutAmount`, transitioning Claim/JobAssurance states) before executing the external `transfer` call.
+**Reentrancy.** `payout()` and `withdrawAvailableAssurance()` transfer [ERC-20](./eip-20.md) tokens to external addresses. Implementations MUST follow the checks-effects-interactions pattern: update all internal accounting state (deducting from `lockedAmount` or `availableAmount`, crediting `paidOutAmount`, transitioning Claim/JobAssurance states) before executing the external `transfer` call.
 
 **Concurrent Claims overcommitment.** If the same AssuranceAccount has multiple active JobAssurances, the sum of all `committedAmount` values may exceed the balance from any single deposit. Implementations MUST synchronously deduct `committedAmount` from `availableAmount` at `commitToJob` time to ensure `lockedAmount` is backed by actual funds.
 
--- a/ERCS/erc-1632.md
+++ b/ERCS/erc-1632.md
@@ -1,5 +1,5 @@
 ---
-eip: XXXX
+eip: 1632
 title: Agent Assurance Protocol
 description: A programmable fulfillment assurance primitive for Agent commerce, supporting on-chain AssuranceAccount, JobAssurance, and Claim lifecycle.
 author: Jacky Wang (@wangbin9953), Henry Chen (@agent-tech)
--- a/ERCS/erc-8210.md
+++ b/ERCS/erc-8210.md
@@ -1,6 +1,6 @@
 ---
-eip: 1632
-title: Agent Assurance Protocol
+eip: 8210
+title: Agent Assurance
 description: A programmable fulfillment assurance primitive for Agent commerce, supporting on-chain AssuranceAccount, JobAssurance, and Claim lifecycle.
 author: Jacky Wang (@wangbin9953), Henry Chen (@agent-tech)
 discussions-to: https://ethereum-magicians.org/t/draft-erc-agent-assurance-protocol/28097
--- a/ERCS/erc-8210.md
+++ b/ERCS/erc-8210.md
@@ -13,7 +13,7 @@ requires: 8183
 
 ## Abstract
 
-This ERC defines a programmable fulfillment assurance primitive for Agent commerce. Building on the Job lifecycle data from [ERC-8183](./eip-8183.md), Agent Assurance Protocol (AAP) introduces three core primitives: `AssuranceAccount`, `JobAssurance`, and `Claim`. AAP MAY optionally integrate with [ERC-8004](./eip-8004.md)-compatible identity and reputation infrastructure but does not depend on that standard.
+This ERC defines a programmable fulfillment assurance primitive for Agent commerce. A programmable fulfillment assurance primitive is an on-chain mechanism that lets an autonomous Agent lock collateral against the outcome of a specific Job, giving counterparties a credible, contract-enforced remedy if the Agent fails to deliver. Building on the Job lifecycle data from [ERC-8183](./eip-8183.md), Agent Assurance Protocol (AAP) introduces three core primitives: `AssuranceAccount`, `JobAssurance`, and `Claim`. AAP may optionally integrate with [ERC-8004](./eip-8004.md)-compatible identity and reputation infrastructure but does not depend on that standard.
 
 An `AssuranceAccount` is a per-Agent, account-level reserve of self-funded collateral. A `JobAssurance` is a commitment allocation made by an Agent against a specific [ERC-8183](./eip-8183.md) Job. Each JobAssurance is associated with at most one Claim. A `Claim` is a payout request initiated by the Beneficiary when coverage conditions are met; upon approval, the payout is debited from the AssuranceAccount.
 
@@ -150,7 +150,7 @@ Implementations MUST support `JobFailure`, `EvaluatorDispute`, and `SettlementDe
 
 `SlashingLoss` is intended for scenarios where an Agent's collateral is simultaneously staked in an external protocol (e.g., a restaking or validation network) and that protocol's slashing rules reduce the Agent's effective backing. Implementations that support `SlashingLoss` SHOULD define how the external slashing event is detected and attested on-chain.
 
-### AssuranceAccount
+### `AssuranceAccount`
 
 Each `AssuranceAccount` corresponds to a single Assured Agent address and maintains its global collateral reserve. An Agent may hold at most one AssuranceAccount.
 
@@ -170,7 +170,7 @@ The invariant `totalFunded == availableAmount + lockedAmount + paidOutAmount` MU
 
 Implementations MAY impose identity requirements on `agent`, for example through an [ERC-8004](./eip-8004.md)-compatible identity system. If identity requirements are imposed, `depositAssurance()` MUST revert when the caller does not satisfy them.
 
-#### AssuranceAccount State Machine
+#### `AssuranceAccount` State Machine
 
 ```solidity
 enum AccountStatus {
@@ -181,15 +181,13 @@ enum AccountStatus {
 
 Valid state transitions:
 
-```text
-(initial) -> Active (Agent calls depositAssurance for the first time)
-Active -> Paused (implementation-defined condition triggered, e.g., AML freeze)
-Paused -> Active (implementation-defined recovery condition met)
-```
+- (initial) → `Active`: Agent calls `depositAssurance` for the first time
+- `Active` → `Paused`: implementation-defined condition triggered, e.g., AML freeze
+- `Paused` → `Active`: implementation-defined recovery condition met
 
 Implementations MUST NOT allow state transitions other than those listed above. All state transitions MUST emit the corresponding event. An `availableAmount` of 0 is a numerical state of the account and does not constitute an independent state transition condition. Implementations SHOULD define a bounded recovery path from `Paused` to `Active` to prevent indefinite account lockout.
 
-### JobAssurance
+### `JobAssurance`
 
 Each `JobAssurance` represents an assurance commitment by an Assured Agent for a specific [ERC-8183](./eip-8183.md) Job. Each JobAssurance covers exactly one CoverageType and is associated with at most one Claim over its entire lifetime.
 
@@ -208,7 +206,7 @@ struct JobAssurance {
 }
 ```
 
-#### JobAssurance State Machine
+#### `JobAssurance` State Machine
 
 ```solidity
 enum AssuranceState {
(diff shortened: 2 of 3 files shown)


ERCS/erc-1632.md
```diff
@@ -0,0 +1,687 @@
+---
+eip: 1632
```
abcoathup:
```suggestion
eip: 8210
```

ERCS/erc-1632.md
```diff
@@ -0,0 +1,687 @@
...
+title: Agent Assurance Protocol
```
abcoathup:
```suggestion
title: Agent Assurance
```
Protocol is superfluous in an ERC title.  Is there a more descriptive title?

Note: I am NOT an editor.

wangbin9953: Thanks, updated.

cmayorga:
Been contributing to the [forum discussion](https://ethereum-magicians.org/t/erc-8210-agent-assurance/28097) on this spec — specifically around the chained workflows gap and the `upstream` field for the companion verification schema that @ThoughtProof is drafting.

A couple of areas where I'd like to contribute directly to this repo:

1. **Test vectors for the claim lifecycle** — specifically the `commitToJob()` → `fileClaim()` → `resolveClaim()` → `payout()` flow, including edge cases like insolvency checks and denial-then-renewal sequencing (Open Question 2 from the spec).

2. **Reference scenarios for multi-hop workflows** — documenting how separate `JobAssurance` commitments compose across agent pipelines, with concrete examples of how the `upstream` verification field (accepted in the forum discussion) enables root-cause attribution without expanding AAP's core scope.

Happy to fork and open a PR scoped to either of these. Which would be most useful to land first?

wangbin9953:
Thanks for offering to contribute directly.
Test vectors first would be the most useful. The spec lists 14 test scenarios but doesn't have concrete implementations yet. Having a third-party author write test vectors also strengthens the case that the spec is implementable by someone other than the original authors, which matters for the Draft → Review transition.
The multi-hop reference scenarios can follow once the core test vectors are solid, since they build on top of the same claim lifecycle flow.
Happy to review PRs against the repo. Let's coordinate on the forum thread for any spec-level questions that come up during implementation.

cmayorga:
Test vectors PR is up: #1647

Covers all 14 scenarios from the Test Cases section as Foundry tests against a minimal IAAP mock. No modifications to the core spec. `forge test -vv` passes 14/14.

Happy to iterate based on review feedback.

cmayorga:
Follow-up to #1647: multi-hop workflow reference scenarios are now up as #1653.

Three composition examples mapped onto the 3-layer architecture you articulated in post #107 of the ERC-8183 thread:

1. Multi-hop dependency tracking via the `upstream` field
2. EvaluatorSlashed → fileClaim (slash event as automatic claim eligibility — agreed with @Bakugo32 from Demsys)
3. Hybrid off-chain scoring with shared `reasoningCID` (inspired by @pablocactus from RNWY)

Self-contained Foundry project under `assets/erc-8210/scenarios/` with its own minimal mocks. Independent of #1647, doesn't require it to merge first. All three tests pass and CI is green.

Happy to align with the Inherited Assumptions section in AAP v2 if useful.

wangbin9953:
Thanks Carlos. Reviewing #1653 in detail now and will leave a longer comment on that PR directly. The 3-layer mapping coming from a third-party contributor is exactly the kind of milestone the spec needed.
For the Inherited Assumptions alignment offer: yes, very much appreciated. Once the v1 spec lands and v2 work begins, the scenarios in #1653 are a natural anchor for the section, assuming the interface alignment we'll discuss shortly. Will coordinate with you before the v2 PR goes up so the references stay accurate.

cmayorga:
## Implementation feedback on the v1 interface from two independent reference implementations

While building two DeFiRe Labs research implementations against the v1 interface — **Lockstep** (non-custodial trading-agent marketplace) and **Catalyst** (autonomous-startup launchpad) — I ran into the same set of small frictions in both. They all trace back to the same root cause, and I think they're worth surfacing here as constructive feedback while v1 is still in Draft.

For context, the multi-hop reference scenarios I upstreamed in #1653 (now using the canonical v1 interface after your review) are a faithful subset of the spec. The frictions described below appeared **on the implementer side**, when wiring real products that consume both ERC-8183 and ERC-8210, not in the scenarios themselves.

**Both products are deployed and publicly verifiable on Base Sepolia:**

- **Lockstep** — 56 tests, deployed on Base Sepolia. Covers Scenarios 1, 2 and 3 of #1653.
  - `LockstepEvaluatorRegistry` → `0xd21B6b55d20Cc3f88A9109db414009290e56f9eE`
  - `CollateralVault` → `0xB255B56bd509A85c27E74bb78d4CF479427880cD`
  - `PerformanceEvaluator` → `0x9c509C0c1146Eeb9dA80B871641D84D98B236EE0`
  - `TradingEscrow` → `0x3e6D79b62ff228FdF62e4286106213Ef3fbd8d0d`
- **Catalyst** — 53 tests, deployed on Base Sepolia. Covers Scenarios 1, 2, 3 plus a Layer 2 off-chain adoption scorer that reuses Scenario 3's `reasoningCID` pattern as a price-manipulation gate.
  - `CatalystEvaluatorRegistry` → `0x81CAfb20788F5B0E0CAb38dc0422e9D8224f812B`
  - `CollateralVault` → `0x68782a318Cd109f349d38429Df530be701BF51e2`
  - `CatalystAdoptionScorer` → `0xeE4331648C8Afff5Ee10c0a2d0DEFcfF4c630544`
  - `PerformanceEvaluator` → `0x44c20d6869eC46B78e76bD9c051518476a9bbdC0`

Both implementations were built independently against the v1 interface over the last couple of days. The convergence of the three findings below across both products is what makes me think they're worth flagging.

---

### Pattern observed: identity-model mismatch between v1 and ERC-8183

In both products, the `CollateralVault` contract ended up declaring its `Claim` struct **locally** rather than reusing `IAAP.Claim` from the v1 spec directly. The reason is the same in both cases and it's not a workaround — it's a structural mismatch:

> The canonical v1 interface addresses claims and scored subjects via `bytes32 claimId` / `address subject`. Both real implementations needed `(uint256 jobId, address claimant)` as a composite key, because the underlying ERC-8183 Job lifecycle in both products uses `uint256` job identifiers throughout.

There's a thin compatibility layer possible (derive `claimId = keccak256(abi.encode(jobId, claimant))` and expose a `getCanonicalClaim` view), but three distinct symptoms of the mismatch showed up during implementation, listed below.

---

### The three findings

**1. `Claim` struct addressing.** Both vaults declare a local `struct Claim { uint256 jobId; address claimant; uint256 amount; bytes32 upstream; bytes32 reasoningCID; bytes32 slashEvidenceHash; uint256 timestamp; }` and key the persistence map by `claims[jobId][claimant]`. The canonical `IAAP.Claim` uses `bytes32`-based addressing instead. Both implementations import `IAAP` but only use it as a documentation anchor — the lint correctly flags it as unused.

**2. `upstream` and `reasoningCID` as first-class fields rather than `bytes evidence`.** The v1 interface carries Scenario 1 and Scenario 3 metadata inside an opaque `bytes evidence` payload. Both implementations independently chose to promote `upstream` and `reasoningCID` to named `bytes32` fields on the `Claim` struct, for two reasons: (a) cheaper read access without ABI decoding, and (b) cleaner indexer events (`event ClaimFiled(uint256 indexed jobId, address indexed investor, uint256 amount, bytes32 upstream, bytes32 reasoningCID, bytes32 slashEvidenceHash)` vs an opaque blob). The patterns from Scenarios 1 and 3 are recurrent enough across our two products that they feel like first-class concerns rather than evidence bytes. I hear the counter-argument — keeping the spec minimal is a real value — but wanted to flag the trade-off based on actual implementation experience.

**3. `OffchainScorer`-style addressing by `address subject`.** Same root cause as #1. In Catalyst's `CatalystAdoptionScorer`, the equivalent contract had to key by `mapping(uint256 jobId => Score)` because that's how startups are identified in the registry. The `ScorePosted` event ends up with `uint256 indexed jobId` instead of `address indexed subject`. Indexers consuming the canonical pattern can't directly listen to this implementation without an adapter, even though the semantics are identical.

---

### Suggestion (constructive, non-breaking)

I'm not proposing to change the core types in v1 — `bytes32 claimId` is a reasonable canonical choice and I don't want to create churn for whatever consumers already exist. Three alternatives that would unblock integer-id implementations without breaking the spec:

1. **Document the integer-id adaptation path explicitly** in the Rationale section: "Implementations using `uint256` job identifiers from ERC-8183 may declare a local `Claim` struct with the same field semantics, computing `claimId = keccak256(abi.encode(jobId, claimant))` if interop with canonical claim-id consumers is needed."
2. **Promote `upstream` and `reasoningCID` to first-class optional fields** in `IAAP.Claim` rather than encoding them inside `bytes evidence`. This change is more invasive but reflects what implementations actually want.
3. **Add a `getCanonicalClaim(bytes32 claimId)` view in a companion `IAAPView` interface** that integer-id implementations can implement to expose canonical-shaped claims to external indexers, without forcing the storage layout to match.

Personally I'd go with option 1 for v1 (cheap, no spec churn) and consider option 2 for v2 once there's more implementation evidence beyond these two products.

---

### Closing

Both products are publicly verified on Base Sepolia (addresses above) and tagged as DeFiRe Labs reference implementations of ERC-8210. Happy to walk through any of the findings in more detail or share test traces if useful. None of them block v1 — the spec works, the scenarios reproduce — but I think folding even just finding #1's documentation note into the spec text would save the next implementer a couple of hours of "wait, do I really have to declare this struct locally?".

wangbin9953:
Carlos, three findings, three responses.

**Finding 1 (Claim struct addressing)**

Root cause: v1's `bytes32 claimId` was chosen for abstraction flexibility, but that abstraction adds friction when the native ERC-8183 identifier is `uint256`. Your Option 1 is the right move. Draft Rationale wording for v2:

> "Implementations integrating with ERC-8183-style Job registries that use `uint256` as the native Job identifier MAY declare local Claim and scorer representations keyed by native integer identifiers (e.g., `(uint256 jobId, address claimant)` for claims, `mapping(uint256 jobId => Score)` for external scorers), deriving the canonical `bytes32 claimId` via `keccak256(abi.encode(jobId, claimant))` and exposing canonical-shape view functions when interoperability with canonical-shape consumers is required."

Not patching v1 PR: a new commit on #1632 would reset the editor review queue, and rushed Rationale wording isn't worth delaying the merge. Goes into v2.

**Finding 2 (`upstream` and `reasoningCID` as first-class fields)**

This sits in tension with the Evidence-First Composability Principle I pushed earlier. My position was keep composition metadata in `bytes evidence` for interface minimalism; your counter from production is that minimalism has real gas and indexability costs. Both hold.

Tracking as an open question through v2. If you know other teams building on ERC-8210, pointing them at this thread for integration feedback would help.

**Finding 3 (OffchainScorer subject addressing)**

Same root cause as Finding 1, and the Rationale draft above already covers the scorer case. No separate entry needed.

**Lockstep and Catalyst**

Addresses noted, will reference both as implementer evidence in the v2 changelog. Spec normative text stays anonymized per ERC convention, Acknowledgments credits you explicitly.

**On continued contribution**

Taken. When the v2 changelog stabilizes (after v1 merges and co-author sync) I'll ping you here. Candidates: Inherited Assumptions, IRiskHook dual-output, IIndependenceSignal, possibly RoleCollusion. Pick what fits.

Appreciate the work that went into both deployments. Shipping two production-shape implementations against a Draft spec is the kind of effort that actually moves it forward.

Jacky

cmayorga:
Jacky, three quick acks plus one note on Finding 2.

**Finding 1.** The Rationale wording you drafted is tighter than what I would have written. Holding for v2 makes sense — no point churning the editor queue.

**Finding 2.** Tracking as open works. The two implementer-side voices most relevant to the `evidence` payload vs first-class fields trade-off are @Demsys (live ERC-8183 implementation, agent-settlement-protocol, integrating the slash-event pattern in this week's redeploy) and @ThoughtProof (verification schema, currently running cycles against Demsys's AgentJobManager). Both already in the Magicians thread, both with firsthand interface composition experience.

**Finding 3.** Agreed.

**On v2.** Of the four candidates, Inherited Assumptions and IRiskHook dual-output are the closest to what Lockstep and Catalyst exercise in production, so that's where I'd bring concrete implementation grounding. RoleCollusion is interesting given Pablo's framing in the Magicians thread, but worth letting their methodology change land first. Whichever fits your sequencing.

Carlos

douglasborthwick-crypto:
The `IRiskHook` extension slot is the right place to host signal inputs that don't belong in the core assurance state machine. Flagging a layer that sits orthogonally to the verification schema ThoughtProof has been developing on the [Magicians thread](https://ethereum-magicians.org/t/draft-erc-agent-assurance-protocol/28097).

**Pre-commitment eligibility** — whether an Agent currently satisfies a set of caller-specified on-chain conditions at a known block — is a distinct verification question from:

- **Post-hoc resolution evidence** (the ThoughtProof verdict/confidence/objections schema proposed for the `resolveClaim` `reason` bytes), which answers *"was this specific claim legitimate at resolution time?"*
- **Behavioral reputation** (RNWY / agent health scoring), which answers *"has this agent behaved reliably over a historical window?"*

Wallet-state attestation answers a third question: *"at block N, does this Agent's wallet satisfy condition C?"* The attestation is a signed boolean verdict, with the condition and the block committed inside the signed bytes, published under a JWKS the Assured Agent does not hold the signing key to — re-runnable by anyone against public chain state.

Natural integration points in AAP as currently specced:

1. **`commitToJob` via `IRiskHook`**: before a `JobAssurance` is created, the hook can query *"does this Agent currently hold at least `amount` of the collateral asset on `chainId`?"* and return a recommended amount based on present on-chain solvency rather than self-reported balance.
2. **Coverage-type precondition gates**: `SlashingLoss` coverage is meaningful only if the Agent is actively staked at commit time; `AMLFreeze` coverage is meaningful only against a compliance-screened wallet. Both are pre-commitment facts, not post-hoc, and compose naturally with `IRiskHook`.

This layer does not replace the verification schema from the Magicians thread or displace reputation scoring — they answer different questions at different lifecycle stages. Pre-commit eligibility, post-hoc evidence, and behavioral reputation compose as three inputs to whatever a Resolver or keeper aggregates. An `IRiskHook` built on all three is stronger than one built on any single layer.

Happy to contribute a concrete hook signature if the extension gets formalized in v2. Reference implementation of the wallet-state attestation primitive across 33 chains, with public JWKS and multi-issuer composition runs, at [insumer-examples#1](https://github.com/douglasborthwick-crypto/insumer-examples/issues/1) — prior art for the hook shape if useful.

wangbin9953:
Douglas, thanks for the PR comment and for pointing me at insumer-examples#1. I've read through the full thread. The three-way decomposition (pre-commitment eligibility / post-hoc resolution evidence / behavioral reputation) is the right framing, and the SlashingLoss and AMLFreeze observation is accurate. Both coverage types carry implicit pre-commitment facts that v1 currently takes on trust rather than verifies.
On the multi-attestation envelope work: four independent issuers (now nine in the SkyeProfile orchestrator trace) converging on a common envelope shape with live JWKS and cross-verified signatures is the most concrete trust-layer composability evidence I've seen in this ecosystem.
This maps cleanly onto where v2 is heading. The current v2 working direction for IRiskHook extends the interface with an optional evidence reference alongside the recommended amount, and the multi-issuer envelope shape your thread has converged on is shaping how v2 thinks about evidence reference composition. The direction your work points to is the direction v2 is moving toward, even if the specific interface details still need to land through co-author and contributor review. The envelope schema itself wouldn't go into the ERC-8210 spec (that's better owned by the envelope spec and its issuer community), but the spec would acknowledge the multi-issuer composition pattern at the interface level. The current working direction is to specify that the evidence reference may point to either a single attestation or a collection from multiple independent issuers, with each issuer's signature independently verifiable, without binding the specific schema. That gives multi-issuer composability a concrete anchor in the spec while keeping the envelope format itself external.
The v2 changelog is currently being drafted. When it's ready I'll post it on the Ethereum Magicians ERC-8210 thread for a review window, and that would be the right moment to bring the multi-attestation envelope perspective into the discussion if you're up for it. For the concrete hook signature offer: I'd rather wait until the IRiskHook extension shape is settled before formalizing a specific signature.
Jacky

SamWilsn:
Friendly reminder to keep discussion about the _content_ of proposals on their Ethereum Magicians threads.

ERCS/erc-8210.md
```diff
@@ -0,0 +1,687 @@
...
+status: Draft
+type: Standards Track
+category: ERC
+created: 2026-03-11
+requires: 8183
+---
+
+## Abstract
+
+This ERC defines a programmable fulfillment assurance primitive for Agent commerce. Building on the Job lifecycle data from [ERC-8183](./eip-8183.md), Agent Assurance Protocol (AAP) introduces three core primitives: `AssuranceAccount`, `JobAssurance`, and `Claim`. AAP MAY optionally integrate with [ERC-8004](./eip-8004.md)-compatible identity and reputation infrastructure but does not depend on that standard.
```
SamWilsn: Avoid creating requirements (with UPPERCASE keywords) outside of the specification section. They're too easy for implementers to miss.

In this particular case, I think you can just use "may" instead of "MAY" since this isn't really a requirement.

SamWilsn: As an outsider to this particular field, I'd appreciate something like:


```suggestion
This ERC defines a programmable fulfillment assurance primitive for Agent commerce. A programmable fulfillment assurance primitive is a... Building on the Job lifecycle data from [ERC-8183](./eip-8183.md), Agent Assurance Protocol (AAP) introduces three core primitives: `AssuranceAccount`, `JobAssurance`, and `Claim`. AAP MAY optionally integrate with [ERC-8004](./eip-8004.md)-compatible identity and reputation infrastructure but does not depend on that standard.
```

ERCS/erc-8210.md
```diff
@@ -0,0 +1,687 @@
...
+
+Valid state transitions:
+
+```text
+Active  -> Claimed  (Beneficiary calls fileClaim)
+Active  -> Released (Assured Agent calls releaseCommitment; Job has completed)
+Active  -> Expired  (block.timestamp > expiry and no Claim filed)
+Claimed -> Paid     (Claim approved and settled)
+Claimed -> Active   (Claim denied; JobAssurance reverts to Active, awaiting subsequent release or expiry)
+```
```
SamWilsn: It's best to reserve code fence blocks for actual code. You could use an SVG diagram, or just plain markdown here instead.

ERCS/erc-8210.md
```diff
@@ -0,0 +1,687 @@
...
+
+Implementations MUST validate eligibility conditions at the time `fileClaim` is called.
+
+AAP integrates with [ERC-8183](./eip-8183.md) through read-only observation; it does not require modifications to [ERC-8183](./eip-8183.md) contracts.
+
+### Optional Extensions
+
+All interfaces below are OPTIONAL. Implementations MAY adopt any combination of these extensions.
+
+#### IRiskHook (Collateral Recommendation)
```
SamWilsn: Code, even in headings, should be in backticks (\`).

ERCS/erc-8210.md
```diff
@@ -0,0 +1,687 @@
...
+
+**AMLFreeze as optional coverage.** AML enforcement is highly dependent on jurisdiction and oracle infrastructure. Making it a core requirement would impose additional burden on implementations targeting permissionless environments.
+
+**Genesis Resolver Set cold-start design.** A decentralized Resolver pool takes time to form. The Genesis Resolver Set is a time-limited transitional arrangement that ensures early adjudication availability while requiring equal staking, challenge, and accuracy rules for all resolvers from day one.
+
+## Backwards Compatibility
+
+This ERC introduces new contracts and interfaces and does not modify the behavior of [ERC-8183](./eip-8183.md), [ERC-8004](./eip-8004.md), or any existing ERC. Whether existing [ERC-8183](./eip-8183.md) Jobs are eligible for Claims is implementation-defined. Implementations SHOULD NOT allow retroactive creation of JobAssurances against already-completed Jobs unless additional verification mechanisms are in place to prevent adverse selection.
```
SamWilsn: Similar comment about UPPERCASE keywords outside of the spec section.

ERCS/erc-8210.md
```diff
@@ -0,0 +1,687 @@
...
+
+AAP integrates with [ERC-8183](./eip-8183.md) through read-only observation and does not require modifications to [ERC-8183](./eip-8183.md) contracts. Integration with [ERC-8004](./eip-8004.md)-compatible systems is accomplished through implementation-defined queries; writing resolver accuracy data back to such systems is an extension concern.
+
+## Test Cases
+
+Test cases will be published alongside the reference implementation. Key scenarios to validate include:
```
SamWilsn: Proposals should be written as if they're already final. It's too easy to forget to update these kinds of notes, and they look really unprofessional in an immutable document. Instead, remove this entire section and replace it with something like `<!-- TODO: add test cases -->`. The linter will ensure you replace the comment before advancing statuses.

wangbin9953:
Thanks for the review. All points addressed:

1. Lowercased RFC keywords outside the Specification section (Abstract, Rationale, Backwards Compatibility, Security Considerations)
2. Added a plain language definition of the assurance primitive to the Abstract
3. Converted the state transition blocks from code fences to plain markdown (including the two other state machines with the same pattern)
4. Backticked code identifiers in headings throughout
5. Replaced the Test Cases and Reference Implementation sections with TODO comments

Also noted on keeping content discussion on the Magicians thread. Future technical exchanges will be directed there.

PR #1638, opened 2026-03-30
Add ERC: Smart Batching

oxshaman:
## Summary

This ERC introduces **Smart Batching** — a composable batch encoding for smart accounts where each parameter declares *how to obtain its value at execution time* (literal, `staticcall`, or balance query) and *what constraints that value must satisfy*, enabling dynamic transaction construction entirely on-chain.

### Key contributions

- **Runtime parameter resolution** — Instead of freezing all calldata at signing time, each `InputParam` specifies a fetcher type (`RAW_BYTES`, `STATIC_CALL`, `BALANCE`) and a routing destination (`TARGET`, `VALUE`, `CALL_DATA`). The EVM constructs calldata from scratch at execution time, eliminating failures caused by stale data (slippage, variable swap outputs, gas shifts).

- **Inline constraints & predicate entries** — Each resolved value can carry `Constraint[]` (`EQ`, `GTE`, `LTE`, `IN`) validated on-chain before routing. An entry with no `TARGET` parameter becomes a pure boolean gate on chain state (a *predicate entry*), enabling execution gating without additional contracts. Relayers simulate batches via `eth_call` and submit when predicates pass.

- **Cross-chain orchestration** — Predicate entries compose with a Merkle-tree authorization structure (defined in a separate ERC) to enable multi-chain flows signed once as a single program. Each chain-local batch is gated by on-chain predicates that observe state changes (e.g., bridged balance arriving), making the mechanism credibly neutral to the interoperability layer (native bridges, ERC-7683, ERC-7786, etc.).

- **Account-standard agnostic** — The standard defines encoding + interface (`IComposableExecution`) + execution semantics. The same encoding works as an ERC-7579 module, ERC-6900 plugin, native account method, or ERC-7702 delegation target. Adapters are thin wrappers over a shared library.

### What's included

| Section | Description |
|---|---|
| Encoding schemes | `ComposableExecution`, `InputParam`, `OutputParam`, `Constraint` structs and enums |
| Core interface | `IComposableExecution.executeComposable()` |
| Execution algorithm | Normative 3-step loop: resolve inputs → execute call → capture outputs |
| Storage contract | Namespaced key-value store for passing return values between steps |
| Adapter guidelines | ERC-7579, ERC-6900, native, ERC-7702 integration patterns |
| Security considerations | Storage isolation, call vs delegatecall context, reentrancy, relayer trust model, runtime value manipulation |

## Motivation

Static batching (ERC-4337 / EIP-5792) forces all parameters to be frozen at signing time. When a swap returns fewer tokens than estimated, gas costs shift, or a bridge delivers with unexpected slippage, the batch reverts. The only workaround today is deploying custom smart contracts for each multi-step flow — expensive, risky, and unscalable.

Smart Batching shifts the paradigm **from transactions to programs**: developers author multi-step, multi-chain flows in TypeScript, compiled to a standard on-chain encoding, signed once, and executed entirely by the EVM. No contract deployment. No per-flow audit cycles.


ERCS/erc-xxxx.md
```diff
@@ -0,0 +1,891 @@
+---
+eip: xxxx
```
lightclient: 8211

ERCS/erc-xxxx.md
```diff
@@ -0,0 +1,891 @@
...
+title: Smart Batching
+description: A smart account batch encoding where each parameter is resolved at execution time from on-chain state, with inline constraints that enable dynamic transactions and predicate-gated cross-chain orchestration.
+author: Mislav Javor (@oxshaman), Filip Dujmušić (@fichiokaku), Filipp Makarov (@filmakarov), Venkatesh Rajendran (@vr16x)
+discussions-to: TBD
```
abcoathup: Please create a discussions topic on Eth Magicians

andrewhong5297: did the discussion get created? I can't find it

andrewhong5297: oh nvm found it https://ethereum-magicians.org/t/erc-8211-smart-batching/28135

ERCS/erc-8211.md
```diff
@@ -0,0 +1,891 @@
...
+    EXEC_RESULT,    // Capture from the return data of the just-executed call
+    STATIC_CALL     // Capture from a separate staticcall (post-execution state read)
+}
+```
+
+**`EXEC_RESULT`** — Captures values directly from the return data of the call that just executed.
+
+```solidity
+// paramData encoding (packed):
+// abi.encode(uint256 returnValueCount, address storageContract, bytes32 storageSlot)
```
ernestognw, line 355: If I'm following correctly, this would be limited to static types, since dynamic return types (bytes, string, arrays) start with an offset pointer at word 0. For example, a function returning `bytes` (`returnValueCount: 1`) would capture the offset pointer 0x20, not the length or content.

Is that intentional?

ERCS/erc-8211.md
```diff
@@ -0,0 +1,891 @@
...
+```solidity
+struct Constraint {
+    ConstraintType constraintType;
+    bytes referenceData;
+}
+
+enum ConstraintType {
+    EQ,     // value == referenceData (as bytes32)
+    GTE,    // value >= referenceData (as bytes32)
+    LTE,    // value <= referenceData (as bytes32)
```
ernestognw, line 460: Have the proposers consider the case where comparison is between signed integers? The spec currently restricts to unsigned integers but I believe some DeFi applications might need signed ones. For example, with `int256`, a negative number like `-1` is `0xffff...ffff`, which is greater than any positive number under unsigned comparison

ERCS/erc-8211.md
```diff
@@ -0,0 +1,891 @@
...
+
+Each input parameter specifies two orthogonal concerns: **where the value goes** (`paramType`) and **how the value is obtained** (`fetcherType`):
+
+```solidity
+struct InputParam {
+    InputParamType paramType;           // Where this value is routed
+    InputParamFetcherType fetcherType;  // How this value is obtained
+    bytes paramData;                    // Fetcher-specific data
+    Constraint[] constraints;           // Conditions the resolved value MUST satisfy
+}
```
ernestognw, line 276: Seems like multiple InputParam constraints are AND-composed (sequential). Also predicate entries in a batch are AND-composed too. Would it be useful to consider having a model where OR is possible. The current workaround would require a smart contract that computes OR logic and returns a bool, then validate it with `EQ`. Does that defeats the purpose of the ERC (leaving constraints in signing time rather than smart contracts?)

ERCS/erc-8211.md
```diff
@@ -0,0 +1,891 @@
...
+       Time passes... constraint satisfied ✓
+       Relayer submits D: composable batch (unwrap → send)
+
+  ─────────────────────────────────────────────────────────────
+
+  Key property: constraints observe STATE, not mechanism.
+  The bridge in step A could be any provider — native bridge,
+  Across, ERC-7683, LayerZero — the constraint doesn't care.
+  It just waits for the balance to appear.
+```
```
ernestognw: I believe Merkle Trees for signature are obscure to the user and not worth the savings.
Maybe you would like to check [my research on ERC-7964](https://eips.ethereum.org/EIPS/eip-7964) about this topic

oxshaman: This is a great observation! In fact - internally we've already migrated to the exact flow you've pointed to with pre-hashed EIP-712 arrays. Try the demo on the website, it uses them even for single chain. I will modify this part to reference your proposal and remove the Merkle Trees!

ERCS/erc-8211.md
oxshaman: We have a utility contract for "unwrapping" custom return types where you can basically query any offset. 
I will add it to the ERC, this is an error by omission - but luckily we already have it.

ERCS/erc-8211.md
ernestognw: Glad to hear it! I tried the demo and it works great. One small observation: the signed structs include a `userOpHash` field which is still opaque, the wallet can't display what's inside it. This is likely a bundler limitation (v0.7 EntryPoint doesn't use EIP-712 structured hashing), since v0.8/v0.9 already support it. Worth noting as a caveat for full wallet transparency once bundler support catches up.

Happy to open a PR against your ERC branch to replace the Merkle tree references with a reference to ERC-7964 if that helps move things along

ernestognw: See https://github.com/oxshaman/ERCs/pull/1

Thanks for considering this!

ERCS/erc-8211.md
```diff
@@ -0,0 +1,895 @@
...
+status: Draft
+type: Standards Track
+category: ERC
+created: 2026-02-11
+requires: 7964
+---
+
+## Abstract
+
+An Ethereum transaction is a single function call on a single contract. ERC-4337 and EIP-5792 (`wallet_sendCalls`) extended this with batch execution — multiple calls under one signature — but every parameter in these batches is static: frozen at signing time, blind to on-chain state at execution. If a swap returns fewer tokens than estimated, gas costs shift, or a bridge delivers with unexpected slippage — the batch reverts. The only workaround is deploying custom smart contracts for each multi-step flow, which introduces new attack surface and demands auditing, testing, and redeployment for every change — a poor security practice and an expensive, time-consuming process.
```
forshtat, line 16: > An Ethereum transaction is a single function call on a single contract. 

This statement may not age well in the context of EIP-8141, may I suggest a better framing, i.e.:
"Today, there are multiple ways to express an operation in Ethereum: legacy single-call Ethereum transactions, ERC-4337 UserOperartions, EIP-5792 batched wallet calls, EIP-8141 multi-frame transactions - but they all share the same issue - every parameter in these batches is static" or something like that?

hilljessica8787-dot (approved):

PR #1648, opened 2026-04-05
Add ERC: Agent NFT Identity Bindings

nxt3d:
## Summary

  Adds a draft ERC for binding ERC-8004 agent identities to external NFT/token contracts.

  The draft standardizes:

  - the reserved metadata key `agent-binding`
  - the binding metadata byte format
  - token standard enum values for ERC-721, ERC-1155, and ERC-6909
  - compact token id encoding with a length byte
  - a minimal verification interface via `bindingOf(uint256 agentId)`

  ## Verification

  Clients verify a binding by:

  1. reading `agent-binding` from the ERC-8004 metadata
  2. decoding `bindingContract`, `tokenStandard`, `tokenContract`, and `tokenId`
  3. calling `bindingOf(agentId)` on `bindingContract`
  4. checking that the returned binding matches the metadata

  Controller semantics are intentionally out of scope.


ERCS/erc-8217.md
```diff
@@ -0,0 +1,241 @@
+---
```
abcoathup, line 1:
```suggestion
---
eip: 8217
```

ERCS/erc-agent-bindings.md
```diff
@@ -0,0 +1,241 @@
+---
+title: Agent NFT Identity Bindings
+description: A standard metadata record for binding ERC-8004 agents to external NFT controllers.
+author: Prem Makeig (@nxt3d)
+discussions-to: https://ethereum-magicians.org/
```
abcoathup: Please create a discussions topic on Eth Magicians

wscdcm:
Great proposal! I've actually deployed ERC-8004 contracts to Sepolia and have some practical insights that might be relevant to ERC-8217's binding design.

**From our deployment experience:**

1. **Identity-agent binding works well as ERC-721**: Our `ERC8004IdentityV2` contract uses NFT-based identity (one address = one Agent), which provides natural transferability. The binding approach in ERC-8217 complements this nicely by enabling cross-contract identity verification.

2. **Metadata key `agent-binding` is a clean design**: Using reserved metadata keys avoids collision with existing ERC-721/1155 metadata. In our implementation, we store capability hashes and endpoint URIs in metadata, and a standardized binding key would make cross-protocol discovery much easier.

3. **Practical concern - binding verification gas cost**: The 4-step verification (read metadata → decode → call `bindingOf` → compare) requires at least 2 contract calls. On L1 this could be expensive. We found that for high-frequency Agent operations, caching binding verification results with TTL (similar to how we handle `trustLevel` in our validation registry) is essential.

4. **Compatibility with reputation/validation registries**: ERC-8004 defines separate Reputation and Validation registries. If ERC-8217 bindings can also encode which reputation/validation contracts an Agent is registered with, it would create a more complete discovery mechanism.

**Deployed contracts for reference:**
- ERC8004IdentityV2: [0xc0D37E6F...](https://sepolia.etherscan.io/address/0xc0D37E6F7B214C92f292FC0534195027CD38AB79#code) (Etherscan verified)
- Discussion: [ethereum/ERCs#1696](https://github.com/ethereum/ERCs/issues/1696)

Would be happy to test ERC-8217 binding compatibility once the spec is more finalized.

nxt3d:
Thanks, also you can see adapter8004.xyz, to see a deployed version. I will make a topic on Ethereum Magicians, and we can continue to discuss.

wscdcm:
Thanks @nxt3d! Just checked out adapter8004.xyz — great to see a live implementation.

Looking forward to the Ethereum Magicians topic. A few thoughts for that discussion:

1. **Cross-implementation compatibility**: Would be valuable to define a test suite that verifies binding compatibility across different ERC-8004 implementations (your adapter vs our V2 contracts). The `bindingOf()` interface is a good start but edge cases (e.g., Agent migration, binding revocation) might need standardization.

2. **Metadata key collision**: As more ERCs build on ERC-8004 (we've seen ERC-8217, ERC-8226, and Self-Sovereign Agent NFTs), having a registry of reserved metadata keys would prevent future conflicts. Similar to how EIP-712 has a domain separator registry.

3. **We're also planning to open a Magicians topic** for our TARS security framework discussion — we can cross-link the threads.

Our deployed demo: https://wscdcm.github.io/erc8004-demo/

wscdcm:
## Interoperability Finding: ERC-8004 Interface Divergence

I've been testing ERC-8217 binding compatibility between our deployment and the official adapter (adapter8004.xyz), and found a significant interface divergence that the community should address.

### The Problem

Our `ERC8004IdentityV2` (deployed on Sepolia at `0xc0D37E6F...`) and nxt3d's official `Adapter8004` use different ERC-8004 identity registry interfaces:

**Official Interface (nxt3d's IERC8004IdentityRegistry):**
```solidity
function register(string memory agentURI, MetadataEntry[] memory metadata) external returns (uint256 agentId);
function setMetadata(uint256 agentId, string memory metadataKey, bytes memory metadataValue) external;
function getMetadata(uint256 agentId, string memory metadataKey) external view returns (bytes memory);
function setAgentWallet(uint256 agentId, address newWallet, uint256 deadline, bytes calldata signature) external;
```

**Our Implementation (ERC8004IdentityV2):**
```solidity
function createIdentity(address owner, string memory name, string memory _metadataURI) external returns (uint256);
function addCapability(uint256 agentId, string memory capability, bytes32 capabilityHash) external;
function updateTrustLevel(uint256 agentId, uint8 newLevel) external;
function setEndpoint(uint256 agentId, string memory endpointType, string memory endpoint) external;
```

Key differences:
1. **No generic metadata API** — Our V2 uses domain-specific methods (addCapability, setEndpoint) instead of `setMetadata(key, value)`
2. **No `agent-binding` key support** — Without `setMetadata()`, we can't write the `agent-binding` metadata that ERC-8217 requires
3. **Registration API mismatch** — `register(uri, metadata[])` vs `createIdentity(owner, name, uri)`

### Impact

ERC-8217's `bindingOf()` verification flow requires reading `agent-binding` from the identity registry's metadata. If implementations don't expose a generic metadata API, the binding is unverifiable through the standard flow.

### Proposal

ERC-8004 should define a **minimal metadata interface** that all compliant identity registries MUST implement:

```solidity
interface IERC8004Metadata {
    function getMetadata(uint256 agentId, string calldata key) external view returns (bytes memory);
    function setMetadata(uint256 agentId, string calldata key, bytes calldata value) external;
}
```

This would ensure ERC-8217 bindings (and future extensions) work across all ERC-8004 implementations, regardless of their domain-specific APIs.

This is especially important as more ERCs build on top of ERC-8004 (we've counted at least 3: ERC-8217, ERC-8226, and the Self-Sovereign Agent NFTs proposal).

Our deployed contracts: [Etherscan verified on Sepolia](https://sepolia.etherscan.io/address/0xc0D37E6F7B214C92f292FC0534195027CD38AB79#code)
Demo: https://wscdcm.github.io/erc8004-demo/

nxt3d:
Thanks for running this interop test and publishing the write-up. A few clarifications so readers of the PR have an accurate picture.

The current ERC-8004 draft already specifies the on-chain key-value metadata surface that `Adapter8004` relies on. The "Onchain metadata" section mandates:

```solidity
function getMetadata(uint256 agentId, string metadataKey) external view returns (bytes memory);
function setMetadata(uint256 agentId, string metadataKey, bytes metadataValue) external;
```

and the Registration section specifies the `register(string agentURI, MetadataEntry[] metadata)` overload and the reserved `agentWallet` key with `setAgentWallet` / `getAgentWallet` / `unsetAgentWallet`. The `IERC8004IdentityRegistry` the Adapter imports is built directly against those signatures, and that is why `ERC-8217` `bindingOf()` verification assumes the same surface: it reads `agent-binding` via `getMetadata(agentId, "agent-binding")`.

Looking at the verified source for `ERC8004IdentityV2` on Sepolia, the contract exposes `createIdentity`, `updateMetadata(tokenId, metadataURI)`, `addCapability`, `updateTrustLevel`, and `setVault`, but does not expose `setMetadata(uint256, string, bytes)`, `getMetadata(uint256, string)`, `setAgentWallet`, `unsetAgentWallet`, or `getAgentWallet`. Metadata is held as a single URI string. That is a valid design, but it is a different specification from the current ERC-8004 draft, which means an ERC-8217 verifier cannot read `agent-binding` from it. The issue is not that the Adapter expects an unusual interface; it is that your `V2` does not implement the ERC-8004 onchain metadata surface at all.

If you want `V2` to interoperate with `bindingOf()` today, adding the two metadata functions (and ideally the `register(string, MetadataEntry[])` overload plus the `agentWallet` functions) is enough. The Adapter will work against any registry that exposes them, no changes required on our side.

ERCS/erc-8217.md
```diff
@@ -0,0 +1,206 @@
...
+=> 0x9c4e8f2a1b7d6e3c0a5f8d2b9e1c4a7f3d6e8b0c
+```
+
+For the same `bindingContract`, `bindingOf(agentId)` might return for example ERC-721 with `tokenId = 0`, ERC-1155 with `tokenId = 5`, or ERC-721 with `tokenId = 0x1234`; those values live only in the `Binding` struct from the binding contract, not in `agent-binding` metadata.
+
+## Security Considerations
+
+Clients MUST NOT assume that decoding `agent-binding` alone is sufficient to determine the current controller of an agent. The metadata reveals only which binding contract to use; the bound token and control semantics come from `bindingOf` and remain implementation-specific.
+
+The trust for this system lies in the contract code of the binding contract. The result of the `bindingOf` function is only as secure and verifiable as the security of the binding contract itself. Clients SHOULD assess that contract (for example audits, reputation, and upgrade risk) before relying on its return values.
```
SamWilsn, line 186: > Clients SHOULD assess [...]

Keep requirements (defined with UPPERCASE keywords) in the specification section, or else they're too easy to miss. You should still discuss their implications in this section though.

SamWilsn (approved):
Looks good for a draft, just move those requirements around before going into review.

OronukpoAra:
Fantastic and fantabulous

PR #1647, opened 2026-04-05
ERC-8210: Add test vectors for Agent Assurance Protocol

cmayorga:
## Summary

Concrete Foundry test vector implementations for the 14 test scenarios defined in ERC-8210's Test Cases section. Written by a third-party contributor to validate the spec is independently implementable — a requirement for the Draft → Review transition.

## What's included

Located in `assets/erc-8210/`:

- `contracts/interfaces/IAAP.sol` — Core interface, enums, structs, events extracted from the spec
- `contracts/interfaces/IERC20.sol` — Minimal ERC-20 interface for settlement asset
- `contracts/mocks/AAPMock.sol` — Minimal IAAP implementation with ERC-8183 job state simulation
- `contracts/mocks/MockERC20.sol` — Minimal ERC-20 mock
- `test/ERC8210TestVectors.t.sol` — 14 test functions, one per spec scenario

## Scenarios covered

| # | Test | Scenario |
|---|------|----------|
| 1 | `test_01_FullHappyPath_AssurancePath` | Deposit → commit → job completes → release |
| 2 | `test_02_FullHappyPath_ClaimsPath` | Deposit → commit → job fails → claim → approve → payout |
| 3 | `test_03_DenialPath` | Claim denied, assurance reverts to Active, re-filing blocked |
| 4 | `test_04_ExpiryPath` | Time exceeds expiry → committedAmount returned |
| 5 | `test_05_PrematureExpiry` | expireCommitment before expiry reverts |
| 6 | `test_06_Insolvency` | Approved claim but payout reverts (insufficient locked) |
| 7 | `test_07_WithdrawalBoundary` | Withdraw up to available; touching locked reverts |
| 8 | `test_08_InvalidStateTransitions` | Claim on non-Active assurance reverts; payout on non-Approved reverts |
| 9 | `test_09_EligibilityGating` | fileClaim reverts if job hasn't reached required terminal state |
| 10 | `test_10_AdverseSelection` | commitToJob reverts if coverage conditions already qualify |
| 11 | `test_11_DuplicateCommitment` | Duplicate (jobId, coverageType) while Active/Claimed reverts |
| 12 | `test_12_AmountValidation` | Zero requestedAmount and zero approvedAmount revert |
| 13 | `test_13_RecusalRuleEnforcement` | Evaluator cannot resolve their own EvaluatorDispute claim |
| 14 | `test_14_InvariantVerification` | totalFunded == available + locked + paidOut after every operation |

## How to run

```bash
cd assets/erc-8210
[...]

PR #1653, opened 2026-04-08
ERC-8210: Add multi-hop workflow reference scenarios

cmayorga:
## Summary

Three reference scenarios demonstrating how ERC-8210 (Agent Assurance Protocol) composes with ERC-8183 and external behavior oracles, mapped onto the **3-layer architecture** articulated by @wangbin9953 in the [ERC-8183 forum thread](https://ethereum-magicians.org/t/erc-8183-agentic-commerce/27902).

This is a follow-up to [PR #1647](https://github.com/ethereum/ERCs/pull/1647) (test vectors), but is **fully self-contained** — it has its own minimal mocks and Foundry setup, and does not depend on the test vectors PR being merged first.

## Architectural Context

The scenarios are organized around the 3-layer architecture for autonomous agent safety:

| Layer | Name | Enforced by |
|-------|------|-------------|
| **1** | **Structure** | ERC-8183 core (auto-assignment, role separation, state machines) |
| **2** | **Behavior** | Risk hooks, evaluator registries, off-chain scoring oracles |
| **3** | **Recovery** | ERC-8210 / AAP (post-hoc redress when layers 1 and 2 don't catch the attack) |

The three scenarios are concrete examples of how Layer 3 (recovery) composes with the other two layers without coupling.

## Scenarios

| # | Scenario | Layers | Key pattern |
|---|----------|--------|-------------|
| 1 | Multi-Hop Dependency Tracking | 3 (+ 1, 2 context) | `upstream` field traces root cause across A→B→C→D pipeline |
| 2 | EvaluatorSlashed → fileClaim | 2 → 3 | Slash event serves as automatic proof for AAP claim — no re-adjudication |
| 3 | Hybrid Off-chain Scoring | 1 + 2 + 3 | Same `reasoningCID` consumed by task rejection and claim filing |

## What's included

Located in `assets/erc-8210/scenarios/`:

- `README.md` — Architectural context and scenario index
- `contracts/interfaces/` — Minimal IAAP and IERC20 interfaces (subset used by scenarios)
- `contracts/mocks/`:
  - `AAPMockMinimal.sol` — AAP mock with just `fileClaim` + `reviewClaim`
  - `MockERC20.sol` — Settlement asset
  - `EvaluatorRegistryMock.sol` — Slash events (Scenario 2). Event signature agreed with @Bakugo32 (Demsys) in the forum thread
  - `OffchainScorerMock.sol` — AHS-style scoring (Scenario 3). Inspired by @pablocactus (RNWY) AHS concept
  - `ChainedJobsMock.sol` — Job pipeline A→B→C→D (Scenario 1)
- `docs/` — Detailed `.md` documentation for each scenario (problem, actors, flow, layer mapping)
- `test/` — Three Foundry end-to-end tests, one per scenario

## How to run

```bash
cd assets/erc-8210/scenarios
forge install foundry-rs/forge-std --no-commit
forge test -vv
```

All three tests pass: `test_MultiHopDependencyTracking`, `test_EvaluatorSlashToClaim`, `test_HybridOffchainScoring`.

## Related

- Spec PR: #1632
- Test vectors PR: #1647
- Forum discussion: https://ethereum-magicians.org/t/erc-8210-agent-assurance/28097
- ERC-8183 thread (composition discussion): https://ethereum-magicians.org/t/erc-8183-agentic-commerce/27902

## Notes

- No modifications to the spec (`ERCS/erc-8210.md`)
- No modifications to the files in PR #1647
- All new content lives under `assets/erc-8210/scenarios/`
- Self-contained Foundry project with its own minimal mocks

### DIFF
--- a/assets/erc-8210/scenarios/README.md
+++ b/assets/erc-8210/scenarios/README.md
@@ -18,12 +18,12 @@ author of ERC-8210, in post #107 of the ERC-8183 Ethereum Magicians thread):
 | **2** | **Behavior** | Evaluates behavioral independence and risk through hooks and oracles (`IRiskHook`, off-chain scorers) | Risk hooks, evaluator registries, scoring oracles |
 | **3** | **Recovery** | Provides post-hoc redress when Layers 1 and 2 do not catch the attack in time | ERC-8210 / AAP |
 
-The three scenarios below demonstrate how Layer 3 (recovery) **composes** with
-the other two layers through concrete, tested patterns. All three scenarios use
+The scenarios below demonstrate how Layer 3 (recovery) **composes** with
+the other two layers through concrete, tested patterns. All scenarios use
 the canonical ERC-8210 v1 interface, encoding their composition metadata
-(upstream references, slash record references, off-chain scoring CIDs) into
-the opaque `bytes calldata evidence` payload that AAP's `fileClaim` already
-defines.
+(upstream references, slash record references, off-chain scoring CIDs,
+solvency snapshots) into the opaque `bytes calldata evidence` payload
+that AAP's `fileClaim` already defines.
 
 ## Scenarios
 
@@ -32,6 +32,7 @@ defines.
 | 1 | [Multi-Hop Dependency Tracking](docs/scenario-1-multi-hop-dependency.md) | 3 (+ 1, 2 context) | Upstream reference encoded in `evidence` traces root cause across A→B→C→D pipeline |
 | 2 | [EvaluatorSlashed → fileClaim](docs/scenario-2-evaluator-slash-claim.md) | 2 → 3 | Slash event serves as automatic proof for AAP claim — no re-adjudication |
 | 3 | [Hybrid Off-chain Scoring](docs/scenario-3-hybrid-offchain-scoring.md) | 1 + 2 + 3 | Same `reasoningCID` consumed by task rejection and claim filing |
+| 4 | [Solvency-Aware Claim Assessment](docs/scenario-4-solvency-aware-claim.md) | 2 → 3 | `EvaluatorSlashed` + `EvaluatorStakeUpdated` in the same tx enable stateless solvency assessment |
 
 ## Project Structure
 
@@ -49,16 +50,19 @@ scenarios/
 │       ├── AAPMockMinimal.sol                 # Minimal AAP implementation (deposit, commit, file, resolve, payout)
 │       ├── MockERC20.sol                      # ERC-20 with mint
 │       ├── EvaluatorRegistryMock.sol          # Slash events (Scenario 2)
+│       ├── EvaluatorRegistryWithStake.sol     # Slash + stake tracking (Scenario 4)
 │       ├── OffchainScorerMock.sol             # AHS-style scoring (Scenario 3)
 │       └── ChainedJobsMock.sol                # Job pipeline A→B→C→D (Scenario 1)
 ├── docs/
 │   ├── scenario-1-multi-hop-dependency.md
 │   ├── scenario-2-evaluator-slash-claim.md
-│   └── scenario-3-hybrid-offchain-scoring.md
+│   ├── scenario-3-hybrid-offchain-scoring.md
+│   └── scenario-4-solvency-aware-claim.md
 └── test/
     ├── Scenario1_MultiHopDependency.t.sol
     ├── Scenario2_EvaluatorSlashClaim.t.sol
-    └── Scenario3_HybridOffchainScoring.t.sol
+    ├── Scenario3_HybridOffchainScoring.t.sol
+    └── Scenario4_SolvencyAwareClaimAssessment.t.sol
 ```
 
 ## Build & Test
@@ -79,6 +83,8 @@ forge test -vv
 [PASS] test_MultiHopDependencyTracking()
 [PASS] test_EvaluatorSlashToClaim()
 [PASS] test_HybridOffchainScoring()
+[PASS] test_SolvencyAwareClaimAssessment()
+[PASS] test_ZeroStakePostSlash_FullWipeout()
 ```
 
 ## Spec Reference
--- a/assets/erc-8210/scenarios/README.md
+++ b/assets/erc-8210/scenarios/README.md
@@ -15,17 +15,21 @@ author of ERC-8210, in post #107 of the ERC-8183 Ethereum Magicians thread):
 | Layer | Name | What it does | Enforced by |
 |-------|------|-------------|-------------|
 | **1** | **Structure** | Enforces what contracts can verify directly: auto-assignment, role separation, state machines | ERC-8183 core |
-| **2** | **Behavior** | Evaluates behavioral independence and risk through hooks and oracles (`IRiskHook`, off-chain scorers) | Risk hooks, evaluator registries, AHS oracles |
+| **2** | **Behavior** | Evaluates behavioral independence and risk through hooks and oracles (`IRiskHook`, off-chain scorers) | Risk hooks, evaluator registries, scoring oracles |
 | **3** | **Recovery** | Provides post-hoc redress when Layers 1 and 2 do not catch the attack in time | ERC-8210 / AAP |
 
 The three scenarios below demonstrate how Layer 3 (recovery) **composes** with
-the other two layers through concrete, tested patterns.
+the other two layers through concrete, tested patterns. All three scenarios use
+the canonical ERC-8210 v1 interface, encoding their composition metadata
+(upstream references, slash record references, off-chain scoring CIDs) into
+the opaque `bytes calldata evidence` payload that AAP's `fileClaim` already
+defines.
 
 ## Scenarios
 
 | # | Scenario | Layers | Key pattern |
 |---|----------|--------|-------------|
-| 1 | [Multi-Hop Dependency Tracking](docs/scenario-1-multi-hop-dependency.md) | 3 (+ 1, 2 context) | `upstream` field traces root cause across A→B→C→D pipeline |
+| 1 | [Multi-Hop Dependency Tracking](docs/scenario-1-multi-hop-dependency.md) | 3 (+ 1, 2 context) | Upstream reference encoded in `evidence` traces root cause across A→B→C→D pipeline |
 | 2 | [EvaluatorSlashed → fileClaim](docs/scenario-2-evaluator-slash-claim.md) | 2 → 3 | Slash event serves as automatic proof for AAP claim — no re-adjudication |
 | 3 | [Hybrid Off-chain Scoring](docs/scenario-3-hybrid-offchain-scoring.md) | 1 + 2 + 3 | Same `reasoningCID` consumed by task rejection and claim filing |
 
@@ -39,10 +43,10 @@ scenarios/
 ├── .gitignore
 ├── contracts/
 │   ├── interfaces/
-│   │   ├── IAAP.sol                           # Minimal AAP interface
+│   │   ├── IAAP.sol                           # Minimal subset of ERC-8210 v1 interface
 │   │   └── IERC20.sol                         # Minimal ERC-20 interface
 │   └── mocks/
-│       ├── AAPMockMinimal.sol                 # AAP — just fileClaim + reviewClaim
+│       ├── AAPMockMinimal.sol                 # Minimal AAP implementation (deposit, commit, file, resolve, payout)
 │       ├── MockERC20.sol                      # ERC-20 with mint
 │       ├── EvaluatorRegistryMock.sol          # Slash events (Scenario 2)
 │       ├── OffchainScorerMock.sol             # AHS-style scoring (Scenario 3)
--- a/assets/erc-8210/scenarios/README.md
+++ b/assets/erc-8210/scenarios/README.md
@@ -1,0 +1,83 @@
+# ERC-8210 — Reference Scenarios for Multi-Hop Workflows
+
+> **Disclaimer:** These scenarios are independent of the test vectors in
+> [PR #1647](https://github.com/ethereum/ERCs/pull/1647) and provide their own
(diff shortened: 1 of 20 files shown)


wangbin9953:
Carlos, this is a really substantive contribution and I want to start by saying so clearly. The 3-layer mapping coming from a third-party contributor as runnable Foundry code is a milestone the spec needed. Each of the three scenarios picks up exactly the right design thread:

- **Scenario 1 (multi-hop)**: tracing causal chains across pipelines without requiring AAP to model dependency graphs at the protocol level is the right boundary. The core spec stays 1:1 Job-scoped, the verification layer carries the provenance.
- **Scenario 2 (slash to claim)**: treating an on-chain slash event as automatic eligibility evidence, with no second adjudication, is going to become a default pattern for any AAP implementation that integrates with a staked evaluator registry. The fact that you coordinated the event signature with @Bakugo32 directly is exactly the kind of cross-team work the spec stack needs.
- **Scenario 3 (hybrid scoring)**: the shared `reasoningCID` between commit-time rejection and claim filing is the cleanest expression of the evidence-first composability principle I've seen articulated. Same evidence object serves prevention and recovery, no re-attestation.

The architectural framing is right and the scenario designs are right. I want to make sure that's on the record before I get into the mechanical part.

---

**The mechanical part: IAAP interface alignment with v1 spec**

In doing a detailed read of the mock contracts, I realized the `IAAP.sol` interface in this PR has drifted from the canonical ERC-8210 v1 spec submitted in #1632. I think this happened naturally because the scenarios are demonstrating concepts that haven't all landed in v1 yet (the `upstream` field in particular, and the `Verdict` enum as a way to express resolution state), so the mock interface evolved to carry those concepts directly.

The drift is small in line count but meaningful in semantic terms. Specifically:

| Area | Current PR mock | ERC-8210 v1 spec |
|---|---|---|
| `fileClaim` signature | `(uint256 amount, bytes32 evidenceHash, bytes32 upstream) returns uint256` | `(bytes32 assuranceId, uint256 requestedAmount, bytes calldata evidence) returns bytes32` |
| Resolution function | `reviewClaim(uint256 claimId, Verdict verdict, bytes32 reason)` | `resolveClaim(bytes32 claimId, bool approved, uint256 approvedAmount, bytes calldata reason)` |
| `claimId` type | `uint256` | `bytes32` |
| `ClaimState` enum | `Filed / UnderReview / Approved / Rejected / Appealed / Resolved` | `None / Filed / Challenged / Approved / Denied / Paid` |
| `Claim.upstream` field | Present | Not present |
| `Verdict` enum | Present | Not present (uses `bool approved`) |
| `JobAssurance` concept | Not modeled | Core to spec, claims must be filed against a JobAssurance |
| `AssuranceAccount` lifecycle | Not modeled | Required before any commitToJob |

The risk of merging the current mock as-is is twofold:

1. Future implementers reading this PR as a reference would build against the mock interface and produce implementations incompatible with the canonical spec
2. If the v2 spec eventually references this PR (which I'd very much like to do), it would be referencing an interface that doesn't match its own normative text

**The good news is that all three scenarios can be expressed faithfully through the v1 interface without losing any of their intent.** The key insight is that the `upstream` reference, the slash record reference, and the `reasoningCID` are all forms of composition metadata that can be carried through the opaque `bytes calldata evidence` parameter that v1 already defines. The composition pattern lives in the evidence payload, not in the interface signature.

---

**Proposed fix**

I went through the three scenarios and rewrote `IAAP.sol` and `AAPMockMinimal.sol` to be a faithful subset of v1 spec, then updated each scenario test to use the new interface. The architectural intent of every scenario is preserved:

- Scenario 1: `upstream` reference is now encoded into the evidence payload as `abi.encode("pipeline_failure", jobD, upstreamRef)`. The `traceToRoot` logic and the assertion that jobC is the root cause are unchanged.
- Scenario 2: the slash record hash is encoded into the evidence payload as `abi.encode("EvaluatorSlashed", registryAddress, jobId, slashHash)`. The resolver still re-derives the slash hash from the registry to verify authenticity, which is the core invariant of the scenario.
- Scenario 3: the `reasoningCID` is encoded as `abi.encode("OffchainScore", scorerAddress, suspectAgent, REASONING_CID)`. The shared-evidence invariant between Layer 2 task rejection and Layer 3 claim filing is preserved.

The three external mocks (`ChainedJobsMock`, `EvaluatorRegistryMock`, `OffchainScorerMock`) are left untouched, since they represent the 8183-side / behavior-layer infrastructure and have nothing to do with the AAP interface itself.

All three tests pass against the rewritten mock with `forge test -vv`:

```
[PASS] test_MultiHopDependencyTracking
[PASS] test_EvaluatorSlashToClaim
[PASS] test_HybridOffchainScoring

3 tests passed, 0 failed
```

I've put the rewritten files in a public gist: https://gist.github.com/wangbin9953/18e14074ba26531fd3a23fec4ff5a393

The 5 files that need to be replaced are:

1. `assets/erc-8210/scenarios/contracts/interfaces/IAAP.sol`
2. `assets/erc-8210/scenarios/contracts/mocks/AAPMockMinimal.sol`
3. `assets/erc-8210/scenarios/test/Scenario1_MultiHopDependency.t.sol`
4. `assets/erc-8210/scenarios/test/Scenario2_EvaluatorSlashClaim.t.sol`
5. `assets/erc-8210/scenarios/test/Scenario3_HybridOffchainScoring.t.sol`

Happy to push these as a commit to your branch directly if that's easier (just let me know and I'll arrange it). Or if you'd prefer to apply them yourself, the gist files are ready to drop in.

---

**On v2 alignment**

Once these updates land and v1 is merged, this PR becomes the canonical reference for cross-layer composition in the AAP v2 Inherited Assumptions section. The design discussions around `upstream` (which connects to ThoughtProof's verification schema work) and around evidence-first composability (which connects to Pablo's RNWY scoring direction) are exactly the kind of forward-looking patterns the spec rationale should anchor on.

A few things that came out of this PR are also going to feed back into the v2 changelog:

- The "spec extension via assets" pattern (putting reference scenarios under `assets/` without touching normative text) is something I want to formalize as a contribution model for future scenarios
- The slash-event-as-eligibility pattern from Scenario 2 will likely show up in v2 as a recommended integration approach when AAP is composed with stake-weighted evaluator registries
- The shared-evidence pattern from Scenario 3 directly motivates the evidence-first composability principle that's already in the v2 changelog

Thanks again Carlos. The architectural work in this PR is what made me realize how important the spec-side alignment is, it's much easier to get the interface right early when there's a concrete implementation to test against.

wangbin9953:
Quick follow-up on the above, on an unrelated attribution point I just learned about:

The `OffchainScorerMock.sol` file comment currently reads "Inspired by @pablocactus (RNWY) AHS concept". P.A. Lopez (the RNWY team lead, @no-one on Ethereum Magicians) reached out to clarify that RNWY and the AHS project are two separate efforts. They happen to overlap conceptually on the multidimensional scoring idea, and the name collision (both principals go by Pablo) made the conflation easy to fall into. I also made the same mistake in my ERC-8210 forum post and have already posted a correction there.

When you apply the IAAP interface updates from the gist, it would be worth also updating the OffchainScorerMock header comment to separate the two attributions, something like:

```
/// @notice Inspired by the conceptual AHS (Agent Health Score) approach.
///         RNWY is a separate project with related but distinct multidimensional
///         scoring (soulbound identity, explicit sybil signal weights, live
///         ERC-8183 hook integration) and will be referenced separately in the
///         AAP v2 IRiskHook section once their methodology document is published.
```

Sorry for the extra churn. Attribution accuracy matters more than it looks at first, especially when multiple independent projects are working in overlapping conceptual space.

cmayorga:
Thanks @wangbin9953 — applied the gist updates as you proposed. All five files (`IAAP.sol`, `AAPMockMinimal.sol`, and the three scenario tests) now use the canonical v1 interface, with composition metadata flowing through the opaque `evidence` payload. `forge test -vv` confirms the three scenarios still pass and CI is green on the new commit.

I also took your second comment into account and updated the `OffchainScorerMock.sol` header with the wording you suggested separating the AHS conceptual approach from the RNWY project. The corresponding scenario doc has been updated to match.

Two additional housekeeping changes in the same commit:

- **Demsys reference:** updated `EvaluatorRegistryMock.sol` and the scenario 2 doc to credit @Demsys (the GitHub handle for Bakugo32) and link the [agent-settlement-protocol](https://github.com/Demsys/agent-settlement-protocol) reference implementation. He confirmed the handle on the forum thread.
- **`foundry.lock`:** added to `.gitignore` (was being created locally by `forge install` but shouldn't be tracked).

Happy to iterate further if anything else needs adjusting before this becomes the canonical reference for v2. :)

wangbin9953:
Thanks so much for pushing all of this through so quickly. This commit fully addresses both review threads:
✅ IAAP interface alignment: the new IAAP.sol and AAPMockMinimal.sol are now a faithful subset of the ERC-8210 v1 spec. All enums, structs, function signatures, and events match the canonical spec. The three scenario tests are updated to the new interface and the full architectural intent of every scenario is preserved.
✅ Attribution fixes: the scenario-2 Demsys reference is now upgraded to the project level (@Demsys + agent-settlement-protocol), confirmed against Bakugo32's forum reply. The handling of AHS / RNWY in OffchainScorerMock and the README is also nicely done. Changing the README's Behavior layer description from "AHS oracles" to "scoring oracles" is a smart wording choice that avoids picking favorites at the architectural level.
✅ The new README paragraph about encoding composition metadata into the evidence payload is exactly the design principle this PR is meant to articulate. Putting it in the README makes the intent visible to anyone reading the scenarios in the future.
From my side, this PR is fully ready. Once CI clears it should be in good shape for editor review.
Thanks again Carlos. This kind of fast back-and-forth is exactly what spec work needs.

cmayorga:
Thank you @wangbin9953, this means a lot. The depth of the review and the speed of the back-and-forth is what made it worth doing properly, being able to converge on the right interface in one round rather than guessing across several is exactly the difference between contributing to a spec and just adding noise around it.

I'm personally invested in seeing this stack land cleanly, because I have two projects currently in beta that are built directly on top of ERC-8210 + ERC-8183. They're not ready for a public introduction yet, but I'll share both in the forum once they reach a state where the community can poke at them and the feedback would be useful in both directions. The work in this PR (especially the evidence-first composability principle and the 3-layer framing) is what's making them tractable to build.

Looking forward to v2 and to whatever Inherited Assumptions / IRiskHook material comes next. Happy to keep contributing scenario-level work if that helps anchor any of it in runnable code.

Demsys:
Hi everyone, thanks @cmayorga for the reference and the updated attribution. The EvaluatorSlashed pattern in Scenario 2 maps cleanly to what we have deployed. We'll comment here once the re-draw pattern with EvaluatorAssignmentFailed is live so the implementation stays current with what AAP v2 will reference.

cmayorga:
Pushed a small fix to `EvaluatorRegistryMock.buildEvidenceHash` adding a `timestamp == 0` guard. Without it, the function returns `keccak256(abi.encode(...zeros...))` for non-existent jobIds, which is a deterministic non-zero value that can collide with a "valid" hash of all-zero fields. The guard lets consumers treat a non-zero return value as proof that a slash actually occurred for the given jobId. Caught while building reference implementations against the mocks.

Separately, posted broader implementer feedback on the v1 spec interface in #1632, with concrete addresses on Base Sepolia for two DeFiRe Labs research implementations (Lockstep and Catalyst) that consume this PR. Nothing in that feedback affects the scenarios in this PR — it's about the canonical interface in the spec itself.

Demsys:
As promised, the bounded re-draw and updated EvaluatorSlashed are now live on Base Sepolia (2026-04-10).

`event EvaluatorSlashed(address indexed evaluator, uint256 indexed jobId, uint256 amount, uint256 remainingStake, bytes32 reason);`

`assignEvaluator()` enforces evaluator ≠ provider, evaluator ≠ client — max 5 attempts before `EvaluatorAssignmentFailed(uint256 jobId, address provider, address client)`.

New contracts : 
`AgentJobManager` `0xB8C41C289AA2D55b7A8ae53003F212AcABEcc597`
`EvaluatorRegistry` `0x454911f476493dcB34273C9c22Ded2CeCec0Dd2c`
→ https://github.com/Demsys/agent-settlement-protocol

Demsys:
@cmayorga redeployment done. EvaluatorSlashed signature is unchanged — same selector. New EvaluatorRegistry address on Base Sepolia : 0x4F4aa58A715B6a5357da0EA067C405803f489BD1

SamWilsn:
Friendly reminder to keep discussion on PRs limited to editorial topics. Keep technical stuff on the Ethereum Magicians thread.

assets/erc-8210/scenarios/README.md
```diff
@@ -0,0 +1,93 @@
...
+|---|----------|--------|-------------|
+| 1 | [Multi-Hop Dependency Tracking](docs/scenario-1-multi-hop-dependency.md) | 3 (+ 1, 2 context) | Upstream reference encoded in `evidence` traces root cause across A→B→C→D pipeline |
+| 2 | [EvaluatorSlashed → fileClaim](docs/scenario-2-evaluator-slash-claim.md) | 2 → 3 | Slash event serves as automatic proof for AAP claim — no re-adjudication |
+| 3 | [Hybrid Off-chain Scoring](docs/scenario-3-hybrid-offchain-scoring.md) | 1 + 2 + 3 | Same `reasoningCID` consumed by task rejection and claim filing |
+| 4 | [Solvency-Aware Claim Assessment](docs/scenario-4-solvency-aware-claim.md) | 2 → 3 | `EvaluatorSlashed` + `EvaluatorStakeUpdated` in the same tx enable stateless solvency assessment |
+
+## Project Structure
+
+```
+scenarios/
```
SamWilsn, line 40: It'd be best to use hyperlinks here. Jekyll doesn't generate directory listings, so there's no way to navigate to these files otherwise.

SamWilsn (commented):
This'll need to be rebased on master once the proposal itself is merged.

PR #1659, opened 2026-04-09
Update ERC-8048: Rename `getMetadata` to `metadata` in ERC-8048

rafael-abuawad:
This PR proposes two key improvements to align the proposal with established Ethereum standards and optimize gas efficiency.

### 1. Standardize Getter Naming Convention
Renamed the metadata retrieval function from `getMetadata()` to `metadata()`. 

**Rationale:** To maintain consistency with established standards like **ERC-20**, **ERC-721**, and **ERC-1155**, which do not use the `get` prefix for view functions (e.g., `name()`, `decimals()`, `symbol()`).

### 2. Event Optimization
Refactored the `MetadataSet` event to reduce redundancy and gas costs.

**Current:**
`event MetadataSet(uint256 indexed tokenId, string key, string indexed indexedKey, bytes value);`

**Updated:**
`event MetadataSet(uint256 indexed tokenId, string indexed key, bytes value);`

**Rationale:** Having duplicate indexed keys in a single event is redundant and increases gas consumption during execution without providing additional utility for off-chain indexers.

### DIFF
--- a/ERCS/erc-8048.md
+++ b/ERCS/erc-8048.md
@@ -2,7 +2,7 @@
 eip: 8048
 title: Onchain Metadata for Token Registries
 description: A key-value store interface that allows registries to store and retrieve arbitrary bytes as metadata directly onchain.
-author: Prem Makeig (@nxt3d), Rafael Abuawad (@rabuawad_)
+author: Prem Makeig (@nxt3d), Rafael Abuawad (@rafael-abuawad)
 discussions-to: https://ethereum-magicians.org/t/erc-8048-onchain-metadata-for-multi-token-and-nft-registries/25820
 status: Draft
 type: Standards Track
--- a/ERCS/erc-8048.md
+++ b/ERCS/erc-8048.md
@@ -43,7 +43,7 @@ interface IOnchainMetadata {
 
 - `metadata(tokenId, key)`: Returns the metadata value for the given token ID and key as bytes
 
-Contracts implementing this ERC MAY also expose a `event MetadataSet(uint256 indexed tokenId, string indexed indexedKey, string key, bytes value);` function to allow metadata updates, with write policy determined by the contract.
+Contracts implementing this ERC MAY also expose a `MetadataSet(uint256 indexed tokenId, string indexed indexedKey, string key, bytes value);` function to allow metadata updates, with write policy determined by the contract.
 
 Contracts implementing this ERC MUST emit the following event when metadata is set:
 
--- a/ERCS/erc-8048.md
+++ b/ERCS/erc-8048.md
@@ -66,7 +66,7 @@ The Diamond Storage pattern provides predictable storage locations for data, whi
 
 ### Examples
 
-The inspiration for this standard was trustless AI agents. The registry extends [ERC-721](./eip-721.md) by adding `metadata` and `setMetadata` functions for optional extra on-chain agent metadata.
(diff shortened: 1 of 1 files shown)


nxt3d:
Having key and indexedKey together actually serves a purpose, because indexedKey is useful for indexing known keys, while key is useful for unknown keys?

rafael-abuawad:
> Having key and indexedKey together actually serves a purpose, because indexedKey is useful for indexing known keys, while key is useful for unknown keys?

Shouldn't all keys be known? I've been playing with the EIP, and, going a step further, its useful to have a list of keys inside of the smart contract implementing it, or something similar. How do you envision the emit of the MetadataSet event? @nxt3d 

I found no usecase to having two topics in an event that point to the key, or is one of them expected to be empty on "X" scenario?

nxt3d:
> Shouldn't all keys be known?

This standard was originally created for ERC-8004, and in that case there are a lot of new metadata field ideas used by various agent types.

The idea of declaring the metadata keys being used was included in ENSIP-24. It could also be used for ERC-8048, but also not sure.

https://docs.ens.domains/ensip/24/

@rafael-abuawad

> I found no usecase to having two topics in an event that point to the key, or is one of them expected to be empty on "X" scenario?

I think the use case is that if you are searching for particular keys, you can just use the indexedKey, and if you want to index all the keys, including keys that you do not know, you can use the key. Also not sure. I have seen this pattern before however.

rafael-abuawad:
@nxt3d I understand, lets keep this interface minimal, restoring the original event:
`event MetadataSet(uint256 indexed tokenId, string key, string indexed indexedKey, bytes value);`
In the thing im working on I didn't find a use-case, but you are right, ERC8004 has a lot of metadata fields, some of them would be unknown compared to the ones that the smart contract developer knowns

What do you think about the first point: Renamed the metadata retrieval function from `getMetadata()` to `metadata()`.

Rationale: To maintain consistency with established standards like ERC-20, ERC-721, and ERC-1155, which do not use the get- prefix for view functions (e.g., `name()`, `decimals()`, `symbol()`).

nxt3d:
> What do you think about the first point: Renamed the metadata retrieval function from `getMetadata()` to `metadata()`.

Two reasons why we might want to keep it the way it is: currently, ERC-8004 uses getMetadata(). ERC-8004 does not "depend" on ERC-8048 now, but in the future, once ERC-8048 is finalized, we could propose it.

Also, it is likely that the single word "metadata" might already be used in a number of projects, but I am not sure.

I am generally positive about changing it because I do think it reads better and is more consistent with existing extensions.

What do you think?

rafael-abuawad:
> What do you think?

I'm strongly in favor of for the `metadata()` function without the get- as is more aligned with other EIPs, also we should add EIP165 with the correct contract ID

I'll add those changes

rafael-abuawad:
Done, added. Calculated the contract ID using [cast](https://www.getfoundry.sh/cast): `cast sig "metadata(uint256,string)"`

nxt3d:
> I'm strongly in favor of for the `metadata()` function without the get- as is more aligned with other EIPs, also we should add EIP165 with the correct contract ID

Should we have more in the interface section, to better specify ERC-165?

rafael-abuawad:
> Should we have more in the interface section, to better specify ERC-165?

I think thats enough looking at other EIPs is done a extremely minimal way

nxt3d (approved):
LGTM!

PR #1661, opened 2026-04-10
Update ERC-8049: Rename `getContractMetadata` to `contractMetadata` in ERC-8049

rafael-abuawad:
This PR proposes two key improvements to align the proposal with established Ethereum standards and optimize compatability.

### 1. Standardize Getter Naming Convention
Renamed the metadata retrieval function from `getContractMetadata()` to `contractMetadata()`. 

**Rationale:** To maintain consistency with established standards like **ERC-20**, **ERC-721**, and **ERC-1155**, which do not use the `get` prefix for view functions (e.g., `name()`, `decimals()`, `symbol()`). Something really similar can be seen in the **ERC-6909** with the `contractURI()`

### 2. Implement ERC-165
Implements the **ERC-165** contract ID, the signature was calculated using [cast](https://www.getfoundry.sh/reference/cast/cast): `cast sig "contractMetadata(string)"`

### DIFF
--- a/ERCS/erc-8049.md
+++ b/ERCS/erc-8049.md
@@ -2,12 +2,13 @@
 eip: 8049
 title: Contract-Level Onchain Metadata
 description: Onchain metadata for arbitrary contracts with optional Diamond Storage for predictable data storage locations.
-author: Prem Makeig (@nxt3d)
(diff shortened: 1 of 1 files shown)


nxt3d (approved):
LGTM! Thx

PR #1701, opened 2026-04-22
Add ERC: Restricted Token with Restriction Codes

rya-sge:
# ERC-1404: Restricted Token with Restriction Codes — modernized draft

## What this PR does

Brings the **ERC-1404 (Simple Restricted Token)** standard up to date and into the
canonical `ERCS/` format, with a reference implementation and tests under
`assets/erc-1404/`.

Rename the ERC to _Restricted Token with Restriction Codes_

## The concept, in short

ERC-1404 lets a token signal *why* a transfer is (or would be) blocked, in a
machine-readable way, so issuers can enforce compliance rules (KYC/AML, lock-ups,
whitelists, jurisdiction limits) on-chain while exchanges and wallets can predict and
explain the outcome off-chain.

It adds just **two functions** to a token:

- `detectTransferRestriction(from, to, value) → uint8` — returns `0` if the transfer is
  allowed, or an issuer-defined restriction code if not. Third parties can call it as a
  view to pre-check a transfer.
- `messageForTransferRestriction(uint8) → string` — maps a code to a human-readable
  reason for UIs.

The core guarantee: **a transfer is rejected if and only if `detectTransferRestriction`
returns a non-zero code** — so the reported code always matches what the token will
actually do.

It also adds an **optional, spender-aware extension** for delegated transfers:

- `detectTransferRestrictionFrom(spender, from, to, value) → uint8` — same restriction
  code space and same `messageForTransferRestriction` lookup, but adds the `spender`
  dimension so a policy can restrict the party *initiating* a `transferFrom` (e.g. a
  frozen or un-KYC'd operator) — something the three-parameter base method cannot see.

The extension is OPTIONAL: it does **not** change the mandatory interface id `0xab84a5c8`
(the two base methods) and is advertised under a separate id `0x78a8de7d` (all three
methods). Because `transferFrom` is a distinct delegated-transfer entry point, the
`spender == from` case SHOULD be evaluated through the spender-aware path: it equals
`detectTransferRestriction(from, to, value)` when the policy does not restrict the
initiating operator, and MAY differ when it does (an implementation MAY collapse the two
only when they are observably equivalent).

## Main differences from the original ERC-1404

The original draft (2018) defined ERC-1404 strictly as `contract ERC1404 is ERC20`. This
revision keeps the original intent and function semantics but modernizes and generalizes
the spec:

| Aspect | Original (2018) | This revision |
|---|---|---|
| **Token coupling** | Hard `requires: 20`; defined as `contract ERC1404 is ERC20` | `requires: 20` removed. ERC-20 is the canonical/primary case, but the interface applies to any token moving a single `uint256` between two addresses (e.g. [ERC-777](./ERCS/erc-777.md)), and can be implemented by a standalone compliance contract that is *not* a token |
| **Enforcement requirement** | `detectTransferRestriction` **MUST be called inside** `transfer`/`transferFrom` | Outcome-based: enforcement **MUST be consistent** with `detectTransferRestriction` (reject ⟺ non-zero code). Calling it directly is RECOMMENDED; re-evaluating equivalent conditions — incl. reverting with typed errors such as ERC-7943 — is allowed if the result matches |
| **Base-interface preservation** | Implicit via `contract ERC1404 is ERC20` (full ERC-20 inheritance) | Token-agnostic: an ERC-1404 token **SHOULD** preserve the function signatures and events of its underlying token interface (e.g. ERC-20) and **SHOULD NOT** remove or alter them — expressed as a SHOULD rather than mandating a specific base standard or requiring "semantics unchanged" (restriction logic changes only whether a transfer succeeds, not the interface) |
| **Introduction / framing** | "Simple Summary" citing "relevant US regulatory bodies"; US-centric framing | Dropped the "Simple Summary"; neutralized jurisdiction-specific wording so the standard reads as generally applicable rather than US-specific |
| **Solidity** | `public`, `returns (string)`, pre-0.8 | `external view`, `returns (string memory)`, `^0.8.x` signatures |
| **Interface discovery** | None | Optional [ERC-165](./ERCS/erc-165.md) support; mandatory interface id `0xab84a5c8`, optional spender-aware extension id `0x78a8de7d` |
| **Spender-aware detection** | Not addressed | Optional `detectTransferRestrictionFrom(spender, from, to, value)` extension to predict/enforce delegated `transferFrom` outcomes on a per-operator basis; separate ERC-165 id. The `spender == from` case SHOULD go through the spender-aware path — it equals the base method when the policy does not restrict the initiating operator, and MAY differ when it does |
| **Mint / burn** | Not addressed | MAY apply analogous restriction checks (optional, non-normative). Where such checks are applied and the extension is exposed, the `spender` parameter of `detectTransferRestrictionFrom` MAY carry the operator initiating the mint/burn, so the policy can restrict that operator directly |
| **Reference impl & tests** | Linked to external GitHub repos | Self-contained Foundry + OpenZeppelin v5 implementation and test suite under `assets/erc-1404/` |
| **Security & test cases** | Minimal | Added Security Considerations and table-driven Test Cases, incl. the reporting-vs-enforcement consistency invariant |
| **Determinism** | Implied | Restriction checks SHOULD be deterministic, so reporting and enforcement stay consistent |
| **Data-source guidance** | None | Documented as **non-normative** Security Considerations (no new conformance bar vs. the original): the `view` result is a prediction and on-chain enforcement at execution time is authoritative; issuer policies should avoid inputs a counterparty can manipulate within the transfer transaction (e.g. a flash-loanable AMM spot price) |

### What is unchanged

- The two function names, parameters, and the meaning of code `0` ("no restriction").
- The intent: standardize *where/whether* a transfer is restricted and *why*, leaving the
  policy itself entirely up to the issuer.
- Interface compatibility with ERC-20 — the standard still only *adds* methods.

## Status

Draft. Author list extended to include the contributor maintaining this revision.


assets/eip-1404/src/ERC1404.sol
```diff
@@ -0,0 +1,134 @@
...
+        override
+        returns (uint8)
+    {
+        if (!whitelist[from]) return SENDER_NOT_WHITELISTED;
+        if (!whitelist[to]) return RECIPIENT_NOT_WHITELISTED;
+        return TRANSFER_OK;
+    }
+
+    /// @notice Returns the human-readable message for a restriction code.
+    function messageForTransferRestriction(uint8 restrictionCode) public pure override returns (string memory) {
```
ivan-atme: It was originally 'view', not 'pure' in https://github.com/ethereum/EIPs/issues/1404#issue-359329364

rya-sge: Thank you for your comment. 
Correct, the original reference used view. In this implementation the function only returns constant values and does not access state, so pure is better than view and more accurately reflects its behavior.
In Solidity, an implementation can be more restrictive than the interface regarding state mutability.

Frank122480 (commented):
Ok

ERCS/erc-1404.pdf
SamWilsn, line 1: PDFs, if required, should go in your assets directory.

assets/erc-1404/.gitmodules
SamWilsn, line 1: Please don't add submodules.

assets/erc-1404/doc/aderyn-feedback.md
SamWilsn, line 1: I'm not sure this adds anything to your proposal.

assets/erc-1404/doc/aderyn-report.md
SamWilsn, line 1: Same here. Not really useful for an ERC.

assets/erc-1404/doc/slither-feedback.md
SamWilsn, line 1: Again, not necessary.

ERCS/erc-1404.md
```diff
@@ -0,0 +1,413 @@
+---
+eip: 1404
+title: Simple Restricted Token
```
SamWilsn, line 3: "Simple" is a bit subjective, like "minimal". It's best to try and pick something objective and unique about your proposal when naming it, but this isn't strictly required. Just a suggestion!

ERCS/erc-1404.md
```diff
@@ -0,0 +1,413 @@
...
+discussions-to: https://ethereum-magicians.org/t/erc-1404-simple-restricted-token-standard/1405
+status: Draft
+type: Standards Track
+category: ERC
+created: 2018-07-27
+---
+
+## Abstract
+
+Current token standards have provided the community with a platform on which to develop a decentralized economy that is focused on building Ethereum applications for the real world. As these applications mature and face consumer adoption, they begin to interface with corporate governance requirements as well as regulations. They must not only be able to meet corporate and regulatory requirements but must also be able to integrate with technology platforms underpinning their associated businesses. What follows is a simple and extendable standard that seeks to ease the burden of integration for wallets, exchanges, and issuers.
```
SamWilsn, line 15: Your abstract is missing a description of what the proposal actually does. This section should give the reader a high-level (but still technical) overview of both the problem the proposal solves and how it solves it.

SamWilsn (commented):
We'll need to grab approval from any one of: Ron Gierlach <@rongierlach>, James Poole <@pooleja>, Mason Borda <@masonicGIT>, Lawson Baker <@lwsnbaker>. If they don't reply in a reasonable amount of time, you'll need to request a new EIP number.

ERCS/erc-1404.md
rya-sge: Thanks for the review! I have kept the original name. Since the ERC is pretty old (2018), I would prefer to keep the original name. But if a new ERC name is required because original authors don't answer, changing the name could be more relevant indeed.

assets/erc-1404/doc/slither-feedback.md
rya-sge: Done, file has been removed

assets/erc-1404/doc/aderyn-report.md
rya-sge: Done, file has been removed

assets/erc-1404/.gitmodules
rya-sge: Done, file has been removed

ERCS/erc-1404.pdf
rya-sge: Done, file has been removed

ERCS/erc-1404.md
rya-sge: Thanks for the feedback, abstract section has been improved. Let me know if it is good or not now

ERCS/erc-1404.md
rya-sge: Finally changing the name for _Restricted Token with Restriction Codes_

PR #1714, opened 2026-04-30
Update ERC-1185: editorial fixes + review feedback

xinbenlv:
# Review: ERC-1185 — Storage of DNS Records in ENS

**Status under review:** Review (advancing toward Final)

This PR contains **editorial fixes only** (typos, RFC links, RFC 2119 boilerplate, backwards-compat clarification). The substantive review feedback is left as **line comments** on the diff so it can be discussed and incorporated by the author rather than landed unilaterally.

This PR is intentionally kept as **Draft** until the substantive items below are triaged.

## Summary

ERC-1185 is conceptually sound and aligned with ERC-137 resolver patterns, but the **specification is incomplete relative to the reference implementation**, lacks **RFC 2119 normative language** (now added in this PR), and underspecifies **events**, **`setZonehash`/`zonehash`**, and **error/return semantics**. The Security Considerations section is too thin given the complexity of DNS wire format parsing on-chain.

## Editorial fixes included in this PR

- Add RFC 2119 keywords boilerplate to **Specification**.
- Fix typo "is as follows" → "are as follows" (`clearDNSZone` args).
- Hyperlink **RFC 1035** and **RFC 4592** references.
- Expand **Backwards Compatibility** with a one-sentence rationale.
- Fix typo "degenenrates" → "degenerates" in **Security Considerations**.
- Trim trailing whitespace after front-matter.

## Substantive items (deferred to line comments)

These are **not** changed by this PR; each is anchored as a line comment:

### Blockers
1. **Spec/impl name mismatch**: `dnsRecords` (spec) vs `dnsRecord` (impl).
2. **`setZonehash` / `zonehash` missing from Specification** — present in reference impl with a separate interface ID.
3. **Events unspecified**: `DNSRecordChanged`, `DNSRecordDeleted`, `DNSZoneCleared`, `DNSZonehashChanged`.
4. **RRset contiguity rule** is "should" but is actually a correctness MUST; deletion-signal edge cases.
5. **Both ERC-165 interface IDs** (`0xa8fa5682`, `0x5c47637c`) need to be declared normatively.

### Spec clarifications needed
- `bytes32 name` parameter hashing rule (keccak256 of wire-format-encoded name).
- "returns nothing" → empty `bytes`.
- `hasDNSRecords` is resource-type-agnostic — make explicit.
- `clearDNSZone` does **not** clear zonehashes — document or change.

### Reference implementation
- `pragma solidity ^0.7.4;` is end-of-life.
- Dead `nameHash` assignments in `setDNSRecords`; inconsistent hashing form.
- Deletion-signal edge case (empty-first-record drops the whole RRset).

### Security Considerations
Expand to cover wire-format parsing, gas-DoS bounds, no-DNSSEC, version-counter storage non-recovery, zonehash content trust, and wildcard semantics.

## Recommendation

Do not advance from **Review** to **Last Call** until the blockers are addressed. The editorial fixes here can land independently if the author prefers.

---

*This PR is Draft on purpose — flip to "Ready for review" once the line comments have been triaged.*


ERCS/erc-1185.md
```diff
@@ -21,6 +21,8 @@ ENS is a highly desirable store for DNS information.  It provides the distribute
 
 ## Specification
 
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119.
```
xinbenlv, line 24: Added RFC 2119 boilerplate. Once added, please audit the spec for lower-case "should"/"must" usages and promote to normative keywords where appropriate (notably L41 — see review body item B5).

ERCS/erc-1185.md
```diff
@@ -57,15 +59,15 @@ The arguments for the function are as follows:
 
   - node: the namehash of the fully-qualified domain in ENS for which to set the records.  Namehashes are defined in [ERC-137](./eip-137.md)
   - name: the `keccak256()` hash of the name of the record in DNS wire format.
-  - resource: the resource record ID.  Resource record IDs are defined in RFC1035 and subsequent RFCs.
+  - resource: the resource record ID.  Resource record IDs are defined in [RFC 1035](https://www.rfc-editor.org/rfc/rfc1035) and subsequent RFCs.
```
xinbenlv, line 62: Hyperlinked RFC 1035. Note that the `name` parameter on L61 still needs its hashing rule clarified — see review body item under "Spec clarifications / L61".

ERCS/erc-1185.md
```diff
@@ -80,7 +82,7 @@ DNS is a federated system of naming, and the higher-level entities control avail
 
 ## Backwards Compatibility
 
-Not applicable.
+Not applicable. This document defines a new resolver profile and does not modify any existing profile or interface.
```
xinbenlv, line 85: Expanded with a one-sentence rationale. Optional: also note that this profile is independently detectable via ERC-165 with interface IDs `0xa8fa5682` (records) and `0x5c47637c` (zonehash) — useful for callers and helps justify "Not applicable". (Also see review body item B4 — both IDs need to be normative in the Specification.)

ERCS/erc-1185.md
```diff
@@ -265,7 +267,7 @@ abstract contract DNSResolver is ResolverBase {
 
 ## Security Considerations
 
-Security of this solution would be dependent on security of the records within the ENS domain.  This degenenrates to the security of the key(s) which have authority over that domain.
+Security of this solution would be dependent on security of the records within the ENS domain.  This degenerates to the security of the key(s) which have authority over that domain.
```
xinbenlv, line 270: Fixed typo "degenenrates" → "degenerates". The section still needs substantial expansion — see the "Security Considerations" section in the review body for the six items I'd suggest adding.

xinbenlv (commented):
# Draft review for ERC-1185

This PR contains editorial fixes only. The substantive review items are below — most anchor on lines that are **not** part of this diff, so GitHub won't let me leave them as inline review comments. They are listed here with line refs to the post-edit file (`ERCS/erc-1185.md`). The handful of items that do fall on changed lines are also left as inline comments.

---

## Blockers

### B1 — `dnsRecords` vs `dnsRecord` (L54 spec ↔ L190 impl)
Spec defines `dnsRecords(bytes32,bytes32,uint16)` (plural) with selector `0x2461e851`. Reference impl declares `dnsRecord(...)` (singular). The selectors will not match. Pick one and align spec + impl. Plural reads better given the function returns concatenated wire-format RRset bytes.

### B2 — `setZonehash` / `zonehash` missing from Specification (L218–231 impl)
The spec body says "two functions to set DNS information and two functions to query DNS information" — but the reference impl exposes four setters/queries plus this zonehash pair. Add a normative subsection for `setZonehash(bytes32,bytes)` and `zonehash(bytes32) view returns (bytes)`, including selectors, the relationship to ERC-1577 contenthash format, and link to the second interface ID. Also add `1577` to the front-matter `requires:` field.

### B3 — Events unspecified (L104–111 impl)
`DNSRecordChanged`, `DNSRecordDeleted`, `DNSZoneCleared`, `DNSZonehashChanged` appear in the reference impl but never in the Specification body. Indexers and DNS bridges depend on these. Each event needs a normative entry: signature, indexed fields, and emission rules. In particular, when `setDNSRecords` is called with empty rdata to delete, does it emit `DNSRecordDeleted` (impl behavior) or `DNSRecordChanged` with empty bytes? Lock that down.

### B4 — Both ERC-165 interface IDs missing from spec (L100–101 impl)
`DNS_RECORD_INTERFACE_ID = 0xa8fa5682` and `DNS_ZONE_INTERFACE_ID = 0x5c47637c` are present in the reference impl but not in the spec. Without them, ERC-165 detection across resolvers is non-portable. List each ID and which functions it covers.

### B5 — RRset contiguity rule should be MUST (L41)
"should be contiguous ... if not then the later RRsets will overwrite the earlier one(s)" is a correctness rule, not a recommendation. Promote to MUST. Also clarify the deletion signal: the reference impl detects deletion via `value.length == 0` of the **first** record in the run only — so an RRset where only the first record has empty rdata will be treated as a full deletion. Either document this exactly, or change semantics to "deletion iff *all* records in the run have empty rdata". I recommend revert-on-non-contiguous rather than silent overwrite.

---

## Spec clarifications

### L61 — `bytes32 name` parameter hashing rule
Clarify: `name` here is `keccak256(<dns-wire-format-encoded-name>)` — the same hashing applied in the impl as `keccak256(abi.encodePacked(name))` where `name` came from `iter.name()` (length-prefixed labels in wire format). A reader cannot derive this rule from current text without reading the implementation.

### L64 — Return value when no records
"will return nothing" is ambiguous in Solidity ABI. Specify: returns an empty `bytes` (length 0). Apply the same clarification to `zonehash(node)` once added.

### L68 — `hasDNSRecords` is resource-type-agnostic
State explicitly that `hasDNSRecords` returns true iff *any* resource type has records for `(node, name)`. The impl backs this via `nameEntriesCount[node][version][name] != 0`.

### L207 — `clearDNSZone` does not clear zonehashes
`clearDNSZone` bumps `versions[node]` (records-only) but does **not** clear `zonehashes[node]`, which is keyed by node alone. Document this asymmetry: callers wanting a full reset must also call `setZonehash(node, "")`.

---

## Reference implementation

### L92 — Stale Solidity pragma
`pragma solidity ^0.7.4;` is end-of-life. Update to `^0.8.20` (or current).

### L164 / L173 / L249 — Dead `nameHash` and inconsistent hashing
`nameHash` is assigned in `setDNSRecords` but never read; `setDNSRRSet` recomputes `keccak256(name)` independently. Remove the dead assignments or thread `nameHash` through. Also: L164 uses `keccak256(abi.encodePacked(name))` while L173 / L249 use `keccak256(name)` — functionally equivalent for `bytes memory` but the inconsistency is confusing.

### L169 — Deletion-signal edge case
`value.length == 0` is captured from the first record of the run only. If a caller passes two records of the same `(name, resource)` where the first has empty rdata and the second has real rdata, the entire RRset is deleted and the real record is dropped silently. Either document this rule precisely in the Specification or change semantics to "delete iff *all* records in the run are empty".

---

## Security Considerations (expansion needed at L270)

The section is one sentence; please expand to cover:
1. **Wire-format parsing on-chain** via `RRUtils` — pin/audit a specific version.
2. **Gas-DoS via large `data`** in `setDNSRecords` — note bounds or warn about per-record cost.
3. **No DNSSEC.** Records are authoritative only insofar as the ENS owner is trusted.
4. **`clearDNSZone` does not free storage** — it only bumps a version counter. No refunds. Storage growth is permanent.
5. **Zonehash content trust.** The zonehash points off-chain (ERC-1577); resolvers must validate the off-chain content against the on-chain hash.
6. **Wildcard semantics (RFC 4592)** that consuming resolvers are expected to follow when answering against this resolver.

---

*This PR is intentionally Draft. Editorial fixes can land independently; substantive items above should be triaged by the author before advancing from Review.*

mcdee (approved):

PR #1715, opened 2026-05-02
Add ERC: Keyed Transfer With Authorization

BrianBland:
Defines an ERC-20 transfer authorization scheme using a `(uint192 nonceKey, uint64 nonceSequence)` split that mirrors ERC-4337's semi-abstracted nonce encoding.

Targets repeat-counterparty payment patterns (loyal-customer): collapses storage to one slot per (payer, key) pair instead of one slot per authorization. Sibling scheme to ERC-3009, not an extension.


ERCS/erc-draft_keyed_transfer_authorization.md
```diff
@@ -0,0 +1,506 @@
+---
```
abcoathup:
```suggestion
---
eip: 8248
```
@BrianBland 

ERCS/erc-draft_keyed_transfer_authorization.md
```diff
@@ -0,0 +1,506 @@
...
+title: Keyed Transfer With Authorization
+description: ERC-20 transfer authorization with two-dimensional sequential nonces
+author: Brian Bland (@brianbland)
+discussions-to: https://ethereum-magicians.org/t/erc-keyed-transfer-with-authorization
```
abcoathup: Please create an Eth Magicians topic with a link to this PR

ERCS/erc-draft_keyed_transfer_authorization.md
```diff
@@ -0,0 +1,506 @@
...
+ * @param from         Payer's address (Authorizer)
+ * @param to           Payee's address
+ * @param value        Amount to be transferred
+ * @param validAfter   The time after which this is valid (unix time)
+ * @param validBefore  The time before which this is valid (unix time)
+ * @param nonceKey     Independent 192-bit sequence selector (e.g. payee address)
+ * @param nonce        Sequential 64-bit counter within nonceKey
+ * @param v            v of the signature
+ * @param r            r of the signature
+ * @param s            s of the signature
```
ilikesymmetry: One critical miss from 3009 was lack of support for 1271 signatures. We should replace `(v,r,s)` with a single `bytes signature` field

ERCS/erc-draft_keyed_transfer_authorization.md
```diff
@@ -0,0 +1,506 @@
...
+
+/**
+ * @notice Execute a transfer with a signed keyed authorization.
+ * @param from         Payer's address (Authorizer)
+ * @param to           Payee's address
+ * @param value        Amount to be transferred
+ * @param validAfter   The time after which this is valid (unix time)
+ * @param validBefore  The time before which this is valid (unix time)
+ * @param nonceKey     Independent 192-bit sequence selector (e.g. payee address)
+ * @param nonce        Sequential 64-bit counter within nonceKey
```
ilikesymmetry: I think we should use `nonceSequence` to describe the auto-incrementing portion of the nonce. Keeps "nonce" as a higher level noun describing the combination of both a key and a sequence. This languages matches ERC-4337 as well.

ilikesymmetry (commented):
I think we should add a `bytes32 witness` field similar to what Permit2 has to make it easier to build other smart contract systems on top of this. ERC-3009 actually can support these workflows with a `bytes32 nonce` and we've done this in production before, but a witness model is better for isolation, evidenced here by replacing the nonce system with something more performant removes our ability for dual-use on the field.

ERCS/erc-draft_keyed_transfer_authorization.md
BrianBland: Agreed, and this also helps with future-proofing

BrianBland:
@ilikesymmetry added witness variants as well, matching Permit2's semantics. Both parallel entrypoints are kept as the witness variant is heavier in terms of gas and calldata.

hilljessica8787-dot (approved):

SamWilsn:
A friendly reminder to keep technical discussion on the Ethereum Magicians thread.

ERCS/erc-8248.md
```diff
@@ -0,0 +1,749 @@
...
+| --- | --- | --- |
+| ERC-3009 (random `bytes32`) | 81,731 | 1× |
+| Bitmap (256 nonces per slot) | 320 | 256× |
+| Keyed sequential (this ERC) | 1 | 81,731× |
+
+Across all 17 loyal-customer buyers in the sample, keyed sequential nonces require fewer than 60 total nonce storage slots for over 780,000 authorizations — compared to over 2,800 slots for a bitmap scheme.
+
+For the explorer pattern a bitmap is a better match, and implementations MAY expose a bitmap entry point alongside this one. Specifying both schemes in a single ERC was considered and rejected as overscoped; this ERC standardizes only the keyed sequential entry point.
+
+The `(uint192, uint64)` partition is identical to the encoding ERC-4337 uses for UserOperation nonces (see its "Semi-abstracted Nonce Support" section), and is a natural fit for any payer that holds independent payment relationships with multiple counterparties. The 192-bit key is wide enough to hold a counterparty address (160 bits) with 32 bits of subkey for channel separation; payers SHOULD use the upper 32 bits as a channel identifier to keep distinct agents acting under the same payee on independent counters.
```
SamWilsn, line 43: You can link to ERC-4337's section here with an anchor if you like:


```suggestion
The `(uint192, uint64)` partition is identical to the encoding ERC-4337 uses for [UserOperation nonces](./eip-4337.md#semi-abstracted-nonce-support), and is a natural fit for any payer that holds independent payment relationships with multiple counterparties. The 192-bit key is wide enough to hold a counterparty address (160 bits) with 32 bits of subkey for channel separation; payers SHOULD use the upper 32 bits as a channel identifier to keep distinct agents acting under the same payee on independent counters.
```

SamWilsn, line 43: > [...] SHOULD [...]

Hiding a requirement in your Motivation section is a sure way to get an implementer to miss it.

ERCS/erc-8248.md
```diff
@@ -0,0 +1,749 @@
...
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
+
+### Storage
+
+A compliant contract MUST maintain a mapping from `(authorizer, nonceKey)` to the next expected sequence value:
+
+```solidity
+mapping(address => mapping(uint192 => uint64)) internal _nonceSequences;
+```
```
SamWilsn, line 55: Requiring a `mapping` specifically is getting into specifying implementation details. Instead, you should specify only the externally visible behaviour of the contract. If someone wanted to implement this using some other data structure and hand rolled assembly, that should be allowed as long as it behaves the same to outside consumers of the API.

ERCS/erc-8248.md
```diff
@@ -0,0 +1,749 @@
...
+
+The OPTIONAL cancellation event:
+
+```solidity
+event KeyedAuthorizationCanceled(
+    address indexed authorizer,
+    uint192 indexed nonceKey,
+    uint64 indexed nonceSequence
+);
+```
```
SamWilsn, line 77: You should specify when these are emitted, and do so in terms of intent (rather than functions called).

Something like:

> `KeyedAuthorizationUsed` MUST be emitted whenever a `(authorizer, nonceKey)` pair is marked as used (i.e. when it goes from valid to cannot be used again).

ERCS/erc-8248.md
```diff
@@ -0,0 +1,749 @@
...
+    uint256 validBefore,
+    uint192 nonceKey,
+    uint64 nonceSequence,
+    bytes32 witness,
+    string calldata witnessTypeString,
+    bytes calldata signature
+) external;
+```
+
+The witness overloads share function names with the no-witness entry points; Solidity dispatches by selector on the differing parameter shape. This follows the precedent in [ERC-7598](./eip-7598.md), which extends [ERC-3009](./eip-3009.md)'s authorization functions through name-overloading rather than coining a parallel surface.
```
SamWilsn, line 212: ERC-7598 and ERC-3009 sound like they might be necessary background reading to understand this proposal. If that's true, they should be in your requires header.

ERCS/erc-8248.md
```diff
@@ -0,0 +1,749 @@
...
+
+OPTIONAL:
```
SamWilsn, line 214: You should probably have a bit more text here explaining what is optional and when you might want/not want it.

ERCS/erc-8248.md
```diff
@@ -0,0 +1,749 @@
...
+
+- `from != address(0)`.
+- `block.timestamp > validAfter`.
+- `block.timestamp < validBefore`.
+- `nonceSequence == _nonceSequences[from][nonceKey]`. (This single equality enforces both replay protection and in-order consumption.)
+- For `receiveWithKeyedAuthorization` only: `to == msg.sender`.
+
+On a successful call, the implementation MUST execute the following steps in order:
+
+1. Set `_nonceSequences[from][nonceKey] = nonceSequence + 1`.
```
SamWilsn, line 246: Again, implementation detail.

ERCS/erc-8248.md
```diff
@@ -0,0 +1,749 @@
...
+2. Validate `signature` against the [EIP-712](./eip-712.md) digest of the appropriate type hash and parameters under the signer `from`. Validation MUST accept ECDSA signatures from EOAs and contract-account signatures per [ERC-1271](./eip-1271.md) (signer's `isValidSignature(digest, signature)` returns the magic value `0x1626ba7e`). The unified verification pattern (e.g., OpenZeppelin's `SignatureChecker.isValidSignatureNow`) is RECOMMENDED. If validation fails, the implementation MUST revert; the revert unwinds step 1.
```
SamWilsn, line 247: > The unified verification pattern (e.g., OpenZeppelin's `SignatureChecker.isValidSignatureNow`) is RECOMMENDED.

You'll need to define "unified verification" here, either by linking to another ERC or by writing it out in full. Unfortunately OpenZeppelin is not a permitted origin for external links.

ERCS/erc-8248.md
```diff
@@ -0,0 +1,749 @@
...
+3. Emit `KeyedAuthorizationUsed(from, nonceKey, nonceSequence)`.
+4. Execute the underlying ERC-20 transfer of `value` from `from` to `to`.
+
+The counter advance in step 1 precedes signature validation in step 2 to defend against reentrancy: `isValidSignature` is an external call that MAY re-enter the entry point. With the counter already advanced, any re-entrant call observes the new sequence value and fails the equality check above. If signature validation reverts, the entire transaction reverts and the counter advance is unwound, preserving denial-of-service protection against garbage signatures.
+
+Implementations MUST NOT accept `nonceSequence` values that skip ahead of or fall behind the stored counter. The 64-bit width of `nonceSequence` admits 2⁶⁴ authorizations per key before saturation; in Solidity 0.8+ the increment in step 1 reverts on overflow. Payers approaching saturation can rotate to a new `nonceKey`.
```
SamWilsn, line 253: Just FYI, we do support limited LaTeX formatting for math.

ERCS/erc-8248.md
```diff
@@ -0,0 +1,749 @@
...
+
+Permit2's alternative pattern — coining `permitWitnessTransferFrom` alongside `permitTransferFrom` — was considered. ERC-7598's overloading model was preferred because the verb-suffix naming convention used here (`transferWith...`) does not have a clean infix slot for a `Witness` keyword, and because ERC-7598 is the more directly analogous prior art (extending an ERC-3009-family authorization scheme).
+
+## Backwards Compatibility
+
+No backward compatibility issues found. This ERC introduces new function selectors, a new EIP-712 type hash, and a new storage region; it does not modify any prior standard.
+
+## Test Cases
```
SamWilsn, line 394: What you have is absolutely fine, but if you like, you could write automated tests and include them in your assets directory.

ERCS/erc-8248.md
```diff
@@ -0,0 +1,749 @@
...
+        );
+    }
+}
+```
+
+## Security Considerations
+
+### Front-running of wrapper contracts
+
+Smart contracts that wrap a keyed authorization in their own logic (e.g. a `deposit(authData)` function that calls `transferWithKeyedAuthorization` internally) are vulnerable to front-running. An observer can submit the raw `transferWithKeyedAuthorization` call directly, advancing the payer's counter and stranding the wrapper's downstream logic. Wrapper contracts MUST use `receiveWithKeyedAuthorization` (which enforces `to == msg.sender`) instead.
```
SamWilsn, line 697: > [...] MUST [...]

Avoid creating requirements outside of the Specification section. They're too easy for implementers to miss. Instead, put all the requirements in the Specification section, and instead discuss their security implications here.

PR #1723, opened 2026-05-06
Add ERC: Agent Tool Registry

ryanio:
## Summary

Adds new **ERC-8257: Agent Tool Registry** - a permissionless onchain registry for AI agent tools with extensible predicate-based access control.

Each registration commits a metadata URI and a `keccak256` content hash. Invocation access is gated by an optional external predicate contract, following the "pluggable external contract" pattern used by Seaport zones, Uniswap v4 hooks, and ERC-4337 paymasters. Registrations are anchored to a canonical off-chain manifest through origin-binding (manifest served at a well-known path on the endpoint's origin) plus creator self-attestation. Pricing hints are protocol-agnostic and live in the manifest; the registry never handles funds.

## Status

Draft. Opening early to gather feedback. Once an editor assigns a number we will rename the file from `erc-xxxx.md` and set the `eip:` field accordingly.

**Discussion:** https://ethereum-magicians.org/t/erc-draft-agent-tool-registry/28457

## Reference implementation

- **Solidity reference impl + example predicates** (ERC-721, ERC-1155, subscription, composite): https://github.com/ProjectOpenSea/tool-registry
- **TypeScript SDK + CLI** for building tools against the registry (manifest authoring, registration, gating middleware): https://github.com/ProjectOpenSea/tool-sdk
- Live deployments on Base mainnet - addresses in the `tool-registry` README

## Author handles

Cody Sears (@CodySearsOS), Ryan Ghods (@ryanio)


ERCS/erc-xxxx.md
```diff
@@ -0,0 +1,1444 @@
+---
+eip: XXXX
```
```suggestion
eip: 8257
```
@ryanio 

ERCS/erc-xxxx.md
```diff
@@ -0,0 +1,1444 @@
...
+title: Agent Tool Registry
+description: Minimal onchain registry for AI agent tools with extensible predicate-based access control
+author: Cody Sears (@CodySearsOS), Ryan Ghods (@ryanio)
+discussions-to: https://ethereum-magicians.org/t/erc-draft-agent-tool-registry/28457
```
```suggestion
discussions-to: https://ethereum-magicians.org/t/erc-8257-agent-tool-registry/28457
```

ERCS/erc-xxxx.md
ryanio: thanks!

ERCS/erc-8257.md
```diff
@@ -0,0 +1,1444 @@
...
+
+**Onchain commitment matters.** By storing a `keccak256` hash of the manifest onchain, consumers can verify that the manifest they fetched has not been tampered with. Combined with origin-binding, this provides a lightweight trust anchor without requiring onchain access checks or gateway infrastructure.
+
+**Pricing is part of discovery.** An agent choosing between two tools that do the same thing needs to know cost. Declaring pricing in the manifest (and committing it by hash onchain) lets consumers compare tools before invocation. The manifest declares what the tool costs; the endpoint enforces payment. The registry itself never handles funds. The pricing schema is deliberately protocol-agnostic: it identifies the payment protocol by an opaque string so the standard does not depend on any specific payment system.
+
+## Specification
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
```
jochem-brouwer:
```suggestion
The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC 2119](https://www.rfc-editor.org/rfc/rfc2119) and [RFC 8174](https://www.rfc-editor.org/rfc/rfc8174).
```

ERCS/erc-8257.md
```diff
@@ -0,0 +1,1444 @@
...
+
+This ERC targets EVM chains. The registry interface uses EVM-native types (`address`, `bytes32`, `uint256`), and tool IDs are scoped to the `(chainId, registryAddress)` tuple of an EVM deployment (see §1). Pricing entries carry their own chain identifiers via CAIP-19 `asset` and CAIP-10 `recipient`, so a tool registered on an EVM chain MAY price itself in assets on any CAIP-10/CAIP-19 namespace; concrete guidance in §3 focuses on `eip155:*` because that is the namespace the reference implementation has been exercised against.
```
jochem-brouwer: Could you link to the CAIPs?

ERCS/erc-8257.md
```diff
@@ -0,0 +1,1444 @@
...
+
+JCS does not normalize Unicode string content, hex-digit case, or byte-order marks, so three extra rules apply to every manifest before JCS is run:
+
+- **Unicode normalization:** all JSON string values MUST be in Unicode Normalization Form C (NFC) per Unicode 16.0 or later. Producers MUST NFC-normalize before JCS serialization; consumers MUST reject a fetched manifest whose strings are not already NFC-normalized (no silent re-normalization, since that would change the bytes that were hashed). This prevents hash divergence between producers that canonicalize with NFC and consumers that do not.
+- **Byte-order mark:** the manifest MUST be served as UTF-8 without a byte-order mark. Consumers MUST reject a fetched response whose bytes begin with `EF BB BF` rather than silently stripping the BOM, because silent stripping would change the bytes fed to `keccak256` and cause a hash mismatch that is difficult to diagnose.
```
jochem-brouwer:
```suggestion
- **Byte-order mark (BOM):** the manifest MUST be served as UTF-8 without a byte-order mark. Consumers MUST reject a fetched response whose bytes begin with `EF BB BF` rather than silently stripping the BOM, because silent stripping would change the bytes fed to `keccak256` and cause a hash mismatch that is difficult to diagnose.
```

ERCS/erc-8257.md
```diff
@@ -0,0 +1,1444 @@
...
+
+All three rules are treated as verification failures (see [§7 Handling Verification Failure](#handling-verification-failure)). Concrete hash-divergence vectors for NFC-vs-NFD and with-vs-without-BOM appear in [Test Cases](#test-cases) alongside canonical reference manifests.
```
jochem-brouwer: NFD is not defined as abbreviation, could you define it?

ERCS/erc-8257.md
```diff
@@ -0,0 +1,1444 @@
...
+
+#### URL Normalization
+
+Because URLs have multiple equivalent representations, both registrants and consumers MUST reduce HTTPS URLs in this ERC to a canonical form before writing, reading, or comparing them. The rules split into two groups: a **general HTTPS-URL** group that applies to every URL field (notably `endpoint` and `image`), and a **well-known-path** group that applies additionally to `metadataURI`.
+
+**General HTTPS-URL normalization** (rules G1‑G3):
+
+- **G1.** Lowercase the scheme and host.
+- **G2.** Omit port 443 (the default for `https`).
+- **G3.** Normalize the host as an A-label (ASCII Compatible Encoding, per [RFC 5891](https://www.rfc-editor.org/rfc/rfc5891)) before lowercasing. Consumers MUST reject a URL whose host is given as a U-label (internationalized) without ACE encoding.
```
jochem-brouwer:
```suggestion
- **G3.** Normalize the host as an A-label (ASCII Compatible Encoding (ACE), per [RFC 5891](https://www.rfc-editor.org/rfc/rfc5891)) before lowercasing. Consumers MUST reject a URL whose host is given as a U-label (internationalized) without ACE encoding.
```

ERCS/erc-8257.md
```diff
@@ -0,0 +1,1444 @@
...
+
+## Security Considerations
+
+### Predicate Gas Consumption
+
+The registry delegates access checks to an external predicate contract via `staticcall`. A malicious or poorly written predicate could consume unbounded gas, causing `hasAccess` calls to revert or become prohibitively expensive. Implementations MUST cap the gas forwarded to predicate calls at no more than 200,000 gas (i.e., `staticcall{gas: <=200_000}`) and MUST treat out-of-gas as "access denied." Without a normative ceiling, onchain consumers (paymasters, gated execution routers, contracts that compose `hasAccess` into their own logic) cannot reason about worst-case gas and the registry becomes unsafe to compose without per-deployment auditing.
```
jochem-brouwer: Review comment: this hardcodes gas sent and this is generally not a good idea because of possible future gas repricings.

ERCS/erc-8257.md
```diff
@@ -0,0 +1,1444 @@
...
+
+Implementations MAY choose a smaller cap. The 200,000-gas value is the upper bound: implementations exceeding it are non-conformant. The cap is chosen so that realistic predicate logic completes comfortably below it:
+
+- An ERC-20 or [ERC-721](./eip-721.md) `balanceOf` check plus a comparison runs in roughly 5,000-10,000 gas.
```
jochem-brouwer: review comment: These specific gas mentions should also be removed IMO because this depends on the current fork and also on the EVM bytecode produced by the compiler used (including compiler settings like optimizations)

ERCS/erc-8257.md
```diff
@@ -0,0 +1,1444 @@
...
+- Merkle proof verification over a depth-20 tree (supporting ~1,000,000 leaves) runs in roughly 40,000-60,000 gas using OpenZeppelin's `MerkleProof` library.
+- An ECDSA signature verification (`ecrecover` plus constant work) runs in roughly 5,000-10,000 gas.
+
+A 200,000 gas cap leaves headroom of at least 2-3× over the most expensive common case (Merkle proof verification), which keeps the happy path well below the cap while still bounding worst-case gas so that a malfunctioning predicate cannot turn `hasAccess` into an expensive storage-depleting call. Implementations that only expect trivial gates MAY choose a smaller cap. Implementations that expect more elaborate predicates (e.g., zero-knowledge proof verification) MUST instead deploy a derivative registry that explicitly opts out of `IToolRegistry` conformance (a new interface ID with a documented higher cap), since a higher cap weakens the worst-case-gas guarantee that downstream callers rely on. In all cases the outcome of a capped `staticcall` is the same: on OOG the call reverts, and the registry treats the revert as "access denied" (or, via `tryHasAccess`, as `(ok=false, granted=false)`).
```
jochem-brouwer: OOG -> out of gas

ERCS/erc-8257.md
```diff
@@ -0,0 +1,1444 @@
...
+
+### Predicate Upgradeability
+
+If the `accessPredicate` is a proxy contract (e.g., an [ERC-1967](./eip-1967.md) transparent proxy or UUPS proxy), the predicate owner can silently change the access logic without the tool creator calling `setAccessPredicate`. This means tool consumers cannot rely solely on the `AccessPredicateUpdated` event to detect changes in access semantics. Consumers SHOULD check whether a predicate address contains proxy patterns (e.g., ERC-1967 storage slots) and SHOULD treat upgradeable predicates as higher risk than immutable ones.
```
jochem-brouwer: UUPS proxy abbreviation not defined

ERCS/erc-8257.md
```diff
@@ -0,0 +1,1444 @@
...
+
+### Predicate Reverting
+
+Implementations MUST treat a reverting predicate call as "access denied" from `hasAccess` rather than bubbling the revert to the caller; this prevents a malfunctioning predicate from breaking the registry's view functions. Consumers that need to distinguish a clean denial from a malfunction SHOULD use `tryHasAccess` instead, which reports the predicate outcome as `(ok, granted)`: a malfunction surfaces as `(false, false)` while a clean denial surfaces as `(true, false)`.
+
+### Predicate Reentrancy
+
+Although `hasAccess` is a `view` function, the predicate it delegates to is an external contract. Implementations MUST call the predicate via `staticcall`, which the EVM forbids from mutating state. State-mutating entrypoints on the registry make no external calls to the predicate at all, so the registry exposes no reentrancy surface for state writes.
```
jochem-brouwer: Nit: I would drop `view` here and instead directly refer the staticcall. View is a solidity concept, staticcall the EVM concept.

ERCS/erc-8257.md
```diff
@@ -0,0 +1,1444 @@
...
+
+Read-only reentrancy is a separate hazard that the staticcall guarantee does not eliminate. A predicate is permitted to read state from any third contract; if a consumer calls `hasAccess` while another protocol's view of its own state is mid-transition (e.g., during a multi-call that updates balances between calls, or via a callback into the consumer mid-execution), the predicate may observe stale or inconsistent state and return a `granted` answer that is no longer true once the outer transaction settles. This is the same class of bug that affected several DeFi protocols whose oracles read view state from contracts mid-mutation. Consumers that gate state changes on `hasAccess` MUST NOT call it between writes whose intermediate state another protocol's view function depends on, and SHOULD apply the same reentrancy guards to `hasAccess` invocations that they apply to any other external view call whose inputs include third-party state.
```
jochem-brouwer, line 1212: This is an interesting pattern, is there some way to already do this? You can share memory between call frames using transient storage and you could thus save the hasAccess on the first time that this contract is being called (top level, but not necessarily call stack top level, which could thus still mean that once tx finishes the hasAccess is now false)

ERCS/erc-8257.md
```diff
@@ -0,0 +1,1444 @@
...
+- requiring the caller to present a signature over a challenge in `data`, verified by the predicate or the endpoint,
+- issuing a short-lived session token after an out-of-band authentication step.
+
+A predicate that gates purely on `account` (e.g., "is this address a holder of NFT X?") is safe to consult but unsafe to act on without such binding. Ignoring this distinction is the most common way that correct-looking access gates become unsound.
+
+#### Concrete AccessProof Pattern
+
+The following challenge-response pattern binds the `account` parameter to a real principal and prevents replay across tools and time windows. Implementations that need authenticated access SHOULD use this pattern or an equivalent that provides the same properties.
+
+1. **Challenge issuance.** The agent or consumer generates a challenge: `challenge = keccak256(abi.encodePacked(toolId, account, block.chainid, deadline))`, where `deadline` is a Unix timestamp after which the proof expires.
```
jochem-brouwer: This might be a security concern. You will have to check if there are no ways someone creates a malicious challenge where the signer signs something which is also valid in another context. For instance: signing a transaction, or something what comes close (this is a list of 4 items which could maybe be tricked here): https://eips.ethereum.org/EIPS/eip-7702 the authorizatoin part where `keccak(MAGIC || rlp([chain_id, address, nonce]))` is signed.
The usual way to mitigate this is to prefix it with a non-used prefix, but these prefix above is used on the core protocol and am not sure if there is an ERC way to safe-prefix these signers to ensure nothing is signed which could be used in another context. ERC-prefixed-signing. 
You could also use a prefixed string for instance used in `eth_sign` to prevent collissions: https://ethereum.org/developers/docs/apis/json-rpc/#eth-sign

(please check with a security expert)

ERCS/erc-8257.md
```diff
@@ -0,0 +1,1444 @@
...
+
+### Zero-Code Access Predicates
+
+Implementations of `IToolRegistry.registerTool` and `IToolRegistry.setAccessPredicate` MUST treat addresses with no deployed code (externally-owned accounts, or CREATE2 addresses that have not yet been deployed) as accepted but unverifiable, per step 2 of [Predicate Validation at Registration](#predicate-validation-at-registration). ERC-165 cannot be queried against empty code, so such a predicate cannot be checked at registration time.
+
+At invocation time, a staticcall to a zero-code address returns empty data, which the registry MUST treat as non-compliant (see [§1 `hasAccess`](#itoolregistry-interface)) and therefore as "access denied." A tool registered with a zero-code predicate is consequently inaccessible until a contract is deployed at that address.
+
+Creators who rely on counterfactual deployment (registering a predicate before deploying it) SHOULD:
+
+1. Verify that the CREATE2 salt and init-code commit to the intended predicate bytecode before registration.
```
jochem-brouwer: The predicate bytecode is here the runtime bytecode (deployed bytecode) I assume? This could be crafted malicious, for instance in the initcode you do SSTORE(0, X). In the runtime bytecode the code looks like:

```
if SLOAD(0) == 0:
- do normal stuff
else
- do malicious predicate
``` 

If we commit to the runtime bytecode above we could thus in practice have two code paths which depend on whatever storage was stored in the initcode. So two different initcodes create the same runtime code, but in practice it will only be able to run one of the two codes

ERCS/erc-8257.md
```diff
@@ -0,0 +1,1444 @@
...
+
+1. If the predicate is `address(0)`, the call is open-access. Skip validation and accept.
+2. If the predicate has no deployed code (`extcodesize == 0`), accept as best-effort. ERC-165 cannot be queried against empty code, so the registration is recorded but the tool is inaccessible until a contract is deployed at the address (see [Zero-Code Access Predicates](#zero-code-access-predicates)).
+3. Probe `IERC165(predicate).supportsInterface(type(IERC165).interfaceId)` with a strict gas cap. The reference implementation uses 30,000 gas, which matches the ERC-165 standard's documented ceiling (the standard itself requires `supportsInterface` to use *less than* 30,000 gas, so a conformant probe completes within the cap and a probe that exhausts it is non-conformant by construction). Implementations MAY choose any cap between 1,000 and 30,000 gas; a cap below 1,000 is too tight for a realistic ERC-165 probe and MUST NOT be used. If the probe reverts or runs out of gas, the predicate is treated as not advertising ERC-165: accept as best-effort.
```
jochem-brouwer: (also has gas cap)

ERCS/erc-8257.md
```diff
@@ -0,0 +1,1444 @@
...
+
+### Malicious Endpoints
+
+Tool endpoints are creator-controlled URLs. The `endpoint` field MUST be an `https://` URL (see §2); consumers MUST reject manifests whose `endpoint` uses any other scheme. Any consumer that may cause a tool to be invoked on a user's behalf (agent frameworks, wallets, invocation proxies, or any surface exposing a "run this tool" affordance, using the same scoping as the 24-hour cache ceiling in [§7 Handling Verification Failure](#handling-verification-failure)) MUST reject endpoints that resolve to private IP ranges (RFC 1918, RFC 6598, loopback, link-local, IPv6 ULA `fc00::/7`, IPv6 link-local `fe80::/10`) to prevent SSRF attacks against internal services, and MUST resolve the host immediately before each invocation rather than caching the resolution, to prevent DNS rebinding. Purely informational surfaces that never invoke endpoints (e.g., indexer digests, historical registries) MAY use SHOULD-level rejection but SHOULD NOT relax it without an explicit policy. Agent frameworks that invoke tool endpoints MUST enforce request timeouts and response size limits.
```
jochem-brouwer: SSRF not defined

ERCS/erc-8257.md
```diff
@@ -0,0 +1,1444 @@
...
+
+### Manifest Parser Hardening
+
+A creator controls the bytes served at `metadataURI`, and a permissionless registration makes every fetched manifest effectively attacker-controlled. A consumer that parses without limits can be DoSed by a single malicious registration. Consumers MUST enforce the following ceilings and MUST reject any manifest that exceeds them:
+
+| Limit | Value | Why |
+| --- | --- | --- |
+| Manifest byte size | 1 MiB (1,048,576 bytes) | Bounds HTTP body and parser memory; a fully-populated honest manifest with ten pricing entries and a moderately rich schema is well under 10 KiB. Consumers MUST truncate at the limit and treat oversize fetches as a verification failure. |
```
jochem-brouwer: Nit: also define 10 KiB here

ERCS/erc-8257.md
```diff
@@ -0,0 +1,1444 @@
...
+| `inputs` + `outputs` total schema nodes | 1,024 nodes | Covers rich schemas; prevents pathological fan-out attacks that create millions of subschemas via shallow-but-wide structures. |
+
+Each row above is an independent upper bound on its own axis; they do not jointly compose. The 1 MiB total-size cap is the binding constraint that limits the joint product. As a worked example, a manifest cannot simultaneously hit `access.requirements.length = 256` and per-entry `data = 4,096` decoded bytes: 256 maxed-out entries serialize to roughly 2.1 MiB of hex envelope (8,194 UTF-8 bytes per entry, plus label, kind, and JSON overhead), well over the 1 MiB ceiling. Consumers MUST enforce every row, but conformant manifests trade off depth for breadth and never hit every row at once.
+
+Regex `pattern` values inside embedded schemas MUST be evaluated by a matcher immune to catastrophic backtracking (e.g., RE2, or an implementation with a bounded step count). A matcher with exponential worst-case behavior on attacker input is unsafe and MUST NOT be used.
```
jochem-brouwer: RE2?

ERCS/erc-8257.md
```diff
@@ -0,0 +1,1444 @@
...
+
+The following normative rules apply to common render paths:
+
+- **Text fields (`name`, `description`, `tags`).** Consumers MUST contextually encode these fields when rendering: HTML-escape for text nodes, attribute-escape for attributes, and JS-escape for script contexts. UIs that inject them into the DOM without encoding are vulnerable to stored XSS from a single malicious registration.
+- **Markdown rendering of `description`.** Consumers that render Markdown MUST disable raw-HTML passthrough or sandbox the rendered output. A compliant Markdown renderer strips `<script>`, `<iframe>`, event handlers, and `javascript:` URLs, or runs under a CSP that neutralizes these surfaces.
+- **`image` URIs.** Consumers MUST NOT accept `javascript:`, `file:`, `data:text/html`, or `vbscript:` URIs for a tool icon: these schemes enable script execution or local-filesystem exposure and have no legitimate icon use case. Consumers SHOULD NOT accept `http:` (MITM risk) or `blob:` (cross-context leakage risk) URIs; consumers that do accept either MUST apply the scheme-appropriate defenses in the Per-Scheme Rendering Guidance below.
```
jochem-brouwer: MITM abbreviation

ERCS/erc-8257.md
```diff
@@ -0,0 +1,1444 @@
...
+- **`endpoint` and extension-namespaced URL fields as clickable links.** Consumers SHOULD NOT render `javascript:`, `file:`, or `blob:` schemes as links; `http:` links are permitted only with explicit user consent because of MITM risk; all external links SHOULD carry `rel="noopener noreferrer"` to prevent tab-nabbing.
+
+#### Non-Normative: Per-Scheme Rendering Guidance for `image`
+
+For any scheme consumers do accept, scheme-appropriate defenses should be layered on the normative rules above:
+
+- `https://`: set `referrerpolicy="no-referrer"` and proxy through consumer-controlled infrastructure where feasible.
+- `ipfs://`: resolve through a trusted gateway the consumer operates, not an arbitrary public gateway. A hostile gateway can inject response headers, set tracking cookies, or serve manipulated content even when the CID is content-addressed.
```
jochem-brouwer: CID abbreviation

jochem-brouwer (changes requested):
Mostly editorial comments, some security points/qs. 

Main points: CAIP/RFCs should have URLs which is handy for the reader. 
All abbreviations should on first-use be defined.

Very interesting ERC! 😄 👍

ryanio:
@jochem-brouwer thanks for the thorough review — both passes have shipped. Below is a summary of what changed, organized by significance.

**Substantive normative changes (`1c1568d8`)**

- **Predicate gas cap removed.** Dropped the normative 200,000-gas `staticcall` ceiling and the illustrative 5k–60k gas table. Safety now relies on `staticcall` semantics alone: any failed sub-call returns `(success=false)` and [EIP-150](./eip-150.md)'s 63/64 rule preserves the caller's gas for the failure path. The gas-bound decision moves to the call site — composers wrap `hasAccess` with `staticcall{gas: budget}` based on their target chain and fork. Cited [EIP-8038](./eip-8038.md) (Glamsterdam SLOAD repricing) as the reason the spec doesn't pin a number. The reference implementation keeps a 200k internal cap framed as defense-in-depth, not a conformance requirement. Downstream gas-cap mentions (ERC-165 probe, recursive-self-reference rationale, `getRequirements` introspection cap) updated to match.
- **AccessProof challenge → EIP-712.** Replaced raw `keccak256(abi.encodePacked(toolId, account, chainid, deadline))` with EIP-712 typed data. The `"\x19\x01"` prefix and domain (`name: "ERC8257-AccessProof"`, `verifyingContract: predicate`) prevent collision with EIP-7702 authorizations, `eth_sign` payloads, or other ERC challenge schemes. Follows the canonical anti-collision pattern; happy to take a dedicated security review on it if you want to flag someone.
- **Init-code commitment.** Acknowledged that runtime bytecode alone doesn't fully commit predicate behavior — init-code can `SSTORE` state that the runtime branches on. Counterfactual-deployment guidance now recommends pinning the **init-code hash** (which determines both runtime bytecode and constructor-planted storage) or inspecting post-deployment storage, and flags pure-logic predicates (no storage reads) as the safer choice for security-sensitive gates.
- **Transient-storage caching note.** Added non-normative pointer for composers: cache `hasAccess` results in [EIP-1153](./eip-1153.md) transient storage keyed by `(toolId, account)` to keep a stable answer across multiple calls within one transaction. Clears automatically at end-of-transaction so it can't leak across txs.

**Editorial (`451a1508` + `1c1568d8`)**

- Linked RFC 2119 / 8174 / CAIP-19 / CAIP-10 / CAIP-2 on first use
- Expanded abbreviations on first use: NFD, ACE, TEE, E2EE, TLS, TCB, IDN, BGP, EOA, UUPS, SSRF, MITM, CID; glossed RE2; defined "10 KiB (10,240 bytes)" inline; spelled out OOG as "out of gas"
- Dropped Solidity `view` framing in §Predicate Reentrancy in favor of `staticcall` (EVM concept, not Solidity)
- Escaped the pipe inside the `amount` regex (`` `^(0\|[1-9][0-9]*)$` ``) so the table cell no longer truncates at `^(0`
- Differentiated the duplicate `// e.g. "eip155:8453"` parsing comments so the reader can see the function intent
- Tagged the `verifyOriginBinding` block as ` ```solidity ` and noted in prose that it's Solidity-flavored pseudocode
- Converted "Handling Verification Failure" bullets to a numbered list; clarified "check 4" as the creator-binding step from §Consumer Verification
- Confined remaining RFC 2119 keywords to §Specification (per EIP-1)

19 threads marked resolved. Particularly happy to discuss the gas-cap rework if you'd like to push back — it's a normative loosening, so worth confirming you're aligned.

ERCS/erc-8257.md
```diff
@@ -0,0 +1,1450 @@
...
+
+This ERC intentionally does not pin a maximum gas value for predicate calls. Gas costs of common predicate building blocks (SLOADs, hashing, signature recovery) change with hard forks (e.g., the SLOAD repricing anticipated under Glamsterdam, see [EIP-8038](./eip-8038.md)) and differ between L1 and the various L2 fee schedules; concrete gas figures also depend on the Solidity compiler version and optimizer settings used to build the predicate. Any fixed ceiling published in this document would either become too tight after a repricing (breaking predicates that worked before) or pre-emptively too loose. The gas-bound decision therefore belongs at the call site:
+
+- **Direct callers** (wallets and agent frameworks invoking `hasAccess` from off-chain or via `eth_call`) supply gas through the normal RPC budget. A predicate that exhausts the budget reverts; the caller treats this as "access denied" and MAY retry with a larger budget.
```
jochem-brouwer:
```suggestion
- **Direct callers** (wallets and agent frameworks invoking `hasAccess` from offchain or via `eth_call`) supply gas through the normal RPC budget. A predicate that exhausts the budget reverts; the caller treats this as "access denied" and MAY retry with a larger budget.
```

ERCS/erc-8257.md
```diff
@@ -0,0 +1,1450 @@
...
+status: Draft
+type: Standards Track
+category: ERC
+created: 2026-04-17
+requires: 165
+---
+
+## Abstract
+
+This ERC specifies a permissionless onchain registry for AI agent tools. Each registration commits a metadata URI and a content hash; invocation access is gated by an optional external predicate contract. Registrations are anchored to a canonical off-chain manifest through origin-binding (the manifest is served at a well-known path on the endpoint's origin) and creator self-attestation (the manifest declares which onchain address is entitled to register it). Pricing and access-model details are deferred: the manifest carries protocol-agnostic pricing hints, and access logic lives in the predicate layer.
```
jochem-brouwer:
```suggestion
This ERC specifies a permissionless onchain registry for AI agent tools. Each registration commits a metadata URI and a content hash; invocation access is gated by an optional external predicate contract. Registrations are anchored to a canonical offchain manifest through origin-binding (the manifest is served at a well-known path on the endpoint's origin) and creator self-attestation (the manifest declares which onchain address is entitled to register it). Pricing and access-model details are deferred: the manifest carries protocol-agnostic pricing hints, and access logic lives in the predicate layer.
```

`offchain` is used in the entire document, so use it here too (another comment later where `off-chain` is used)

ERCS/erc-8257.md
```diff
@@ -0,0 +1,1450 @@
...
+
+### Malicious Endpoints
+
+Tool endpoints are creator-controlled URLs. The `endpoint` field MUST be an `https://` URL (see §2); consumers MUST reject manifests whose `endpoint` uses any other scheme. Any consumer that may cause a tool to be invoked on a user's behalf (agent frameworks, wallets, invocation proxies, or any surface exposing a "run this tool" affordance, using the same scoping as the 24-hour cache ceiling in [§7 Handling Verification Failure](#handling-verification-failure)) MUST reject endpoints that resolve to private IP ranges (RFC 1918, RFC 6598, loopback, link-local, IPv6 ULA `fc00::/7`, IPv6 link-local `fe80::/10`) to prevent Server-Side Request Forgery (SSRF) attacks against internal services, and MUST resolve the host immediately before each invocation rather than caching the resolution, to prevent DNS rebinding. Purely informational surfaces that never invoke endpoints (e.g., indexer digests, historical registries) MAY use SHOULD-level rejection but SHOULD NOT relax it without an explicit policy. Agent frameworks that invoke tool endpoints MUST enforce request timeouts and response size limits.
```
jochem-brouwer:
```suggestion
Tool endpoints are creator-controlled URLs. The `endpoint` field MUST be an `https://` URL (see §2); consumers MUST reject manifests whose `endpoint` uses any other scheme. Any consumer that may cause a tool to be invoked on a user's behalf (agent frameworks, wallets, invocation proxies, or any surface exposing a "run this tool" affordance, using the same scoping as the 24-hour cache ceiling in [§7 Handling Verification Failure](#handling-verification-failure)) MUST reject endpoints that resolve to private IP ranges ([RFC 1918](https://www.rfc-editor.org/rfc/rfc1918), [RFC 6598](https://www.rfc-editor.org/rfc/rfc6598), loopback, link-local, IPv6 ULA `fc00::/7`, IPv6 link-local `fe80::/10`) to prevent Server-Side Request Forgery (SSRF) attacks against internal services, and MUST resolve the host immediately before each invocation rather than caching the resolution, to prevent DNS rebinding. Purely informational surfaces that never invoke endpoints (e.g., indexer digests, historical registries) MAY use SHOULD-level rejection but SHOULD NOT relax it without an explicit policy. Agent frameworks that invoke tool endpoints MUST enforce request timeouts and response size limits.
```

jochem-brouwer: (add RFC urls)

jochem-brouwer (commented):
Two suggestions to change off-chain to offchain, one clarification (BOM) and link RFCs

ryanio:
@jochem-brouwer thanks, all four applied in `ba1bb6ca`:

- L16, L1180: `off-chain` → `offchain`
- L411: BOM bound on first use
- L1317: RFC 1918 / 6598 URLs added

A few extra cleanups while I was in there — bare `§N` are now anchor links, `requires:` widened to `150, 165, 712, 1967`, DoS / SGX / SEV-SNP / TDX expanded on first use, and the `erc8257` CAIP-19 namespace is called out as proposed rather than implied-registered.

jochem-brouwer (approved):
Great, thanks a lot, lets merge!! 🚀 😄 
LGTM

PR #1741, opened 2026-05-13
Update ERC-7683: Redesign around resolvers

frangio:
Based on the analysis and design presented in [ERC-7683 Redux: Programmable Fillers](https://ethereum-magicians.org/t/erc-7683-redux-programmable-fillers/27674).

### DIFF
--- a/ERCS/erc-7683.md
+++ b/ERCS/erc-7683.md
@@ -392,7 +392,9 @@ The payload allows implementations to take arbitrary specifications for these be
 
 A compliant intent protocol must implement and deploy a resolver, along with a payload format that encodes orders for that protocol. The resolver contract must validate and interpret order payloads, and produce instructions for solvers to follow.
 
-Resolution happens offchain via `eth_call`, even though the resolver is published onchain. This is useful because the resolver is the point of trust: solver operators whitelist resolvers, and once vetted, the instructions they produce are trusted. Vetting is done through the usual means: security audits, bounties, lindiness, etc.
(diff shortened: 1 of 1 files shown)


nicholaspai (commented):
This is a great rewrite, i think it covers a wide range of use cases and allows the intent creator to set important requirements for their order.

One area I think it might be missing is how to handle JIT setting of the `target` and `selector`from the intent fulfiller. It appears that you already handle the case where the intent creator wants to let the filler set a function argument at fill time, using the `NeedsVariable` attribute. 

Imagine that one of the steps is a "DEX swap" where the intent creator wants to leave it up to the filler which DEX they choose to swap on.

The difficulty of implementing this is that the `Call.arguments` list now can be variable if the `selector` can change.

Great work, really supportive of this direction!

nicholaspai (approved):

frangio:
> you already handle the case where the intent creator wants to let the filler set a function argument at fill time, using the `NeedsVariable` attribute

To clarify, setting custom arguments is supported without `NeedsVariable` by just setting a variable index as the argument:
```
`arguments`: A list of function arguments, each a constant or a variable.
```
In EVM resolvers:
```
Each element is either an ABI-encoded variable index, or a framed ABI encoding of a value.
```

---

> Imagine that one of the steps is a "DEX swap" where the intent creator wants to leave it up to the filler which DEX they choose to swap on.

Yeah this is a good case to consider, and I think it may need at least spec clarifications, if not a new attribute or kind of step. The issue is that I think there's an assumption currently that the filler uses its own liquidity for fills, and there is no clear way to use the funds that the order pays out as the liquidity, since we claim those are only guaranteed once all of the order is fulfilled.

I don't think the solution is to allow the filler to specify a custom target and function selector, because the entry point for the call should probably remain a protocol contract that can atomically validate the output funds are received. Perhaps the solution is a callback mechanism where an attribute specifies that a (filler chosen) contract will be called with a certain calldata and token allowance, and that the protocol will check it has received some funds after the callback.

PR #1743, opened 2026-05-13
Update ERC-6357: Fix delegatecall to use individual data elements

red-swan:
The reference implementation didn't actually split out the different calls. It fed the entire call data array into the delegate call each time.


Pandapip1 (approved):
This seems probably right? Been a while

PR #1783, opened 2026-06-01
Update ERC-7730: Introduce string internationalization process

forshtat:
I am worried that leaving internationalization to be an afterthought in a Clear Signing protocol is too risky.
Once the standard is adopted in its English-only version, it may be very hard to move the needle and get hardware wallets to update their implementation to address less lucrative markets.

The proposal seems to me to be quite simple, with a JSON translation files being linked directly from the core ERC-7730 spec.


manuelwedler (commented):
I agree that internationalization is quite important and we might want to include it in the original ERC. About this proposal, I mainly wonder how it relates to ERC-8176. Should there be separate attestations for translation files? Or should we have a hash of each of the translation files in the descriptor, such that the descriptor's attestation also includes the translation files?

forshtat:
> Should there be separate attestations for translation files? Or should we have a hash of each of the translation files in the descriptor, such that the descriptor's attestation also includes the translation files?

I don't have a strong opinion one way or another, but if it was up to me I would put the translation hashes into the main descriptor to create a single immutable multilingual Clear Signing attestation. There is only a limited attention that can go into validating  the localizations by attesters, and the last thing we should want is someone sneaking in a malicious Luxembourgish localization to a well known contract.

hilljessica8787-dot (approved):

ankurdotb:
There's a bit of a precedent that I want to mention here: [ERC-1155 has the concept of a `localization` parameter](https://eips.ethereum.org/EIPS/eip-1155#:~:text=%7D%0A%09%7D%0A%7D-,Localization,-Metadata%20localization%20should) that allows defining a URI to fetch the details, what the default locale is, and what the list of locales are.

Might be worth mirroring instead of creating a bespoke `i8n` map of our own, since ERC-1155 is a more well-known standard that wallets might already have built-in support and parsing for.

assets/erc-7730/erc7730-v3.0.0-next.schema.json
```diff
@@ -36,6 +36,33 @@
...
+            "description": "An optional map from BCP-47 language tags to URI references pointing to translation resources for this ERC-7730 document. The internal format of the referenced resource, key strategy, and attestation mechanism are defined by a companion localization specification, not by this schema.",
+            "propertyNames": {
+                "pattern": "^[a-zA-Z]{2,3}(-[a-zA-Z0-9]{2,8})*$",
+                "description": "BCP-47 language tag (e.g. 'fr', 'zh-Hans', 'pt-BR')"
+            },
+            "additionalProperties": {
+                "title": "Candidate references",
+                "type": "array",
+                "minItems": 1,
+                "description": "One or more candidate references to the translation resource for this locale, in preference order. This specification does not mandate or favor any particular URI scheme or resolution mechanism: entries may use HTTP(S), IPFS, a relative reference resolved against this document's own location, or any other transport a wallet supports. A wallet MUST use the first entry it is able to resolve.",
```
forshtat:
```suggestion
                "description": "One or more candidate references to the translation resource for this locale. This specification does not mandate or favour any particular URI scheme or resolution mechanism. A wallet may use any entry it is able to resolve.",
```

assets/erc-7730/example-main.json
```diff
@@ -1,5 +1,5 @@
 {
-    "$schema": "./erc7730-v2.schema.json",
+    "$schema": "./erc7730-v3.0.0-next.schema.json",
```
forshtat:
```suggestion
    "$schema": "./erc7730-v2.schema.json",
```

assets/erc-7730/erc7730-v3.0.0-next.schema.json
```diff
@@ -36,6 +36,42 @@
 
         "display": {
             "$ref": "#/$display/main"
+        },
+        "$localization": {
+            "title": "Localization",
+            "type": "object",
+            "description": "Optional localization metadata for this ERC-7730 document, loosely mirroring ERC-1155's 'localization' extension. This is a non-authoritative discovery hint: wallets may also obtain translation resources from other sources, and validity of a translation resource does not depend on being listed here. The internal format of referenced translation resources, key strategy, and attestation mechanism are defined by a companion localization specification, not by this schema.",
```
forshtat:
```suggestion
            "description": "Optional localization metadata for this ERC-7730 document. This is a non-authoritative discovery hint and wallets may also obtain translation resources from other sources. Validity of a translation resource does not depend on being listed here. The internal format of referenced translation resources, key strategy, and attestation mechanism are defined by a companion localization specification ERC-8346.",
```

ERCS/erc-7730.md
```diff
@@ -1526,6 +1526,66 @@ List of currently supported encryption schemes hints for fields
...
 
+### Internationalization
+
+All display strings in ERC-7730 files — including intents, labels, field names, and enumeration values — MUST be authored in English (`en-US`).
+
+Translation of these strings into other languages can be provided based on the English document.
+
+#### `$localization` field
+
+An optional object holding localization metadata for the document, loosely mirroring [ERC-1155](./eip-1155.md)'s `localization` extension. Its `references` property maps [BCP-47](https://www.rfc-editor.org/rfc/rfc5646) language tags to an array of candidate references to a translation resource; a wallet may use any entry it is able to resolve. Language tags MUST use canonical BCP-47 casing (e.g. `fr`, `zh-Hant`, `pt-BR`).
```
```suggestion
An optional object holding localization metadata for the document. Its `references` property maps [BCP-47](https://www.rfc-editor.org/rfc/rfc5646) language tags to an array of candidate references to a translation resource; a wallet may use any entry it is able to resolve. Language tags MUST use canonical BCP-47 casing (e.g. `fr`, `zh-Hant`, `pt-BR`).
```

forshtat:
> There's a bit of a precedent that I want to mention here: [ERC-1155 has the concept of a `localization` parameter](https://eips.ethereum.org/EIPS/eip-1155#:~:text=%7D%0A%09%7D%0A%7D-,Localization,-Metadata%20localization%20should) that allows defining a URI to fetch the details, what the default locale is, and what the list of locales are.
> 
> Might be worth mirroring instead of creating a bespoke `i8n` map of our own, since ERC-1155 is a more well-known standard that wallets might already have built-in support and parsing for.

Thanks @ankurdotb! There are some small differences in our case, but I have update this PR to something quite close to the way ERC-1155 `localization` field works, I think. The PR is much smaller now and the actual translation specification (very much WIP) is moved to https://github.com/ethereum/ERCs/pull/1906.
I understand internationalization is not a blocker, but still I would love to get this done eventually, so I would appreciate any feedback.

PR #1771, opened 2026-05-26
Add ERC: AI Inference Proof Verification Interfaces

JimmyShi22:
## ERC-8274 v0.2 — Two-layer inference verification interfaces

This PR introduces ERC-8274, defining three minimal abstract interfaces organized into two composable layers for on-chain AI inference proof verification.

### What this ERC defines

**`IProofVerifier`** — inner algorithm layer, implemented by proof system providers (ZK teams, TEE vendors, oracle networks, attestation gateways). Stateless. Answers: "is this proof cryptographically valid for this input and output?"

```solidity
function verify(bytes32 inputHash, bytes32 outputHash, bytes calldata metadata, bytes calldata proof) external returns (bool);
function proofSystem() external view returns (string memory);  // e.g. "zk/sp1", "attestation/judgment"
function proofProfile() external view returns (bytes32);       // deployment-specific fingerprint
```

**`IAgentVerifier`** — outer application layer, implemented by agent developers. Stateful. Wraps one or more `IProofVerifier`(s). Answers: "for this task, was this agent authorized and does the proof confirm it?"

```solidity
function verify(bytes32 taskId, bytes32 agentId, bytes32 inputHash, bytes32 outputHash, bytes calldata proof)
    external returns (bool valid, bytes32 verificationDigest);

event VerificationCompleted(bytes32 indexed taskId, bytes32 indexed agentId, bytes32 indexed verificationDigest,
    bool valid, bytes32 inputHash, bytes32 outputHash, bytes32 agentProofProfile);
```

**`IAgentVerifiable`** — declaration layer for settlement contracts; declares which `IAgentVerifier` is in use and emits `AgentVerifierUpdated` on any change.

### Key design decisions

**Two-layer separation.** Proof system providers implement `IProofVerifier` without knowing the application context. Agent developers implement `IAgentVerifier` without being coupled to a specific proof system. Settlement contracts call `IAgentVerifier` without knowing the backend. Each layer is independently deployable and auditable.

**`verificationDigest` commitment.** `IAgentVerifier.verify()` returns a `bytes32` digest computed from `keccak256(abi.encode(taskId, agentId, inputHash, outputHash, valid, agentProofProfile))`. `VerificationCompleted` carries all preimage fields alongside the digest, making every record independently verifiable by any observer without querying live contract state — following OCP's recompute → compare → confirm model.

[...]

PR #1788, opened 2026-06-03
Add ERC: Observation Commitment Protocol

damonzwicker:
Adds a draft ERC for the Observation Commitment Protocol (OCP) — a minimal proof envelope format, extraction rule, and verification invariant for independently confirming that a specific byte sequence was committed on-chain.

Discussions-to: https://ethereum-magicians.org/t/draft-erc-observation-commitment-protocol-ocp-chain-agnostic-cryptographic-commitment-primitive/28399


JimmyShi22:
ERC-8274 (AI Inference Proof Verification Interfaces) explicitly names OCP as the unifying verification primitive. The Rationale section describes all `IProofVerifier` implementations as designed in accordance with OCP's `recompute → compare → confirm inclusion` primitive — each backend executes that primitive for a specific proof system (zkML, TEE, oracle, multisig).

The composability is direct: `IProofVerifier.verify()`'s `inputHash` parameter is the same digest OCP anchors on-chain for the same inference event. The anchoring layer (ERC-8263) and the verification layer (ERC-8274) both reduce to OCP's core invariant at the commitment level.

Happy to see this formally in the ERC process. This is the right foundation.

Echo-Merlini:
ERC-8004 (Validation Network Interface) and the wyriwe attestation pipeline compose directly with OCP. In our live implementation — a 3-node ccip-router mesh running on mainnet — every CCIP-Read response generates a wyriwe EIP-712 attestation where `inputHash = keccak256(sender || calldata)`. That digest is what OCP anchors. `AttestationIndex` (`0xc7BCCD785Fb994e570d0ca10D0F7899d87C82210`, Ethereum mainnet) stores `commitmentHash → signerOf` and `inputHash → commitmentHash`, which is OCP's `confirm inclusion` step implemented permissionlessly.

What's notable from the implementation side: OCP's minimal footprint is precisely what makes this work. It doesn't mandate a proof system or a chain — it just defines the commitment shape. That's what allows wyriwe (EIP-712 multisig), ERC-8274 (zkML/TEE), and ERC-8263 (on-chain anchoring) to all reduce to the same primitive without any of them absorbing the others.

The right foundation. Happy to see it formalised.

ERCS/erc-9999.md
```diff
@@ -0,0 +1,130 @@
+---
+eip: 9999
```
```suggestion
eip: 8281
```
@damonzwicker 

ERCS/erc-9999.md
```diff
@@ -0,0 +1,130 @@
...
+title: Observation Commitment Protocol
+description: A minimal proof envelope and verification invariant for independently confirming that a specific byte sequence was committed on-chain.
+author: Damon Zwicker (@damonzwicker) <damon@ocp-labs.org>
+discussions-to: https://ethereum-magicians.org/t/draft-erc-observation-commitment-protocol-ocp-chain-agnostic-cryptographic-commitment-primitive/28399
```
```suggestion
discussions-to: https://ethereum-magicians.org/t/erc-8281-observation-commitment-protocol-ocp/28399
```
Updated Eth Magicians title with assigned number and ERC title

ERCS/erc-9999.md
damonzwicker: Done — updated to eip: 8281 and renamed file to erc-8281.md. Thank you.

ERCS/erc-9999.md
damonzwicker: Done — updated to the ERC-8281 URL. Thank you.

TruthAnchor-AI:
Endorsing this from the ERC-8263 side.

OCP isolates the right primitive — a chain-agnostic cryptographic commitment with a clean observation / commitment / verification boundary — and keeps it deliberately narrow. That narrowness is exactly what makes it composable instead of overlapping with the specs that build on top of it.

ERC-8263 (on-chain action commitments for AI agents) composes with OCP directly. An ERC-8263 anchor records an opaque 32-byte commitment to an agent action; OCP's commitment shape is one valid instantiation of that field. ERC-8263 v0.2 cites OCP's boundary statement verbatim for precisely this reason — it lets the write side (anchoring an action) and the commitment primitive remain separately governed while still interoperating cleanly.

The composition is descriptive, not normative — each spec governs itself. From the ERC-8263 side the boundary is clean and the dependency direction is correct.

— Vincent Wu, author of ERC-8263

damonzwicker:
Requesting editor review when bandwidth allows.

This revision addresses the following before moving to Review status:

- Removed `timestamp` from the `Recorded` event (breaking change from v0, documented)
- Added `receipt_log_position` to the proof envelope for unambiguous log selection
- Added `chain_id` endpoint verification to the extraction rule
- Constrained `hash_function` to a closed, named registry (multihash identifiers)
- Added extraction profile registry with two registered profiles: `recorded` (default) and `erc8263-anchorproof`
- Added ERC-165 interface ID (`0xb5c645bd`), printed topic-0 literal (verified by cast)
- Added finality, reorg, RPC trust, and low-entropy observation security considerations
- All asset files in `assets/erc-8281/`: interface, contract, reference verifier, 21-vector test suite, test runner (26/26 pass)
- Reference verifier published as `ocp-verify@2.0.0` on npm

All 9 CI checks passing. Happy to address any feedback.

hilljessica8787-dot (approved):

damonzwicker:

ERC-8281 — Observation Commitment Protocol (OCP) [#1788]: the shared verification primitive the others lean on. CI green, 21/21 test vectors, reference verifier ocp-verify@2.0.0 on npm, deployed on Ethereum mainnet + Arbitrum One; initial-pass comments addressed.
ERC-8274 — AI Inference Proof Verification Interfaces [#1771]: the verification layer. CI green, editor-ready.
ERC-8299 — WYRIWE (Input Provenance) [#1810]: input-provenance layer; composes onto ERC-8004 (identity) and ERC-8274 (verification). All 9 CI checks green, open.
ERC-8301 — AI Agent Execution [#1815]: execution-attestation layer. CI green, editor-ready.
ERC-8323 — Source-Token Agent Binding [#1851]: source-binding for ERC-8004 agents. CI 9/9 green; one editor review outstanding.
Endorsements across the set note clean composition (8263 / 8274 / 8004 authors on OCP as the shared primitive). We’re happy to make any changes — just surfacing the cluster in case individual PRs fell off the queue. Thank you for your time.

damonzwicker:
Quick note: I’ll be offline for a health-related reason starting Friday, 7/17/2026 and may be slow to respond for a bit — nothing needed from me right now, just flagging so a quiet stretch doesn’t read as me dropping this. Back to it as soon as I can.

babyblueviper1:
Appreciate the heads up, Damon -- take the time you need, nothing here is pressing. Hope it goes smoothly.

giskard09:
The envelope and verification invariant are tightly scoped, and the "no revocation" / "application-layer semantics" boundary is explicit (Out of Scope, §No revocation) — which is the right call for a single-shot commitment.

One thing worth flagging for implementers building on top: the spec proves that *one* byte sequence was committed at *one* point in time, but says nothing about linking a sequence of commitments to the same subject as it's superseded or amended over multiple `record()` calls — no parent/successor reference in the envelope, no ordering rule. That's a real gap for anyone whose observation isn't a one-shot fact.

We hit this building a permissionless on-chain anchor around the same recompute→compare→confirm invariant (byte-exact recompute verified live against this working group's own `/ledger #236`, PR#3/#4 open in `trustless-ai/cross-reference-console`) — ended up adding a parent/root-reference layer above the anchor, outside this spec's boundary. Not proposing a change to the envelope, just noting it might be worth a one-line pointer in §Out of Scope alongside identity/authorship, so implementers know that layer isn't accidentally omitted.

babyblueviper1:
Same shape we hit building `verdict_outcome` on top of our own single-shot commitment layer (`/ledger`, `POST /review sign=true`) -- worth naming since it converged independently.

We solved it the same way you describe: no parent/successor field on the original commitment. Instead, a *new*, separate entry type (`verdict_outcome`) carries a one-directional `cites_decision_ref` pointing back at the original -- the cited entry is never touched, never re-signed, never gains a field. Multiple citations can point at the same decision_ref (e.g. two parties disputing whether an outcome held), and the read side derives a `current` view by walking citations forward, never by mutating the thing being cited.

The layer-boundary discipline matters more than it sounds: this only works because "record a fact" and "cite a fact" are structurally different operations from day one. If OCP's `Recorded` event ever grew a `parent`/`successor` field, you'd be pulling the sequencing/ordering problem INTO the commitment primitive -- exactly the kind of scope creep the spec's own "no revocation, out of scope" boundary is protecting against. A pointer, one layer up, that references the commitment without the commitment knowing about it, keeps OCP itself unchanged while still closing the gap.

One place this bit us that might be relevant to your implementers note: a citation can itself be contested (two citations claiming different things about the same subject) -- worth deciding up front whether "sequence of commitments to the same subject" needs a single ordering rule, or a diverge-and-let-the-reader-resolve model. We went the second route (citations are permanent and additive, "current" is a derived read, never a mutation) specifically so a bad-faith or simply-wrong citation can't suppress or overwrite the earlier honest record.

giskard09:
Same layer-boundary discipline on our end — `delegation-chain-ref-v1` uses `parent_delegation_ref` the same way: a pointer one layer up, the cited entry never mutated, never re-signed. The thing that clicked reading your `cites_decision_ref`: our chain is single-parent (each hop points back one), yours is many-to-one (multiple citations converging on the same `decision_ref`, reader derives `current` by walking forward). That's a real generalization — worth keeping in mind if OCP implementers ever need "N parties disputing one commitment" instead of "a linear chain of amendments."

Your diverge-and-let-the-reader-resolve call matches a principle we landed on independently: `source_class` should never silently upgrade trust by walking hops (min-rule, not max). Good to see it converge again, different shape, same reason.

babyblueviper1:
Right characterization, and the reason it's many-to-one rather than single-parent: a verdict can be disputed by more than one independent party, and we don't want the SECOND dispute to silently supersede the first the way a linear chain would -- each `cites_decision_ref` citation stands on its own, the reader sees every citation and reconciles (or doesn't), rather than the protocol picking a winner by "whoever cited last." That's the same instinct as your min-rule on `source_class`: don't let structure quietly promote confidence past what any single hop actually earned.

Real test for whether "N parties disputing one commitment" ever bites `delegation-chain-ref-v1`: does a delegation ever get re-derived by two principals independently (e.g. two different authorities both issuing a delegation that traces back to the same root grant, and a reader needs to see both rather than have one silently override)? If a delegation chain is always single-issuer by construction, single-parent is the right and sufficient shape -- the many-to-one generalization only earns its complexity when genuine multi-party disagreement over the SAME antecedent is possible, not just because it's more general in the abstract.

giskard09:
Confirmed against the spec (`delegation-chain-ref-v1.0`): single-parent is right and sufficient as designed today. Chain continuity (invariant 1) is strictly linear — `hops[i].delegatee == hops[i+1].delegator`, and each hop's `delegator` is a single string, not a set. There's no field or mechanism for two independent authorities to both claim a delegation tracing to the same `root_delegator` and require reconciliation — if the root issues two separate grants, they surface as two distinct `chain_artifact`s (different `chain_id`), not a dispute over the same antecedent, since nothing compares or ranks chains against each other (cross-chain scope narrowing is explicitly out of scope). So the answer to your test: always single-issuer by construction, at least until we add a notion of a contested/re-derived grant — which doesn't exist yet.

damonzwicker:
@giskard09 — apologies for the slow reply; I was offline for a health reason and this got buried under the CI work on return. Your original read was right, and the subsequent thread here spelled out exactly the generalization you named: single-parent chains and many-to-one citation graphs as two shapes of one rule, the cited artifact inert in both, sequencing and current-state interpretation living in the layer that references it.

Added to §Out of Scope alongside identity and authorship: "This ERC does not define linkage between successive commitments — supersession, amendment, or parent/successor ordering across multiple record() calls. A committed digest is a single-shot fact; sequencing and current-state interpretation are the responsibility of the layer that references it."

Thanks for the careful read and for confirming single-parent is right and sufficient for delegation-chain-ref-v1 by construction. The many-to-one generalization earns its complexity exactly where Fede's test named it: when genuine multi-party dispute over the same antecedent is possible, not just because it's more general in the abstract.

giskard09:
No apology needed — glad the framing held up and landed as a clean §Out of Scope line. Good place to draw it.

Frank122480:
> Confirmed against the spec (`delegation-chain-ref-v1.0`): single-parent is right and sufficient as designed today. Chain continuity (invariant 1) is strictly linear — `hops[i].delegatee == hops[i+1].delegator`, and each hop's `delegator` is a single string, not a set. There's no field or mechanism for two independent authorities to both claim a delegation tracing to the same `root_delegator` and require reconciliation — if the root issues two separate grants, they surface as two distinct `chain_artifact`s (different `chain_id`), not a dispute over the same antecedent, since nothing compares or ranks chains against each other (cross-chain scope narrowing is explicitly out of scope). So the answer to your test: always single-issuer by construction, at least until we add a notion of a contested/re-derived grant — which doesn't exist yet.

Why are we talking diverting I am not quite understanding now. If transaction gets diverted. Where does it get diverted to
Plz let me know thoughts or is the protocol to open for discussion
Frank

babyblueviper1:
Hey Frank — fair question, that got dense fast. Nothing about a transaction being redirected or funds moving anywhere here. "Diverge"/"dispute" in this thread means something narrower: if two different parties each record a claim about the same earlier fact (e.g. two people each saying "here's what actually happened with X"), does the protocol pick one as the winner, or does it keep both and let whoever's reading later decide which one to trust? We're arguing for the second option (keep both, don't auto-pick a winner) specifically so a later claim can't quietly erase or override an earlier honest one. "Diverting" isn't the right mental model — think "two witnesses disagreeing," not "a payment getting rerouted."

And yes — this is a public ERC spec repo, genuinely open to anyone weighing in, no special standing needed. Happy to walk through any specific part again in plainer terms if that helps.

damonzwicker:
@SamWilsn — quick follow-up on ERC-8281. The earlier editorial comments have been addressed, all checks are green, and it appears the only remaining merge requirement is the editor/code-owner review. If this is in your review queue, happy to address anything else you find. Thanks.

PR #1790, opened 2026-06-03
Add ERC: Wallet-Scoped Token Pull Execution

wenzhenxiang:
The common `approve`-then-call pattern is also cumbersome and gas-expensive. It often requires a separate approval transaction before settlement, plus a later revocation if the user wants to remove the allowance.

This ERC standardizes a narrow pattern:

- the wallet wants to call an external target;
- the target may need to collect one specific ERC-20 token during that call;
- the wallet wants temporary, target-bound, asset-bound, and amount-capped authorization; and
- the wallet does not want to leave behind a reusable allowance.


ERCS/erc-8284.md
```diff
@@ -0,0 +1,243 @@
+---
```
abcoathup, line 1:
```suggestion
---
eip: 8284
```
@wenzhenxiang 

ERCS/erc-draft_wallet-scoped-token-pull-execution.md
```diff
@@ -0,0 +1,243 @@
+---
+title: Wallet-Scoped Token Pull Execution
+description: Wallet primitive for temporary ERC-20 pull authorization during one external call.
+author: Xiang (@wenzhenxiang)
+discussions-to: https://ethereum-magicians.org/t/erc-wallet-scoped-token-pull-execution/28691
```
abcoathup:
```suggestion
discussions-to: https://ethereum-magicians.org/t/erc-8284-wallet-scoped-token-pull-execution/28691
```

hilljessica8787-dot (approved):

PR #10, opened 2023-10-25
ERC-7540: Asynchronous ERC-4626 Tokenized Vaults

Joeysantoro:
ported from https://github.com/ethereum/EIPs/pull/7876

PR #1808, opened 2026-06-10
Add ERC: Validation Network for ERC-8004

Traciecmyers:
This PR adds a draft ERC extending ERC-8004's Validation Registry with a 
standard contract interface for permissionless, operator-diverse validator 
networks.

Discussion: https://ethereum-magicians.org/t/erc-8004-validation-network-interface-extension-for-multi-validator-networks/28669

Reference implementation and full spec history: 
https://github.com/pokt-network/erc-8004-vni

Authors: Chris "Jinx" Jenkins (Pocket Network Foundation), Luis Correa de León 
(Synaptika), Bryan White (Pocket Network Foundation), Tiago Merlini.

The spec is strictly additive to ERC-8004 — single-address validators continue 
to work unchanged. Substrate-neutral by design; admits permissionless RPC 
networks, restaking AVSs, TEE consortia, and decentralized oracle networks on 
equal footing.

Open as Draft for review by EIP editors and continued discussion on Ethereum 
Magicians.


ERCS/erc-1808.md
```diff
@@ -0,0 +1,463 @@
+---
+eip:
```
abcoathup:
```suggestion
eip: 8294
```
@Traciecmyers 

ERCS/erc-1808.md
```diff
@@ -0,0 +1,463 @@
...
+title: Validation Network Interface for ERC-8004
```
abcoathup:
```suggestion
title: Validation Network for ERC-8004
```
Interface is superfluous in an ERC title

ERCS/erc-1808.md
```diff
@@ -0,0 +1,463 @@
...
+description: A standard interface for permissionless, operator-diverse validator networks plugging into ERC-8004's Validation Registry
+author: Chris "Jinx" Jenkins (@TheFeloniousMonk), Luis Correa de Leon (@luyzdeleon), Bryan White (@bryanchriswhite), Tiago Merlini (@TMerlini)
+discussions-to: https://ethereum-magicians.org/t/erc-8004-validation-network-interface-extension-for-multi-validator-networks/28669
```
abcoathup:
```suggestion
discussions-to: https://ethereum-magicians.org/t/erc-8294-validation-network-interface-for-erc-8004/28669
```
Updated Eth Magicians title with assigned number and current ERC title

Traciecmyers:
Hi all — friendly check-in here, approaching two weeks since submission. PR remains as-is from abcoathup's review pass (suggestions applied, all CI green). Discussion on the Magicians thread has continued substantively, with engagement from several reviewers across the broader agent-verification stack. Available for any further feedback or to address questions whenever the second editor review fits the queue.

Thanks

hilljessica8787-dot (approved):

MarkovianProtocol:
Nice work, and the substrate-neutral framing is the right call. One class worth checking the interface against: every validator type listed here (RPC networks, restaking AVSs, TEE consortia, oracle networks) derives assurance from operator diversity, you trust a set. There is a categorically different validator whose assurance comes from reproducibility instead: a deterministic-replay validator, where validation is recomputing the outcome from public data so any party reproduces the same result and checks it against an external anchor. It needs no operator set, because there is nothing to trust beyond the public inputs and the function.

Does the ERC-8294 interface admit a single verifier whose output is independently recomputable (assurance from reproducibility, not operator-set diversity)? If it does, that seems to strengthen the substrate-neutral claim rather than complicate it. Honest scope: this only applies to outcomes decidable from public data by a deterministic function; subjective or private-input validation still needs the operator-diversity models you describe.

Traciecmyers:
@MarkovianProtocol  — thanks, and this is a useful clarification. Yes, the interface admits exactly the case you describe.

The `SelectionPolicy` explicitly allows `selectionSize` = 1 and `minOperators` = 1 (the Assurance Tiers table calls this Tier 1: "cheap signal, equivalent to single-validator"). That configuration is the entry point for any single-verifier network — including a deterministic-replay validator where assurance comes from reproducibility rather than diversity. The interface doesn't require a validator set; it requires a validator to be selected, and N=1 satisfies that.

The `verificationProfile()` return value is the surface for exactly the class distinction you're drawing. A deterministic-replay network would return something like keccak256("deterministic-replay-v1") — announcing to clients that assurance derives from public-input reproducibility rather than signature aggregation over a diverse set. That's already the intended semantics of the field: not a taxonomy of "how many validators," but a taxonomy of "what trust model backs the result." A client verifying an attestation from a deterministic-replay profile knows to recompute the function against the anchored public inputs rather than to weight signatures across an operator set.

Agree this strengthens rather than complicates the substrate-neutral framing. Two validation-model families coexist under the same interface — reproducibility-backed for deterministic public-input cases, operator-diversity-backed for subjective or private-input cases — and the honest scope you note (deterministic function over public data) is exactly the right boundary. The Rationale prose currently underemphasizes the first family; folding an explicit call-out into the next revision is a fair observation, and we'll queue it alongside the other refinements from this discussion.

MarkovianProtocol:
Thanks, that mapping is clear, and `verificationProfile()` is the right surface for it.

One refinement worth putting on the record. At N=1, a diversity-based validator and a reproducibility-based validator sit at opposite ends of the assurance spectrum, not at the same tier. A single diversity validator is a genuinely cheap signal: you are trusting one operator. A single deterministic-replay validator is the opposite: assurance is maximal, because the result is not attested, it is recomputed. Any consumer runs the same public inputs through the same function and checks the output against an external anchor. There is no operator to collude with and nothing to trust beyond the inputs.

The taxonomy risk is that `verificationProfile()` lets a replay validator read as "cheap Tier 1" when it belongs near the strongest tier. It would help for the profile to carry an explicit assurance-basis field (diversity vs reproducibility), so a consumer can distinguish "one operator said so" from "anyone can re-derive this," rather than inferring assurance from validator-set size alone.

On the concrete side: a reference implementation of this shape is public and re-derivable, with the input pre-commitment step canonicalized (RFC 8785) and Bitcoin-confirmed via OpenTimestamps (github.com/MarkovianProtocol/eval-anchor-vector). Happy to contribute the full `validationResponse` shape as a reference for the reproducibility profile if useful.

MarkovianProtocol:
@Traciecmyers thanks, that resolves it cleanly. The N=1 `SelectionPolicy` plus a `verificationProfile()` that names the trust model is exactly the seam that lets both families coexist under one interface without special-casing either.

One point worth making explicit for the reproducibility-backed family, because it changes what the client must do. Under a deterministic-replay profile, the attestation is only as trustworthy as the immutability of the inputs it was replayed against. A signature-diversity profile draws its assurance from the operator set at verification time. A replay profile draws it from the inputs being fixed *before* the result was produced. So the profile's contract to the client is really two clauses: (1) recompute the function over the referenced inputs, and (2) verify those inputs were committed prior to the claimed evaluation. Without (2), replay proves determinism but not honesty, since an operator could select inputs after seeing the output it wanted.

That mechanism doesn't belong in the interface. Input pre-commitment is out of scope and substrate-neutral, a TSA timestamp or a chain anchor, whatever the deployment uses. But a one-line note in the Rationale that a replay profile presumes input pre-commitment would keep implementers from shipping the weaker version.

Happy to draft the Rationale call-out for both families if it saves a round, and to contribute a minimal deterministic-replay `verificationProfile` reference so the two-family claim has a runnable example in-tree.

MarkovianProtocol:
Delivering on the offer from the last note: a drop-in Rationale call-out and the reference shape for the reproducibility-backed family. Both kept minimal and non-normative so they ride with the existing revisions rather than expanding scope.

**Rationale call-out (drop-in, in the style of the existing "Why..." entries):**

**Why reproducibility-backed validation is a distinct assurance basis.** The verification profiles enumerated above (signature-only, indexer-attested, light-client-backed, zk execution proof, TEE execution proof) all locate residual trust in a set: after signatures verify, a client still relies on the diversity or honesty of the operators, provers, or enclaves that produced the attestation. A second family locates assurance in reproducibility. Under a deterministic-replay profile the attested outcome is a pure function of referenced public inputs, and the relying party does not trust the attestation, it recomputes the function over those inputs and compares. This is the one profile under which the "network of independent validators does not mean end-to-end trustlessness" caveat above does not bind the client, because the client re-derives the verdict rather than accepting the tally. The Assurance Tiers table measures operator diversity and is orthogonal to this axis: a deterministic-replay network at `selectionSize = 1` is not a Tier 1 cheap signal but the strongest assurance the interface admits, since one recomputation settles it. `verificationProfile()` is the correct surface for the distinction, and a relying party branches on it: for a reproducibility profile it recomputes, for a diversity profile it weights signatures. A reproducibility profile is sound only if the referenced inputs were committed before the outcome was produced, otherwise replay proves determinism but not honesty. That pre-commitment is out of scope for this interface and substrate-neutral (a chain anchor or an RFC 3161 timestamp), and it composes with the `wyriwe-input-provenance-v1` challenge kind, which already returns a committed `inputHash` as evidence.

**Minimal deterministic-replay profile (reference shape):**
- `verificationProfile()` returns `keccak256("deterministic-replay-v1")`.
- Response URI evidence (JCS canonical JSON) carries: the function identifier and version, a content-addressed reference to the inputs, the committed `inputHash` (or a `wyriwe-input-provenance-v1` chain), the recomputed output, and the pre-commitment anchor. `responseHash` binds this file.
- Client contract, two clauses: (1) recompute the named function over the referenced inputs and check the output hash equals the attested `responseHash`; (2) verify the `inputHash` was anchored at or before evaluation, via the referenced challenge or anchor. If both hold the verdict stands with no operator to trust; if either fails, reject.

A reference implementation of this shape is public and re-derivable: the input pre-commitment step is canonicalized (RFC 8785) and Bitcoin-confirmed via OpenTimestamps at github.com/MarkovianProtocol/eval-anchor-vector. Wiring it as a `validationResponse` against the ERC-8004 Validation Registry is the PR I am offering to open, with the testnet response linked as the runnable reference.

If useful I can open that PR: a one-line reproducibility example in the Verification Profile list, the Rationale call-out folded in, and a `deterministic-replay-v1` note in the evolving challenge-kind / profile registry.

MarkovianProtocol:
Following up with a runnable reference for the reproducibility-backed profile, now live on testnet.

A deterministic-replay `validationResponse` is recorded against the ERC-8004 Validation Registry on Base Sepolia (`validatorAddress` = the replay validator, agentId 7816, tag `markovian.resolve.v1`). The claim is decidable only from public data: whether Bitcoin block 957110's hash is even. It was canonicalized (RFC 8785), hashed to the `requestHash`, and OpenTimestamps-anchored before that block existed, so the inputs were fixed before the outcome (the pre-commitment clause). When the block mined, the function was re-run, the verdict written on-chain, and the evidence pinned at a stable URI.

The point is that nobody trusts the validator. A short `verify.py` reads only public endpoints (the registry via RPC, the evidence over HTTPS, the block hash from a public explorer) and recomputes every step:
- on-chain `requestHash` == SHA256(canonical claim): PASS
- on-chain `responseHash` == SHA256(canonical evidence): PASS
- verdict reproducible from block 957110: PASS

Registry read (`getValidationStatus`): https://sepolia.basescan.org/tx/0x7670aceb570cc418aed54e99801d06fb50bce36268a3e1912bb6b84b55814a3b
Verifier + evidence: https://github.com/MarkovianProtocol/erc8004-resolution-validator/tree/main/live-demo

This is what a `verificationProfile()` of `keccak256("deterministic-replay-v1")` announces to a client: recompute, do not trust the tally. Offering it as the runnable example behind the Rationale call-out from the earlier note, and happy to shape it into an in-tree reference if that helps the revision.

hilljessica8787-dot (approved):

Traciecmyers:
@MarkovianProtocol — strong contribution, and the two-family framing is the right one. The substance here lands.

On the Rationale call-out. This is well drafted — you've got the distinction exactly right, and it slots naturally into the existing "Why…" pattern. The enumerated profiles all locate residual trust in a set; a reproducibility profile relocates assurance to recomputation, which is why the "independent validators ≠ end-to-end trustlessness" caveat doesn't bind a client that re-derives the verdict. And framing the Assurance Tiers table as orthogonal to this axis rather than in tension with it is the cleanest reconciliation available. We'll work it into the next revision — some shaping for voice and to keep the Rationale scoped to merged, citable surfaces, so the wyriwe-input-provenance-v1 composition note is the one piece we'd leave out as a deployment-side detail. The pre-commitment condition itself we're carrying.

On the two-clause client contract. Recompute over referenced inputs, and verify the inputs were anchored at or before evaluation — that's the right decomposition, and the second clause is the load-bearing one. Replay without pre-commitment proves determinism, not honesty. We read that as belonging in Security Considerations as much as Rationale, since it's really a description of how a replay profile gets gamed if an implementer skips it.

On the live reference. Useful, and the anchoring construction is the interesting part — canonicalized, timestamped before the outcome existed, verdict re-derivable from public endpoints. We'll run the verifier ourselves before citing it anywhere; that's the whole premise of the profile, so it'd be odd to take it on trust.

On the PR offer. Yes — open it against the companion repo (pokt-network/erc-8004-vni) rather than the ethereum/ERCs PR. That keeps #1808 at the state the editors have already reviewed, which matters for the merge path, and lets the profile work land in the next spec revision. Scope as you outlined is fine, with one preference: keep verify.py and the live demo in your repo and cite it, rather than co-mingling implementation with spec.

— Pocket Network VNI Group

Traciecmyers:
Hi all — follow-up to my earlier check-in, since there's been substantive movement worth noting.

Since my last comment, MarkovianProtocol contributed a materially significant refinement plus a live testnet reference implementation for the reproducibility-backed validation profile class:

- A Rationale drop-in (in the style of the existing "Why..." entries) articulating reproducibility-backed validation as a distinct assurance basis from operator-diversity-backed validation
- A minimal deterministic-replay profile shape with a two-clause client contract
- A live `validationResponse` on Base Sepolia against the ERC-8004 Validation Registry, with a canonicalized claim (whether Bitcoin block 957110's hash is even), OpenTimestamps pre-commitment anchoring, and a `verify.py` that reproduces every step from public data alone: registry via RPC, evidence over HTTPS, block hash from a public explorer
- A companion PR being coordinated against the Pocket Network VNI companion repo, folding the Rationale call-out, a Verification Profile entry, and a `deterministic-replay-v1` challenge-kind note into the next spec revision

This is the reference implementation the spec anticipated. It's independent of the Pocket implementation, which is exactly the substrate-neutral demonstration the spec framing intended.

Follow-up commitments now tracked for the next spec revision (post-merge):
1. MUST-carry-complete-set language for aggregated response (TMerlini)
2. ERC-8299 citation in restored composition note (Damon Zwicker + TMerlini)
3. Shared operatorDiversity derivation methodology (coordination with Brooks1003)
4. Reproducibility-backed model class Rationale call-out (MarkovianProtocol, delivered)
5. Assurance qualifier semantics for verificationProfile() (MarkovianProtocol, delivered)
6. deterministic-replay-v1 profile registry entry (MarkovianProtocol, delivered)

PR remains as-is from abcoathup's review pass with hilljessica8787-dot's approval. All CI green. Available for any further editor feedback whenever the second code owner review fits the queue.

Thanks!

giskard09:
Hi @Traciecmyers — the reproducibility-profile framing in the recent revisions is a clean addition, and the pre-commitment caveat in the rationale ("that pre-commitment is out of scope for this interface and substrate-neutral — a chain anchor or an RFC 3161 timestamp, whatever the deployment uses") is the right place to be honest about what's load-bearing there.

One concrete option for the "chain anchor" case: `AnchorRegistry`, a single-function permissionless contract (`anchor(bytes32 ref)`, no owner, no access control) deployed at the same address via CREATE2 on Arbitrum, Base, and Ink (`0x49fEcA52bC634a9Ab773226D16619deC547794aa`). Any deployment can anchor a commitment hash and get a queryable `Anchored(ref, anchoredBy, timestamp)` event back — no relationship with us required, no fee beyond gas.

Worth being precise about what that does and does not buy you, since it's exactly the gap that matters for a reproducibility profile: an anchor proves a hash existed at or before a given block. It does not by itself prove the anchorer did not wait to see how things would resolve before anchoring — that still depends on the anchored value being derived from inputs the anchorer could not have controlled (e.g., a not-yet-mined block height, as in the deterministic-replay example upthread). The anchor is the timestamp primitive; the "committed before the outcome was knowable" property still has to come from the claim's own structure.

Reference: `giskard-payments/src/AnchorRegistry.sol`, and two worked examples anchoring external artifacts on Base (commits `cd74d55`, `c4a1b17` in giskard09/*).

Traciecmyers:
@giskard09 — appreciated, and the precision in the second half is the useful part. You've drawn exactly the right line: an anchor proves a hash existed at or before a block, but not that the anchorer didn't wait to see how things resolved before anchoring. The "committed before the outcome was knowable" property has to come from the claim's own structure — a not-yet-mined block height, an unresolved external event — not from the anchor primitive itself. That's the right sharpening of the pre-commitment note, and it's the distinction implementers most need to hear.

On AnchorRegistry specifically: we're keeping the Rationale's anchor language substrate-neutral rather than naming a particular deployment — which is the framing you're already endorsing ("a chain anchor or an RFC 3161 timestamp, whatever the deployment uses"). The spec's job is to state the property a reproducibility profile presumes — inputs committed before the outcome is knowable — and leave the mechanism to the deployment. Privileging any single contract or address in-spec would cut against exactly that neutrality. A permissionless, ownerless CREATE2 anchor is a perfectly reasonable thing for a deployment to reach for; it just lives at the implementation layer, not in the interface or its Rationale.

Thanks for the careful framing — the "what it does and doesn't buy you" precision is the right note for this whole class.

— Pocket Network VNI Group

giskard09:
Makes sense, and it's the right call for the spec — a Rationale that named a contract would be a worse Rationale, not a better-credited one. The property (committed before the outcome is knowable) living in the interface while the mechanism stays at the deployment layer is exactly the right cut — same reason our own draft treats an on-chain anchor and a SCITT receipt as interchangeable at the mechanism layer without touching the derivation itself. Appreciate you spelling out why, not just the decision.

Traciecmyers:
@giskard09 exactly, and the SCITT/anchor interchangeability on your side is the same cut in a different substrate: the receipt and the on-chain anchor are both just the timestamp primitive, and the derivation is what carries the "committed before knowable" property regardless of which one a deployment picks. That's the shape the profile class wants. Appreciated the careful framing throughout; this is the kind of exchange that makes the neutrality claim more defensible, not just asserted.

— Pocket Network VNI Group

giskard09:
Likewise — good exchange.

0rkz:
Disclosure: I work on PayPerByte/BYTEDev, which runs live x402 oracle feeds; one of them is the example below.

TMerlini's `IValidatorMetadata` proposal in the discussions-to thread (Magicians 28669, post #10), specifically the `ValidatorType.Judgment` case and "a judgment score's credibility is inseparable from the track record behind it", matches what we hit shipping one.

We run merchant-screen, a Judgment-type verifier: an opinionated ALLOW/WARN/BLOCK over `(domain, address, observed_price)`, returned before an agent settles an x402 payment. Making that output usable by someone who didn't make the call forced three things into the response:
- a pinned ruleset id (`methodology: "ms-v1"`) rather than an opaque score, with each signal's method disclosed per field
- a signer distinct from the party being judged (our own per-feed key, never the counterparty)
- a signature over the exact response bytes, so the verdict is tamper-evident and replayable to a third party

Two smaller ones: a 0–100 `score` beside the verdict, where an unverifiable core input reports `unverified` and can at most pull the verdict to WARN (it never produces a BLOCK by itself), and the scope written into the schema's normative text: authenticity of the verdict bytes, never correctness of the judgment.

Question for the spec: should `Judgment`'s `getSummary()` require a method/ruleset id plus a signed-bytes commitment, so reputation accrues to a `(validator, ruleset)` pair rather than the validator alone? That's the gap between `recordPointer` meaning "this validator, generally" and "this validator running this exact method."

Live: `POST https://x402.payperbyte.io/feeds/merchant-screen` (x402-gated; returns a 402 challenge unpaid, no signup wall). Schema: https://x402.payperbyte.io/openapi.json

Traciecmyers:
@orkz — thanks for this. Production operational input on the judgment-verifier case is exactly the empirical grounding the design conversation needed. The three requirements you surface (pinned ruleset id, distinct signer, signature over exact response bytes) aren't accidents of your architecture — they're the necessary conditions for judgment output to be portable to a third party. Names them well.

**On the (validator, ruleset) pair as the reputation unit.** This is a genuinely stronger frame than where we'd previously landed. Our thinking on TMerlini's IValidatorMetadata proposal (Magicians 28669 #10) had been to redirect the judgment-specific concern away from a generic metadata surface and toward either (a) an off-chain aggregated-response field or (b) a separately-scoped IJudgmentValidator interface, defaulting toward the lighter-weight option. Your point that a single validator running two methodologies produces two distinct reputational surfaces is exactly the kind of concrete evidence that shifts the calculus toward the interface option: a normative shape that forces (ruleset, signature) to travel with the verdict, so reputation accrues to the pair rather than the validator alone.

**On the getSummary() shape.** If we do go with a judgment-scoped interface, requiring a method/ruleset id plus a signed-bytes commitment is the right normative shape — otherwise implementers each invent their own version of these fields and clients need per-verifier code, which is exactly the fragmentation the spec exists to prevent. Two implementation questions worth thinking through:

1. **Ruleset identifier format.** Free-form string vs bytes32 vs URI-resolvable identifier. Your `ms-v1` convention is human-readable; the tradeoff is stability guarantees when methodology evolves. We'd probably want a monotonic-version discipline baked into the spec so `ms-v2` can never silently replace `ms-v1`.
2. **Signed-bytes scope.** Whether the signature covers only the verdict bytes, or the verdict bytes plus the input tuple. Your language ("signature over the exact response bytes") suggests the former. That's cleaner for replayability but requires the input tuple to be carried in the signed payload rather than implied by context. Worth explicit language either way.

**On the authenticity vs. correctness scoping.** This is the observation that most cleanly generalizes. "The signature attests to who said what, not to whether they were right" is the honest scope of any judgment surface — and it's the same distinction MarkovianProtocol drew for reproducibility profiles earlier in the thread, from the opposite direction. Reproducibility gets you correctness verifiability directly; judgment can only get you authenticity, and correctness has to come from something else (reputation, methodology audit, external validation). Naming this scope in the normative text of a judgment-scoped interface is worth doing.

**On the "unverified → cap at WARN" pattern.** Generalizable, and the right conservatism default: a Judgment verifier can degrade confidence when a core input can't be resolved but shouldn't manufacture a stronger negative signal than what the input actually supports. Whether the spec should encode this normatively or leave it to implementer discretion is a fair question; we'd lean toward normative encoding for the same fragmentation-avoidance reason above.

Live endpoint and schema noted — thank you for making it hit-able without a signup wall. Between merchant-screen as a live production Judgment reference and MarkovianProtocol's Base Sepolia deterministic-replay reference, the two-verifier-family framing has concrete grounding rather than being aspirational. All of the above joins the queue of items being folded into the next spec revision, including revisiting the IJudgmentValidator interface question with this concrete input as the shape argument.

— Pocket Network VNI Group

0rkz:
@Traciecmyers — both good questions; answering from what we actually run rather than what I'd like the spec to say.

**On the ruleset identifier format.** Ours are free-form strings — `ms-v1` for merchant-screen, `ar-v1` for address-reputation — and I'd argue for leaving the *encoding* implementer-defined while making the *discipline* normative. The identifier's whole job is to let a third party say "this exact method produced this verdict", and that needs the meaning to be immutable, not the encoding to be a particular width. A monotonic-version rule — an id, once published, never changes behaviour; a behaviour change means a new id — gets you the stability guarantee you're after regardless of whether it's a string, a bytes32, or a URI.

I'd be wary of *requiring* URI-resolvability specifically. It makes verification depend on an endpoint being reachable at verification time, which is the dependency a signed verdict exists to remove — someone checking a six-month-old verdict shouldn't need my server to be up. Resolvable-if-present is useful; resolvable-as-a-precondition inverts the property.

Worth being honest about the limit on our side: monotonicity for us is a convention, not something enforced cryptographically. Nothing stops us reusing `ms-v1` for changed behaviour except our own discipline. That's an argument for the spec encoding the rule normatively rather than trusting each implementer to hold the line — you'd be making explicit the thing that is currently just good manners.

**On signed-bytes scope.** Concretely, in ours: a paid response carries `answer`, `attestation` and `broadcast`, and the verdict signature — made with a per-feed key, distinct from the gateway's transport receipt over the whole body — commits to `keccak256` of the `answer` object's exact bytes plus their length. `answer` carries the input tuple next to the verdict, the ruleset id, and per-signal input hashes, so verdict and input are covered together — but as one opaque byte range, not as separately-typed fields. The signed struct itself is only `(publisher, payloadHash, payloadLength, deadline)`. That's the distinction I'd suggest the spec draw explicitly, because "signed-bytes" currently reads both ways:

- *Signature over a payload that happens to contain the input* — what we do. Byte-exact replayable, and the verifier never has to understand the verdict schema. One cost worth the spec knowing: the verifier has to take that byte range exactly as delivered, and a JSON parse-then-re-serialize round trip breaks the recompute — one writer rendering a number `10.0` where another renders `10` is enough to do it. We hit that in production and had to spell it out in our own verify recipe, so it's a real cross-implementation footgun rather than a hypothetical one.
- *Signature over (verdict, input) as structured fields* — more legible to a verifier that wants to check the input binding without parsing the payload, at the cost of the spec having to fix the field layout.

We picked the former because byte-exact replay was the property we needed, and because it keeps the verifier independent of our schema evolving. But a spec choosing the latter is making a defensible different trade, and the thing that would actually hurt implementers is leaving it ambiguous — that's how you get two verifiers that both claim conformance and can't check each other's output.

**On the unverified → WARN cap.** We do it today, and I'd vote normative for the reason you gave: left to discretion, some implementer will read an unresolvable input as suspicious, and the same verdict string ends up meaning two different things across verifiers.

Worth drafting the rule precisely, though, because ours is narrower than it first sounds. In our ruleset an unverified core signal caps the verdict at WARN and never produces a BLOCK on its own — but it isn't inert. An unresolved domain age and a domain measured as young take the same branch in one of our rules, so an unknown input can still sit inside a BLOCK that a different signal drives. "Cap confidence at WARN" and "an unverified input never counts against the subject" are two different normative rules, and it's worth the text saying which one it means — we'd pass the first today and fail the second.

**On the scoping.** Agreed, and I'd put it in the normative text plainly, in your phrasing: the signature attests to who said what, not to whether they were right. A judgment surface can carry authenticity and tamper-evidence; correctness has to come from somewhere else. Saying so in the spec is what stops the receipt being read as an endorsement.

Disclosure: I work on PayPerByte/BYTEDev — merchant-screen and address-reputation are the live implementations behind these answers, and `https://x402.payperbyte.io/openapi.json` is public if the response shape is useful as a reference.

Prepared with AI assistance, human-reviewed.

0rkz:
Rehearsal artifact. This text exists only to give the dispatcher a delimited SEND region so case F can reach a real would-send. It is never posted: every rehearsal run is --dry.

PR #1810, opened 2026-06-11
Add ERC: WYRIWE — Input Provenance for AI Inference

TMerlini:
## Summary

WYRIWE defines a triple-hash commitment scheme and EIP-712 attestation profile that proves the input a model received matches the input the user intended.

### The gap this fills

ERC-8126 (AI Agent Verification, Final) establishes agent-level trust but explicitly leaves the execution receipt layer out of scope. ERC-8004 attests agent identity — but no standard commits to the *input* before inference runs.

Without a committed input record, a settlement contract cannot verify the output corresponds to the funded input, and a dispute mechanism has no ground truth for what the model was actually asked to do.

### What it specifies

1. **Triple-hash construction** — `raw_input_hash → sanitization_pipeline_hash → input_hash` — forms an auditable, reproducible chain of custody verifiable by any third party without trusting the agent or gateway.
2. **IDENTITY_SENTINEL_CID** — explicit, verifiable no-sanitization case (pinned at `QmccvoM6aRVgZ2dtFWvT6Wm3DmTvoAUHHotK7uQufnStVR`).
3. **WyriweAttestation** — EIP-712 typed struct (8 fields, all required, field ordering normative).
4. **EIP-712 domain** — `ERC8004AttestationGateway`, dynamic `chainId`.
5. **Verification procedure** — 5 deterministic steps.
6. **Gateway query interface** — `GET /agent/verify/:inputHash`.

### Live implementation

Reference gateway: `https://gateway.ensub.org/agent/verify/:inputHash`  
Source: https://github.com/Echo-Merlini/ccip-router  
Mainnet WyriweProofVerifier: `0xd8a09d830b27697e1b24e8c9800e562d20318a09`  
Live interop tx: `0xc32b66ae9446e0d5282a6fc813ba106126a8da05bced638b83840d9c2510e4d0` (mainnet block 25289963)

### Note on ERC number

`eip: 11796` is a plausible placeholder based on the current highest PR in the queue. We defer the actual number to the editors and will rename the file on request.

### Discussion thread

https://ethereum-magicians.org/t/wyriwe-what-you-read-is-what-you-execute-input-provenance-for-verifiable-ai-inference/28655

### DIFF
--- a/ERCS/erc-11796.md
+++ b/ERCS/erc-11796.md
@@ -332,7 +332,7 @@ Slot mapping against WYRIWE's L3 scheme:
 
 The `snapshotRoot` is derived from rows of `(address contributor, uint256 score, uint256 timestamp)` sorted by contributor address ascending, ABI-encoded. The `commitmentHash` is what a node submits during the commit phase; the full snapshot rows are revealed and verified against it during the reveal phase. The commit-reveal scheme closes the "committed → executed" gap by construction: a node can only reveal data whose hash matches its commit.
 
-Reference implementation: `POST /contributions/snapshot/freeze` in [ccip-router v0.6.0](https://github.com/Echo-Merlini/ccip-router).
+Reference implementation: `POST /contributions/snapshot/freeze` in ccip-router v0.6.0.
 
 ### L4 Judgment validator binding (@babyblueviper1)
 
@@ -420,7 +420,7 @@ Domain separator: `ERC8004AttestationGateway` / version `"1"` / `block.chainid` 
 - Entries predating the wiring carry a partial block with `executed_action_hash: null` and an explicit `"not backfilled by design"` status. A commitment you did not make at the time is not one you get to manufacture later.
 - Where a production system records a single timestamp per governance cycle, `executedTimestamp` is `null` with an ordering note rather than a fabricated reveal time. The strict `verdictTimestamp < executedTimestamp` invariant belongs to the on-chain attestation; an off-chain production mapping should record what it actually measured.
 
-Reference implementation: [api.babyblueviper.com/ledger](https://api.babyblueviper.com/ledger) — live production ledger. Entry `/ledger/3` shows a pre-wiring partial block; new entries carry the full `judgment_execution` block. Running against real capital.
+Reference implementation: api.babyblueviper.com/ledger — live production ledger. Entry `/ledger/3` shows a pre-wiring partial block; new entries carry the full `judgment_execution` block. Running against real capital.
 
 5. **`string recordPointer` is the correct type for EIP-712.** The `RecordPointer` struct (Appendix B) is the resolved payload schema — it is NOT inlined into the signed type. The attestation is signed once and frozen at verdict time; the record it points to is alive and grows over time (`outcomeEvidence` does not exist when the verdict attestation is signed). Inlining `RecordPointer` into the EIP-712 struct would require signing a permanently incomplete field. The on-chain anchoring of the verdict artifact (Nostr event ID or IPFS CID in `verdictHash`) already secures commitment integrity independently of the pointer. The 9-field type string is therefore stable and MUST NOT be changed pre-review.
 
@@ -428,7 +428,7 @@ Reference implementation: [api.babyblueviper.com/ledger](https://api.babybluevip
 
 ### Appendix A — Cross-system settlement: GenericCommitRevealSettler integration
 
-This appendix documents the integration pattern for judgment attestations with `GenericCommitRevealSettler` as a reference for any L4 producer. First demonstrated in the cross-system settlement of ledger entry 19 at [api.babyblueviper.com/ledger/19](https://api.babyblueviper.com/ledger/19) — commit block 11030402, reveal block 11030403.
+This appendix documents the integration pattern for judgment attestations with `GenericCommitRevealSettler` as a reference for any L4 producer. First demonstrated in the cross-system settlement of ledger entry 19 at api.babyblueviper.com/ledger/19 — commit block 11030402, reveal block 11030403.
 
 **Contract:** `GenericCommitRevealSettler` on Sepolia: `0xFe7Ab6d95f7567a311B98D029373d0fc1511aCCe`
 
--- a/ERCS/erc-11796.md
+++ b/ERCS/erc-11796.md
@@ -13,7 +13,7 @@ requires: 712, 8004
 
 ## Abstract
 
-This ERC defines a triple-hash commitment scheme and [EIP-712](./eip-712.md) attestation profile for proving that the input a model received is the input the user intended. It introduces three linked fields — `raw_input_hash`, `sanitization_pipeline_hash`, and `input_hash` — that together form a verifiable chain of custody for AI inference inputs. A verifier can confirm input integrity using only the committed hashes and the public sanitization specification, without trusting the agent, gateway, or execution environment. This standard occupies the input-provenance layer of the AI inference trust stack, complementing [ERC-8004](./erc-8004.md) (agent identity), [ERC-8126](./erc-8126.md) (agent verification), TruthAnchorV1 (on-chain proof commitment and anchor layer), and OCP (observation commitment protocol).
+This ERC defines a triple-hash commitment scheme and [EIP-712](./eip-712.md) attestation profile for proving that the input a model received is the input the user intended. It introduces three linked fields — `raw_input_hash`, `sanitization_pipeline_hash`, and `input_hash` — that together form a verifiable chain of custody for AI inference inputs. A verifier can confirm input integrity using only the committed hashes and the public sanitization specification, without trusting the agent, gateway, or execution environment. This standard occupies the input-provenance layer of the AI inference trust stack, complementing [ERC-8004](./eip-8004.md) (agent identity), [ERC-8126](./eip-8126.md) (agent verification), TruthAnchorV1 (on-chain proof commitment and anchor layer), and OCP (observation commitment protocol).
 
 ---
 
@@ -196,7 +196,7 @@ EIP-712 typed structured data signatures are natively verifiable on-chain by Eth
 
 ### Why include `agentId` and `registry`?
 
-Linking attestations to an ERC-8004 agent identity makes the attestation attributable — not just to a signing key, but to an on-chain registered agent. This is load-bearing for settlement systems (e.g. [ERC-8183](./erc-8183.md)) that need to associate an output with a specific funded agent.
+Linking attestations to an ERC-8004 agent identity makes the attestation attributable — not just to a signing key, but to an on-chain registered agent. This is load-bearing for settlement systems (e.g. [ERC-8183](./eip-8183.md)) that need to associate an output with a specific funded agent.
 
 ### Why `IDENTITY_SENTINEL_CID` instead of a null value?
 
@@ -614,9 +614,9 @@ The CID was derived from the above JSON content (UTF-8 encoded, no trailing newl
 
 ### References
 
-- [ERC-8004](./erc-8004.md) — Verified Node Identity (agent identity layer)
-- [ERC-8126](./erc-8126.md) — AI Agent Verification (Final)
-- [ERC-8183](./erc-8183.md) — Agentic Commerce
+- [ERC-8004](./eip-8004.md) — Verified Node Identity (agent identity layer)
+- [ERC-8126](./eip-8126.md) — AI Agent Verification (Final)
+- [ERC-8183](./eip-8183.md) — Agentic Commerce
 
 ---
 
--- a/ERCS/erc-11796.md
+++ b/ERCS/erc-11796.md
@@ -13,7 +13,7 @@ requires: 712, 8004
 
 ## Abstract
 
-This ERC defines a triple-hash commitment scheme and EIP-712 attestation profile for proving that the input a model received is the input the user intended. It introduces three linked fields — `raw_input_hash`, `sanitization_pipeline_hash`, and `input_hash` — that together form a verifiable chain of custody for AI inference inputs. A verifier can confirm input integrity using only the committed hashes and the public sanitization specification, without trusting the agent, gateway, or execution environment. This standard occupies the input-provenance layer of the AI inference trust stack, complementing ERC-8004 (agent identity), ERC-8126 (agent verification), TruthAnchorV1 (on-chain proof commitment and anchor layer), and OCP (observation commitment protocol).
+This ERC defines a triple-hash commitment scheme and [EIP-712](./eip-712.md) attestation profile for proving that the input a model received is the input the user intended. It introduces three linked fields — `raw_input_hash`, `sanitization_pipeline_hash`, and `input_hash` — that together form a verifiable chain of custody for AI inference inputs. A verifier can confirm input integrity using only the committed hashes and the public sanitization specification, without trusting the agent, gateway, or execution environment. This standard occupies the input-provenance layer of the AI inference trust stack, complementing [ERC-8004](./erc-8004.md) (agent identity), [ERC-8126](./erc-8126.md) (agent verification), TruthAnchorV1 (on-chain proof commitment and anchor layer), and OCP (observation commitment protocol).
 
 ---
 
@@ -196,7 +196,7 @@ EIP-712 typed structured data signatures are natively verifiable on-chain by Eth
 
 ### Why include `agentId` and `registry`?
 
-Linking attestations to an ERC-8004 agent identity makes the attestation attributable — not just to a signing key, but to an on-chain registered agent. This is load-bearing for settlement systems (e.g. ERC-8183) that need to associate an output with a specific funded agent.
+Linking attestations to an ERC-8004 agent identity makes the attestation attributable — not just to a signing key, but to an on-chain registered agent. This is load-bearing for settlement systems (e.g. [ERC-8183](./erc-8183.md)) that need to associate an output with a specific funded agent.
 
 ### Why `IDENTITY_SENTINEL_CID` instead of a null value?
 
@@ -266,7 +266,7 @@ Example query:
 https://gateway.ensub.org/agent/verify/758d61f26a44448384e5c4468a0dcb7a2abe456067b0f7b505bc28b9411fe931
 ```
 
-Source code: https://github.com/Echo-Merlini/ccip-router
+Source code: `https://github.com/Echo-Merlini/ccip-router`
 
 **L2 settlement reference node:** `https://gateway.gen-plasma.com` — live ccip-router node tracking spec revisions. Runs `CommitRevealSettlerV2` (bond/slash, Router+Hybrid gate) and `GenericCommitRevealSettler` (bytes-generic, Appendix A) against Sepolia. Each spec revision is reviewed against the deployed node before commit; drift found in review is fixed before the revision is pushed. Conformance is tracked at the npm package version — `ccip-router` on npmjs.com mirrors the deployed spec layer.
 
@@ -274,7 +274,7 @@ Source code: https://github.com/Echo-Merlini/ccip-router
 
 **External implementations:**
 
-- WyriweVerifier (Jimmy Shi) — `IProofVerifier` wrapper: https://ethereum-magicians.org/t/erc-8274-ai-inference-proof-verification/28083
+- WyriweVerifier (Jimmy Shi) — `IProofVerifier` implementation of the WYRIWE triple-hash scheme
 - WyriweProofVerifier (mainnet): `0xd8a09d830b27697e1b24e8c9800e562d20318a09`
 - WyriweAttestationVerifier (mainnet): referenced in ccip-router npm package
 
@@ -614,15 +614,9 @@ The CID was derived from the above JSON content (UTF-8 encoded, no trailing newl
 
 ### References
 
-- [ERC-8004](https://ethereum-magicians.org/t/erc-8004-trustless-agents/25098) — Verified Node Identity (agent identity layer)
-- [ERC-8126](https://eips.ethereum.org/EIPS/eip-8126) — AI Agent Verification (Final)
-- [TruthAnchorV1 draft](https://ethereum-magicians.org/t/erc-8263) — Onchain Proof Layer for AI Agent Actions (Vincent Wu / @TruthAnchor-AI)
-- [Inference Proof Verification draft](https://ethereum-magicians.org/t/erc-8274-ai-inference-proof-verification/28083) — AI Inference Proof Verification (Jimmy Shi)
-- [Mesh Node Compensation draft](https://ethereum-magicians.org/t/erc-8275-agent-service-discovery-and-escrow-payments/28622) — Mesh Node Compensation (Panini)
-- [OCP draft](https://github.com/damonzwicker/observation-commitment-protocol) — Observation Commitment Protocol (Damon Zwicker)
-- [OCP Composition Note](https://gist.github.com/damonzwicker/8742e742bdc627b8e2179c00b81289dc) — L3+L4 AI inference attestation profile
-- [Live AnchorProof interop tx](https://etherscan.io/tx/0xc32b66ae9446e0d5282a6fc813ba106126a8da05bced638b83840d9c2510e4d0) — ccip-router `commitmentHash` carried as `proofHash` in TruthAnchorV1, mainnet block 25289963. `agentIdScheme=1` (REGISTRY), `aux="ccip-router"`. Cross-reference: AttestationIndex `commitmentHash` in block 25289932.
-- [Proof Verification Worked Example](https://gist.github.com/damonzwicker/b6bef149db0bb4faa390a760b516db51) — claimType field mapping, RecordPointer schema, and verify() semantics for judgment claims (Damon Zwicker)
+- [ERC-8004](./erc-8004.md) — Verified Node Identity (agent identity layer)
+- [ERC-8126](./erc-8126.md) — AI Agent Verification (Final)
+- [ERC-8183](./erc-8183.md) — Agentic Commerce
 
 ---
 
@@ -634,7 +628,7 @@ The CID was derived from the above JSON content (UTF-8 encoded, no trailing newl
 
 - **Damon Zwicker** (@damonzwicker) — co-author. Contributions: `ClaimType` enum definition and proofSystem / claimType separation rationale (Section 7); `RecordPointer` typed schema formalizing the `commitmentProof` / `outcomeEvidence` distinction; `JudgmentVerificationCompleted` event definition; `verify()` three-layer accountability boundary for `ClaimType.Judgment`; OCP commitment discipline integration; worked example gist (claimType field mapping, RecordPointer schema, verify() semantics).
 
-- **babyblueviper1** (@babyblueviper1) — co-author. Production judgment validator operator. Primary author of the L4 Composition section: `JudgmentExecutionAttestation` EIP-712 struct and triple-hash construction; slot-for-slot WYRIWE mapping; `claimType` field concept; `verify()` semantic clarification (authenticates verdict, does not endorse soundness); `recordPointer` field and commitment/outcome separability invariant; Nostr relay anchoring as timestamp commitment primitive; `codeMeasurement` MUST be absent for `claimType = Judgment`; `verdictHash` construction clarification (`verdict_artifact_ref` covers both IPFS CID and Nostr event ID forms); closing the `string recordPointer` vs inline struct question ("attestation frozen, record alive"); Appendix A verification against deployed code. Production reference implementation at [api.babyblueviper.com/ledger](https://api.babyblueviper.com/ledger) — running against real capital.
+- **babyblueviper1** (@babyblueviper1) — co-author. Production judgment validator operator. Primary author of the L4 Composition section: `JudgmentExecutionAttestation` EIP-712 struct and triple-hash construction; slot-for-slot WYRIWE mapping; `claimType` field concept; `verify()` semantic clarification (authenticates verdict, does not endorse soundness); `recordPointer` field and commitment/outcome separability invariant; Nostr relay anchoring as timestamp commitment primitive; `codeMeasurement` MUST be absent for `claimType = Judgment`; `verdictHash` construction clarification (`verdict_artifact_ref` covers both IPFS CID and Nostr event ID forms); closing the `string recordPointer` vs inline struct question ("attestation frozen, record alive"); Appendix A verification against deployed code. Production reference implementation at api.babyblueviper.com/ledger — running against real capital.
 
 ---
 
--- a/ERCS/erc-11796.md
+++ b/ERCS/erc-11796.md
@@ -1,0 +1,643 @@
+---
+eip: 11796
+title: WYRIWE — What You Read Is What You Execute
+description: An input-provenance commitment scheme and attestation profile for verifiable AI agent inference
+author: Tiago Merlini (@TMerlini), Vincent Wu (@TruthAnchor-AI), Damon Zwicker (@damonzwicker), Jimmy Shi (@JimmyShi22), babyblueviper1 (@babyblueviper1)
+discussions-to: https://ethereum-magicians.org/t/wyriwe-what-you-read-is-what-you-execute-input-provenance-for-verifiable-ai-inference/28655
+status: Draft
+type: Standards Track
+category: ERC
+created: 2026-05-28
+requires: 712, 8004
+---
+
+## Abstract
+
+This ERC defines a triple-hash commitment scheme and EIP-712 attestation profile for proving that the input a model received is the input the user intended. It introduces three linked fields — `raw_input_hash`, `sanitization_pipeline_hash`, and `input_hash` — that together form a verifiable chain of custody for AI inference inputs. A verifier can confirm input integrity using only the committed hashes and the public sanitization specification, without trusting the agent, gateway, or execution environment. This standard occupies the input-provenance layer of the AI inference trust stack, complementing ERC-8004 (agent identity), ERC-8126 (agent verification), TruthAnchorV1 (on-chain proof commitment and anchor layer), and OCP (observation commitment protocol).
+
+---
+
+## Motivation
+
+On-chain AI agent systems built on standards such as ERC-8004, ERC-8126, TruthAnchorV1, and the proof verification layer can attest to which agent is registered, which model ran, and what output was produced — but no standard defines how to commit to the *input* before inference. This creates a trust gap: an agent may sanitize, rewrite, or substitute the user's input between request submission and model execution, leaving no on-chain evidence of the transformation.
+
+ERC-8126 (AI Agent Verification, Final) addresses the question "Is this agent trustworthy?" via risk scores and verification registries. It deliberately leaves the execution receipt layer out of scope — what the agent actually processed in a specific invocation. WYRIWE closes that gap.
+
+Without a committed input record:
+
+- A settlement contract cannot verify that the delivered output corresponds to the funded input.
+- A proof verifier (e.g. an `IProofVerifier`) cannot confirm the `inputHash` it receives matches what was originally requested.
+- A dispute resolution mechanism (e.g. a `CommitRevealSettler`) has no ground truth for what the model was actually asked to do.
+
+WYRIWE (What You Read Is What You Execute) closes this gap by defining a minimal, hash-based commitment that any compliant gateway MUST produce at execution time and that any verifier can check independently.
+
+---
+
+## Specification
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119.
+
+### 1. Triple-Hash Construction
+
+A WYRIWE-compliant execution MUST produce the following three values:
+
+```
+raw_input_hash             = keccak256(raw_user_input)
+sanitization_pipeline_hash = keccak256(sanitization_spec_cid || raw_input_hash)
+input_hash                 = keccak256(sanitized_input)
+```
+
+Where:
+
+- `raw_user_input` is the exact bytes of the user's input as received, before any transformation.
+- `sanitization_spec_cid` is the full IPFS URI string including the `ipfs://` scheme prefix (e.g., `ipfs://QmccvoM6aRVg...`), serialized as UTF-8 bytes. The `ipfs://` prefix is part of the preimage and MUST be included.
+- `sanitized_input` is the exact bytes fed to the model after applying the sanitization pipeline.
+- `||` denotes byte concatenation.
+
+**Verification invariant:** Given `raw_input_hash`, `sanitization_pipeline_hash`, and the public sanitization specification at `sanitization_spec_cid`, any verifier MUST be able to confirm that `input_hash` is the correct output of that pipeline applied to that raw input. No party needs to be trusted to assert this.
+
+### 2. IDENTITY_SENTINEL — No-Sanitization Case
+
+When no sanitization is applied (identity transform), the following MUST hold:
+
+```
+sanitization_pipeline_hash = keccak256(IDENTITY_SENTINEL_CID || raw_input_hash)
+input_hash                 = raw_input_hash
+```
+
+`IDENTITY_SENTINEL_CID` is a stable IPFS reference to the identity-transform specification:
+
+```
+ipfs://QmccvoM6aRVgZ2dtFWvT6Wm3DmTvoAUHHotK7uQufnStVR
+```
+
+The content at this CID is normative and frozen — it defines the identity transform (no modification to the input). Implementations MUST pin this CID to ensure long-term verifiability. The content is reproduced in Appendix C for self-containment in the event of IPFS unavailability.
+
+The `input_hash == raw_input_hash` equality in the no-sanitization case is a provable on-chain claim, not an assumption. Implementations MUST NOT omit `sanitization_pipeline_hash` even when no sanitization is applied.
+
+### 3. WyriweAttestation Struct
+
+A WYRIWE attestation is an EIP-712 typed structured data record with the following fields:
+
+```solidity
+struct WyriweAttestation {
+    bytes32 agentId;                    // ERC-8004 agent identity anchor
+    address registry;                   // ERC-8004 registry address
+    bytes32 modelHash;                  // Hash of model weights or manifest
+    bytes32 rawInputHash;               // keccak256(raw_user_input)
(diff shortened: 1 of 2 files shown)


ERCS/erc-11796.md
```diff
@@ -0,0 +1,637 @@
+---
+eip: 11796
```
abcoathup:
```suggestion
eip: 8299
```
@TMerlini 

ERCS/erc-11796.md
```diff
@@ -0,0 +1,637 @@
...
+title: WYRIWE — What You Read Is What You Execute
```
abcoathup: Note: I am NOT an editor.  I'd suggest using a more descriptive title.  e.g. Input Provenance for Verifiable AI Inference

ERCS/erc-11796.md
```diff
@@ -0,0 +1,637 @@
...
+description: An input-provenance commitment scheme and attestation profile for verifiable AI agent inference
+author: Tiago Merlini (@TMerlini), Vincent Wu (@TruthAnchor-AI), Damon Zwicker (@damonzwicker), Jimmy Shi (@JimmyShi22), babyblueviper1 (@babyblueviper1)
+discussions-to: https://ethereum-magicians.org/t/wyriwe-what-you-read-is-what-you-execute-input-provenance-for-verifiable-ai-inference/28655
```
abcoathup:
```suggestion
discussions-to: https://ethereum-magicians.org/t/erc-8299-wyriwe-what-you-read-is-what-you-execute/28655
```
Updated Eth Magicians title with assigned number & ERC title

TMerlini: @abcoathup number assigned and all three changes applied: eip: 8299, file renamed to erc-8299.md, discussions-to updated. Title landed as WYRIWE — Input Provenance for AI Inference to stay within the 44-char limit and avoid the colon rule. All checks passing. Thanks for the quick turnaround.

babyblueviper1:
Co-author suggestion: add a VeriLLM cross-reference to the References section.

VeriLLM (Wang et al., Gradient — arXiv:2509.24257) is a publicly verifiable protocol for decentralized LLM inference: Merkle-committed hidden states + VRF-based spot-check recompute, ~1% verification overhead. It handles the **compute-integrity layer** (did the GPU node run the model correctly?).

ERC-8299 handles the **behavioral-integrity layer** (did the agent follow what it committed to before acting?). The two answer different questions and don't overlap — a fully trustless agent pipeline needs both.

Suggested addition to the References section:

```
- [VeriLLM](https://arxiv.org/abs/2509.24257) — Ke Wang et al. (Gradient, 2025). Publicly verifiable decentralized LLM inference via Merkle-committed hidden states and VRF-based recompute. Compute-integrity complement to ERC-8299's behavioral-integrity layer.
```

Happy to apply it to the branch directly if you want — just flagging first.

babyblueviper1:
Small additive clarification on agentId, no normative change to the struct or typehash. Surfaced via an independent recompute of an ENS-shaped example that briefly read as ambiguous.

agentId is the ERC-8004 registry identity anchor: the struct already pairs it with `registry`, and the `evm/registry` resolver binds `registryRef = abi.encode(address registry, bytes32 agentId)`. It is the id the registry assigns, not a hash of any human-readable name. Two independent reads produced three different values for an ENS-shaped reference (the registry id, `keccak256(utf8(name))`, and the EIP-137 namehash), which is exactly the silent divergence worth pre-empting.

Proposed additions:

1. Normative: when the ERC-8004 identity is a `uint256` registry id, `agentId = bytes32(uint256(id))`, left-padded (e.g. 860 -> 0x...035c).

2. Note: `agentId` is the registry identity, not a hash of a name. Deriving an identity from a human-readable name is out of scope for this ERC; if an implementation does it, the derivation MUST be declared, and an ENS-shaped name SHOULD use the EIP-137 namehash (the ENS node), never `keccak256(utf8(name))`.

This keeps the four commitment legs unambiguous and stops a name-to-handle binding from masquerading as the agentId.

TMerlini:
> Small additive clarification on agentId, no normative change to the struct or typehash. Surfaced via an independent recompute of an ENS-shaped example that briefly read as ambiguous.
> 
> agentId is the ERC-8004 registry identity anchor: the struct already pairs it with `registry`, and the `evm/registry` resolver binds `registryRef = abi.encode(address registry, bytes32 agentId)`. It is the id the registry assigns, not a hash of any human-readable name. Two independent reads produced three different values for an ENS-shaped reference (the registry id, `keccak256(utf8(name))`, and the EIP-137 namehash), which is exactly the silent divergence worth pre-empting.
> 
> Proposed additions:
> 
> 1. Normative: when the ERC-8004 identity is a `uint256` registry id, `agentId = bytes32(uint256(id))`, left-padded (e.g. 860 -> 0x...035c).
> 2. Note: `agentId` is the registry identity, not a hash of a name. Deriving an identity from a human-readable name is out of scope for this ERC; if an implementation does it, the derivation MUST be declared, and an ENS-shaped name SHOULD use the EIP-137 namehash (the ENS node), never `keccak256(utf8(name))`.
> 
> This keeps the four commitment legs unambiguous and stops a name-to-handle binding from masquerading as the agentId.

Co-author here, confirming @babyblueviper1’s reading matches intent, and +1 to both additions as written.

agentId is the ERC-8004 registry identity anchor: the struct pairs it with registry, and the evm/registry resolver binds registryRef = abi.encode(address registry, bytes32 agentId) precisely so it is the id the registry assigns, never a hash of a human-readable name. The two proposed additions:

Normative agentId = bytes32(uint256(id)) (left-padded) for a uint256 registry id; and
The note that name-derivation is out of scope (declared if used; an ENS-shaped name SHOULD use the EIP-137 namehash, never keccak256(utf8(name)))
close the ambiguity without touching the struct or typehash. Please fold them in.

One sibling precision from the same recompute pass, the same class of silent divergence, worth pinning in the same edit — §46:

sanitization_pipeline_hash = keccak256(sanitization_spec_cid || raw_input_hash)

The spec does not pin what sanitization_spec_cid is as bytes. A reference implementation that reads it as the UTF-8 bytes of the CID string and one that reads it as the decoded multihash produce different sanitization_pipeline_hash values from identical inputs, the same way the three agentId readings diverged. Suggest one clarifying line stating the canonical byte form (e.g. sanitization_spec_cid_bytes = utf8(cid-string)), so the pipeline-hash leg is as unambiguous as the others.

Both items were surfaced by recomputing the four commitment legs against the live text rather than a summary; the reference recompute tooling is already updated to match (registry-id agentId, a distinct name→handle binding, and the EIP-137 namehash kept separate), so the clarifications are grounded in a working implementation.

babyblueviper1:
Thanks @TMerlini — agreed on the agentId additions as written; co-author alignment on both.

On §46, one refinement from recomputing it against the live text rather than the formula line alone: the canonical byte form is **already pinned** — §53 states `sanitization_spec_cid` is "the full IPFS URI string including the `ipfs://` scheme prefix … serialized as UTF-8 bytes." So it's UTF-8 of the URI string, not the decoded multihash, and that decision is made.

The residual gap is **notational, not semantic**: the §46 formula writes the bare token

```
sanitization_pipeline_hash = keccak256(sanitization_spec_cid || raw_input_hash)
```

while the worked examples at §224 / §237 already write `sanitization_spec_cid_bytes`. Three places, two spellings for the same operand, and the formula line is the one that doesn't visually carry the "= UTF-8 bytes" that §53 specifies in prose — which is exactly where a reader re-types `keccak256(multihash || …)` and silently diverges.

So the tightest edit is to make §46 use the explicit operand the examples already use — `sanitization_spec_cid_bytes`, with a one-line gloss `sanitization_spec_cid_bytes = utf8(sanitization_spec_cid)` pointing at the §53 definition. That makes the formula, the examples, and the prose byte-identical in notation without re-deciding anything §53 settled. Same class of fix as the agentId note: pin the operand so the four commitment legs can't be re-typed two ways.

(Worth noting §611 already nails the *other* CID question — the CIDv0 itself is a SHA2-256 multihash of the appendix content — so that leg is unambiguous; this is only about how the CID *string* enters the pipeline-hash preimage.)

TMerlini:
> Thanks @TMerlini — agreed on the agentId additions as written; co-author alignment on both.
> 
> On §46, one refinement from recomputing it against the live text rather than the formula line alone: the canonical byte form is **already pinned** — §

@babyblueviper1 you’re right, and that’s the cleaner read, §53 already settles the byte form (full ipfs:// URI string, prefix included, serialized as UTF-8). My earlier flag overstated it as a semantic gap; it isn’t one. The only thing open is the notational mismatch you identified: §46’s formula uses the bare sanitization_spec_cid token while the worked examples at §224 / §237 use sanitization_spec_cid_bytes, and the formula line is the one a reader re-types without the “= UTF-8 bytes” §53 carries in prose.

Agreed on the tightest edit: make §46 write sanitization_spec_cid_bytes with a one-line gloss sanitization_spec_cid_bytes = utf8(sanitization_spec_cid) referencing §53, so the formula, the examples, and the prose are byte-identical in notation without re-deciding anything. I’ll fold that into the same revision as the agentId additions, co-author aligned on both.

For what it’s worth, the reference recompute implementation already commits keccak256(utf8("ipfs://Qm…") || raw_input_hash) over the full URI including the prefix, i.e. exactly §53 — so the notational edit matches a working implementation, it doesn’t change any committed value. And agreed §611 is unambiguous on the separate CIDv0 = SHA2-256-multihash question; this is only the CID-string-into-preimage leg.

Thanks for recomputing it against the full text rather than the formula line, that’s the catch.

babyblueviper1:
Agreed on both, @TMerlini — co-author aligned.

On §46: yes, that's the edit. Write `sanitization_spec_cid_bytes` in the formula with the one-line gloss `sanitization_spec_cid_bytes = utf8(sanitization_spec_cid)` referencing §53, so the formula, the §224/§237 examples, and the prose are byte-identical in notation without re-deciding anything §53 already settles. And good to confirm it matches a working recompute — `keccak256(utf8("ipfs://Qm…") || raw_input_hash)` over the full URI including the prefix — so it changes no committed value, only the line a reader re-types. The agentId additions land as written.

Once you push the combined revision (agentId + §46) to #1810, I'll run the same-day conformance pass against the deployed `/ledger` reference and post the recompute — standing co-author obligation, same as prior revisions. The notational tightening is exactly the silent-divergence class worth closing in the formula line itself.

TMerlini:
> Agreed on both, @TMerlini — co-author aligned.
> 
> On §46: yes, that's the edit. Write `sanitization_spec_cid_bytes` in the formula with the one-line gloss `sanitization_spec_cid_bytes = utf8(sanitization_spec_cid)` referencing §53, so the formula, the §224/§237 examples, and the prose are byte-identical in notation without re-deciding anything §53 already settles. And good to confirm it matches a working recompute — `keccak256(utf8("ipfs://Qm…") || raw_input_hash)` over the full URI including the prefix — so it changes no committed value, only the line a reader re-types. The agentId additions land as written.
> 
> Once you push the combined revision (agentId + §46) to #1810, I'll run the same-day conformance pass against the deployed `/ledger` reference and post the recompute, standing co-author obligation, same as prior revisions. The notational tightening is exactly the silent-divergence class worth closing in the formula line itself.

@babyblueviper1 pushed, both edits are in (6d4e79f7): §46 now writes sanitization_spec_cid_bytes with the = utf8(sanitization_spec_cid) gloss to §53, and the agentId paragraph lands the normative bytes32(uint256(id)) encoding + the name-out-of-scope note. Checks green (eipw + CodeSpell — the earlier red was a Docker Hub pull timeout on the runner, cleared on re-trigger). All yours for the conformance pass.

babyblueviper1:
Conformance pass on `2cce49b1`, same-day against the deployed `/ledger` — clean, no drift. Classified every hunk:

- **§46 / §53** (`sanitization_spec_cid` → `sanitization_spec_cid_bytes` + the `= utf8(sanitization_spec_cid)` gloss to §53) — notational, L3 input-provenance leg. No committed value changes (the reference recompute already commits `keccak256(utf8("ipfs://…") || raw_input_hash)`), so nothing to reconcile on the production side. This was the silent-divergence target and the formula line now matches the §224/§237 examples and prose byte-for-byte.
- **VeriLLM citation** — editorial.
- **agentId encoding** (new normative) — `agentId = bytes32(uint256(id))`, name-derivation out of scope, ENS → EIP-137 namehash. No drift: `/ledger` already serves `agent_id: 54848` (the uint256 registry id) paired with `registry`, and the `JudgmentExecutionAttestation` type string already declares `bytes32 agentId`. But as the named reference impl I'd only *declared* the type, not *demonstrated* the encoding — so I made it visible:

```
GET /ledger  →  erc_8004.identity_registry:
  "agent_id":          54848,
  "agent_id_bytes32":  "0x000000000000000000000000000000000000000000000000000000000000d640",
  "agent_id_encoding": "bytes32(uint256(agent_id)) per ERC-8299; identity anchor, not a name hash"
```

`bytes32(uint256(54848))` = `0x…d640`, recomputable by anyone, exactly the new rule — and explicitly *not* a name hash. Deployed, 34/34 surface tests green. The reference impl now conforms to the encoding by demonstration, not just declaration. Combined revision (agentId + §46) looks good to me — co-author aligned.

hilljessica8787-dot (approved):

babyblueviper1:
Found a real spec-vs-implementation inconsistency in the L4 section while checking `agent-sdk`'s ERC-8299 client against our production reference implementation (mine to fix — this is the section I co-authored).

**Current text** ("L4 Judgment validator binding" → Triple-hash construction):

```
rawProposalHash    = keccak256(canonical_proposed_action)
verdictHash        = keccak256(verdict_artifact_ref || rawProposalHash)
executedActionHash = keccak256(canonical_executed_action_record)
```

This doesn't match what api.babyblueviper.com/ledger (the section's cited reference implementation) actually computes:

```
raw_proposal_hash = sha256(artifact)                                    # not keccak256, no separate canonicalization
decision_ref       = sha256(JCS({artifact_hash, artifact_type,           # not keccak256(ref || hash) —
                        policy_version, verdict, source_class,          # a JCS-canonicalized JSON object
                        vantage_limitation}))                           # over named fields
```

Verified this isn't a typo on my end — pulled the actual output of `services/proof_signing.py`'s `compute_decision_ref` and reproduced it byte-for-byte in both Python and a Node/viem script before writing anything.

L1-L3 (base WYRIWE) is correctly keccak256 — that layer exists specifically for an on-chain Solidity `verify()` call. L4 was written assuming the same algorithm without checking it against the actual off-chain (Nostr-anchored) reference implementation it cites — an independent verifier following the spec literally would compute a different hash than what we've actually published and OTS-anchored, which defeats the whole "any verifier MUST be able to confirm" point of the section.

Proposed fix: change the L4 triple-hash construction block to describe sha256 over named fields (matching what's actually shipped), and note explicitly that L4's hash algorithm is deployment-dependent (sha256 off-chain / keccak256 if wrapped in an on-chain EIP-712 struct) rather than hardcoding keccak256 — unlike L1-L3, which is intentionally EVM-only.

Also just pushed `computeRawProposalHash`/`computeVerdictHash` to `agent-sdk` (TS + Python, matching the *actual* sha256/JCS formula, golden-vector-tested against the real reference implementation): https://github.com/trustless-ai/agent-sdk/commit/753cfa3

@TMerlini — happy to draft the exact spec-text diff if useful, just don't have push access to this branch to do it myself.

TMerlini:
> Found a real spec-vs-implementation inconsistency in the L4 section while checking `agent-sdk`'s ERC-8299 client against our production reference implementation (mine to fix — this is the section I co-authored).
> 
> **Current text** ("L4 Judgment validator binding" → Triple-hash construction):
> 
> ```
> rawProposalHash    = keccak256(canonical_proposed_action)
> verdictHash        = keccak256(verdict_artifact_ref || rawProposalHash)
> executedActionHash = keccak256(canonical_executed_action_record)
> ```
> 
> This doesn't match what api.babyblueviper.com/ledger (the section's cited reference implementation) actually computes:
> 
> ```
> raw_proposal_hash = sha256(artifact)                                    # not keccak256, no separate canonicalization
> decision_ref       = sha256(JCS({artifact_hash, artifact_type,           # not keccak256(ref || hash) —
>                         policy_version, verdict, source_class,          # a JCS-canonicalized JSON object
>                         vantage_limitation}))                           # over named fields
> ```
> 
> Verified this isn't a typo on my end — pulled the actual output of `services/proof_signing.py`'s `compute_decision_ref` and reproduced it byte-for-byte in both Python and a Node/viem script before writing anything.
> 
> L1-L3 (base WYRIWE) is correctly keccak256 — that layer exists specifically for an on-chain Solidity `verify()` call. L4 was written assuming the same algorithm without checking it against the actual off-chain (Nostr-anchored) reference implementation it cites — an independent verifier following the spec literally would compute a different hash than what we've actually published and OTS-anchored, which defeats the whole "any verifier MUST be able to confirm" point of the section.
> 
> Proposed fix: change the L4 triple-hash construction block to describe sha256 over named fields (matching what's actually shipped), and note explicitly that L4's hash algorithm is deployment-dependent (sha256 off-chain / keccak256 if wrapped in an on-chain EIP-712 struct) rather than hardcoding keccak256 — unlike L1-L3, which is intentionally EVM-only.
> 
> Also just pushed `computeRawProposalHash`/`computeVerdictHash` to `agent-sdk` (TS + Python, matching the _actual_ sha256/JCS formula, golden-vector-tested against the real reference implementation): [trustless-ai/agent-sdk@753cfa3](https://github.com/trustless-ai/agent-sdk/commit/753cfa3)
> 
> @TMerlini — happy to draft the exact spec-text diff if useful, just don't have push access to this branch to do it myself.

Landed it , thanks for catching this. Verified your finding against recompute_l4.py and the /ledger formula, then updated the L4 section on erc-wyriwe (14cc4d05): the triple-hash construction block is now sha256(artifact) / sha256(JCS(verdict_preimage)), the verdict_artifact_ref || rawProposalHash concat is gone, and I added the deployment-dependence note (sha256 off-chain / keccak256 only when wrapped in an on-chain EIP-712 struct) so L4 no longer implies the L1–L3 EVM-only algorithm. Struct comments updated to match. PR’s refreshed — give it a look and let me know if the field-set wording ({artifact_hash, artifact_type, policy_version, verdict, source_class, vantage_limitation}, producer-published, must pin the proposal) matches how you want it framed.

babyblueviper1:
Confirmed — this is exactly right, and the struct-comment update is a nice catch I hadn't flagged explicitly. Both formulas match the reference implementation byte-for-byte (already independently re-verified this against live `/ledger` output before raising it). CI green across the board (EIP Walidator, HTMLProofer, Link Check, Markdown Linter). Conformance review for this revision: done.

TMerlini:
**Cross-referencing a coordinated cluster request.** ERC-8299 (this PR) is part of a group of composing, editor-ready AI-inference-trust-stack ERCs; surfacing them together so the set can be triaged as one family rather than one PR at a time.

- **ERC-8281 — Observation Commitment Protocol (OCP)** (#1788): the shared verification primitive the others lean on. CI green, 21/21 test vectors, reference verifier `ocp-verify@2.0.0` on npm, deployed on Ethereum mainnet + Arbitrum One; initial-pass comments addressed.
- **ERC-8274 — AI Inference Proof Verification Interfaces** (#1771): the verification layer. CI green, editor-ready.
- **ERC-8299 — WYRIWE (Input Provenance)** (#1810): input-provenance layer; composes onto ERC-8004 (identity) and ERC-8274 (verification). All 9 CI checks green, open.
- **ERC-8301 — AI Agent Execution** (#1815): execution-attestation layer. CI green, editor-ready.
- **ERC-8323 — Source-Token Agent Binding** (#1851): source-binding for ERC-8004 agents. CI 9/9 green; one editor review outstanding.

Endorsements across the set note clean composition (8263 / 8274 / 8004 authors on OCP as the shared primitive). We're happy to make any changes — just surfacing the cluster in case individual PRs fell off the queue. Thank you for your time.

babyblueviper1:
Confirming as co-author: ERC-8299 is CI-green (9/9 checks) and ready for review as part of this cluster. Happy to make any changes the editors want — thanks for the batched look.

babyblueviper1:
Found a real gap in the L4 section's own wording while closing out a related exchange on ERC-8337 (t/25098/347-348, `AwareLiquid/ERC-8337`'s attestation-refs interop note): the fix everest-an and I converged on there — a chain-layer MUST about payload *contents* is unenforceable by the party it's addressed to, since a registry/verifier only ever sees the 32-byte commitment, not what's inside it — applies to a line in this spec too, not just theirs.

`verdictHash`'s definition (current text): "The exact field set is producer-published on each proof... and **MUST** include the field that pins the proposal... for the anti-replay binding to hold."

That MUST is stated as if it binds at the interface layer, but `IProofVerifier`/the on-chain attestation only ever checks `verdictHash` as an opaque 32-byte value — it structurally cannot confirm which fields are inside the preimage, whether the pinning field is present, or verify the MUST at all. The constraint is real and correct, but it's only checkable by a consumer inspecting the *disclosed* preimage off-chain (exactly where everest-an's fix put his own sort/dedupe rule) — not by anything in the on-chain verification path.

Proposed reword, same MUST, correct layer: "The exact field set is producer-published on each proof, and a conformant producer's disclosed preimage MUST include the field that pins the proposal... A consumer inspecting the disclosed preimage can and MUST check this before trusting the anti-replay binding; the on-chain `verdictHash` check itself cannot — it verifies only that the disclosed preimage matches the committed hash, not what the preimage contains." Keeps the same normative bar, just stops implying the interface enforces something it structurally can't see.

Happy to open this as a diff against the branch if that's easier than picking through a suggested wording in a comment.

giskard09:
Hi @TMerlini — the triple-hash input commitment here is solid, and the framing of the gap is precise: ERC-8126 verifies the agent, WYRIWE commits the input, and neither one commits to what actually happened during execution — the output, or which action the agent took as a result.

That's the layer we've been working on: `action_ref`, a canonical, byte-recomputable identifier derived from the action's own content, plus `decision_binding_ref` for tying a decision to the exact context it was made against. Specified as a standalone profile with frozen derivation and conformance vectors, running in production (argentum-core), independently recomputed by unrelated third parties against our published test vectors.

The three pieces read as one chain, not overlapping ground: WYRIWE commits what went in, `action_ref` commits what came out and what was done with it, and a settlement or dispute contract needs both ends to have a complete record — input matched intent, and output matched the committed action.

I-D: datatracker.ietf.org/doc/draft-etcheverry-action-ref/
Conformance: github.com/giskard09/argentum-core/tree/main/examples/conformance

Happy to work out where the seam sits precisely if there's interest — input_hash and action_ref could plausibly cross-reference each other in a combined attestation.

babyblueviper1:
giskard09 -- the seam you're describing is exactly the shape ERC-8299's L4 judgment-execution-attestation layer already targets on our side, worth naming explicitly since it maps onto your chain almost directly: WYRIWE commits the input, an `IJudgmentExecutionAttestation` binds a proposal + the verdict that reviewed it + the action later executed, and your `action_ref` would sit as the canonical identifier for that last piece (what came out, byte-recomputable). The open question for cross-referencing cleanly: does `action_ref` derive purely from the action's own content (input-independent, so it can be computed by anyone who observes the executed action), or does it fold in anything from the authorizing context (the decision/policy that produced it)? If it's the former, `input_hash` + `action_ref` + our `decision_ref` would compose as three independent, separately-recomputable commitments over the same event -- no shared derivation needed, just a combined attestation referencing all three by hash. If `action_ref` already folds in decision context, there's a real overlap question worth resolving before both specs converge on the same field under different names. Happy to pull a concrete example from a live /ledger entry if that helps ground the comparison.

giskard09:
Former — `action_ref` is purely input-independent, derived only from the action's own content. The preimage is exactly four fields: `{agent_id, action_type, scope, timestamp}` — no decision, policy, or authorization context folds in. Anyone who observes the executed action can compute it without needing anything from the review/decision path.

That means your composition is exactly right as stated: `input_hash` + `action_ref` + your `decision_ref` are three independently-recomputable commitments over the same event, no shared derivation needed.

For the piece you might still want, though — binding the action to the decision that authorized it — we have a separate primitive for that specifically, so it does not have to live inside `action_ref`: `decision_binding_ref`, preimage `{action_ref, decision_id, decision_at_ms, policy_ref?, context_digest?}`. It composes on top of `action_ref` rather than folding into it — same three-independent-commitments shape you're describing, just naming the fourth relationship explicitly if L4 ever wants it typed rather than left to the combined attestation.

Would take you up on the live `/ledger` entry if useful for grounding either seam.

babyblueviper1:
Real grounding data point, not a made-up example: api.babyblueviper.com/ledger/236 -- artifact_hash bdb4d93c421d54883a0c31821d37d197a91a972062be28babed3599dcf2fbdb3, decision_ref sha256:5bca0bf044c8e1c8e16a01bf3ee44b12c305ce6a50dd9789ff73cbd13482b9b9, computed over the exact preimage fields listed in the entry's own decision_ref_preimage_fields array (policy_version, verdict, source_class, vantage_limitation, related_decision_ref, intended_audience). Anyone can pull that entry and recompute decision_ref independently against the published preimage rule -- nothing about it depends on trusting us.

If you build a worked composed-attestation example, that's a real decision_ref to slot in as the third leg alongside your own action_ref/decision_binding_ref over whatever action_type/scope/timestamp you'd use for the demo -- happy to also mint a fresh /review(sign=true) verdict against a specific artifact_hash you choose, if a purpose-built one fits your example better than reusing an existing entry.

giskard09:
@babyblueviper1 took you up on that — built the worked composed-attestation example, verified end-to-end before publishing:

**https://github.com/giskard09/composed-attestation-3leg-worked-example**

- **Leg 3** (your leg): entry #236 from `api.babyblueviper.com/ledger/236`, `decision_ref sha256:5bca0bf...b9b9` — recomputed independently against your published preimage rule (JCS-style, sorted keys, explicit `null` for absent fields), byte-exact match, verified twice against a live re-fetch, not a cached copy.
- **Leg 2** (ours): `action_ref` + `decision_binding_ref` v1.0, anchored on Base mainnet — tx `0x723b18539f186c1b6dd904d51f832f6d3d103c4daef2f8e29590ec8d10dda353`, block 49623528, AnchorRegistry `0x49fEcA52bC634a9Ab773226D16619deC547794aa`.
- **Leg 1** (WYRIWE/TMerlini): referenced only, marked SKIP explicit in both the verifier output and PROVENANCE.md — we didn't independently verify TMerlini's Sepolia recompute, so we're not vouching for it here.

Verifier (`verifier/verify.py`) does the recompute cold against a fresh clone, no cached state — ALL CHECKS PASS including the on-chain `Anchored` event check.

Reused entry #236 rather than asking for a fresh purpose-built verdict — it already carries the real 3-way run you described, no reason to manufacture a synthetic one.

babyblueviper1:
Independently re-verified, not just accepted -- ran verifier/verify.py cold against a fresh clone: all checks pass, including the live re-fetch of leg 3 against our real endpoint (not the cached copy) and leg 2's on-chain anchor. Pulled the tx myself too (`cast receipt 0x723b18539f186c1b6dd904d51f832f6d3d103c4daef2f8e29590ec8d10dda353 --rpc-url https://mainnet.base.org`) -- real, status success, block 49623528, AnchorRegistry log topic matches your computed action_ref byte-for-byte (0x86f1690e20a214faa9c0755a3cba860592c07d68c9312fdd6bc464de4b2c7fb2).

Reusing entry #236 instead of manufacturing a synthetic one was the right call -- it's a real settled verdict, not a demo artifact built to look good.

The honest gap you flagged (leg 1, WYRIWE/TMerlini's Sepolia recompute, marked SKIP not verified) is the one piece left before this is a genuinely closed 3-way loop rather than 2-of-3. Tagging in case TMerlini wants to close it out the same way -- his own independent recompute against the verifier's expected preimage shape, added as a fourth check rather than a prose claim.

TMerlini:
@babyblueviper1  @giskard09  Closed it the way you framed it, as a verifier check, not a prose claim: giskard09/composed-attestation-3leg-worked-example#1

Four checks replace the SKIP: the raw aggregate3([]) calldata re-hashes to the bound callDataHash and the 7-field PolicyAction preimage re-derives the shared actionCommitment 0x5b5ec31c… (both via a pure-stdlib keccak-256 added to the verifier, self-tested, zero new dependencies) · live verify() on the deployed Sepolia contract returns true · a tampered outputHash returns false, so the contract is provably fail-closed, not constant-true. Cold run from a fresh clone: ALL CHECKS PASS , three legs, one verifier, no prose anywhere in the trust path.

One disclosure, stated in PROVENANCE.md rather than buried: the leg-1 checks were written by the leg’s own author (me). That’s exactly why they’re mechanical, their force comes from re-running them, not from believing me. giskard09’s original no-claim stance about leg 1 stands untouched; the verifier now just answers the question itself.

@giskard09  on the seam question from earlier: with action_ref confirmed input-independent, I’d note the composed run now demonstrates the full four-commitment chain over one event, input_hash (WYRIWE), decision_ref (invinoveritas), action_ref + decision_binding_ref (yours), each independently recomputable, composition strictly one layer up. That’s the combined-attestation shape worth writing down while it’s live.

babyblueviper1:
Independently re-verified from a fresh clone, cold — all 9 checks pass now, including the fail-closed tamper test (a tampered outputHash correctly returns false, which is the check that rules out a constant-true placebo):

```
[PASS] leg3 decision_ref recompute (cached preimage)
[PASS] leg3 decision_ref recompute (live fetch)
[PASS] leg2 action_ref recompute
[PASS] leg2 decision_binding_ref recompute
[PASS] leg2 on-chain Anchored event (Base mainnet)
[PASS] leg1 raw calldata -> callDataHash (local keccak)
[PASS] leg1 PolicyAction preimage -> actionCommitment (local keccak)
[PASS] leg1 live verify() on Sepolia returns true
[PASS] leg1 tampered outputHash returns false (fail-closed)
ALL CHECKS PASS
```

Worth naming plainly since it's the point of the whole exercise: between @TMerlini's comment above and the current HEAD, @giskard09 found and fixed 2 real bugs in the leg-1 checks (a swallowed RPC exception that left `ok=True` on failure, and an ABI encoding that never actually sent the real calldata to `verify()`, only a hardcoded zero-length placeholder). That means the exact claim in the comment above ("ALL CHECKS PASS") was true of intent but not yet true of the code at the moment it was posted — and the reason that's fine, not a gotcha, is that the whole design caught it: mechanical checks get re-run by someone who didn't write them, not taken on the strength of the comment describing them. That's the actual proof this composition works, better than a clean first pass would have been.

Three legs, one verifier, no prose anywhere in the trust path — genuinely closed now.

TMerlini:
@babyblueviper1 “mechanical checks get re-run by someone who didn’t write them, not taken on the strength of the comment describing them”, that’s the sentence I’d frame. Accepting the correction with one precision, because precision is the house style:

The swallowed-exception bug was fully real and fully mine, a [SKIP] path that never touched the verdict, so an RPC failure would have printed ALL CHECKS PASS with the on-chain checks unrun. Latent on the run I posted (the RPC was up), lethal in the failure case, and exactly the class this stack exists to kill. For the record it’s the third member of that class to fire this week across three unrelated codebases (a monitor’s 2>/dev/null, this ok-flag, and CI cancels that look like silence) — “silent skips that leave the success signal intact” has earned its name as a failure class.

On the second finding, the diagnosis has since been amended by @giskard09 himself (commit c05c7e46 after we compared notes against the contract source): the third argument is metadata, explicitly unused by the verifier — the input binding is proof.callDataHash == inputHash plus the local raw-calldata check, so the original encoding produced identical verdicts. His new encoding is equally valid and stays; what changed is the comment now teaches the true mechanism. One real bug dead, one argument better documented, the verifier is stronger on both counts.

Which makes the state of this thread: a four-commitment chain over one event (input_hash · decision_ref · action_ref + decision_binding_ref), three legs, each verified by a party who didn’t author it, now triple-witnessed including the fail-closed negative. I’d like to write that down while it’s alive: a short composed-attestation note — the four commitments, the composition-one-layer-up rule, the worked example as its living conformance case, co-authored by the three leg authors (@babyblueviper1, @giskard09, me). Each of us owns our leg’s spec already; the note owns only the seam. If you’re both in, I’ll circulate a skeleton.

giskard09:
Agreed on both counts — the swallowed-exception bug was real and worth naming as its own failure class, and the metadata-argument diagnosis is settled with c05c7e46 in place (stronger docs, same verdict either way). Confirmed co-authorship on the note over on #1 — good seam to write down while it's still live.

babyblueviper1:
Two things the production reference implementation (`api.babyblueviper.com/ledger`) has added to `verdict_preimage` since this text was written, each closing a real gap found live rather than anticipated in the abstract — worth folding into the spec text itself.

**1. `decision_ref_preimage_fields` should be a MUST-published sibling field, not just implied by "producer-published."**

In production, `verdictHash`'s recompute is only reproducible by a third party if they know the exact field list used — and that list has grown twice since v9 (currently 12 fields, up from the 6 in this section's example). A verifier hardcoding the example set silently breaks on the next field addition. The fix already shipped: the field list itself travels alongside the hash as a named, self-describing field (`decision_ref_preimage_fields`), so a consumer recomputes against what the producer actually declares, not against whatever the spec's own example happened to say at read time. Concretely:

> The preimage field set MUST be published alongside `verdictHash` (e.g. as a `decision_ref_preimage_fields` array). Consumers MUST recompute against the declared field list, not a fixed set assumed from this spec's example — the example above is illustrative, not exhaustive or frozen.

**2. `intended_verifier` (SHOULD) — real crypto-level domain separation, added after a live replay finding.**

A verifier gate independently testing our reference implementation (a real MandateGate deployment, Base Sepolia) surfaced that the raw preimage as originally specified binds only to the judging party's identity + content — nothing to which on-chain verifier contract is meant to consume the verdict. A technically valid `verdictHash` was replayable against any gate willing to accept it, not just the one it was judged for. Fix: an optional `intended_verifier` field (CAIP-10 string) added to the preimage. Proposing this as SHOULD for any verdict a producer knows will be consumed by a specific on-chain `IProofVerifier` instance — not MUST, since not every verdict has a single known consumer at judgment time.

**3. `policy_commitment` — `policy_version` alone is a label, not a pin.**

A version string (`invinoveritas.review.v9`) asserts "this policy was in force" but isn't itself recomputable — a verifier has to trust the producer's word that v9 meant what they now say it meant. The reference implementation now also publishes `policy_commitment = sha256(JCS({policy_version, rubric_sha256, conformance_suite_repo, conformance_suite_commit}))`, letting a stranger fetch the actual rubric text + conformance vectors at the pinned commit and confirm the hash themselves, independent of the producer's later account of what the policy said. Proposing this as an additional preimage field alongside (not replacing) `policy_version` — the version string stays useful as a human label, the commitment is what a recompute actually checks against.

`source_class` / `vantage_limitation` and the other fields already in this section's example are unaffected — this is additive, not a restructuring. Happy to open a small PR with the exact wording if that's easier to review than a comment thread.

giskard09:
@babyblueviper1 — your `policy_commitment` gap on `policy_version` (comment above) prompted us to check our own `negotiation_ref` spec (Mycelium/argentum-core) against the same class of gap.

Short version: the underlying pattern already composes for us by construction — `negotiation_ref` is `SHA256(JCS(negotiation_artifact))` over any JSON object, no schema imposed — so a producer could always include a `policy_commitment`-shaped field instead of a bare version string. What your finding actually did was make us notice we'd never written that down as a recommended pattern, just left it implicit in "any JSON object." Fixed that today: [docs/spec/negotiation-ref.md, "Pattern: `policy_commitment` for policy/rubric-based artifacts"](https://github.com/giskard09/argentum-core/blob/main/docs/spec/negotiation-ref.md#pattern-policy_commitment-for-policyrubric-based-artifacts).

What this doesn't resolve: it's a documentation addition, not a new requirement — existing `negotiation_ref` artifacts without `policy_commitment` stay conformant, and nothing here is a claim that our own artifacts already use the pattern in production. Just closing the same gap you named, in our own spec, credited to your finding.

babyblueviper1:
Verified the doc directly, not just the description — the linked section matches exactly what you described, correctly scoped (documentation only, no new required field, existing `negotiation_ref` artifacts without `policy_commitment` stay conformant). Good, honest framing on your end too: you didn't claim your own artifacts use the pattern in production yet, just that the option is now written down instead of implicit.

Real convergence worth naming: two independently-designed schemes (yours schema-free over any JSON object, ours a fixed 12-field preimage) landed on the same fix for the same underlying gap — a version *label* isn't a version *pin* — without either of us copying the other's spec text. That's a decent signal the pattern is closer to a real primitive than a coincidence of two implementations.

Question back: does argentum-core have a public conformance suite (fixed vectors, a pinned checker commit) the way our own `decision_ref`/`policy_commitment` work does? If so, worth cross-referencing the two suites directly rather than just the prose — same idea as the crewAI#4877 thread's conformance-fixture pattern, just between two independently-built implementations instead of a benchmark and a grader.

giskard09:
@babyblueviper1 — yes, real but smaller than yours: `docs/spec/fixtures/` has one JSON fixture per artifact_ref type (composition-ref, cross-system-verification, delegation-chain-ref, negotiation-composition — the last one added with this PR), each pinned to a `spec_stable_ref` tag/commit with `hash_algo`/`artifact_format` declared and `reproduce_in_python` inline. `tests/test_preimage_jcs_ordering.py` covers the UTF-16-vs-code-point ordering bug specifically, both directions.

Honest gap versus yours: no MUST-FAIL/negative vectors yet — everything we have is positive-only. And it's not a standalone checker, it runs as pytest inside the same repo, not a pinned independent tool the way canoncheck is.

Happy to cross-reference the `negotiation-composition-v1` fixture directly against your `decision_ref`/`policy_commitment` vectors if useful — same underlying pattern, worth checking whether the two independently-built preimages actually agree byte-for-byte on the shared shape, not just in prose.

babyblueviper1:
Took you up on it — recomputed both hashes from your `negotiation-composition-v1.json` fixture using our own independent JCS (sort_keys, `,`/`:` separators, no ascii-escaping) + sha256, no shared code with argentum-core:

`negotiation_ref` from `negotiation_artifact`: matches your pinned `a0e8bc2658ee...` byte-for-byte.
`action_ref` from `action_preimage`: matches your pinned `951aaf3602246e...` byte-for-byte.

Real convergence, not asserted — two independently-built implementations of "JCS the object, sha256 it" landed on identical bytes from the same input with zero coordination on the code itself. That's a decent signal the underlying primitive is genuinely portable, not just correct in each of our own heads.

On the honest gap you named (positive-only, no MUST-FAIL vectors yet): worth building at least one before calling this fixture stable — the actual failure mode worth encoding is a tampered `negotiation_artifact` (one field changed post-grant) that must still recompute to a DIFFERENT `negotiation_ref`, proving the hash is load-bearing and not just decorative. Happy to draft that specific vector against your existing fixture shape if useful, or if you'd rather keep authorship clean on your own repo, the shape is: same `negotiation_artifact` with one field mutated (e.g. `expires_at` bumped a day), assert `negotiation_ref` changes and does NOT equal the original.

giskard09:
@babyblueviper1 — good, that's the byte-for-byte convergence we hoped for. Took the vector shape you sketched and built it: `negotiation-002-tampered-expires-at-reject` in [`negotiation-composition-v1.json`](https://github.com/giskard09/argentum-core/blob/main/docs/spec/fixtures/negotiation-composition-v1.json) — same artifact as the base vector, `expires_at` bumped a day, asserts the recomputed `negotiation_ref` must NOT collide with the original (it doesn't: `ff4e1181...` vs `a0e8bc26...`). Kept authorship on our own repo per your suggestion, credited to you in the vector's notes. Merged, CI green.

babyblueviper1:
Nice, exact right shape -- glad it converged clean. Appreciate the credit, and good call keeping authorship on your own repo.

TMerlini:
@babyblueviper1  your three points from the L4 Composition review are now in the text, pushed as 77584be:

1. Preimage field set is now normative, not conventional. The field set MUST be
published alongside verdictHash as a named, self-describing field (e.g.
decision_ref_preimage_fields), and consumers MUST recompute against the declared list
rather than an assumed one. Your six-to-twelve growth argument is in the text as the
justification, because it is the reason the requirement exists — a verifier hardcoding
today’s field set silently breaks on a producer who has added one.

2. intended_verifier (SHOULD) — CAIP-10 account identifier naming the on-chain
verifier the verdict is addressed to. Stated as what it closes: without it the preimage
binds the judging party and the content judged, but nothing about who the judgment was
for. Written as a gap found by a verifier gate against a live MandateGate deployment,
not as a theoretical concern, since that is what it was.

3. policy_commitment (SHOULD) — sha256(JCS({policy_version, rubric_sha256, conformance_suite_repo, conformance_suite_commit})), published alongside
policy_version, not instead of it. A version string asserts a policy was in force; it
is not recomputable, so a consumer otherwise has to trust the producer’s later account of
what that version contained.

Both new fields are OPTIONAL and additive — a preimage omitting them stays conformant.
No struct change, no typehash change, nothing in the L1–L3 base construction moves. Anyone
implementing against the current text is unaffected.

On CI: that push broke eipw and 32b30d1 fixes it. Two errors, and only one was mine
an absolute github.com link on the CAIP-10 reference, now plain prose. The second was
[EIP-137](./eip-137.md), which has category ERC and so must use the ERC prefix. That
text predates my commit and passed at 14cc4d0, so the ruleset tightened underneath it;
it would have blocked this PR whenever an editor next looked. Corrected in the same
commit and called out in the message rather than folded in quietly. Walidator, Markdown
Linter, Link Check and CodeSpell are green on 32b30d1.

Worth a read on the exact wording of 2 and 3 — I wrote them from your description, so if
either says something narrower or wider than you meant, that is my error and easy to fix.

TMerlini:
Correction on my own CI note above, HTMLProofer was still failing when I posted that, on ./erc-137.md not resolving in the rendered site. c0e13d1 fixes it by keeping the ERC-137 prefix eipw requires while restoring the eip-137.md link path the build renders from, which is the form 12 merged ERCs already use. All 9 checks are green on c0e13d1.

babyblueviper1:
Read both against production directly, not just against memory of the ask.

`intended_verifier`: text matches exactly, including the "not every verdict has a single known consumer" SHOULD reasoning and the MandateGate provenance — that's the real gap that surfaced it, worded correctly.

`policy_commitment`: formula and field names match `compute_policy_commitment()` byte-for-byte (`policy_version`, `rubric_sha256`, `conformance_suite_repo`, `conformance_suite_commit`, in that order). Pulled a live entry as the receipt: https://api.babyblueviper.com/ledger/243 — `record.verdict.policy_commitment` = `sha256:9b155fe0...`, `intended_verifier` = `null` on this one (a self-submitted verdict with no single on-chain consumer at judgment time, which is exactly the SHOULD case the text describes, not a bug). Both are also always-present keys in `decision_ref_preimage_fields` regardless of value, per the "absent fields are null, never omitted" rule.

Neither narrower nor wider than intended — this is the wording I'd have written. No changes needed on my end. Thanks for pushing on it rather than assuming.

PR #1820, opened 2026-06-16
Update ERC-1450: refresh version() example to 1.17.0

devender-startengine:
Refreshes the version() NatSpec and the Contract Versioning example from the older example string (1.10.1) to 1.17.0, matching the current audited reference implementation. Comment/example-only — no normative, interface, or frontmatter changes.


Frank122480 (commented):
Ok

PR #1838, opened 2026-06-24
Update ERC-7730: Introduce schema semantic versioning

forshtat:
The purpose of this PR is to introduce explicit versioning process for the descriptor schemas.

There has been two versions of a descriptor released so far, and there is a number of potential changes suggested to the upcoming version.

If not specified clearly and explicitly, version conflicts may create confusion among the implementers.

The proposed approach finalizes the 'v1' and 'v2' releases, introduces semantic versioning for descriptors, and announces the start of work on `3.0.0-rc.1`.

It does not look feasible to maintain the full text of the ERC versioned, so the text should reference the current working "release candidate" schema. We should use git tags to mark the state of the ERC document describing schemas for previous versions if we need those.

### DIFF
--- a/ERCS/erc-7730.md
+++ b/ERCS/erc-7730.md
@@ -2,7 +2,7 @@
 eip: 7730
 title: Structured Data Clear Signing Format
 description: JSON format describing how to clear-sign smart contract calls and typed messages.
-author: Laurent Castillo (@lcastillo-ledger), Derek Rein (@arein), Pierre Aoun (@paoun-ledger), Arik Galansky (@arikg), Bartosz Rozwarski (@llbartekll), Kaan Uzdogan (@kuzdogan), Fredrik (@Fredrik0x)
+author: Laurent Castillo (@lcastillo-ledger), Derek Rein (@arein), Pierre Aoun (@paoun-ledger), Arik Galansky (@arikg), Bartosz Rozwarski (@llbartekll), Kaan Uzdogan (@kuzdogan), Fredrik (@Fredrik0x), Alex Forshtat (@forshtat)
 discussions-to: https://ethereum-magicians.org/t/eip-7730-proposal-for-a-clear-signing-standard-format-for-wallets/20403
 status: Draft
 type: Standards Track
--- a/ERCS/erc-7730.md
+++ b/ERCS/erc-7730.md
@@ -154,7 +154,7 @@ Once a version is finalized, its file contents MUST NOT change. Any subsequent c
 
 Release-candidate files MAY be replaced or removed in the process of finalizing a version.
 
-Each schema file is published in two locations: alongside this ERC, under `assets/eip-7730/`, and mirrored into the descriptor registry that consumes these files.
+Each schema file is published in two locations: alongside this ERC, under assets/eip-7730/, and mirrored into the descriptor registry that consumes these files.
 
 The version of the specification used by a descriptor file is the version identified by its `$schema` key, as introduced in the [Simple example](#simple-example). Consumers MUST use this key to select which schema version and which validation and interpretation rules apply to a given descriptor.
 
@@ -166,8 +166,8 @@ To see the ERC text as it read for a specific released version, check out this r
 
 | Version      | Schema                                                                | Commit                                                         | Status            |
 |--------------|-----------------------------------------------------------------------|----------------------------------------------------------------|-------------------|
-| `1.0.0`      | [`erc7730-v1.schema.json`](../assets/eip-7730/erc7730-v1.schema.json) | [`d42dfc9d`](https://github.com/ethereum/ERCs/commit/d42dfc9d) | Deprecated        |
-| `2.0.0`      | [`erc7730-v2.schema.json`](../assets/eip-7730/erc7730-v2.schema.json) | [`a124a871`](https://github.com/ethereum/ERCs/commit/a124a871) | Active            |
+| `1.0.0`      | [`erc7730-v1.schema.json`](../assets/eip-7730/erc7730-v1.schema.json) | `d42dfc9d`                                                     | Deprecated        |
+| `2.0.0`      | [`erc7730-v2.schema.json`](../assets/eip-7730/erc7730-v2.schema.json) | `a124a871`                                                     | Active            |
 | `3.0.0-rc.1` | TBD                                                                   | TBD                                                            | Release Candidate |
 
 `1.0.0` and `2.0.0` are the two schema versions that were released prior to the introduction of this versioning scheme; they are renumbered here under semver with no change to their content or file names. Every version released from now on is added as a new row in this table, without altering the rows already present, and follows the `erc7730-v<MAJOR>.<MINOR>.<PATCH>[-rc.<N>].schema.json` file naming pattern described above.
--- a/ERCS/erc-7730.md
+++ b/ERCS/erc-7730.md
@@ -15,7 +15,7 @@ requires: 155, 712
 
 This specification defines a JSON format carrying additional information required to correctly display structured data for human verification on wallet screens or for machine consumption e.g. by transaction simulation engines.
 
-The [ERC-7730](./erc-7730.md) specification enriches type data contained in the ABIs and schemas of structured messages (structures like the calldata of an EVM transaction, an [EIP-712](./eip-712.md) message or an [ERC-4337](./erc-4337.md) User Operation) with additional formatting information and dynamic value interpolation, enabling both human-readable display with contextual intent descriptions and machine-interpretable data processing. For instance, a solidity field containing an amount, encoded as an uint256, can be converted to the right magnitude and appended with the correct ticker for display, or parsed programmatically for transaction simulation. Fields containing encrypted values (such as FHE-encrypted amounts in confidential token standards like [ERC-7984](./erc-7984.md)) can also be annotated with decryption context, enabling wallets to either decrypt and display the plaintext value or present a meaningful fallback.
+The [ERC-7730](./eip-7730.md) specification enriches type data contained in the ABIs and schemas of structured messages (structures like the calldata of an EVM transaction, an [EIP-712](./eip-712.md) message or an [ERC-4337](./eip-4337.md) User Operation) with additional formatting information and dynamic value interpolation, enabling both human-readable display with contextual intent descriptions and machine-interpretable data processing. For instance, a solidity field containing an amount, encoded as an uint256, can be converted to the right magnitude and appended with the correct ticker for display, or parsed programmatically for transaction simulation. Fields containing encrypted values (such as FHE-encrypted amounts in confidential token standards like [ERC-7984](./eip-7984.md)) can also be annotated with decryption context, enabling wallets to either decrypt and display the plaintext value or present a meaningful fallback.
 
 Wallets and automated systems will use curated ERC-7730 files alongside the raw data to sign in order to construct appropriate interfaces for their respective use cases.
 
@@ -28,7 +28,7 @@ Properly validating a transaction on a hardware wallet's screen (also known as C
 - Function name or main message type is often a developer oriented name and does not translate to a clear intent for the user
 - Fields in the data to sign are tied to primitive types only, but those can be displayed in many different ways. For instance, integers can be displayed as percentages, dates, etc...
 - Some fields require additional metadata to be displayed correctly, for instance token amounts require knowledge of the decimals and the ticker, as well as where to find the token address itself to be correctly formatted.
-- Some values can be encrypted, such as encrypted amounts in confidential tokens following [ERC-7984](./erc-7984.md). These values cannot be directly interpreted, so additional context must be provided to enable wallets to decrypt and display them for users.
+- Some values can be encrypted, such as encrypted amounts in confidential tokens following [ERC-7984](./eip-7984.md). These values cannot be directly interpreted, so additional context must be provided to enable wallets to decrypt and display them for users.
 
 This specification intends to provide a simple, open standard format to provide wallets with the additional information required to properly format a structured data to sign for review by users.
 
@@ -181,9 +181,9 @@ To see the ERC text as it read for a specific released version, check out this r
 
 | Version      | Schema                                                                    | Commit ID  | Status     |
 |--------------|---------------------------------------------------------------------------|------------|------------|
-| `1.0.0`      | [`erc7730-v1.schema.json`](../assets/erc-7730/erc7730-v1.schema.json)     | `d42dfc9d` | Deprecated |
-| `2.0.0`      | [`erc7730-v2.schema.json`](../assets/erc-7730/erc7730-v2.schema.json)     | TBD        | Active     |
-| `3.0.0-next` | [`erc7730-v3.0.0-next.schema.json`](../assets/erc-7730/erc7730-v3.0.0-next.schema.json) | N/A | Draft |
+| `1.0.0`      | [`erc7730-v1.schema.json`](../assets/eip-7730/erc7730-v1.schema.json)     | `d42dfc9d` | Deprecated |
+| `2.0.0`      | [`erc7730-v2.schema.json`](../assets/eip-7730/erc7730-v2.schema.json)     | TBD        | Active     |
+| `3.0.0-next` | [`erc7730-v3.0.0-next.schema.json`](../assets/eip-7730/erc7730-v3.0.0-next.schema.json) | N/A | Draft |
 
 `1.0.0` and `2.0.0` are the two schema versions that were released prior to the introduction of this versioning scheme; they are renumbered here under semver with no change to their content or file names. Every version released from now on is added as a new row in this table, without altering the rows already present, and follows the `erc7730-v<MAJOR>.<MINOR>.<PATCH>[-rc.<N>].schema.json` file naming pattern described above.
 
@@ -210,7 +210,7 @@ Displaying structured data is often done by wallets to review its content before
 - The container structure does not necessarily follow the same type system as the structured data.
 - Wallets receive the full container structure and uses the signature scheme to generate a signature on the overall structure.
 
-![Structured Data and Container](../assets/erc-7730/structured-data.svg)
+![Structured Data and Container](../assets/eip-7730/structured-data.svg)
 
 Current specification covers EVM smart contract calldata:
 - Defined in Solidity
@@ -657,7 +657,7 @@ Maps references can be used anywhere a parameter with constant value would be us
 
 A wallet MUST replace a path to a map using the value matching the `keyPath` parameter. In case no values matches the wallet MUST consider the 7730 file invalid for the transaction. 
 
-An example can be found in [example-maps.json](../assets/erc-7730/example-maps.json) and [example-maps-pools.json](../assets/erc-7730/example-maps-pools.json).
+An example can be found in [example-maps.json](../assets/eip-7730/example-maps.json) and [example-maps-pools.json](../assets/eip-7730/example-maps-pools.json).
 
 *Examples*
 
@@ -1051,7 +1051,7 @@ Optional object containing formatter-specific parameters. Available parameters a
 
 When a formatter applies to an array, the parameters CAN also reference a path to an array. In that case, wallets should format each element of the array using the parameter value at the same index as the current element being formatted. Parameter arrays should be the same length as the formatted array else the wallet SHOULD raise an error.
 
-See [example-array-iteration.json](../assets/erc-7730/example-array-iteration.json) for examples on array iteration.
+See [example-array-iteration.json](../assets/eip-7730/example-array-iteration.json) for examples on array iteration.
 
 **`$id`** 
 
@@ -1132,7 +1132,7 @@ Recursive path references work by concatenating the paths of the parents *struct
 
 This recursivity allows structuring the ERC-7730 file itself, but is NOT RECOMMENDED. It is expected that resource limited wallets will only support very limited recursivity in the ERC-7730 file itself, so the initial intent of the spec is to "flatten" the list of fields to display using the *path* mechanics. 
 
-See [example-array-iteration.json](../assets/erc-7730/example-array-iteration.json) for examples on grouping and group iteration control.
+See [example-array-iteration.json](../assets/eip-7730/example-array-iteration.json) for examples on grouping and group iteration control.
 
 #### Slices in paths
 
@@ -1451,7 +1451,7 @@ Formats applicable to `address` solidity type.
 
 | **`raw`**       |                                                                                              |
 |-----------------|----------------------------------------------------------------------------------------------|
-| *Description*   | Display address as an [ERC-55](./erc-55.md) formatted string. Truncation is device dependent |
+| *Description*   | Display address as an [ERC-55](./eip-55.md) formatted string. Truncation is device dependent |
 | *Parameters*    | None                                                                                         |
 | *Examples*      | ---                                                                                          |
 | `0x5aAe...eAed` | Field Value 0x5aAeb6053F3E94C9b9A09f33669435E7Ef1BeAed                                       |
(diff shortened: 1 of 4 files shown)


ERCS/erc-7730.md
```diff
@@ -126,9 +126,38 @@ In this example, the `to` parameter and the `value` parameter SHOULD both be dis
...
+
+This standard may deprecate a version of the schema explicitly. Wallets SHOULD maintain support for all non-deprecated versions of the schema declared in the [Released Version](#released-versions) section.
+
+#### Version numbers
+
+Schema versions follow Semantic Versioning with an optional `-rc.N` pre-release suffix.
+
+#### Schema files and the `$schema` key
+
+Each released and finalized version, including the current pre-release one, is published as its own immutable file named `erc7730-v<VERSION>[-rc.<N>].schema.json`.
```
kuzdogan: Where? In the registry or alongside this EIP?

kuzdogan (commented):
Small question, lgtm in general. Thank you!

But also another bigger picture question is, if we are sure any change in the "spec" is going to be reflected by the "schema"? Like the descriptor has a lot of recommendations and implementation details like the formatters. Do we maybe need to version them separately? (I hope not)

ERCS/erc-7730.md
forshtat: I believe it must be published alongside the EIP in the repository to serve as a single source of truth. But we can find a place in the registry for all supported schemas, say, if that helps with schema validation.

forshtat:
> Small question, lgtm in general. Thank you!
> 
> But also another bigger picture question is, if we are sure any change in the "spec" is going to be reflected by the "schema"? Like the descriptor has a lot of recommendations and implementation details like the formatters. Do we maybe need to version them separately? (I hope not)

From my point of view,  versioning specifications is not as useful at a level of recommendations and implementation details. We can keep git tags for a snapshot of the exact wording of an ERC document at the time the spec "version" was created, and we can have a separately versioned "compatibility test suite", but I don't see a way to maintain the entire body of an ERC with multiple versions.
What do you think?

ERCS/erc-7730.md
```diff
@@ -126,9 +126,38 @@ In this example, the `to` parameter and the `value` parameter SHOULD both be dis
...
+
+The version of the specification used by a descriptor file is the version identified by its `$schema` key, as introduced in the [Simple example](#simple-example). Consumers MUST use this key to select which schema version and which validation and interpretation rules apply to a given descriptor.
+
+Outside of this section, this document's text always describes the current release candidate version of the schema only.
+
+#### Released versions
+
+| Version               | Schema                                                                |
+|-----------------------|-----------------------------------------------------------------------|
+| `1.0.0`               | [`erc7730-v1.schema.json`](../assets/eip-7730/erc7730-v1.schema.json) |
```
manuelwedler: I think v1 should already be explicitly deprecated. Apart from Ledger, no tooling supports it.

ERCS/erc-7730.md
manuelwedler: I think we need it in both places. The registry CI relies on the schema being there.

manuelwedler:
> From my point of view, versioning specifications is not as useful at a level of recommendations and implementation details. We can keep git tags for a snapshot of the exact wording of an ERC document at the time the spec "version" was created, and we can have a separately versioned "compatibility test suite", but I don't see a way to maintain the entire body of an ERC with multiple versions. What do you think?

Maybe we can add in the versioning section that you should find the specific release in the git history if you need the ERC document for an older schema version?

Also, the problem with tagging in the ERC repo is that it would need to be done by someone with write access. There are no git tags at all currently.

ERCS/erc-7730.md
```diff
@@ -126,9 +126,38 @@ In this example, the `to` parameter and the `value` parameter SHOULD both be dis
...
+This ERC will periodically release a new, immutable, explicitly numbered version that may be considered "Final" for all intents and purposes.
+
+Descriptors conforming to this versioned schema are produced and consumed by supporting software and hardware wallets.
+When encountering a descriptor of a yet unsupported version, the wallet MUST consider it invalid.
+
+This standard may deprecate a version of the schema explicitly. Wallets SHOULD maintain support for all non-deprecated versions of the schema declared in the [Released Version](#released-versions) section.
+
+#### Version numbers
+
+Schema versions follow Semantic Versioning with an optional `-rc.N` pre-release suffix.
```
manuelwedler: It's not clear when the rc version is incremented? Like should every change to the schema include an increment to the rc version?

manuelwedler (commented):
The process of releasing is not entirely clear to me after reading. I mainly wonder how it works with the ERC process. Should the version be incremented for every PR in the ERC repo that changes the schema? Do we then keep this updated schema as a new file? Basically any merged PR goes to https://eips.ethereum.org/ and people might build against that released version.

forshtat:
> The process of releasing is not entirely clear to me after reading. I mainly wonder how it works with the ERC process. Should the version be incremented for every PR in the ERC repo that changes the schema? Do we then keep this updated schema as a new file? Basically any merged PR goes to https://eips.ethereum.org/ and people might build against that released version.

I think every PR that alters the schema de-facto creates a new "release candidate" version (i.e. `v3.0.0-rc.3`).
It is the authors' responsibility to make sure the new R.C. files are created, and it's okay to immediately drop the old R.C. files.
People should not build against a "release candidate" schema unless they really know what they are doing.
Once the authors make a decision to finalize and release the version, it gets a "major" number (`v3.0.0-rc.3` -> `v3.0.0`).
If there is a new PR being merged into the ERC, it goes to a "release candidate" for the next major version (i.e. `v4.0.0-rc.1`)

> Maybe we can add in the versioning section that you should find the specific release in the git history if you need the ERC document for an older schema version? Also, the problem with tagging in the ERC repo is that it would need to be done by someone with write access. There are no git tags at all currently.

I can add a "commit ID" column for releases, and this way we can have a back reference to a specific ERC text version without needing a git tag write access or manual versioning of plain text. What do you think?

manuelwedler (approved):
Thanks! The changes make the process clear and resolve my comments. From my side, this looks good to be merged.

llbartekll:
I'm actually thinking that $commit feels redundant. $schema version already maps 1:1 to a commit, so the descriptor doesn't carry any new info. Maybe drop it?

kuzdogan:
Ah yeah what I meant was if we can include the `commit` in the JSON schema itself, like if there's a metadata field "this schema is based on the `a32efac1` version of the ERC spec" but then you wouldn't know the commit hash ahead so maybe just a date would work? 

I also think `$commit` is too much

The goal is to say "this schema is based on this human readable spec". 

Otherwise it's not a big deal and not a blocker to me

hilljessica8787-dot (approved):

manuelwedler:
> Ah yeah what I meant was if we can include the `commit` in the JSON schema itself, like if there's a metadata field "this schema is based on the `a32efac1` version of the ERC spec" but then you wouldn't know the commit hash ahead so maybe just a date would work?

json schemas allow a `$comment` key that could be used for this. See https://json-schema.org/understanding-json-schema/reference/comments

But yeah it would need to be added in an additional commit or something.

kuzdogan:
Ok as the spec and the schema changes will happen with commits, they will be in sync implicitly. So I'm fine without extra fields etc. Just versioning the schema is enough. We can add a comment mentioning the schema reflects the human readable spec of the same date.

forshtat:
> Ok as the spec and the schema changes will happen with commits, they will be in sync implicitly. So I'm fine without extra fields etc. Just versioning the schema is enough. We can add a comment mentioning the schema reflects the human readable spec of the same date.

Noted - I pushed a change that makes the "$commit" optional (`MAY`) in case someone needs to target a very specific, in-between version of ERC-7730 for some reason. Also stated that the schema's `$comment` MUST contain the date and the commit ID of the finalized ERC version.

Let me know if we are getting close 😀

ERCS/erc-7730.md
kuzdogan: I believe so, below it says 

> A merged pull request against this ERC that changes the schema produces a new release-candidate version of that schema.
It must publish the corresponding `-rc.N+1` file

ERCS/erc-7730.md
kuzdogan: Would it complicate things if we point the CI to here instead? 

If so we can have an external periodical test to check the spec vs. registry schema equivalence

ERCS/erc-7730.md
```diff
@@ -126,9 +126,76 @@ In this example, the `to` parameter and the `value` parameter SHOULD both be dis
...
+
+This binds a descriptor to an exact, independently verifiable point in this document's history, removing any remaining ambiguity beyond the schema validation.
+
+*Example*
+
+```json
+{
+    "$schema": "https://eips.ethereum.org/assets/eip-7730/erc7730-v3.0.0-rc.1.schema.json",
+    "$commit": "a1b2c3d4e5f6a7b8c9d0e1f2a3b4c5d6e7f8a9b0"
+}
```
kuzdogan: I'd say let's get rid of this and just have the implicit binding as we describe above

ERCS/erc-7730.md
```diff
@@ -126,9 +126,76 @@ In this example, the `to` parameter and the `value` parameter SHOULD both be dis
...
+|--------------|-----------------------------------------------------------------------|------------|-------------------|
+| `1.0.0`      | [`erc7730-v1.schema.json`](../assets/eip-7730/erc7730-v1.schema.json) | `d42dfc9d` | Deprecated        |
+| `2.0.0`      | [`erc7730-v2.schema.json`](../assets/eip-7730/erc7730-v2.schema.json) | `a124a871` | Active            |
+| `3.0.0-rc.1` | TBD                                                                   | TBD        | Release Candidate |
+
+`1.0.0` and `2.0.0` are the two schema versions that were released prior to the introduction of this versioning scheme; they are renumbered here under semver with no change to their content or file names. Every version released from now on is added as a new row in this table, without altering the rows already present, and follows the `erc7730-v<MAJOR>.<MINOR>.<PATCH>[-rc.<N>].schema.json` file naming pattern described above.
+
+The `Commit ID` column contains the acceptable [`$commit`](#the-commit-key) values for descriptors, and the value that a schema file's `$comment` key SHOULD record when finalizing that version.
```
kuzdogan: Similarly, remove `$commit`

ERCS/erc-7730.md
```diff
@@ -126,9 +126,76 @@ In this example, the `to` parameter and the `value` parameter SHOULD both be dis
...
-The version of the specification used by the 7730 file is specified by linking to the reference schema file under the `$schema` key.
+The descriptor schema is versioned as part of this standard.
 
-The current version of the specification is `2` at the time of writing. The reference schema for version `2` is [here](../assets/eip-7730/erc7730-v2.schema.json).
+This ERC will periodically release a new, immutable, explicitly numbered version that may be considered "Final" for all intents and purposes.
+
+Descriptors conforming to this versioned schema are produced and consumed by supporting software and hardware wallets.
+When encountering a descriptor of a yet unsupported version, the wallet MUST consider it invalid.
+
+This standard may deprecate a version of the schema explicitly. Wallets SHOULD maintain support for all non-deprecated versions of the schema declared in the [Released Version](#released-versions) section.
```
kuzdogan:
```suggestion
This standard may deprecate a version of the schema explicitly. Wallets and tooling SHOULD maintain support for all non-deprecated versions of the schema declared in the [Released Version](#released-versions) section.
```

ERCS/erc-7730.md
```diff
@@ -126,9 +126,76 @@ In this example, the `to` parameter and the `value` parameter SHOULD both be dis
...
+
+Once a version is finalized, its file contents MUST NOT change. Any subsequent change, including error fixes, is published as a new version with a new file.
+
+Release-candidate files MAY be replaced or removed in the process of finalizing a version.
+
+Each schema file is published in two locations: alongside this ERC under the "assets" folder, and mirrored into the descriptor registry that consumes these files.
+
+Since the schema and the spec text are always changed together in the same pull request, a finalized schema file implicitly stays in sync with the ERC text as it read at that time. To make this traceable without introducing a separate versioning scheme, a finalized schema file's top-level `$comment` key MUST record the date and the commit ID at which that version was finalized, pointing back to the exact revision of this document the schema corresponds to.
+
+The version of the specification used by a descriptor file is the version identified by its `$schema` key, as introduced in the [Simple example](#simple-example). Consumers MUST use this key to select which schema version and which validation and interpretation rules apply to a given descriptor.
```
kuzdogan, line 174: I feel we should have the version inside the schema file as the source of truth for the version, rather than the file name:

```json
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "$id": "https://eips.ethereum.org/assets/eip-7730/erc7730-v3.0.0-rc.1.schema.json",
  "version": "3.0.0-rc.1",
  "title": "ERC-7730 Clear Signing Descriptor Schema",
  ...
}
```

So that the consumers don't rely on file name parsing for version parsing.
But still keep the file naming as a convention, not as the version identification.

ERCS/erc-7730.md
```diff
@@ -126,9 +126,76 @@ In this example, the `to` parameter and the `value` parameter SHOULD both be dis
...
+
+Each released and finalized version, including the current pre-release one, is published as its own immutable file named `erc7730-v<VERSION>[-rc.<N>].schema.json`.
+
+Once a version is finalized, its file contents MUST NOT change. Any subsequent change, including error fixes, is published as a new version with a new file.
+
+Release-candidate files MAY be replaced or removed in the process of finalizing a version.
+
+Each schema file is published in two locations: alongside this ERC under the "assets" folder, and mirrored into the descriptor registry that consumes these files.
+
+Since the schema and the spec text are always changed together in the same pull request, a finalized schema file implicitly stays in sync with the ERC text as it read at that time. To make this traceable without introducing a separate versioning scheme, a finalized schema file's top-level `$comment` key MUST record the date and the commit ID at which that version was finalized, pointing back to the exact revision of this document the schema corresponds to.
```
kuzdogan: Sorry for this :D but let's keep it simple and remove this. Anyone who needs to read the corresponding descriptor can come here and see the commit history.

ERCS/erc-7730.md
```diff
@@ -126,9 +126,76 @@ In this example, the `to` parameter and the `value` parameter SHOULD both be dis
...
 ### Versioning
 
-The version of the specification used by the 7730 file is specified by linking to the reference schema file under the `$schema` key.
+The descriptor schema is versioned as part of this standard.
 
-The current version of the specification is `2` at the time of writing. The reference schema for version `2` is [here](../assets/eip-7730/erc7730-v2.schema.json).
+This ERC will periodically release a new, immutable, explicitly numbered version that may be considered "Final" for all intents and purposes.
+
+Descriptors conforming to this versioned schema are produced and consumed by supporting software and hardware wallets.
+When encountering a descriptor of a yet unsupported version, the wallet MUST consider it invalid.
```
kuzdogan: This should be different in MAJOR vs. MINOR or PATCH. It should only reject the MAJOR versions.

kuzdogan (commented):
Couple more comments, now I have a better understanding, we're getting closer :) 

1. Let's remove all of the commit binding. As I said it's implicit. Let's keep it simple
2. Unless there's a reason not to, I feel the schema's own version should be inside the schema itself, not derived from the file name. But we can keep the convention
3. The tests also have their own schema `erc7730-tests.schema.json` how should we handle that @manuelwedler ? Not necessarily in this spec but how?
4. Also not spec but registry: There could be multiple descriptors with two different versions like `aave-erc7730-v1` + `aave-erc7730-v2`+ `aave-erc7730-v2.2.1`. Bu we have a single index in the registry. I guess we need indexes for each schema version? cc @manuelwedler

ERCS/erc-7730.md
forshtat: Removed

ERCS/erc-7730.md
forshtat: Removed

ERCS/erc-7730.md
forshtat: Done

ERCS/erc-7730.md
forshtat: Added `"version"` field to the schema, and also to the existing v1 and v2 schema files (this should not be an issue - this is just adding a field)

ERCS/erc-7730.md
forshtat: Got it, rephrased the entire paragraph.

ERCS/erc-7730.md
forshtat: Specified MAJOR as "must reject", MINOR as "must continue parsing", right?

forshtat:
> Couple more comments, now I have a better understanding, we're getting closer :)
> 
> 1. Let's remove all of the commit binding. As I said it's implicit. Let's keep it simple

Done.

> 2. Unless there's a reason not to, I feel the schema's own version should be inside the schema itself, not derived from the file name. But we can keep the convention

Added.

> 3. The tests also have their own schema `erc7730-tests.schema.json` how should we handle that @manuelwedler ? Not necessarily in this spec but how?

As far as I understand, these tests are not currently part of ERC-7730 itself, right? Should they be?

> 4. Also not spec but registry: There could be multiple descriptors with two different versions like `aave-erc7730-v1` + `aave-erc7730-v2`+ `aave-erc7730-v2.2.1`. Bu we have a single index in the registry. I guess we need indexes for each schema version? cc @manuelwedler

Sounds like it, or maybe a nested index (`context` -> `versions` -> `descriptor`)? One approach makes each index files smaller, the other makes falling back to older descriptor version easier.


UPD: @kuzdogan I invited you to have a full write access to this fork if you want to make any changes yourself

kuzdogan (commented):
Thank you! lgtm as is. Just still not approving yet to avoid auto-merge. 

A short review from the Ledger side @lcastillo-ledger needed IMO since the changes are decisive moving forward.

manuelwedler:
> 3\. The tests also have their own schema `erc7730-tests.schema.json` how should we handle that @manuelwedler ? Not necessarily in this spec but how?

The test schema should not be part of the spec. The test schema is relatively simple and only describes how expected texts should be written. I don't expect that the test schema needs to change with spec updates.

> 4\. Also not spec but registry: There could be multiple descriptors with two different versions like `aave-erc7730-v1` + `aave-erc7730-v2`+ `aave-erc7730-v2.2.1`. Bu we have a single index in the registry. I guess we need indexes for each schema version? cc @manuelwedler

That is getting more complicated than I hoped. Ideally, descriptors can be updated and we don't need to keep multiple versions of it. But depends on how the spec evolves. I think it at least it does not makes sense to have the same descriptor for different minor versions of the same major version as in your example. minor bumps should be backwards compatible, so no need to keep two versions. 
For major updates, we could add the version information in the index. Preferred by me to not duplicate the data across a lot of index files. In general, we could rethink how we handle the index. I think it was not designed with all edge cases in mind. Also eip712 and calldata indexes differ in their structure.

forshtat:
> For major updates, we could add the version information in the index. Preferred by me to not duplicate the data across a lot of index files. In general, we could rethink how we handle the index. I think it was not designed with all edge cases in mind. Also eip712 and calldata indexes differ in their structure.

@manuelwedler @kuzdogan It seems like the index files, and their format, are not currently standardised by ERC-7730 itself but are part of the registry repository directly, so I did not address them in this PR.
However, I am wondering if there is a good reason not to document the format for the index files in the ERC directly?

kuzdogan:
> However, I am wondering if there is a good reason not to document the format for the index files in the ERC directly?

The spec is explicitly not about the registry. I think it should not be documented here: "Curation and registry of ERC-7730 files are out of scope for this specification"

I just wanted to bring it up as it came to my mind.

ERCS/erc-7730.md
```diff
@@ -126,9 +126,64 @@ In this example, the `to` parameter and the `value` parameter SHOULD both be dis
...
+
+The schema and this ERC text are always changed together in the same pull request. A finalized schema file remains in sync with the ERC text as it read at the time of finalization.
+
+The version of the specification used by a descriptor file is the version identified by its `$schema` key, as introduced in the [Simple example](#simple-example). Consumers MUST use this key to select which schema version and which validation and interpretation rules apply to a given descriptor.
+
+Outside of this section, this document's text always describes the current release candidate version of the schema only.
+
+To see the ERC text as it read for a specific released version, check out this repository at the commit referenced for that version in the [Released versions](#released-versions) table below.
+
+#### Released versions
```
ankurdotb, line 180: The PR itself suggests using git tags, which don't exist. I'd suggest adding them.

Commit ID is a useful add-on but it's useless without also providing a checksum and checksum of what while should be specified.

ERCS/erc-7730.md
```diff
@@ -126,9 +126,64 @@ In this example, the `to` parameter and the `value` parameter SHOULD both be dis
...
+
+| Version      | Schema                                                                | Commit ID  | Status            |
+|--------------|-----------------------------------------------------------------------|------------|-------------------|
+| `1.0.0`      | [`erc7730-v1.schema.json`](../assets/eip-7730/erc7730-v1.schema.json) | `d42dfc9d` | Deprecated        |
+| `2.0.0`      | [`erc7730-v2.schema.json`](../assets/eip-7730/erc7730-v2.schema.json) | `a124a871` | Active            |
```
ankurdotb: I think the commit IDs are incorrect here, especially after the JSON schema change.

ankurdotb (commented):
Can I make a suggestion here?

I think having git tags and a separate file per RC (`erc7730-v3.0.0-rc.1.schema.json`, `-rc.2`, etc., created and deleted as candidates churn) could become a nightmare to manage, especially once there's real churn while a version is being built.

It'll also make the table in this PR hard to keep accurate. For example, the 2.0.0 row already points at a commit (`a124a871`) that predates the file actually reaching its current shape, and this PR is itself editing the "Active" 2.0.0 file's content while marking it finalized, which is exactly the kind of drift I think we should avoid.

I also think it's worth naming that no other ERC does versioning this way. I checked: the closest thing is ERC-5139, which versions a data document via a self-declared version field and MUST-bump rules, but never mints separate files per version and has no `rc` concept at all. So this is a genuinely new mechanism for this repo, not something borrowed from precedent. Maybe that's fine and we all agree to it, but the current suggestion is complex.

So here's what I'd propose instead:
1. Keep a `erc7730-draft.schema.json` (or any other name like `erc7730-next.schema.json`: One mutable file, always representing whatever's currently being worked on. Its internal version field can say 3.0.0-rc.1 as an informal courtesy annotation, but nothing depends on that being accurate. If someone needs to pin a specific in-progress state such as for testing, for the registry, whatever, they use a git tag (or just the commit SHA) against the commit that touched this file. No file to create, modify, or delete for each RC.
2. Stable versions become `erc7730-vN.schema.json`: Created exactly once, at the moment a version is declared final. Copy next's current content, bump the internal version field to the clean number (no -rc), add the table row. Never touched again, no exceptions. This is the only thing the table should point to.
3. **We need to say who creates the git tag** and I think this should be **whoever merges the PR should create the git tag**. Ideally I think this could/should be automated as there is scope to forget doing it. Git tags should be creatred after the PR is merged, not before, since we use squash-merges on this repo, so tagging within a PR branch will create an orphaned tag. So: the tag should to be created by whoever merges the PR (an EIP editor), as an immediate next step after merging, against the commit that actually lands on `master` branch. We should capture this in the PR.
4. Drop or clarify the "Commit ID" column in the table: The `$commit-key` mechanism from earlier in this branch's history got reverted, including the sentence explaining what the "Commit ID" column is for, but the column itself and its values stayed in the table. Right now nothing in the text says what a reader's supposed to do with those hashes. Worth either restoring a defining sentence, or (my suggestion) dropping the column from the table and instead cover it with git tags.

kuzdogan:
@ankurdotb Keeping a single `-next` file is a great idea and I had missed that situation having potentially too many `-rc` files. But I don't see how and where it's suggested to use git tags. 

I don't think we really need tags plus I'm not sure if we can write tags in this repo at all. I see none: https://github.com/ethereum/ERCs/tags

hilljessica8787-dot (approved):

kuzdogan:
lgtm :) Just @forshtat pls add yourself as an author. Thank you!

kuzdogan (approved):

PR #1815, opened 2026-06-15
Add ERC: AI Agent Execution

JimmyShi22:
## Abstract

Defines three minimal standard interfaces for on-chain AI agent task dispatch — `AgentTask`, `IAgentCaller`, and `IAgentHandler` — along with a composable workflow model that scales from a simple single-agent call to distributed multi-round consensus, without changing the base interface.

## Motivation

Every dApp that wants to invoke an AI agent today defines its own task format, and every agent must implement a separate adapter per dApp — the same N×M fragmentation that ERC-8274 solved at the verification layer. This ERC adds the missing execution primitive to the on-chain AI agent base layer:

| Primitive | ERC |
|-----------|-----|
| Identity | ERC-8004 |
| **Execution** | **This ERC** |
| Verify | ERC-8274 |
| Anchor | ERC-8263 |
| Settlement | ERC-8275 |
| Prove | ERC-8281 + ERC-8299 |

## Key Design Decisions

- **`agentWorkflowHash`**: a `keccak256` commitment to a workflow definition document, committed at dispatch time so the agent cannot substitute an unauthorized transform after seeing the input. Replaces the earlier `systemPromptHash` concept by encompassing the full execution protocol (behavior, model, stage definitions).
- **`onAgentStep(stage, isFinal)`**: a single unified callback replacing separate pre-reply / reply callbacks. Any workflow position — from a simple single-step result to a multi-round BFT consensus vote — maps onto `(stage, isFinal)`.
- **`onAgentProve(stage)`**: proof submission for any stage, decoupled from step submission. Can be called by a third party, not only the agent.
- **Minimum path**: `callAgent()` → `onAgentStep(stage=0, isFinal=true)`. No additional infrastructure required for the simplest case.
- **Workflow format**: intentionally left open in this version; standardizing the document format warrants separate community discussion.

## Relationship to Other ERCs

- **ERC-8274**: `AgentVerifier` is the recommended home for stage state management; `onAgentProve` passes a `verifier` address directly.
- **ERC-8299 (WYRIWE)**: Stage 1 input preparation integrates WYRIWE attestation via `onAgentStep(stage=1, data=<attestation>)`.
- **ERC-8004**: `agentId` in `onAgentStep` carries the ERC-8004 agent identity.
- **ERC-8275**: Stage 2+ output consensus patterns (commit-reveal, BFT) are designed for distributed inference settlement.

## Discussion

https://ethereum-magicians.org/t/draft-erc-ai-agent-execution/28785

PR #85, opened 2023-11-02
Add ERC: Crypto Security Token Smart Contract Interface

itinance:
This ERC was created by the members of the German Federal Association of Crypto Securities Registry Administrators to comply with the requirements of the so-called German Electronic Securities Act (German: "Elektronisches Wertpapiergesetz" = "eWpG") and to establish a standard for Security Tokens that adhere to the law.

PR #1851, opened 2026-07-03
Add ERC: Source-Token Agent Binding for ERC-8004

babyblueviper1:
## Summary

Companion standard to [ERC-8004](https://eips.ethereum.org/EIPS/eip-8004) (Trustless Agents). ERC-8004 assumes an agent identity is minted fresh with no relationship to a pre-existing asset. In practice the most natural agents are derived from NFTs people already hold — a PFP, a membership token, a game character — and there's no standard way to say, verifiably and on-chain, "this agent *is* that NFT" without conflating two facts that should stay separate:

- **Provenance** (`getSourceNFT` + `SourceNFTLinked`): which source token an agent was derived from. Immutable, permanent.
- **Live ownership** (`isSourceNFTOwnershipValid`): whether the current agent holder still owns that source token. Mutable, re-checked at query time, never cached.

Non-custodial by design — `ownerOf` is checked once at mint, the source token is never locked, escrowed, or transferred. Explicitly compatible with [ERC-6551](https://eips.ethereum.org/EIPS/eip-6551) token-bound-account and binding-custody (sovereignty) patterns: `isSourceNFTOwnershipValid` treats the agent's canonical TBA or the binding contract itself as valid holders, not just the literal agent owner — a bare `ownerOf(source) == ownerOf(agentId)` check is called out as non-conformant since it force-fails those patterns. Re-mint is explicitly permitted (a source token changing hands lets each successive holder mint an agent from it), and the spec is explicit that track record never travels with the source token — a re-minted agent re-earns its record from zero, which is what keeps source-token resale Sybil-neutral at the trust layer.

## Status

Reference implementation is deployed and live on Ethereum mainnet across 4 source NFT collections (the `dinamic.eth` / `ens-dynamic-kit` stack) — every contract address in the Reference Implementation section is independently verifiable via a single `eth_call`. Discussion thread: https://ethereum-magicians.org/t/28920.

Co-authored with @TMerlini.

## Test plan

- [ ] Editors assign the `eip:` number (left as `<TBD — editors assign>` per convention)
- [ ] CI / eipw preamble checks pass once the number is assigned


TMerlini:
Co-author here (@TMerlini) — thanks for opening it, @babyblueviper1.

For reviewers/editors: every address in the Reference Implementation is a live `eth_call`, verifiable without trusting us —

- **Mainnet** (Pixel Goblins registry `0xe0454dFA17a57a84c3E0E2DBFdA5318CBBE91e2c`): agent #1 → `isSourceNFTOwnershipValid(1)` = **true** (still holds source Goblin #85); agent #14 → **false** (source #577 changed hands, while `getSourceNFT(14)` still resolves). That's the immutable-provenance / mutable-ownership separation the standard draws, observable live.
- **Sepolia** fully-conformant reference `0x01B014fc69d823FfEb80C313C29Bb6ead3978CEe`: `supportsInterface(0x27eba962)` = **true**, with an agent whose `isSourceNFTOwnershipValid(1)` = true.
- The interface id `0x27eba962` is the XOR of the five function selectors — both authors re-derived it independently.

`eip:` left TBD for the editors as agreed; discussion is linked in the frontmatter. The one design point we'd most welcome eyes on is the **sovereignty / TBA carve-out** in `isSourceNFTOwnershipValid`: a bare `ownerOf(source) == ownerOf(agentId)` check would force-fail the legitimate case where the source token is held by the agent's own ERC-6551 token-bound account — spelled out in Security Considerations.

babyblueviper1:
EIP Walidator is green. The remaining HTMLProofer red is 100% dependency-ordering, not a content issue: two normative refs (ERC-8299, ERC-8274) link to sibling specs that haven't merged into this repo yet (still open as #1810 and #1771). Both links follow this repo's established convention of linking sibling refs even pre-merge rather than leaving them as unlinked plain text (which the `markdown-link-first` lint requires). This resolves itself as each sibling merges — nothing in this file can fix it in the meantime.

ERCS/erc-8330.md
```diff
@@ -0,0 +1,181 @@
+---
+eip: 8330
```
SamWilsn: You can't pick your own number. Please update this to:


```suggestion
eip: 8323
```

Don't forget to change the file name, and your assets directory.

ERCS/erc-8323.md
```diff
@@ -0,0 +1,181 @@
+---
+eip: 8323
+title: Source-Token Agent Binding for ERC-8004
+description: Derive an ERC-8004 agent identity from a pre-existing ERC-721, recording permanent provenance while exposing live ownership separately.
+author: Tiago Merlini (@TMerlini) <merloproductions@gmail.com>, babyblueviper1 (@babyblueviper1)
+discussions-to: https://ethereum-magicians.org/t/source-token-agent-binding-for-erc-8004/28920
```
abcoathup:
```suggestion
discussions-to: https://ethereum-magicians.org/t/erc-8323-source-token-agent-binding-for-erc-8004/28920
```

babyblueviper1:
Applied — discussions-to now points at the erc-8323-prefixed slug (verified the old URL 301s to it, so this is just making the canonical link explicit). Thanks for catching it.

TMerlini:
@babyblueviper1 — first pass at the `IAgentSourceBindingView` companion, as a diff-able patch rather than prose so you can review against it and take the second pass on the conformance MUSTs.

Cut against this PR's head (`2c4af98`); applies clean with `patch -p1` (verified, zero fuzz). Additive only — no change to `IAgentSourceBinding` (`0x27eba962`), no frontmatter change.

**What it adds**
- `assets/erc-8323/IAgentSourceBinding.sol`: the `IAgentSourceBindingView` interface (the 3 read functions), id `0x8b3597c9`, `@dev` noting it's an *independent* id — the interfaces aren't inheritance-linked, so neither drifts.
- Spec §"Query-only source binding": the interface + the four conformance rules (advertise `0x8b3597c9` ⇒ implement all three; MUST NOT claim `0x27eba962` without the two write fns; a full impl MAY/SHOULD also advertise the view; consumer guidance), closing on the framing — the honest limit of what a self-sourced agent can attest, not a lesser interface.
- Rationale bullet ("Honest ERC-165 over convenient ERC-165"), incl. the `type(...).interfaceId` no-drift point.
- Reference-Implementation note: the live self-sourced `GenesisAgentRegistry` (`0xe919…4963`), `eth_call`-verifiable like the rest of that section — `0x8b3597c9`→true, `0x27eba962`→false on-chain.

**Two calls for your eyes**
1. I made all three view functions explicitly *"same semantics as `IAgentSourceBinding`"* so the view can't read as a semantic fork, only a selector subset — that's the one substantive editorial call, trim if you want it terser.
2. The Reference-Impl address line is optional — it commits the ERC to citing the examples contract. Kept it because the section already lists every reference address, but it's the first thing to drop if you'd rather leave Reference Implementation as yours.

The TBA-sovereignty clause in `isSourceNFTOwnershipValid` I left pointing back at your existing wording rather than restating it — that one's yours to sharpen.

```diff
--- a/assets/erc-8323/IAgentSourceBinding.sol
+++ b/assets/erc-8323/IAgentSourceBinding.sol
@@ -65,3 +65,37 @@
     ///      no longer exists.
     function isSourceNFTOwnershipValid(uint256 agentId) external view returns (bool);
 }
+
+/// @title IAgentSourceBindingView — query-only subset of IAgentSourceBinding
+/// @notice The read side of source binding: prove where an agent's source lives
+///         and whether it is still owned, WITHOUT source-binding registration.
+///         For agents that can attest source ownership but do not implement
+///         `boundCollection` / `registerWithSource` — the canonical case being a
+///         SELF-SOURCED agent whose source is the agent contract itself
+///         (`getSourceNFT(id) == (address(this), id)`), which has no external
+///         collection to bind to.
+/// @dev ERC-165 interface id: 0x8b3597c9
+///      (getSourceNFT ^ hasSourceNFT ^ isSourceNFTOwnershipValid).
+///      This is an INDEPENDENT id, not derived from `IAgentSourceBinding`
+///      (0x27eba962): the three functions share signatures with the full
+///      interface, but the interfaces are intentionally not linked by
+///      inheritance, so each id stays fixed to the functions it declares.
+///      Function semantics are identical to `IAgentSourceBinding`, including the
+///      live-ownership 3-case rule of `isSourceNFTOwnershipValid`.
+interface IAgentSourceBindingView is IERC165 {
+    /// @notice The immutable source token an agent was derived from.
+    /// @dev MUST revert if `agentId` does not exist or has no source binding.
+    function getSourceNFT(uint256 agentId)
+        external
+        view
+        returns (address sourceContract, uint256 sourceTokenId);
+
+    /// @notice Whether `agentId` has a recorded source binding.
+    function hasSourceNFT(uint256 agentId) external view returns (bool);
+
+    /// @notice Whether the source token is still under the control of `agentId`.
+    /// @dev Same live-ownership semantics as
+    ///      `IAgentSourceBinding.isSourceNFTOwnershipValid` (the 3-case holder /
+    ///      canonical ERC-6551 TBA / binding-contract rule).
+    function isSourceNFTOwnershipValid(uint256 agentId) external view returns (bool);
+}

--- a/ERCS/erc-8323.md
+++ b/ERCS/erc-8323.md
@@ -88,6 +88,39 @@
 }
 ```
 
+### Query-only source binding
+
+Not every agent that can prove source ownership implements source-binding *registration*. The canonical case is a **self-sourced** agent, whose source is the agent contract itself — `getSourceNFT(agentId)` returns `(address(this), agentId)` — so there is no external collection and `boundCollection` / `registerWithSource` have no meaning for it. Such an implementation MUST NOT advertise `IAgentSourceBinding` (`0x27eba962`): it does not implement all five functions, and advertising the full id is a false [ERC-165](./eip-165.md) positive that a consumer would trust.
+
+For these implementations this standard defines a query-only companion interface, `IAgentSourceBindingView` — the read side of source binding, with [ERC-165](./eip-165.md) id `0x8b3597c9` (the XOR of the `getSourceNFT`, `hasSourceNFT`, and `isSourceNFTOwnershipValid` selectors):
+
+```solidity
+interface IAgentSourceBindingView /* is IERC165 */ {
+    /// @notice The immutable source token an agent was derived from.
+    /// @dev MUST revert if `agentId` does not exist or has no source binding.
+    function getSourceNFT(uint256 agentId)
+        external
+        view
+        returns (address sourceContract, uint256 sourceTokenId);
+
+    /// @notice Whether `agentId` has a recorded source binding.
+    function hasSourceNFT(uint256 agentId) external view returns (bool);
+
+    /// @notice Whether the source token is still under the control of `agentId`.
+    ///         Same semantics as in `IAgentSourceBinding`.
+    function isSourceNFTOwnershipValid(uint256 agentId) external view returns (bool);
+}
+```
+
+The three functions carry the **same semantics** as in `IAgentSourceBinding`, including the live-ownership rule of `isSourceNFTOwnershipValid` (the holder / canonical [ERC-6551](./eip-6551.md) TBA / binding-contract cases). `IAgentSourceBindingView` is an independent interface, **not** a base of `IAgentSourceBinding`: the two ids stay fixed to the functions each declares.
+
+- An implementation that returns `true` from `supportsInterface(0x8b3597c9)` MUST implement all three functions with the semantics defined in this standard.
+- Such an implementation MUST return `false` from `supportsInterface(0x27eba962)` unless it also implements `boundCollection` and `registerWithSource`.
+- A full `IAgentSourceBinding` implementation MAY, and SHOULD, also return `true` from `supportsInterface(0x8b3597c9)` — the three functions are a subset it already implements — so a view-only consumer can detect it without probing the full interface.
+- A consumer that only needs to read and re-verify source ownership SHOULD accept either `0x8b3597c9` or `0x27eba962`. A consumer that requires source-binding registration MUST check `0x27eba962`.
+
+The boundary is drawn here deliberately: `IAgentSourceBindingView` is not "a smaller interface for lesser contracts" but the **honest limit of what a self-sourced agent can attest**. Whether a source exists and is still owned is universal; registration presumes an external collection. An implementation MUST advertise only the interface it implements.
+
 ### Registration
 
 - `registerWithSource` MUST verify live ownership of the source token by the caller and MUST revert otherwise.
@@ -136,6 +169,7 @@
 - **Agent identity as ERC-721:** makes agents themselves transferable, marketplace-compatible, and royalty-bearing ([ERC-2981](./eip-2981.md)), reusing existing infrastructure rather than inventing custody.
 - **Metadata inheritance preserves identity:** mirroring the source token's metadata means the agent presents as its source NFT, making the "this agent *is* that token" claim legible to wallets and marketplaces, not just to contracts reading `getSourceNFT`.
 - **Fees are non-normative.** `registerWithSource` is `payable` to accommodate implementations that charge a mint fee, but the standard does not define a fee, treasury, or pricing mechanism — these are implementation concerns. The reference implementation's per-registry `mintPrice` → `treasury` forwarding is illustrative only. A conformant registry MAY mint for free.
+- **Honest ERC-165 over convenient ERC-165.** A self-sourced or read-only-source agent implements the read side of binding but not registration. Rather than let it advertise the full `IAgentSourceBinding` and misreport `supportsInterface`, the standard gives that profile its own id (`0x8b3597c9`). Advertising exactly what is implemented is the purpose of [ERC-165](./eip-165.md); the companion view interface makes the honest choice expressible instead of forcing a false positive. Deriving the id via `type(IAgentSourceBindingView).interfaceId` (rather than a hardcoded constant) keeps it from drifting from the functions the interface declares.
 
 ## Backwards Compatibility
 
@@ -165,6 +199,8 @@
 
 Source: `AgentIdentityRegistry.sol` in the `ens-dynamic-kit` stack. The reference implementation additionally provides [EIP-712](./eip-712.md)-verified agent wallets (`setAgentWallet`), ERC-2981 royalties, source-metadata inheritance, and an `{agentId}` URI template — only the `IAgentSourceBinding` surface is normative.
 
+**Self-sourced profile (`IAgentSourceBindingView`).** The query-only companion is exercised by a self-sourced registry — an agent whose source is the registry itself. Reference: `GenesisAgentRegistry` at `0xe91934aB1f6A40cc1Bb4cD530FEFF56dFE524963` (mainnet), for which `getSourceNFT(agentId)` returns `(0xe91934aB1f6A40cc1Bb4cD530FEFF56dFE524963, agentId)`, `supportsInterface(0x8b3597c9)` returns `true`, and `supportsInterface(0x27eba962)` returns `false` — advertising the read side it implements without the registration side it does not. Source: `GenesisAgentRegistry.sol` in the same stack.
+
 ## Security Considerations
 
 - **Provenance ≠ current ownership.** Consumers MUST NOT treat `getSourceNFT` as proof the agent owner currently holds the source token; use `isSourceNFTOwnershipValid` for that. Conversely, a `false` from the live check does not retroactively invalidate provenance.
```

babyblueviper1:
Got the patch -- reviewing the conformance MUSTs and the TBA-sovereignty clause properly before applying and pushing. Back shortly.

babyblueviper1:
Applied and pushed (cd004397 on top of your patch, 2c4af98 base) — thanks for cutting it as a diff, made this fast to verify.

Both editorial calls, decided:
1. Kept the explicit "same semantics" statement, but strengthened it: rather than a descriptive line, it's now a MUST that names the specific trap — a self-sourced implementation isn't exempt from the TBA/binding-contract sovereignty cases just because it lacks registration, so a naive `ownerOf(sourceToken) == ownerOf(agentId)` check is non-conformant here for the same reason it is in the full interface. That's the TBA-sovereignty sharpening you left for me.
2. Kept the `GenesisAgentRegistry` reference-implementation citation — re-verified live on-chain just now (`supportsInterface(0x8b3597c9)` → true, `supportsInterface(0x27eba962)` → false, via a fresh `cast call`), so it's independently reproduced evidence, not just cited.

Also independently recomputed both interface ids via `type(...).interfaceId` in a throwaway Foundry test rather than trusting the hex: `0x8b3597c9` for the view interface, `0x27eba962` unchanged for the full one. eipw (pinned action version, 0.9.0) reports zero findings on the changed file; Solidity compiles clean under 0.8.26.

TMerlini:
**Cross-referencing a coordinated cluster request.** ERC-8323 (this PR) is part of a group of composing, editor-ready AI-inference-trust-stack ERCs; surfacing them together so the set can be triaged as one family rather than one PR at a time.

- **ERC-8281 — Observation Commitment Protocol (OCP)** (#1788): the shared verification primitive the others lean on. CI green, 21/21 test vectors, reference verifier `ocp-verify@2.0.0` on npm, deployed on Ethereum mainnet + Arbitrum One; initial-pass comments addressed.
- **ERC-8274 — AI Inference Proof Verification Interfaces** (#1771): the verification layer. CI green, editor-ready.
- **ERC-8299 — WYRIWE (Input Provenance)** (#1810): input-provenance layer; composes onto ERC-8004 (identity) and ERC-8274 (verification). All 9 CI checks green, open.
- **ERC-8301 — AI Agent Execution** (#1815): execution-attestation layer. CI green, editor-ready.
- **ERC-8323 — Source-Token Agent Binding** (#1851): source-binding for ERC-8004 agents. CI 9/9 green; one editor review outstanding.

Endorsements across the set note clean composition (8263 / 8274 / 8004 authors on OCP as the shared primitive). We're happy to make any changes — just surfacing the cluster in case individual PRs fell off the queue. Thank you for your time.

babyblueviper1:
Confirming as co-author: ERC-8323 is CI-green (9/9 checks) with one editor review outstanding. Ready as part of this cluster — happy to make any changes the editors want.

PR #1859, opened 2026-07-06
Website: Fix internal links

jwahdatehagh:
# Fix internal links

**Description:** Rewrite proposal, asset, and code-comment references so they resolve after the EIP/ERC repository split.

## Abstract

This repository stores proposals as `ERCS/erc-N.md`, but the documents still cross-reference each other with `./eip-N.md` style links inherited from ethereum/EIPs. Those targets do not exist here, so every internal proposal link (and every `../assets/eip-N` asset link) is broken on ercs.ethereum.org. This PR rewrites all internal references to their correct targets and adjusts CI so both the standalone site and the merged eips.ethereum.org build keep resolving.

## Motivation

After the EIP/ERC repository split, files in this repository were renamed to `erc-N.md` and asset directories to `assets/erc-N`, but link targets were never updated. As a result:

- ~2,800 relative proposal links and 525 asset links 404 on ercs.ethereum.org.
- The markdown-link-check CI job fails for any PR that touches a file containing such a link, because the referenced files do not exist in this repository.

internal-links-404

## Specification

Links are rewritten according to these rules:

1. **Proposal lives in this repository** → relative link to the actual file: `[ERC-721](./erc-721.md)` (~2,600 links, anchors preserved).
2. **Proposal lives in ethereum/EIPs** → canonical URL: `[EIP-712](https://eips.ethereum.org/EIPS/eip-712)` (~250 links). There is no file in this repository to link relatively.
3. **References inside code blocks** (Solidity natspec and similar) → live absolute URLs — `https://ercs.ethereum.org/ERCS/erc-N` for ERCs, `https://eips.ethereum.org/EIPS/eip-N` for core EIPs (106 references). Repo-relative paths are meaningless to readers of deployed contract code.
4. **Asset links** → `../assets/eip-N` becomes `../assets/erc-N`, matching the local directory names (509 links). `assets/eip-1` keeps its name, and `erc-7730`'s reference to `assets/eip-712` stays: those assets live in ethereum/EIPs.
5. `_includes/eipnums.html` generates matching hrefs: `./erc-N` for local ERCs, `./eip-N` for local non-ERC files (EIP-1), canonical eips.ethereum.org URL otherwise.
6. CI companions:
   - The HTMLProofer job's *Merge Repos* step renames `erc-*` files to `eip-*`; it now also rewrites the link targets inside the merged files the same way, so the merged-site build keeps resolving.
   - `config/eipw.toml` allows `https://eips.ethereum.org/EIPS/eip-N` in the `markdown-rel-links` lint (rule 2 is otherwise impossible).

## Rationale

- Relative links are kept wherever a local file exists, matching EIP-1 style and keeping eipw's relative-link lint meaningful.
- Canonical URLs are used only where no local file can exist (cross-repository references), mirroring how `eipnums.html` already resolves unknown proposals.
- Bare-text citations such as `- ERC-20 Token Standard. ./erc-20.md` in References sections were updated to the real file names; they render as plain text and are not checked by any tooling.

## Backwards Compatibility

eips.ethereum.org renders these documents through a merge that renames `erc-*` files and asset directories to `eip-*`. Without a rewrite of link targets, this PR would break every converted link there. The *Merge Repos* step in this repository's CI is updated accordingly; **ethereum/EIPs applies the same merge in `.github/actions/merge-repos` and needs the equivalent change before this lands**, or the production site will show dangling links.

## Test Cases

Verified programmatically over the full diff (523 files):

- Every relative `./erc-N.md` target exists in `ERCS/`; every canonical URL points at a proposal that exists in ethereum/EIPs and not here (checked against the upstream file list). Zero misclassifications.
- A normalized diff (links reduced to tokens) confirms no text changed other than link targets.
- A simulation of the merged-site rename + rewrite leaves zero unresolved links or asset references.
- Sampled `https://ercs.ethereum.org/ERCS/erc-N` and asset URLs return HTTP 200.

## Security Considerations

No specification semantics change. Note for reviewers: URLs inside code blocks of **Final** proposals (e.g. ERC-721, ERC-1155, ERC-6093, ERC-1820) are edited by rule 3 — the interface text is otherwise untouched.

## Copyright

Copyright and related rights waived via CC0.


frangio:
Confusingly the links are like this by design.

https://ercs.ethereum.org is broken and should probably be taken down to avoid confusion.

The correct site to visualize all ERCs is https://eips.ethereum.org.

3esmit:
Approved.

Pandapip1:
> Confusingly the links are like this by design.

This is because the website split was a bit of a rush job. If this genuinely fixes it (I haven't tested it and haven't been involved in quite a while) and doesn't break the website, this is worth fixing.

devender-startengine (approved):
Reviewed the erc-1450.md changes — link-target-only rewrites, no normative text touched, all targets resolve. LGTM for ERC-1450.

jwahdatehagh:
> Confusingly the links are like this by design.

imho if the website should be taken down, take it down. but i thought there is a clear delineation between what constitutes an ERC vs an EIP, so a more focused site to browse application level standards is helpful / good i think.

another oddity about the current setup is having the "EIPS / ERCS" namespace in both the subdomain and the path: (e.g. https://eips.ethereum.org/EIPS/eip-4361).

that looks like a smell tbh. and points to a unified setup where both EIPs and ERCs live on the same domain, at distinct sub paths.

PR #1862, opened 2026-07-07
Fix RSS feed self links and non-ERC feed filtering

weifanglab:
## Summary

Fix several RSS feed template copy-paste issues.

The `atom:link rel="self"` values for a few RSS templates pointed to another feed instead of the feed being rendered. This updates those self links so each RSS template points to its own `/rss/...` path.

This also fixes `rss/nonerc.xml`, which described itself as the feed for all non-ERC EIPs but was still using the Last Call-only filter and Last Call item description. The template now mirrors the regular update-feed behavior while keeping the non-ERC category filter.

## Testing

- Ran a local RSS template consistency check to verify each `atom:link rel="self"` matches its feed path.

### DIFF
--- a/rss/all.xml
+++ b/rss/all.xml
@@ -7,7 +7,7 @@ layout: null
     <title>Ethereum EIPs</title>
     <description>A feed of all EIPs</description>
     <link>{{ site.url }}</link>
-    <atom:link href="{{ site.url }}/all.xml" rel="self" type="application/rss+xml" />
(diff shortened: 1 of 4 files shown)


weifanglab:
@g11tech, @jochem-brouwer, @lightclient, @samwilsn, @xinbenlv  Hi, Could you please review this PR at your convenience? Thank you very much.

PR #1867, opened 2026-07-08
Update ERC-8196: Move to Final

Cybercentry:
Changed status to Final.


jochem-brouwer:
This ERC refers to ERC-8004 which is not Final, we have to move this to Final first. EIP-8126 is Final and requires EIP-8004, this should not have been moved into Final and should have been flagged by the bot.

Cybercentry:
Many thanks, @jochem-brouwer. Can ERC-8126 be amended? On ERC-3009, this could be removed. ERC-8004, we'll link in with the team.

SamWilsn (approved):

PR #1877, opened 2026-07-14
Update ERC-7730: Escape literal Liquid delimiters in ERC-7730

minksy-ux:
## Summary
- escape literal `{{` and `}}` examples in ERC-7730 with Jekyll raw tags
- wrap the JSON example that contains `{{function}}` so Liquid does not parse it during site builds

## Why
ERC-7730 is merged into the EIPs site during CI via the merge-repos action. The literal Liquid delimiter examples in the markdown trigger Jekyll/Liquid parsing warnings in the combined build.

## Validation
- reviewed the rendered markdown diff locally
- confirmed the fix only touches the literal-brace examples


forshtat (approved):
Let's see...

PR #1883, opened 2026-07-15
Add ERC: Transaction Metadata Encoding

chunter-cb:
## Abstract

[EIP-8130](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-8130.md) adds an optional, opaque `metadata` field to its account-abstraction transaction type but leaves the byte layout to a companion specification. This ERC defines that layout.

`metadata` is a single deterministic [CBOR](https://www.rfc-editor.org/rfc/rfc8949) value: a text string (memo), a byte string (commitment digest), a map of reserved keys (attribution, memo, commitment, scope), or an array of these. The map keys are interoperable with ERC-8021 schema 2, extended with keys for offchain commitments and call scoping. Any value MAY be replaced by a salted commitment, recursively, giving selective disclosure with a single primitive; the ERC also specifies the offchain proof package, verification, and delivery protocol, plus an [ERC-5792](https://github.com/ethereum/ERCs/blob/master/ERCS/erc-5792.md) `metadata` capability that supersedes `dataSuffix`.

The encoding is a transport-agnostic standard: EIP-8130's `metadata` field is the target carrier and the source of the scope model, but the same value MAY be carried by another transport (e.g. a calldata data suffix on other transaction types).

## Motivation

EIP-8130's `metadata` field is opaque bytes; without a shared structure every producer invents its own framing and no indexer can read across them. This ERC provides a compact, deterministic, ERC-8021-compatible encoding that covers attribution, memos, per-call scope, and privacy-preserving commitments to offchain data.

## Notes

- New draft ERC; a single file is added under `ERCS/`.
- Depends on ERC-5792 and EIP-8130 (`requires: 5792, 8130`).
- Discussion: https://ethereum-magicians.org/t/erc-transaction-metadata-encoding/29022

The file uses the `erc-draft_` placeholder name pending an editor-assigned number; I will rename it to `erc-<number>.md` and add the `eip:` header once assigned.


ERCS/erc-8340.md
```diff
@@ -0,0 +1,465 @@
+---
```
abcoathup, line 1:
```suggestion
---
eip: 8340
```

ERCS/erc-draft_transaction_metadata_encoding.md
```diff
@@ -0,0 +1,465 @@
+---
+title: Transaction Metadata Encoding
+description: A CBOR encoding for transaction metadata, targeting the EIP-8130 metadata field, for attribution, memos, and offchain commitments.
+author: Chris Hunter (@chunter-cb) <chris.hunter@coinbase.com>
+discussions-to: https://ethereum-magicians.org/t/erc-transaction-metadata-encoding/29022
```
abcoathup:
```suggestion
discussions-to: https://ethereum-magicians.org/t/erc-8340-transaction-metadata-encoding/29022
```

PR #1898, opened 2026-07-20
Update ERC-7730: make owner field optional

melanciani:


PrisionMike:
I think the  `owner` key was deliberately kept as required since it is pivotal to the objective of Clear signing.

The user should know if they are interacting with a contract whose owner is known. More specifically,how the auditor has approved the contract owner to be identified as.

kuzdogan:
@PrisionMike I tend to disagree. The objective of clear signing is to decode tx's into human readable messages. It's not about the provenance, whitelisting, contract labeling etc. It's not a one-stop-shop for secure contract interactions.

There are infinite number of things we can do for "security". The objective should not extend beyond proper decoding IMO.

kuzdogan (approved):

kuzdogan (approved):
Approving as discussed in the call  https://github.com/ethereum/clear-signing-erc7730-registry/issues/2655

PR #1901, opened 2026-07-21
Update ERC-7730: Fix editorial issues from the versioning PR

ankurdotb:
Small follow-up to #1838, cleaning up a few loose ends noted during review that landed anyway:

- The "Released versions" table claimed `1.0.0`/`2.0.0` were renumbered "with no change to their content or file names" — but #1838 itself added a `version` key to both files, so that claim was already false the moment it merged. Reworded to say what actually happened.
- Restored the "Commit ID" column, explicitly labeled as informational only (not a versioning mechanism — the `version` key inside each schema file remains authoritative). Corrected `2.0.0`'s entry: the previous value predated two later commits that changed `erc7730-v2.schema.json`'s content, so it no longer matched the file. Both entries now use full 40-char SHAs rather than short hashes, since short hashes carry a growing collision risk over the unbounded life of this document as the repo accumulates more commits.
- Removed a bare `MUST` from the Backwards Compatibility section — RFC 2119 keywords are reserved for the Specification section per this repo's process doc, and this one was a redundant restatement of a rule already stated normatively in Versioning.

Note: an earlier version of this PR also changed the relative links from `assets/eip-7730/` to `assets/erc-7730/`, on the theory that the former was a stale/broken path. That was wrong and has been reverted — this repo's CI renames every `erc-*` file and asset folder to `eip-*` at build time (see the "Merge Repos" step in `ci.yml`), so `assets/eip-7730/` is the path that actually resolves in the published site. Thanks to @forshtat for catching it on the previous version of this PR.

No content/schema changes, purely editorial.


forshtat:
Regarding relative links, I think there is some conflict in HTMLProofer that requires using "incorrect" names but the links end up working when published on `eips.ethereum.org`, or am I missing something?

ankurdotb:
> Regarding relative links, I think there is some conflict in HTMLProofer that requires using "incorrect" names but the links end up working when published on `eips.ethereum.org`, or am I missing something?

Good catch, there's a CI I missed that renames the file paths, reverted that

ERCS/erc-7730.md
```diff
@@ -175,17 +175,19 @@ The version of the specification used by a descriptor file is the version identi
...
 | `3.0.0-next` | [`erc7730-v3.0.0-next.schema.json`](../assets/eip-7730/erc7730-v3.0.0-next.schema.json) | N/A | Draft |
 
-`1.0.0` and `2.0.0` are the two schema versions that were released prior to the introduction of this versioning scheme; they are renumbered here under semver with no change to their content or file names. Every version released from now on is added as a new row in this table, without altering the rows already present, and follows the `erc7730-v<MAJOR>.<MINOR>.<PATCH>[-rc.<N>].schema.json` file naming pattern described above.
+`1.0.0` and `2.0.0` are the two schema versions that were released prior to the introduction of this versioning scheme; they are renumbered here under semver, with the addition of the top-level `version` key to each file being the only change made to bring them in line with this section. Every version released from now on is added as a new row in this table, without altering the rows already present, and follows the `erc7730-v<MAJOR>.<MINOR>.<PATCH>[-rc.<N>].schema.json` file naming pattern described above.
```
kuzdogan: Is it necessary to add the information about the `version` field?

ankurdotb: I've trimmed it, please take a look if this is better?

kuzdogan (approved):

PR #1905, opened 2026-07-22
Update ERC-8056: fix broken image asset paths (eip-8056 -> erc-8056)

jumpboxtech:
The two diagrams in the Rationale section render as broken images. The image paths point to `../assets/eip-8056/…` but the asset files are located at `assets/erc-8056/` — both `token_value_repr.jpeg` and `token_arch_layers.jpeg` return 404 at the referenced path and 200 at `assets/erc-8056/`. This corrects the two paths.

Editorial only — no normative changes.

### DIFF
--- a/ERCS/erc-8056.md
+++ b/ERCS/erc-8056.md
@@ -248,9 +248,9 @@ Alternative Approaches Considered:
 
 4. Off-chain Solutions: Purely off-chain solutions lack standardization and require trust in centralized providers.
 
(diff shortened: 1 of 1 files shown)


jumpboxtech:
Closing this — I was mistaken. The `eip-8056` asset path is correct: the rendered site (eips.ethereum.org) serves assets under the `eip-` namespace, matching every other ERC (e.g. erc-7208 references `assets/eip-7208/`), and both images resolve there (200). My check looked at the raw source directory rather than the rendered convention. Apologies for the noise.

PR #1906, opened 2026-07-23
Add ERC: Translation Files for ERC-7730 Descriptors

forshtat:
## Summary

Companion specification to [ERC-7730](https://github.com/ethereum/ERCs/blob/master/ERCS/erc-7730.md), defining the translation file format referenced by its `$i18n` field:

- Translation file schema (superseding the draft schema that briefly lived under ERC-7730's own assets during v3 development).
- Dot-namespaced, snake_case translation keys, replacing raw-English-string keys.
- A reserved `common.` namespace for a future shared vocabulary package, with resolution precedence.
- Subresource-Integrity-style hashes on translation resource references.
- Required-languages list, BCP-47 lookup/fallback semantics, and placeholder-preservation rules.

## Test plan

- [ ] `erc7730-i18n-v1.schema.json` validated as well-formed Draft 2020-12 JSON Schema
- [ ] `example-main.fr.json` validated against the schema


ERCS/erc-8346.md
```diff
@@ -0,0 +1,214 @@
+---
```
abcoathup, line 1:
```suggestion
---
eip: 8346
```

ERCS/erc-8346.md
```diff
@@ -0,0 +1,238 @@
...
+
+A translation file is a JSON document with the following fields:
+
+| Field            | Required  | Description                                                        |
+|------------------|-----------|--------------------------------------------------------------------|
+| `$schema`        | Yes       | URI of the translation file schema.                                |
+| `$locale`        | Yes       | Canonical BCP-47 language tag.                                     |
+| `descriptorHash` | See below | Hash binding this file to the descriptor it translates.            |
+| `includes`       | No        | Map from namespace to candidate references to a shared package.    |
+| `translations`   | Yes       | Flat map from translation key to translated value.                 |
```
PrisionMike, line 40: Let's specify UTF-8 normalization as well. [NFC](https://unicode.org/reports/tr15/) maybe?

ERCS/erc-8346.md
```diff
@@ -0,0 +1,238 @@
...
+    "$schema": "./erc8346-v3.0.0-next.schema.json",
+    "$locale": "fr",
+    "descriptorHash": "0x14da251e322186245d812ff3fd0b1e9b404c5896964f9dbce2bd2bd898abbe84",
+    "includes": {
+        "erc20": [
+            { "uri": "./example-erc20.fr.json" }
+        ]
+    },
+    "translations": {
+        "mytoken.transfer.interpolated_intent": "Envoyer {value} à {to}",
```
PrisionMike, line 53: For a generic contract, the indexing now requires names of contract and function names, instead of cannonical addresses and function signatures maybe.
i.e. `<contract-address>.<func-sig>.interpolated_intent`

And given that the descriptor hash, maybe `<func-sig>.intent` etc. would be less redundant, and more precise.

ERCS/erc-8346.md
```diff
@@ -0,0 +1,238 @@
...
+    "staking.withdraw.intent": {
+        "Native Staking": { "label": "Staking natif", "value": "Retirer" },
+        "Rewards":        { "label": "Récompenses", "value": "Consensus et exécution" }
+    }
+}
+```
+
+The translation object's key set MUST exactly equal the key set of the English intent object.
+
+*Enumerations.* The display values of an ERC-7730 `metadata.enums` entry are translated as an object mapping each raw enumeration value to its translated display string; the key set MUST exactly equal the key set of the English enumeration.
```
PrisionMike, line 130: Why can't we translate the enumerations?

ERCS/erc-8346.md
```diff
@@ -0,0 +1,238 @@
...
+* It carries no `descriptorHash`.
+* It MUST NOT contain an `includes` field of its own: includes resolve one level deep.
+* It is subject to the same [integrity requirements](#translation-resource-integrity) as any translation resource.
+
+A key `k` with namespace `ns` resolves as follows:
+
+1. If the including file's own `translations` map defines `k`, use that value. This lets a file override a shared term when its descriptor's context requires different wording.
+2. Otherwise, if `includes` has an entry for `ns`, look up `k` in the resolved shared package's `translations` map. Only entries whose namespace is `ns` are considered.
+
+There is no further fallback: a key that resolves through neither step renders the whole resource invalid, per [Validity](#validity).
```
PrisionMike, line 99: Consider falling back to English in want of a translation. It is less likely to happen but a standard safe fallback.

Say someone updated the contract descriptor to add a new term. Or change some labels. Instead of no localization, the user will only see that one new label unlocalized that can further be revised.

Not all translators would be able to catch up with all updates at all times.

PR #1907, opened 2026-07-23
Add ERC: Financial Lease

javierpmateos:
## New ERC: Financial Lease Standard

Jurisdiction-neutral interface for tokenized financial lease agreements.

- **Discussion:** https://ethereum-magicians.org/t/erc-financial-lease-standard-jurisdiction-neutral-interface-for-tokenized-lease-agreements/29076
- Lessor position as ERC-721 (`tokenId == leaseId`); assignment and securitization via standard NFT transfers
- Schedules denominated in units of account with explicit conversion, covering both fixed and inflation-indexed leases (UVA/UF/IPCA)
- Two-tier delinquency (objective `InArrears` / formal `InDefault`)
- Optional `IFinancialLeaseAssetBound` extension for on-chain leased assets with atomic purchase-option settlement
- Reference implementation in `assets/` (Foundry, 8 passing tests including fuzz coverage of rounding invariants)

Semantics follow the UNIDROIT Convention on International Financial Leasing (1988) and IFRS 16 terminology. Composes with ERC-3643/7943 (compliance), ERC-4907 (usage rights), and the titled-asset family (ERC-8325..8330).


ERCS/erc-1907.md
```diff
@@ -0,0 +1,168 @@
+---
+eip: 1907
```
abcoathup:
```suggestion
eip: 8348
```

ERCS/erc-1907.md
```diff
@@ -0,0 +1,168 @@
...
+title: Financial Lease Interface
```
abcoathup:
```suggestion
title: Financial Lease
```
Interface is superfluous in an ERC title

ERCS/erc-1907.md
```diff
@@ -0,0 +1,168 @@
...
+description: Jurisdiction-neutral interface for tokenized financial leases with indexed-denomination schedules.
+author: Javier Mateos (@javierpmateos)
+discussions-to: https://ethereum-magicians.org/t/erc-financial-lease-standard-jurisdiction-neutral-interface-for-tokenized-lease-agreements/29076
```
abcoathup:
```suggestion
discussions-to: https://ethereum-magicians.org/t/erc-8348-financial-lease/29076
```
Updated Eth Magicians title with assigned number & ERC title

ERCS/erc-1907.md
javierpmateos: Done ... number assigned across preamble, filename and assets folder. Thanks!

ERCS/erc-1907.md
javierpmateos: Done ... Magicians thread renamed to match and preamble updated. Thanks!

ERCS/erc-1907.md
javierpmateos: Done, thanks!

javierpmateos:
Added two optional extensions based on discussion in the Ethereum Magicians thread (composition with the ERC-8325..8330 titled-asset family):

- `IFinancialLeaseAnchored` ... explicit `(chainId, registry, anchorId)` binding to an ERC-8325 registry, as an optional extension. Core interface unchanged.
- Input-freshness reporting (`inputFreshness` / `ServicingInput`) — per-input staleness as observable data, not enforced. Isolated from the payment path.

Reference implementation updated accordingly: 14 passing tests, including an isolation test asserting servicing updates don't affect lease state. Spec and implementation are in the same commits.

javierpmateos:
Second round of updates from thread feedback:

- Published explicit ERC-165 interface identifiers for the core interface and both optional extensions (a reviewer noted the spec required ERC-165 support without naming the values).
- Extracted `IFinancialLease` and `IFinancialLeaseAssetBound` into standalone interface files — previously the core interface only existed inside the implementation contract.
- Added a normative rule for conversion rate scaling: base units of the payment asset per 1e18 units of account.
- Reference implementation now includes three oracle adapters (Chainlink AggregatorV3, composed two-feed, and attested for indices without on-chain feeds), with tests.

25 passing tests. Spec and implementation remain in the same commits.

javierpmateos:
Third round of updates. Ran a security audit against the reference implementation; two exploitable findings, both fixed, and in both cases the root cause was the specification permitting the wrong implementation.

**Frequency-dependent penalty accrual.** Penalties were accrued by re-applying a periodic rate to a balance that already included accrued penalty, so the total depended on how often accrual state was persisted (+4.58% between one accrual at 90 days and ninety daily accruals). Since `pay()` is permissionless, the lessor could push accrual toward the higher figure at no net cost. The spec now requires reported arrears to be a function of the schedule, payment history and elapsed time, without prescribing simple versus compound interest.

**Lessee reassignment could take the purchase option.** Reassignment accepted the default-declaration role as caller, and that role defaults to the originator. The spec now requires authorization from the outgoing lessee or a separately granted authority, and prohibits conflating it with roles held for other purposes.

**`Terminated` is now reachable.** It was declared in `LeaseStatus` but no code path assigned it, and a lease in `InDefault` had no terminal exit. Two causes in core (mutual agreement, default after a configurable non-zero delay), with domain-separated `bytes32` reasons for extension. Post-termination settlement is explicitly out of scope.

The Specification section also now includes a state transition table and an operation-by-status matrix, replacing behaviour that was previously left to inference.

74 passing tests, including regression coverage for both findings. Spec and reference implementation are in the same commits.

javierpmateos:
Fourth round of updates, from further review of the reference implementation.

- **Penalty crystallization underflow**: fractioning a payment within one block leaked up to one unit per fraction, eventually causing `pay()` and `arrears()` to revert. Beyond the panic, the leak violated the normative requirement that arrears not depend on transaction count or timing. Fixed by crystallizing the difference of floored grosses; regression test plus fuzz coverage over fraction count.
- **Test robustness**: two tests, including the frequency-invariance one, failed under the optimizer due to `block.timestamp` locals being rematerialized across `vm.warp`. Fixed, and optimizer settings are now committed so the suite is verified under a deployable configuration.
- **Reproducibility**: `foundry.toml` and `remappings.txt` were missing from `assets/erc-8348/`, so the README instructions didn't build. Both added and verified from a clean checkout.

76 passing tests.

PR #1908, opened 2026-07-24
Update ERC-7730: Improve duration format

melanciani:
## Abstract

This change amends the ERC-7730 `duration` field format. A `duration` value (a number
of seconds) is rendered as `<days>d<hh>h<mm>m<ss>s`, where `days` is unbounded and
`hh`/`mm`/`ss` are zero-padded to two digits. This replaces the previous `HH:MM:ss`
rendering, whose unbounded hours field (e.g. `240:00:00` for ten days) is ambiguous and
easily misread on a clear-signing display.

## Motivation

Durations in clear-signing are frequently authorization or validity windows (permit and
delegation expiries, operator authorizations). The prior `HH:MM:ss` format collapsed
multi-day windows into large hour counts (`240:00:00`), which users can misread — a
meaningful risk when the value governs how long access is granted. A fixed, days-inclusive
format makes the magnitude legible at a glance without adding optional styles or parameters
that would fragment wallet behavior.

## Rationale

Alternatives considered:

- **Keep `HH:MM:ss`** — rejected: ambiguous for multi-day durations (`240:00:00`).
- **ISO-8601 durations (`P10D`, `PT2H17M30S`)** — the formal standard, but rejected: cryptic
  for end users on a hardware-wallet screen.
- **Opt-in `style`/`base` parameters** — rejected: adds configuration and lets the same value
  render differently across wallets. A single imposed format is simpler and consistent.

The chosen `Dd HHh MMm SSs` form is days-inclusive, self-describing, and unambiguous, and keeps
`duration` parameter-free.

## Backwards Compatibility

This is a breaking change to the rendered output of `duration` (previously `HH:MM:ss`), so it
affects renderers, not descriptor documents (no schema-structure change; documents that
validated before still validate). It is introduced in the v3.0.0 schema line — the current
mutable `-next` major draft — and does NOT alter the released v2 schema. Wallets and libraries
implementing ERC-7730 MUST update their duration renderer to the new format when adopting v3.

## Test Cases

| Field value (seconds) | Rendered      |
|-----------------------|---------------|
| `0`                   | `0d00h00m00s` |
| `45`                  | `0d00h00m45s` |
| `3600`                | `0d01h00m00s` |
| `8250`                | `0d02h17m30s` |
| `360000`              | `4d04h00m00s` |
| `864000`              | `10d00h00m00s`|

## Reference Implementation

```ts
function formatDuration(totalSeconds: bigint): string {
  const t = totalSeconds < 0n ? -totalSeconds : totalSeconds;
  const days = t / 86400n;
  const hh = ((t % 86400n) / 3600n).toString().padStart(2, "0");
  const mm = ((t % 3600n) / 60n).toString().padStart(2, "0");
  const ss = (t % 60n).toString().padStart(2, "0");
  return `${days}d${hh}h${mm}m${ss}s`;
}
```

## Security Considerations

`duration` commonly renders a validity or authorization window (e.g. a decryption-delegation
or operator expiry). A misread duration could lead a user to authorize longer-lived access
than intended. The days-inclusive format directly reduces that misread risk versus large
unbounded hour counts. The change is rendering-only and introduces no new attack surface.

### DIFF
--- a/ERCS/erc-7730.md
+++ b/ERCS/erc-7730.md
@@ -1386,21 +1386,42 @@ Formats applicable to `uint`/`int` solidity types.
 |-----------------|------------------------------------------------------------------------------------------|
 | *Description*   | Display an int (a number of seconds) as a time duration, rendered as `<days>d<hh>h<mm>m<ss>s` — `days` is unbounded and `hh`/`mm`/`ss` are zero-padded to two digits. This days-inclusive form avoids ambiguous large hour counts (e.g. `240:00:00`). |
 | *Parameters*    | ---                                                                                      |
-| `unboundedMessage` | Optional. Message shown when the value equals `type(uint256).max` (2²⁵⁶−1), a sentinel for an unbounded duration. Defaults to `Forever`. |
+| `specialValues` | Optional. Maps sentinel values to a message displayed in place of the formatted duration, e.g. `{"0": "Immediate"}`. Keys are written in decimal or `0x`-prefixed hexadecimal and compared numerically. `type(uint256).max` (2²⁵⁶−1) maps to `Forever` unless an explicit entry overrides it. |
 | `approximation` | Optional boolean, defaults to `false`. When `true`, an approximate calendar duration is appended in parentheses after the exact duration. Only relevant for values of at least one month; below that no suffix is appended. |
 | *Examples*      | ---                                                                                      |
 | `10d00h00m00s`  | Field Value = 864000                                                                     |
 | `0d02h17m30s`   | Field Value = 8250                                                                       |
-| `Forever`       | Field Value = `type(uint256).max`                                                        |
+| `0d00h00m00s`   | Field Value = 0 <br> no `specialValues` entry for `0`                                    |
+| `Forever`       | Field Value = `type(uint256).max` <br> no `specialValues` (implied entry)                 |
+| `Immediate`     | Field Value = 0 <br> `specialValues` = `{"0": "Immediate"}` (a timelock delay)            |
+| `No expiry`     | Field Value = 0 <br> `specialValues` = `{"0": "No expiry"}` (an expiry date)               |
+| `Max`           | Field Value = `type(uint64).max` <br> `specialValues` = `{"0xffffffffffffffff": "Max"}`   |
 | `1856d00h00m00s (approx. 5 years and 1 month)` | Field Value = 160358400 <br> `approximation` = true      |
 | `210d00h00m00s (approx. 7 months)` | Field Value = 18144000 <br> `approximation` = true                  |
 | `10d00h00m00s`  | Field Value = 864000 <br> `approximation` = true (below one month, no suffix appended)    |
 
+The `specialValues` map is resolved before the value is formatted:
+
+1. Keys are parsed as integers, decimal or `0x`-prefixed hexadecimal, and compared numerically, so
+   `"0"`, `"0x0"` and `"0x00"` all designate zero. A file MUST NOT contain two keys of equal
+   numeric value.
+2. An entry mapping `type(uint256).max` (2²⁵⁶−1) to `Forever` is implied unless explicitly
+   overridden.
+3. On a match, the message is displayed alone, without the exact duration and without an
+   approximation suffix; otherwise the value is formatted as a duration as described above.
+
+No default message is assigned to `0`, whose meaning is contract specific: a zero timelock delay
+reads as `Immediate`, a zero expiry usually the opposite. Authors SHOULD map `0` explicitly where
+the contract gives it a distinguished meaning; absent an entry it renders as `0d00h00m00s`. Because
+a message replaces the duration rather than annotating it, authors MUST only map values whose
+message is faithful to the contract's behaviour, and SHOULD map only the sentinels the contract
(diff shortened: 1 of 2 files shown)


forshtat:
Just a quick question - should an upper bound for `duration` be defined in the ERC? I mean, if `totalSeconds: bigint` is 2²⁵⁶, is `threshold` applicable for `duration` fields?

melanciani:
> Just a quick question - should an upper bound for duration be defined in the ERC? I mean, if totalSeconds: bigint is 2²⁵⁶, is threshold applicable for duration fields?

truncating duration seems dangerous for clear signing as it could hide genuine large values , no ? Still, we could include some kind of `unboundedMessage` parameter (with default `Forever`) for `type(uint256).max`, wdyt ?

forshtat:
> truncating duration seems dangerous for clear signing as it could hide genuine large values , no ? Still, we could include some kind of `unboundedMessage` parameter (with default `Forever`) for `type(uint256).max`, wdyt ?

Well, for me personally having a `Forever` value for durations (similar to existing `Unlimited` value for amounts) makes total sense.
Will be happy to hear if it was not added for a reason that I might be missing here.

melanciani:
I was not part of the initial specs regarding this so can't say, this PR is suggestion toward improving this ! I've pushed the unboundedMessage parameter change

kuzdogan:
While we are on it, shouldn't we also add "months" and "years" too? I assume there will be use cases to set couple years of deadlines etc.

Edit: I guess months will be tricky because of variable days of month (30 or 31). Same for years with 366 and 365 day years?

forshtat:
> Edit: I guess months will be tricky because of variable days of month (30 or 31). Same for years with 366 and 365 day years?

Yeah, the wallet will need to know if the duration field means "a countdown that starts right now" or not, and to know the current time (most wallets don't) to render it precisely (could even give you the end date then).
Should we allow rendering durations as "Approximately 5 years and 7 months"?

melanciani:
@kuzdogan @forshtat following your comments, I agree years/months would make sense but I think we should avoid displays approximation by default. Still, I feel it would make sense to introduce a new optional parameter that would enable a more readable view like "Approximately 5 years and 7 months" yes, I'll try to suggest something soon

melanciani:
hey again @kuzdogan @forshtat what do you think of my suggestion ? I'm mostly not sure about the need to specify the exact way we want to display approximation:

**Add `approximation` parameter to the `duration` format**

Optional `approximation` boolean (default `false`) on `durationParameters`. When enabled, an approximate calendar duration is appended after the exact one:

```
1856d00h00m00s (approx. 5 years and 1 month)
210d00h00m00s (approx. 7 months)
```

It's appended, never a replacement, so no precision is hidden.

The algorithm is fully specified so all wallets render identically: round to the nearest month (`2629800` s = 30.4375 d, so 12 months = 1 Julian year), split into years/months, omit zero components, singular/plural per value. Values below one month get no suffix, and `type(uint256).max` still renders `unboundedMessage` alone.

Only years and months — stopping there keeps the output a deterministic function of the value, with no clock, locale or calendar needed. Default `false`, so existing descriptors are unaffected

melanciani:
CI is failing on something unrelated: I've opened an issue to fix this: https://github.com/ethereum/ERCs/pull/1933

melanciani:
ci is green again  ! cc @forshtat @kuzdogan

ERCS/erc-7730.md
```diff
@@ -1382,12 +1382,38 @@ Formats applicable to `uint`/`int` solidity types.
...
+   appended.
+3. Otherwise, let `months` be `value / 2629800` rounded to the nearest integer, halves away from
+   zero, where `2629800` is the number of seconds in one month (30.4375 days). Twelve such months
+   are exactly one Julian year (365.25 days).
+4. Let `years` be `months / 12` (integer division) and `remainingMonths` be `months % 12`.
+5. The suffix is `" (approx. "`, then the non-zero values among `years` and `remainingMonths` — in
+   that order, joined by `" and "`, each rendered as the integer followed by `" year"` / `" years"`
+   or `" month"` / `" months"`, singular when the value is exactly `1` — then `")"`. A component
+   equal to zero MUST be omitted.
+
```
forshtat: Well, if we are actually doing this "approximate duration" thing, we can simplify the rules a little, something like this

```suggestion
3. Otherwise, define the following:
- `months := value / 2629800`
- `approxYears := months / 12`
- `approxMonths := months % 12`
- `approxTemplate := "(approximately {approxYears} years and {approxMonths} months)"`

A consumer SHOULD use `approxTemplate` when rendering the signature request in English. It MAY substitute a proper localized alternative conveying the same `approxYears`/`approxMonths`.

```

melanciani: that's fair yes

melanciani:
while working again on https://github.com/ethereum/clear-signing-erc7730-registry/pull/2595, I suggest a new improvement: instead of just adding `Forever`, we generalize to special values (not threshold), to also support `0` values (without enforcing any labels)

For example: we have a setter that gives right for a period of time, and revoke path uses `0` as the time input: in this case, we'd prefer not to display the 01/01/1970 date and instead have a custom label like `Immediately`

PR #1919, opened 2026-07-30
Add ERC: Confidential Agent Policy Verdicts

mzf11125:
A pre-execution allow/deny interface for agent actions, proven in zero knowledge against a policy that is never disclosed on-chain.

This ERC bridges ERC-8004 (agent identity) and ERC-7812 (evidence registry) with a verdict envelope that a guard contract verifies locally, gating execution without revealing the rules that permitted it.

## Summary of changes
- Defines the `Verdict` envelope and `IConfidentialPolicyVerdict` core interface
- Specifies the canonical `actionCommitment` that binds a verdict to a specific action
- Provides a companion `IPolicyDomainRegistry` for domain and root management
- Includes concrete code examples from the [reference implementation](https://github.com/zexoverz/confidential-agent-policy-verdicts)
- Composes with ERC-8004 for identity and ERC-8312 for mandate metering


ERCS/erc-1919.md
```diff
@@ -0,0 +1,483 @@
+---
+eip: 1919
```
abcoathup:
```suggestion
eip: 8354
```
@mzf11125 

ERCS/erc-1919.md
```diff
@@ -0,0 +1,483 @@
...
+title: Confidential Agent Policy Verdicts
+description: A pre-execution allow/deny interface for agent actions, proven in zero knowledge against a policy never disclosed on-chain.
+author: Muhammad Zidan Fatonie (@mzf11125), Faisal Firdani (@zexoverz), Maulana Asykari Muhammad (@WeissCurry)
+discussions-to: https://ethereum-magicians.org/t/draft-idea-confidential-agent-policy-verdicts/29088
```
abcoathup:
```suggestion
discussions-to: https://ethereum-magicians.org/t/erc-8354-confidential-agent-policy-verdicts/29088
```

zexoverz:
Pushed a Security Considerations addition (`23e04ea`): a subsection distinguishing action-level integrity from interpreter-level fidelity.

It names a gap that bites implementers quietly. A verdict proves the committed interpreter ran over the action and returned ALLOW. It does not prove that interpreter faithfully implements the policy the domain intends. A deviant interpreter is faithful to itself, so its proofs verify perfectly while it judges wrong. The subsection states that boundary and places fidelity out of band, in a content-addressed append-only companion registry in the ERC-7812 pattern, read as a contestable assertion rather than a proof.

This came out of the discussion on Ethereum Magicians, where interpreter authors and attester authors converged on the same two-layer split from both sides:
https://ethereum-magicians.org/t/erc-8354-confidential-agent-policy-verdicts/29088

No interface or normative-core changes, Security Considerations only.

cc @abcoathup

zexoverz:
This is on the Meeting 110 agenda, so a quick state of the PR for the editors.

ERC-8354 (Confidential Agent Policy Verdicts) is a view, non-reverting interface for a pre-execution allow or deny verdict. It proves an action clears a policy that is never revealed on-chain, and burns a single-use nullifier so a verdict cannot be replayed. It is app-layer and changes nothing at consensus. interfaceId is 0x6c832e88.

CI is green (9/9). The reference implementation is linked in the spec and passing, a Noir circuit with an UltraHonk verifier. discussions-to points at the Magicians thread, which carries the full design history including the recent Security Considerations addition that separates action-level integrity from interpreter-level fidelity.

eip-review-bot currently has it at one more editor review needed. I would really appreciate that final review to move it along, and I will be on the call to answer anything on the spec or the proof system. Thanks for the time.

giskard09:
§Security Considerations → "Interpreter-level fidelity" names the gap precisely and even sketches the shape of the fix — a content-addressed, append-only registry for interpreter provenance, kept beside the standard for the same reason the Domain registry is a companion rather than core.

That shape (permissionless commit, append-only, content-addressed) is the same primitive category we run in production for a different provenance problem — worth flagging only because if other ERCs in this catalog converge on wanting the same companion mechanism, it might be worth designing once rather than per-ERC. Not proposing this standard adopt a specific implementation — just noting the convergence.

zexoverz:
@giskard09 this is the right read, and the convergence is real. On the Magicians thread Helmy deployed a ProvenanceRegistry as a companion, Baby Blue is building the same for an off-chain signing verifier, and the Domain registry here is a companion for the same reason. Three separate needs, one primitive shape. Permissionless, append-only, content-addressed provenance, read as a contestable assertion rather than a proof.

I kept that out of 8354's normative core on purpose, so it can point at a shared mechanism instead of freezing one. The interface proves action-level integrity and says nothing about the registry. If the catalog wants to design the companion once, 8354 should reference it rather than carry its own copy.

Glad to help push that. The design questions are already live on the Magicians thread, namespace isolation on register and what the content hash has to commit to for an off-chain interpreter, so that is a good place to converge before anyone writes a second registry. What are you running in production, and is it close enough to the append-only content-addressed shape to generalize?

giskard09:
What we run in production is `AnchorRegistry` — a single-function permissionless contract, `anchor(bytes32 ref)`, no owner or roles, emits `Anchored(ref indexed, anchoredBy indexed, timestamp)`. Same address via CREATE2 across Base/Arbitrum/Ink. It generalizes cleanly to this shape: append-only comes from the event log itself, content-addressed is enforced by convention (callers pass a content hash as `ref`, never an opaque ID), and permissionless means anyone can anchor without going through an operator. We use it for a different provenance problem (action/decision receipts, not interpreter fidelity), but the primitive is identical — happy to compare notes on the Magicians thread if useful.

zexoverz:
That AnchorRegistry shape is close to ideal for this. A single `anchor(bytes32 ref)` with no owner and no roles, append-only from the event log, content-addressed by convention, CREATE2 to the same address across chains. That is the minimal version of what the fidelity companion needs.

The one place the fidelity use case wants more is lineage. Helmy's ProvenanceRegistry carries parent DAGs and answers shareLineage on-chain, which a pure `anchor(ref)` cannot without the structure being readable on-chain. But that structure does not have to live in the contract. If `ref = keccak(record)` where the record is the interpreter's spec commit, implementation commit, parents, and review method, then anchor stays a one-function primitive and the schema is an off-chain convention anyone can recompute. You trade on-chain lineage queries for maximal minimalism, and resolve lineage by walking the content-addressed records off-chain.

So there are really two points on one spectrum. A pure anchor, yours, when you only need existence and ordering. A structured register, Helmy's, when you want on-chain lineage traversal. Same primitive category, they just move the schema across the on-chain and off-chain line.

Would like to take you up on the Magicians thread. Helmy and Baby Blue are both there building registries for adjacent problems, so that is the right room to see whether one companion can serve all of these instead of one per ERC.

giskard09:
Running AnchorRegistry in production on Base mainnet — single `anchor(bytes32 ref)`, no owner, no roles, append-only via event log, CREATE2 to the same address across chains. Your keccak(record) approach for lineage is exactly how we handle it too: schema lives off-chain, contract stays one function, anyone recomputes the record independently instead of trusting a stored DAG. Happy to compare notes on the Magicians thread if useful — curious how programKey would need to shift to cover a signing-server verifier like Baby Blue's, since that's the same "nothing on-chain to hash" gap we'd hit generalizing this.

zexoverz:
The shift is the same move you make with AnchorRegistry, applied to the verifier's identity instead of a record. programKey stops meaning keccak(deployed bytecode) and starts meaning a commitment to the pinned, publicly-verifiable verification logic. Bytecode is one instance of that, not the definition.

For a signing-server verifier there is no bytecode to hash, but there is still something recomputable that defines what it will accept: the verifying key it signs against, plus the policy version it is bound to. So programKey becomes keccak over (verifying key, policy commitment), where the policy commitment is itself a content hash and not a label, which is exactly the direction Baby Blue took policy_version, a hash of the published rubric and the conformance-suite commit rather than a string like invinoveritas.review.v7.

The invariant that has to hold is the one AnchorRegistry already enforces by convention. Every element of the preimage is independently recomputable from public data, never an opaque id. If any element is a bare label the gap reopens one level down. So the rule generalizes: programKey MUST commit to logic a third party can reconstruct and check, whether that logic is on-chain bytecode, a verifying key, or a versioned policy the server proves it evaluated. "Nothing on-chain to hash" is only a gap if you insist the thing hashed lives on-chain. Hash the off-chain artifacts that pin the behavior and a signing-server verifier is in scope with the same honesty as a bytecode one.

Would be good to take this to the Magicians thread. Baby Blue is there and it is his verifier, so we can nail the exact preimage for his case in the open.

giskard09:
Agreed, that's the right venue for this — pinning Baby Blue's exact preimage needs his verifier in front of us, not a hypothetical. Heading over to Magicians now.

SamWilsn:
Please keep discussion about the content of the proposal on Ethereum Magicians. We use PRs only for formatting/style-related talk.

ERCS/erc-8354.md
```diff
@@ -0,0 +1,487 @@
...
+
+### Why the registry is a companion, not the standard
+
+The reusable, chain-agnostic contribution is the verdict envelope and its verification. Domain and root management is deployment policy, and reasonable operators will differ (a single hub, a per-tenant registry, an existing access-control system). Binding the core to one registry would force those operators to fork the standard. So `IPolicyDomainRegistry` is specified as the default that makes independent implementations interoperate, while the core interface depends only on the three semantic hooks it actually calls.
+
+### Why ERC-7812 rather than a new registry
+
+[ERC-7812](./eip-7812.md) is deployed on Mainnet and Sepolia at a deterministic address, is in Review, and was designed as a singleton specifically so that only a single `bytes32` root needs to cross chains to prove registry state. A hub-and-spoke policy deployment needs exactly that and nothing more. It also states that it was written abstractly to let subsequent ERCs build specific use cases on top. This is one. Defining a parallel registry would fragment the trust anchor for no benefit.
```
SamWilsn: Mentioning things about the state of ERC-7812 (like that it is in Review) is likely to go out of date. There's nothing _wrong_ with that, just something to be aware of while you're working on this proposal.

ERCS/erc-8354.md
```diff
@@ -0,0 +1,487 @@
...
+2. **Replay** -- re-submitting a consumed verdict reverts `VerdictReplayed`.
+3. **Expiry** -- `block.timestamp >= v.expiry` reverts `VerdictExpired`.
+4. **Executor mismatch** -- submission by any address other than `v.executor` reverts `ExecutorMismatch`, even with an otherwise valid proof.
+5. **DENY not consumable** -- a verdict with `decision == 0` reverts `VerdictDenied`.
+6. **Action binding** -- a verdict whose `actionCommitment` does not match the action the Guard is about to execute reverts (Guard-level), and is not consumable against a different `(target, value, callData, actionNonce)`.
+7. **Cross-chain replay** -- a verdict minted for `chainid` A is not consumable on chain B (the `actionCommitment` binds `block.chainid`).
+8. **Stale-root grace** -- a verdict against a root superseded less than `maxRootAge` ago succeeds. One older than `maxRootAge` reverts `PolicyRootRejected`.
+9. **Revocation** -- after `DomainRevoked`, every root reverts `PolicyRootRejected` immediately, with no grace window.
+10. **Malformed proof** -- `verify` returns `false` (does not revert) for malformed `proof` bytes.
```
SamWilsn: Test cases should be written either as simple input/expected output pairs or as unit tests in code. You can't just list the shapes of tests you want.

ERCS/erc-8354.md
```diff
@@ -0,0 +1,487 @@
...
+
+## Reference Implementation
+
+A reference implementation is maintained in the repository at
+`@zexoverz/confidential-agent-policy-verdicts` (CC0).
```
SamWilsn: This is an external link. If you'd like to include a reference implementation, it'll have to go in your assets directory.

ERCS/erc-8354.md
```diff
@@ -0,0 +1,487 @@
...
+- `ConfidentialPolicyVerdict` Guard (`verify` / `consume` / `isConsumed`) with the checks in the order specified, and a `PolicyDomainRegistry` companion (root rotation with `maxRootAge` grace, immediate revocation).
+- A proving-system-agnostic `IVerifier` boundary (`verifyProof(programKey, publicInputs, proof)`) so `SP1` / Groth16 / `RISC0` slot in behind one interface.
+- `PolicyAction` -- the canonical, domain-separated (`chainId` + `domainId`) action-commitment preimage, hashed identically on-chain and in-circuit.
+- Cryptographic executor binding: direct submission, or an [EIP-712](./eip-712.md) signed relay verified via `SignatureChecker` (ECDSA and [ERC-1271](./eip-1271.md)).
+- [ERC-165](./eip-165.md) support. `IConfidentialPolicyVerdict` `interfaceId` is `0x6c832e88`.
+
+Still in progress:
+
+- The `SP1` interpreter program and its Groth16 wrapper for on-chain verification (skeleton present, documenting the in-circuit obligations).
+- A worked cross-chain deployment with the root mirrored to spoke chains.
```
SamWilsn: Proposals should be written as if already Final. In this case, I'd put this note into an HTML-style comment:


```suggestion
<!-- Still in progress:

- The `SP1` interpreter program and its Groth16 wrapper for on-chain verification (skeleton present, documenting the in-circuit obligations).
- A worked cross-chain deployment with the root mirrored to spoke chains.
- 

-->
```

ERCS/erc-8354.md
```diff
@@ -0,0 +1,487 @@
...
+
+### Root staleness is a revocation delay
+
+`maxRootAge` exists because a spoke chain's mirrored root lags the hub. Setting it to zero makes verdicts unusable during any sync gap. Setting it high means a rule removed from the policy keeps authorizing actions for `maxRootAge` seconds on every spoke. Domains MUST treat `maxRootAge` as their revocation service-level agreement and MUST document it. Domains handling adversarial agents SHOULD keep it under one block time on the slowest spoke, and accept the liveness cost. `DomainRevoked` is the escape hatch and bypasses the window entirely. It is coarse by design, stopping the whole domain rather than one rule.
+
+### Zero knowledge proves execution, not judgment
+
+A valid proof establishes that the committed ruleset was evaluated faithfully. It establishes nothing about whether the ruleset is correct, fair, or non-malicious. A domain that commits to `always allow` produces proofs that verify. implementers MUST NOT read a verdict as a safety property. It is an integrity property over a policy whose merit is a separate question, answered socially rather than cryptographically.
```
SamWilsn: Avoid creating requirements (defined with UPPERCASE keywords) outside of the specification section. They're too easy for implementers to miss. Instead, put the requirements in the Specification section and discuss their implications here.

mzf11125:
Understood, thanks @SamWilsn. Moving the interpreter fidelity and registry discussion over to the Magicians thread, and keeping this PR to formatting and style only.

One housekeeping note. I have just pushed the reference implementation into assets/erc-8354 and updated the spec to point at it, so the file list has grown. Status stays Draft.

mzf11125:
Thanks @zexoverz for the update on assets to better reflect implementation

mzf11125:
CI is all green and ready for review @SamWilsn @eip-review-bot

zexoverz:
@mzf11125 pushed two consistency fixes, and both follow-ups they left open are closed now as well.

`VerdictKindMismatch` is check 1, since it has to run before `DomainInactive` can make the disagreement unobservable. The ERC-8004 check is conditional on a domain declaring an `identityRegistry`, as check 2, so chains without one stay usable. `PolicyDomainRegistry` keeps every root still inside its own `maxRootAge` instead of one previous slot. `requires` now declares 712, 721 and 1271, which the document already linked.

`interfaceId` unchanged at `0xd6da8150`. 31 assets tests, 9 checks green. ready for next office hour call

SamWilsn (approved):

PR #1923, opened 2026-08-02
Update ERC-7730:  Generalize "mustMatch" feature with `boundaries` object with constraints leading to automatic signing refusal

forshtat:
## Summary

Early-stage idea, opened as a draft to gather feedback before finalizing.

Sometimes a parameter only makes sense within a bounded range (e.g. `pauseAccount(uint256 time)` intended to take a duration of at most ~10 days). A call with a wildly out-of-range value (`time = 864000000`, ~27 years) is very likely a mistake or malicious intent. Clear signing shows the raw value, but there's currently no way for a spec author to tell a wallet "this value is implausible, don't sign it without extra friction."

This PR adds an optional `boundaries` property to the field formatter (`v3.0.0-next` schema only):

```json
"boundaries": {
  "min": "0",
  "max": "864000",
  "excludes": ["1337"],
  "mustMatch": [0, 1, 2],
  "errorMessage": "You are locking your account for {time} seconds — this may be a mistake."
}
```

Wallets SHOULD refuse to sign by default when the underlying value violates the declared boundaries, and MAY offer an explicit override. It's independent of `format` and of `visible` — the field still displays normally.

As part of this, `mustMatch` moves out of `visible` into `boundaries`. It was never really a display concern — it forced the field to always be hidden and had nothing to do with visibility, it just didn't have anywhere else to live. `visible` now covers display only (`always`/`never`/`optional`/`ifNotIn`); `boundaries` covers refusal only, and no longer requires hiding the field to validate it.

## Open questions (feedback welcome)

- Should `tokenAmount`'s existing `threshold`/`message` (the "Unlimited" display swap) be related to or folded into `boundaries`? They're triggered by a similar condition (crossing a limit) but take different actions (display substitution vs. refusal) — leaning towards keeping them separate to avoid mixing display and refusal concerns again, but open to discussion.
- Byte-slice/sub-range bounds on `bytes` values, array-valued fields, and per-field severity levels (e.g. warn vs. refuse) are intentionally out of scope for now.

## Test plan

- [x] Schema changes validated as syntactically valid JSON
- [x] New `example-boundaries.json` validates against the updated schema
- [x] Existing `v3.0.0-next`-targeted examples still validate (no regressions)
- [ ] Community feedback on the shape of `boundaries` and the open questions above

### DIFF
--- a/ERCS/erc-7730.md
+++ b/ERCS/erc-7730.md
@@ -1071,7 +1071,8 @@ Optional rule to apply before displaying a field. Rules allow defining under whi
   * `optional`: Wallets MAY display this field if possible or sensible.
 * An object with keys representing complex rules, with the following complex rules supported:
   * `ifNotIn` key with an array of values to make the field visible only when the field value does NOT match any of the specified values
-  * `mustMatch` key with an array of values to make the field always hidden AND check its value against the list. Wallet should raise an error if the value does not match the list.
+
+Value validation independent of display (e.g. requiring an exact match) is expressed with `boundaries` instead.
 
 **`separator`** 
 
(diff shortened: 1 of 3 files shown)


ERCS/erc-7730.md
```diff
@@ -1071,7 +1071,8 @@ Optional rule to apply before displaying a field. Rules allow defining under whi
   * `optional`: Wallets MAY display this field if possible or sensible.
 * An object with keys representing complex rules, with the following complex rules supported:
   * `ifNotIn` key with an array of values to make the field visible only when the field value does NOT match any of the specified values
-  * `mustMatch` key with an array of values to make the field always hidden AND check its value against the list. Wallet should raise an error if the value does not match the list.
+
+Value validation independent of display (e.g. requiring an exact match) is expressed with `boundaries` instead.
```
```suggestion
```

PR #1950, opened 2026-08-15
Website: Fix the Auto Stagnant Bot, which has never run in this repository

Cybercentry:
The Auto Stagnant Bot has never moved a single ERC to Stagnant. This restores it by porting the EIPs bot into this repository.

## The problem

Two separate faults, one hiding the other.

**1. The job was gated to the wrong repository.** From the workflow's introduction until 4 August 2026 the condition was `if: github.repository == 'ethereum/eips'`, so in this repository every scheduled run since 2021 was skipped. That was corrected in 2b5baad9 and the job began executing on 9 August 2026.

**2. The bot cannot work here.** `ethereum/EIP-Bot` looks for proposals in a hardcoded directory:

```ts
// EIP-Bot/src/lib.ts, getEIPs()
.getContent({ owner, repo, path: "EIPS" })
```

This repository keeps its proposals in `ERCS/`, so the call 404s. From the first run that actually executed ([run 31285609418](https://github.com/ethereum/ERCs/actions/runs/31285609418)):

```
	====== CLEANUP COMPLETE =====
UnhandledPromiseRejectionWarning: HttpError: Not Found
    at async Object.getEIPs (.../build/src/lib.js:54:22)
    at async run (.../build/src/index.js:26:21)
```

The bot dies before evaluating a single proposal.

**3. The failure was invisible.** The step reports success anyway, because `node-version: '14'` downgrades an unhandled rejection to a warning instead of a non-zero exit. The run log says so itself: *"Unhandled promise rejections are deprecated... will terminate the Node.js process with a non-zero exit code"*. So the workflow has been reporting a green check every Sunday while doing nothing.

This cannot be fixed upstream: `ethereum/EIP-Bot` was archived in January 2023, so it accepts no pull requests.

## What this does

Adds `.github/scripts/mark-stagnant-ercs.js`, a port of EIP-Bot's `src/lib.ts` and `src/index.ts` parameterised for `ERCS/`, and points the workflow at it via `actions/github-script`. There is no `package.json` here, so the script uses only Node built-ins and the Octokit instance `github-script` provides.

Behaviour is deliberately identical to what runs in `ethereum/EIPs` today:

| | EIPs bot | This port |
|---|---|---|
| Cutoff | 6 months since last edit | same |
| Eligible statuses | Draft, Review | same |
| Branch | `mark-eip-<n>-stagnant-(2026-Apr-12th@00.41.33)` | `mark-erc-<n>-stagnant-(…)`, same date format |
| PR title | `EIP-8025 stagnant (…)` | `ERC-<n> stagnant (…)` |
| PR body | "has not been active since …; which, is greater than the allowed time of 6 months" + author mentions | same wording |
| Labels | `created-by-bot`, `1272989785` | same |
| Draft then ready for review | yes | same |
| Auto squash merge | after 2 weeks | same |
| Cleanup | close non-self bot PRs, close repeats, close obsolete, delete orphaned branches | same |
| Per-run limit | none | none |

Two intentional deviations, both marked `DEVIATION` in the source:

- **Last edited dates come from `git log` instead of one `listCommits` call per proposal.** Equivalent result, no API quota, and it removes 600+ requests per run. This is why the checkout now uses `fetch-depth: 0`.
- **Open pull requests are listed via `pulls.list` instead of the search API.** EIP-Bot's `fetchNonBotPRs` recurses into `fetchBotCreatedPRs` when paginating, so from page two onward it returns bot pull requests instead of the remaining non-bot ones. This repository has over 250 open pull requests, so that bug would bite here; listing directly paginates correctly and avoids the search API's tighter rate limit.

The `node-version: '14'` step is removed, so a future failure fails the job instead of passing silently.

## Please read this before merging

There are **611 ERCs, of which 221 are Draft and 63 are Review. 214 of those have not been edited in over six months** — the oldest have been untouched since the repository was split out in October 2023.

Because the bot has never run here, that entire backlog is eligible at once. With EIPs parity and no per-run cap, **the first scheduled run will open roughly 214 pull requests, and squash merge them two weeks later.** That is the correct outcome of the EIP-1 process applied to a backlog that has been accumulating for three years, but it is a large one-off event on a repository that already has ~274 open pull requests, so it should be a deliberate decision rather than a surprise on a Sunday morning.

`workflow_dispatch` therefore takes a `dry-run` input, **defaulting to true**, which writes the full list of affected ERCs to the job summary without creating, merging or closing anything. Running that once before the first live run is recommended. Scheduled runs are unaffected and always run live, exactly as in EIPs.

If you would prefer to drain the backlog gradually instead, say so and I will add a per-run limit.

## Testing

Run against a full clone with the API mocked, asserting that:

- selection matches a hand-computed list — 214 eligible, dropping to 213 when a mocked open pull request touches one of them
- every selected ERC really is Draft or Review
- the status rewrite changes exactly one line, `status: Draft` to `status: Stagnant`, leaving the rest of the file byte-identical
- the date formatter reproduces `(2026-Apr-12th@00.41.33)` from real EIPs PR ethereum/EIPs#11517, plus 1st/2nd/3rd/11th ordinal edge cases
- the orphaned-branch regex round-trips against a branch name the script itself generates, and does not match `master`

I have not been able to exercise the mutating paths against a live repository, since they need `secrets.TOKEN`. The `dry-run` mode exists partly so that a maintainer can validate selection here before anything is created.

### DIFF
--- a/.github/scripts/mark-stagnant-ercs.js
+++ b/.github/scripts/mark-stagnant-ercs.js
@@ -1,0 +1,510 @@
+// @ts-check
+/**
+ * Moves ERCs that have gone six months without an edit to Stagnant, opening one
+ * pull request per ERC and squash merging it two weeks later.
+ *
+ * This is a port of ethereum/EIP-Bot, which drives the identical process in
+ * ethereum/EIPs. That repository was archived in January 2023 and looks for
+ * proposals in a hardcoded `EIPS/` directory, so it cannot be fixed upstream and
+ * has never done anything here beyond failing silently.
+ *
+ * Behaviour is intentionally identical to the EIPs bot: same six month cutoff,
+ * same stagnatable statuses, same branch names, pull request titles, bodies and
+ * labels, and the same two week merge delay. Deviations are marked "DEVIATION"
+ * and explained.
+ */
+
+const { execFileSync } = require("child_process");
+const fs = require("fs");
+const path = require("path");
+
+const PROPOSALS_DIR = "ERCS";
+const PROPOSAL_PREFIX = "erc";
+const STAGNATION_CUTOFF_MONTHS = 6;
+const MERGEABLE_CUTOFF_WEEKS = 2;
+const MERGEABLE_CUTOFF_DESCRIPTION = "2 weeks";
+const STAGNATABLE_STATUSES = ["draft", "review"];
+const STAGNANT = "Stagnant";
+
+/** Pull requests are opened against this. */
+const DEFAULT_BRANCH = "master";
+
+/**
+ * The marker label the EIPs bot uses to recognise its own pull requests. Kept
+ * identical so tooling and saved searches behave the same across both repos.
+ */
+const BOT_ID = "1272989785";
+const PR_KEY_LABELS = ["created-by-bot", BOT_ID];
+
+/** EIPs excludes EIPS/eip-1.md here; this repository has no equivalent file. */
+const PATHS_TO_ALWAYS_EXCLUDE = [];
+
+/** Matches the branches this script creates, for orphan cleanup. */
+const STAGNANT_BRANCH_RE = new RegExp(
+  `mark-${PROPOSAL_PREFIX}-(\\d)+-stagnant-\\(\\d+-[a-zA-Z]+-.+@\\d+\\.\\d+\\.\\d+\\)`
+);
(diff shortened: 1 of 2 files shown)


Cybercentry:
Closing in favour of a cleaner branch; reopening the same change shortly.

PR #100, opened 2023-11-10
Add ERC: Embedded Accounts as Smart Modules

GregTheGreek:
NOTE: This EIP requires [SSO for Account Discovery](https://github.com/ethereum/ERCs/pull/99) which has not yet been merged.

PR #1951, opened 2026-08-15
Website: Fix the Auto Stagnant Bot, which has never run in this repository

Cybercentry:
The Auto Stagnant Bot has never moved a single ERC to Stagnant. This restores it by porting the EIPs bot into this repository.

## The problem

Two separate faults, one hiding the other.

**1. The job was gated to the wrong repository.** From the workflow's introduction until 4 August 2026 the condition was `if: github.repository == 'ethereum/eips'`, so in this repository every scheduled run since 2021 was skipped. That was corrected in 2b5baad9 and the job began executing on 9 August 2026.

**2. The bot cannot work here.** `ethereum/EIP-Bot` looks for proposals in a hardcoded directory:

```ts
// EIP-Bot/src/lib.ts, getEIPs()
.getContent({ owner, repo, path: "EIPS" })
```

This repository keeps its proposals in `ERCS/`, so the call 404s. From the first run that actually executed ([run 31285609418](https://github.com/ethereum/ERCs/actions/runs/31285609418)):

```
	====== CLEANUP COMPLETE =====
UnhandledPromiseRejectionWarning: HttpError: Not Found
    at async Object.getEIPs (.../build/src/lib.js:54:22)
    at async run (.../build/src/index.js:26:21)
```

The bot dies before evaluating a single proposal.

**3. The failure was invisible.** The step reports success anyway, because `node-version: '14'` downgrades an unhandled rejection to a warning instead of a non-zero exit. The run log says so itself: *"Unhandled promise rejections are deprecated... will terminate the Node.js process with a non-zero exit code"*. So the workflow has been reporting a green check every Sunday while doing nothing.

This cannot be fixed upstream: `ethereum/EIP-Bot` was archived in January 2023, so it accepts no pull requests.

## What this does

Adds `.github/scripts/mark-stagnant-ercs.js`, a port of EIP-Bot's `src/lib.ts` and `src/index.ts` parameterised for `ERCS/`, and points the workflow at it via `actions/github-script`. There is no `package.json` here, so the script uses only Node built-ins and the Octokit instance `github-script` provides.

Behaviour is deliberately identical to what runs in `ethereum/EIPs` today:

| | EIPs bot | This port |
|---|---|---|
| Cutoff | 6 months since last edit | same |
| Eligible statuses | Draft, Review | same |
| Branch | `mark-eip-<n>-stagnant-(2026-Apr-12th@00.41.33)` | `mark-erc-<n>-stagnant-(…)`, same date format |
| PR title | `EIP-8025 stagnant (…)` | `ERC-<n> stagnant (…)` |
| PR body | "has not been active since …; which, is greater than the allowed time of 6 months" + author mentions | same wording |
| Labels | `created-by-bot`, `1272989785` | same |
| Draft then ready for review | yes | same |
| Auto squash merge | 2 weeks later, in a subsequent run | **same run, once required checks pass** |
| Cleanup | close non-self bot PRs, close repeats, close obsolete, delete orphaned branches | same |
| Per-run limit | none | none |

Three intentional deviations, all marked `DEVIATION` in the source:

- **Pull requests are merged in the run that opens them**, rather than sitting open for two weeks. Each is polled until `mergeable_state` leaves `blocked`, then squash merged. Opening and merging are separate phases: merging each pull request immediately after opening it would idle the job for the minutes its checks take, and 200+ of those would exceed the six hour job limit, so all the pull requests are opened first and the earliest are already green when the merge pass starts. Anything that conflicts, fails its checks, or has not settled within 30 minutes is left open, and the next run's cleanup merges it — EIP-Bot's merge-old-PRs routine is retained as that safety net, with its cutoff set to zero.
- **Last edited dates come from `git log` instead of one `listCommits` call per proposal.** Equivalent result, no API quota, and it removes 600+ requests per run. This is why the checkout now uses `fetch-depth: 0`.
- **Open pull requests are listed via `pulls.list` instead of the search API.** EIP-Bot's `fetchNonBotPRs` recurses into `fetchBotCreatedPRs` when paginating, so from page two onward it returns bot pull requests instead of the remaining non-bot ones. This repository has over 250 open pull requests, so that bug would bite here; listing directly paginates correctly and avoids the search API's tighter rate limit.

The `node-version: '14'` step is removed, so a future failure fails the job instead of passing silently.

## Please read this before merging

There are **611 ERCs, of which 221 are Draft and 63 are Review. 214 of those have not been edited in over six months** — the oldest have been untouched since the repository was split out in October 2023.

Because the bot has never run here, that entire backlog is eligible at once. There is no per-run cap, matching EIPs, so **the first run will open roughly 214 pull requests and merge all of them in the same run.** That is the correct outcome of the EIP-1 process applied to a backlog that has accumulated for three years, but it is a large one-off event and worth calling out explicitly:

- It will trigger roughly 214 × 4 = **850 or so check runs** in a short window, since every pull request runs CodeSpell, EIP Walidator, Link Check and HTMLProofer. That is a real load on the runner pool.
- It will notify the authors of ~214 proposals, and unlike in EIPs there is no two week window in which they can object first — the status change lands in the same run. It remains reversible: EIP-1 lets an author resurrect a Stagnant proposal at any time by opening a pull request moving it back to Draft.
- The run itself will take a while: ~214 pull requests at a five second spacing, plus the merge pass.

**Every run after the first is small.** Steady state is however many ERCs cross the six month line in a given week, typically a handful or none, opened and merged the same morning.

`workflow_dispatch` takes a `dry-run` input, **defaulting to true**, which writes the full list of affected ERCs to the job summary without creating, merging or closing anything. Running that once before the first live run is strongly recommended. Scheduled runs always run live.

If you would rather drain the backlog gradually, or keep the two week author notice window, say so — both are small changes.

## Testing

Run against a full clone with the API mocked, asserting that:

- selection matches a hand-computed list — 214 eligible, dropping to 213 when a mocked open pull request touches one of them
- every selected ERC really is Draft or Review
- the status rewrite changes exactly one line, `status: Draft` to `status: Stagnant`, leaving the rest of the file byte-identical
- the date formatter reproduces `(2026-Apr-12th@00.41.33)` from real EIPs PR ethereum/EIPs#11517, plus 1st/2nd/3rd/11th ordinal edge cases
- the orphaned-branch regex round-trips against a branch name the script itself generates, and does not match `master`

And a second suite driving the live path with the whole API mocked and timers stubbed, asserting that:

- all 214 proposals get exactly one branch, one commit, one pull request and one label set each, with no per-proposal errors
- every pull request is opened as a draft and then marked ready for review
- each is polled until its checks report clean, and only then merged, by squash
- a pull request reporting a conflict is skipped and left open rather than force merged — 213 of the 214 merge
- the commit flips `status:` to `Stagnant` and passes the correct blob sha
- titles, commit messages and merge titles match the EIPs formats, e.g. `(bot 1272989785) moving ERCS/erc-4972.md to stagnant (#100)`

What is **not** covered: no mutating call has run against a live repository, since they need `secrets.TOKEN`. The mocks assert the calls this script makes, not that GitHub accepts them. `dry-run` exists partly so a maintainer can validate selection here before anything is created.

### DIFF
--- a/.github/scripts/mark-stagnant-ercs.js
+++ b/.github/scripts/mark-stagnant-ercs.js
@@ -1,17 +1,18 @@
 // @ts-check
 /**
  * Moves ERCs that have gone six months without an edit to Stagnant, opening one
- * pull request per ERC and squash merging it two weeks later.
+ * pull request per ERC and squash merging it once its required checks pass.
  *
  * This is a port of ethereum/EIP-Bot, which drives the identical process in
  * ethereum/EIPs. That repository was archived in January 2023 and looks for
  * proposals in a hardcoded `EIPS/` directory, so it cannot be fixed upstream and
  * has never done anything here beyond failing silently.
  *
- * Behaviour is intentionally identical to the EIPs bot: same six month cutoff,
- * same stagnatable statuses, same branch names, pull request titles, bodies and
- * labels, and the same two week merge delay. Deviations are marked "DEVIATION"
- * and explained.
+ * Behaviour follows the EIPs bot closely: same six month cutoff, same stagnatable
+ * statuses, same branch names, pull request titles, bodies and labels, and the
+ * same cleanup routines. Deviations are marked "DEVIATION" and explained; the
+ * significant one is that pull requests are merged in the run that opens them
+ * rather than two weeks later.
  */
 
 const { execFileSync } = require("child_process");
@@ -21,8 +22,23 @@ const path = require("path");
 const PROPOSALS_DIR = "ERCS";
 const PROPOSAL_PREFIX = "erc";
 const STAGNATION_CUTOFF_MONTHS = 6;
-const MERGEABLE_CUTOFF_WEEKS = 2;
-const MERGEABLE_CUTOFF_DESCRIPTION = "2 weeks";
+
+/**
+ * DEVIATION: the EIPs bot leaves its pull requests open for two weeks and merges
+ * them during a later run's cleanup. Here they are merged in the same run, as
+ * soon as the required status checks pass.
+ *
+ * The cleanup merge is kept as a safety net with a cutoff of zero, so anything
+ * that could not be merged in its own run — checks still running when the
+ * timeout expired, a transient failure — is merged by the next run instead of
+ * being stranded.
+ */
+const MERGEABLE_CUTOFF_WEEKS = 0;
+const MERGEABLE_CUTOFF_DESCRIPTION = "a previous run";
+
+/** How long to wait for a new pull request's required checks before giving up. */
+const CHECKS_TIMEOUT_MS = 30 * 60 * 1000;
+const CHECKS_POLL_MS = 15 * 1000;
 const STAGNATABLE_STATUSES = ["draft", "review"];
 const STAGNANT = "Stagnant";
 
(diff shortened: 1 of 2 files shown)


Cybercentry:
Superseded by a simpler implementation.

PR #1953, opened 2026-08-16
Add ERC: Unclonable Agent Execution Credentials

mzf11125:
This proposal defines a single use capability token for delegated agent execution. An orchestrator issues a capability bound to an agent identity and a per issuance secret salt. Spending it requires a zero-knowledge proof, which forces a nullifier derived from that salt into the open, and the Guard burns the nullifier as part of performing the authorized action. A cloned agent holds the same salt, produces the same nullifier, and is rejected.

The idea was discussed on Ethereum Magicians before this PR was opened:
https://ethereum-magicians.org/t/idea-draft-erc-unclonable-agent-execution-credentials-via-zero-knowledge-nullifiers/29274

That thread ran for two weeks and changed the design three times. The substantive points and where each landed:

- Unclonability and authorization soundness are orthogonal (https://ethereum-magicians.org/t/idea-draft-erc-unclonable-agent-execution-credentials-via-zero-knowledge-nullifiers/29274/2). This is now the leading subsection of Security Considerations rather than a footnote. The proposal states plainly that it supplies one of several independent layers and that adopting it alone does not make an agent safe.
- `chainId` must not be in the nullifier preimage, or the nullifier forks per chain and the credential is spendable once on each. Chain binding is now an acceptance check on `homeChainId` instead.
- A separable spend primitive lets anyone satisfying the executor check burn a capability without performing the work, so exactly once holds and the deployment still loses. Issuance is now mandatory and the burn is fused to the authorized call. This was the largest change and it came directly from adversarial test vectors contributed on the thread.
- A revert cannot emit, so observability is a burn event on the accepting path plus a named error on the rejecting one, with `highestIssuedIndex` to classify a collision as a clone versus an issuer reissue bug.
- The security claim is stated as at most once with no ordering, not exactly once. Two holders of the same salt race and the Guard cannot rank them.
- Aggregate spend per identity is named as a separate layer and explicitly out of scope.

The reference implementation lives at https://github.com/mzf11125/unclonable-agent-execution-credentials and the Solidity sources are also vendored into `assets/erc-1953/`, so the proposal body carries no external links. Nineteen Foundry tests cover the cases in the Test Cases section, including the adversarial vectors from the thread.

Known open questions, stated rather than hidden:

1. Cross chain nullifier mirroring. A capability deliberately issued as spendable on either of two chains with no designated home needs real consensus on spentness. The draft requires a home domain and declares the other case out of scope rather than solving it.
2. Proof generation latency. Proving sits in the execution path, and an attacker holding cloned memory can start proving at the same moment. No mitigation is specified, because a time lock moves the race later without changing who wins it.
3. Relayed submission. The Guard requires `msg.sender == executor`, so an agent holding no gas cannot spend through a relayer. Adding an EIP-712 path needs a signature argument on `execute`, which changes the interface, so it is deferred.

All authors listed in the preamble have consented to CC0 licensing and to being named.


babyblueviper1:
Good to see this land as a formal PR -- the thread's arc (separable `consume` → Cedric's adversarial vectors → mandatory issuance + fused burn) is exactly the right sequence, and CI's green on the current head.

One place worth pinning down: "authorization soundness" is named as the layer this proposal explicitly does *not* supply (§ Security Considerations, "Unclonability and authorization soundness are orthogonal"). Is that meant to cover both (a) whether the issuing policy was actually followed, and (b) whether the resulting authorized action was itself a *good* decision? Those are different failure modes with different fixes -- a confidential policy verdict (CAPV / ERC-8354) can prove (a) without ever touching (b). We ran into the same split on the CAPV thread (https://ethereum-magicians.org/t/29088/54, replying to a similar "does ZK add anything over signing" question): ZK removes trust in the *definer's* honesty about execution, but that's compliance-with-policy, not soundness-of-the-outcome -- a policy can be followed to the letter and still authorize something bad.

Not proposing a change to scope, just flagging that if "authorization soundness" reads as one property to an implementer, a deployment wiring in ERC-8354 alone could believe it's covered both axes when it's only closed the first. Might be worth a clause distinguishing them, the same way the doc already separates unclonability from soundness.

ERCS/erc-1953.md
```diff
@@ -0,0 +1,385 @@
+---
+eip: 1953
```
abcoathup:
```suggestion
eip: 8380
```

ERCS/erc-1953.md
```diff
@@ -0,0 +1,385 @@
...
+title: Unclonable Agent Execution Credentials
+description: Single use agent capability tokens that burn a zero-knowledge nullifier so a cloned agent cannot replay them
+author: Muhammad Zidan Fatonie (@mzf11125), Faisal Firdani (@zexoverz), Maulana Asykari Muhammad (@WeissCurry), Venkata ramana Komari (@Venkat5599)
+discussions-to: https://ethereum-magicians.org/t/idea-draft-erc-unclonable-agent-execution-credentials-via-zero-knowledge-nullifiers/29274
```
abcoathup:
```suggestion
discussions-to: https://ethereum-magicians.org/t/erc-8380-unclonable-agent-execution-credentials/29274
```
Updated Eth Magicians title with assigned number and ERC title.

mzf11125:
@zexoverz the PR is up here. Feel free to push directly to the branch if you have technical changes, especially anything on the coupling question from the forum thread. If you'd rather keep it to a review, that works too, just flag the specific sections and I'll fold it in.

zexoverz:
@babyblueviper1 two properties, and only the first is reachable here. A proof shows the verdict came from the committed policy, never that the policy was worth committing to.

@mzf11125 keeping it to review rather than pushing, since the nullifier point moves the public input list. `pipavlo82` is right in post 19 and I will answer the coupling question on the thread. Separately the ref impl links point at `../assets/eip-1953/` while the directory is `assets/erc-1953/`, and both want `erc-8380`.

babyblueviper1:
@zexoverz agreed, and worth being explicit about since it'll come up again on other proposals in this cluster: property (b) isn't fixable by making the ZK circuit bigger or the policy commitment tighter — no proof over a committed policy can tell you the policy was the right one to commit to, because that's a judgment call about the world outside the circuit, not a computation over its inputs. It needs an independent party actually evaluating the decision, before or after the fact, not another cryptographic binding on the same artifact. Different tool, not a stronger version of this one.

mzf11125:
Pushed a fix for all three points.

pipavlo82 was right in post 19. nullifier is now a ninth public input and the circuit must prove its own H(NULLIFIER_TAG, salt) output equals it, so cap.nullifier is checked against a value the circuit actually derived instead of trusted from calldata. executor and expiry now sit inside the capabilityCommitment preimage too, since neither was constrained by the circuit before this.

Also fixed the asset path mismatch Zexo flagged. Renamed assets/erc-1953 to assets/erc-8380 and updated every eip-1953 link in the doc to eip-8380, plus the domain separation tag strings.

babyblueviper1, added a clause in Security Considerations splitting authorization soundness into whether the action matches what was authorized versus whether the authorization itself was sound. This proposal only reaches the first, same framing Zexo gave above.

Diff is on the branch now.

zexoverz:
Checked all three, they land. Input 8 plus the circuit proving its own nullifier closes the calldata-trust gap, and executor/expiry in the preimage kills the swap. Good to go from me.

babyblueviper1:
Read the current head. The Security Considerations clause reads correctly — the action-matches-authorization vs. authorization-was-sound split is stated cleanly under "Unclonability and authorization soundness are orthogonal," and it's honest about reaching only the first property.

The binding fix also checks out: nullifier is now public input 8 with the circuit proving it equals its own `keccak256(NULLIFIER_TAG, salt)` output, and executor/expiry both sit inside `capabilityCommitment`'s preimage now rather than being trusted calldata. That closes the gap pipavlo82 flagged (post 19) and matches what zexoverz confirmed landed.

One thing worth naming since it's adjacent to the soundness point: the "at most once, with no ordering" section is doing real work — it's explicit that a compromised agent's clone can still *win* the race for the one authorized action, and that no ordering guarantee exists without keeping the salt off cloneable memory. That's the honest complement to the soundness disclaimer above it: even a perfectly-authorized, perfectly-consumed credential doesn't guarantee the *legitimate* party is the one who spent it. Good to see both stated plainly rather than implied.

No objection from this side.

Frank122480:
U can view my banks and Financials i frank122480 haven't spent any crypto or able to get any out other than little things with coinbase. But haven't been member of coinbase for over years. Had to go to kraken

babyblueviper1:
Verified the domain-scoping fix (7068f2a2) against the real diff, not just the commit message: `highestIssuedIndex` moving to `[agentId][homeDomainId]`, `issue` taking `homeDomainId` and resolving authorization through `domainRegistry.orchestratorOf`, and test_16 rewritten to actually exercise cross-domain isolation (asserts domain B's ceiling stays 0 after domain A issues to 10, then arms a genuine collision at index 4 in domain B and confirms it reverts). 17/17 as claimed. Real fix for a real inversion — a clone colliding at an index another domain issued was reading as the orchestrator's own reissue bug.

One adjacent question this raised while reading it, distinct from today's fix (which is correct as far as it goes): line 75 says `capabilityIndex` MUST be monotonically increasing per `(agentId, homeDomainId)`, not that it must be consecutive. `issue()` doesn't enforce either — it only conditionally raises the ceiling (`if (capabilityIndex > highestIssuedIndex[...])`), which reads as intentional given the orchestrator is already the trusted party elsewhere in the spec (Issuance section: "Whether the Guard checks the commitment against an issuance record is a security property," i.e., against forged issuance, not against a misbehaving orchestrator).

Given that, if an orchestrator issues 1, then jumps to 100 (both satisfy "monotonically increasing," nothing requires 2..99 to have been used), the ceiling becomes 100. A clone colliding at index 50 — never actually issued — is `<= highestIssuedIndex`, the same read `test_16` uses to mean "the orchestrator's own reissue bug." Does the spec assume/require the RECOMMENDED derivation (`HKDF(issuerSecret, agentId || homeDomainId || capabilityIndex)`) is walked without gaps in practice, making this a non-issue by convention rather than by any on-chain guarantee? Or is a skipped-index collision meant to be a known residual (same shape as the outage/liveness tradeoffs already named elsewhere in Observability), just not spelled out for this specific case yet?

chugarchugarr:
Following up on the skipped-index question here: https://github.com/ethereum/ERCs/pull/1953#issuecomment-5515589275

I made that case executable against the current `7068f2a2` head before proposing a change.

The `1 → 100`, unissued `50` case confirms that `highestIssuedIndex` is a high-water mark rather than exact issuance membership. With monotonic but non-consecutive indices, an index can be `<= highestIssuedIndex` without ever having been issued.

While testing that, a second issue became clearer: on the current collision path, `consumed[cap.nullifier]` is checked before the attempted capability has passed proof/issuance authentication. So caller-supplied capability metadata should not itself determine what the collision proves.

I tested a candidate that records the capability commitment that first consumed each nullifier and authenticates any later opening before classifying it.

That gives three distinguishable states:

- valid opening + unissued commitment → outside recorded issuance state;
- valid opening + different issued commitment sharing the nullifier → issuer salt reuse across distinct issuances;
- valid opening + same issued commitment → replay of that capability.

The last case appears to expose an important boundary in the current observability claim: replay is observable, but “clone” versus “same holder retrying” is not established by replay alone.

This also suggests `highestIssuedIndex` is useful bookkeeping, but should not carry collision-provenance semantics.

I added regressions for the skipped-index case, forged collision metadata, issued salt reuse, and exact replay. Against the current head, the existing 17 tests plus these 4 are green: **21/21**.

I have the corresponding spec/interface/Guard change built, but I have **not pushed it yet**. Before I turn it into an upstream diff, does this match the intended invariant, or is there an assumption in the current design that I am missing?

babyblueviper1:
Independently re-read the current head (7068f2a2) against both claims before saying anything -- both check out precisely.

`highestIssuedIndex` (L54-55): `issue()` only bumps it when `capabilityIndex > highestIssuedIndex[...]`, so it's genuinely a max-seen counter over the values that happened to be issued, not a membership set. A monotonic-but-gapped issuance sequence (your 1→100, skip 50 case) leaves it structurally unable to distinguish "50 was issued" from "50 was skipped" -- confirmed, `issued[commitment]` (the real membership check) is keyed by commitment, not index, so nothing else in the contract fills that gap either.

The ordering issue (L95 vs L98/105) is real too: `consumed[cap.nullifier]` reverts on caller-supplied, entirely unauthenticated calldata -- `issued[]` and the ZK proof check both happen after. So `CredentialAlreadySpent` fires (and reveals that specific nullifier's state) before anything in `cap` has been shown to correspond to a real issued capability at all.

Your three-state classification is the right fix shape -- it needs the collision path to authenticate before it discloses anything, same principle as gating on proof-then-classify rather than classify-then-proof. Nothing to add beyond confirming it; 21/21 including your four new regressions is a solid bar for landing this.

chugarchugarr:
Implemented and revalidated cleanly from the exact current head (`7068f2a2853504475d966c06c8e04ec6d546dbb8`).

Clean commit:
https://github.com/chugarchugarr/ercs/commit/5d216decc41b92264e40f29c852eeeca05ec65fa

Exact base → candidate diff:
https://github.com/chugarchugarr/ercs/compare/7068f2a2853504475d966c06c8e04ec6d546dbb8...5d216decc41b92264e40f29c852eeeca05ec65fa

Validation run in the ERC fork:
https://github.com/chugarchugarr/ercs/actions/runs/33828137295

Foundry 1.8.1 / Solc 0.8.28: **21 passed, 0 failed, 0 skipped** — the existing 17 plus the 4 new regressions.

The clean branch is one commit ahead of `7068f2a2` and changes exactly four files: the ERC text, `CoupledCredentialGuard.sol`, `IUnclonableCredential.sol`, and `AuthenticatedCollision.t.sol`.

The implementation follows the three-state shape discussed above: proof-authenticate a consumed-nullifier opening first, then distinguish unissued commitment / issued salt reuse / exact replay, while leaving `highestIssuedIndex` as bookkeeping rather than provenance evidence.

@​mzf11125 the commit is ready to cherry-pick or fold into the ERC branch.

PR #1958, opened 2026-08-18
Update ERC-8167: Move to Last Call

wjmelements:
#### Changes
* move ERC-8167 to Last Call


wjmelements:
@SamWilsn ptal

wjmelements:
I hope to make this ERC "Final" before the next Filecoin hard fork, which will enshrine two contracts that follow this standard.

ERCS/erc-8167.md
SamWilsn, line 160: You should avoid creating requirements (defined with UPPERCASE keywords) outside of the specification section. They're way too easy for implementers to miss. Instead, you should put the requirements in the Specification section and discuss their security implications here.

That said, there's no official rule about this, so I'll go ahead and approve this, but if you could fix it before going into final, I'd appreciate it!

SamWilsn (approved):

ERCS/erc-8167.md
wjmelements: That makes sense to me.

PR #1963, opened 2026-08-23
Update ERC-7573: Fix key parameter descriptions in ILockingContract m…

mtoub:
## Summary
Editorial fixes to the `ILockingContract` method descriptions. No normative or semantic change.

1. **`inceptTransfer`**: the description of `keyEncryptedSeller` said the key "can be used by the buyer to claim the token". Both parameters of this call describe the single seller key (hash and ciphertext of the same secret), as shown by the parameter naming, the symmetric wording of `confirmTransfer`, and the reference paper's sequence diagram, `inceptTransfer(H(F), E(F))`. Corrected to the seller (re-)claim key.
2. **`confirmTransfer`**: the buyer claims for the first time, so the "(re-)" prefix is removed, and `keyEncryptedBuyer` is clarified as encrypting the same key hashed by `keyHashedBuyer`.


Frank122480:
Lock contract plz
Frank122480 
Frankrking36
Victim of fraud

PR #1966, opened 2026-08-24
Update ERC-7730: Clarify metada.maps key

melanciani:
`metadata.maps` has been in the ERC-7730 spec since November 2025 but is not yet used by
any descriptor in the registry. Implementing it in two wallet libraries surfaced
a few underspecified points; this clarifies them. All schema changes are to the
`3.0.0-next` draft only — `2.0.0` is untouched.

- **Key resolution.** The spec never defined how the value at `keyPath` becomes a
  map key. Adds a canonical key form per resolved type, and requires keys to be
  matched case-insensitively — descriptor authors write ERC-55 checksummed
  addresses, while a value resolved from calldata or the container carries no
  casing guarantee.
- **No match.** Scopes the existing MUST: the wallet must not use the descriptor
  for that transaction and must fall back as if no descriptor were available,
  rather than substituting a default or rendering the format's other fields.
  Adds a SHOULD to surface a diagnostic, since a miss is normally a deployment
  added without a matching map entry.
- **Schema completeness.** The prose allows map references for constant
  parameters, but the schema only declared them on parameters having a `*Path`
  sibling. Adds `threshold`, `nativeCurrencyAddress`, `senderAddress` and the
  `unit` scale, and states the inclusion rule in the text: constant operands of
  a format accept a map reference, parameters selecting how a field is decoded
  (`encoding`, `encryption`, `$ref`) do not.
- **New example** `example-maps-deployments.json`, keyed on the called contract
  address. Both existing examples key on the chain id, so the address-keyed case
  — the one that motivated the case-insensitivity rule — had no coverage.
- Minor: the map definition used `"unresolvedProperties"`, which is not a JSON
  Schema keyword and was therefore inert. Corrected to `unevaluatedProperties`.
```

### CODE REVIEW

File: assets/erc-7730/example-maps-deployments.json [hunk 1]
```diff
@@ -0,0 +1,63 @@
...
+    "info": {
+      "url": "https://example.io/"
+    },
+    "maps": {
+      "underlyingToken": {
+        "$keyType": "Address of the called wrapper contract",
+        "values": {
+          "0x00112233445566778899AABBCCDDEEFF00112233": "0x0000000000000000000000000000000000deadbeef",
+          "0xAABBCCDDEEFF00112233445566778899AABBCCDD": "0x0000000000000000000000000000000000feedface"
+        }
```
forshtat: Maybe an `address => address` mapping is not the best example - and also there is an example in the ERC text already, could we maybe reuse that one?

```json
"$keyType": "Chain ID",
                "values" : {
                    "1": "0xaabbccddeeff...",
                    "17000": "0x112233445566..."
                }
```

ERCS/erc-7730.md
```diff
@@ -655,13 +655,40 @@ A map is a json object with two keys:
 * The `$keyType` key contains an non normative indication of the expected type of the map key passed when referencing the map.
 * The `values` key contains a list of key / value pairs, each key being used to match the data pointed to in the reference `keyPath` key. The corresponding value is used as the context-dependent constant when referenced in the `display` section.
 
-Maps references can be used anywhere a parameter with constant value would be used. A map reference is a json object with two keys:
+Maps references can be used in place of a constant parameter value. The parameters accepting a map reference are the ones declared as such in the schema, following two rules:
+
+* A parameter accepts a map reference when its value is a constant *operand* of the format: a value the format compares the field against, or uses to render it. `token`, `nativeCurrencyAddress`, `senderAddress`, `collection`, `callee`, `threshold` and the `unit` scale are of this kind.
```
forshtat: I am not sure I was able to understand this sentence.

ERCS/erc-7730.md
```diff
@@ -655,13 +655,40 @@ A map is a json object with two keys:
...
+* A parameter does not accept a map reference when it selects *how* the field is decoded, or when it points at another definition. `encoding`, the `encryption` parameters and `$ref` are of this kind: making them context-dependent would let identical data be interpreted differently per deployment, which the bound contract's ABI already fixes.
+
+Purely human-readable parameters, such as the `message` accompanying a `threshold`, are deliberately left out: they carry no context-dependent constant. A map value is a scalar, so parameters taking an array of values, such as `types` and `sources`, cannot be resolved through a map either.
```
forshtat: `A map value is a scalar`

Is there a reason for that?

ERCS/erc-7730.md
```diff
@@ -655,13 +655,40 @@ A map is a json object with two keys:
...
+
+A wallet MUST replace a path to a map using the value matching the `keyPath` parameter.
+
+**Resolving the map key**
+
+The value pointed to by `keyPath` is rarely a string in the structured data, so a wallet MUST convert it to its *canonical key form* before matching it against the keys of the `values` object:
+
+| Resolved value type | Canonical key form                                             |
+|---------------------|----------------------------------------------------------------|
+| `address`           | `0x`-prefixed hexadecimal, no checksum casing applied          |
```
```suggestion
| `address`           | `0x`-prefixed hexadecimal          |
```

At least if we say:
``A wallet MUST compare the canonical key form against the keys of the `values` object case-insensitively.`` - checksum should not matter, should it?

ERCS/erc-7730.md
```diff
@@ -655,13 +655,40 @@ A map is a json object with two keys:
...
+| `bool`              | `true` or `false`                                              |
+| `string`            | The string itself                                              |
+
+A wallet MUST compare the canonical key form against the keys of the `values` object case-insensitively. This makes address keys robust to checksum casing: descriptor authors typically write [ERC-55](./eip-55.md) checksummed addresses, whereas the value resolved from the structured data or from the transaction container carries no casing guarantee. It also means a map MUST NOT declare two keys that differ only by case.
+
+**When no value matches**
+
+If no key of the `values` object matches, the descriptor does not describe the transaction it is being applied to: the map enumerates the contexts its author accounted for, and this is not one of them. The wallet MUST NOT use the descriptor to display the transaction, and MUST fall back to the same behaviour as when no descriptor is available at all. It MUST NOT substitute a default, and MUST NOT render the remaining fields of the format, as unrelated constants in the same descriptor may be equally out of date.
```
forshtat: `It MUST NOT substitute a default`

Is there a way to define a default value for a mapping? Should there be?

assets/erc-7730/example-maps-deployments.json
melanciani: I've removed the new example and updated the initial one to include address => address, as I think it's still good to show different usage, especially since we mention case sensitivity checks in the specs !

ERCS/erc-7730.md
melanciani: agree, tried to clarify !

ERCS/erc-7730.md
melanciani: no ! removed it

ERCS/erc-7730.md
melanciani: agree, removed it

ERCS/erc-7730.md
melanciani: I don't think it should no, I clarified the sentence !

PR #1968, opened 2026-08-26
Update ERC-8354: state why DomainIdentityRegistryUpdated must be monitored

zexoverz:
One sentence in Security Considerations. It told implementers to monitor `DomainIdentityRegistryUpdated` for the same reason they monitor `DomainProgramUpdated`. The action is the same and the reason is not, and flattening them loses the part that matters. A current `programKey` interprets itself, so a reader who missed the event can compare it against what they last saw, and a reader with no baseline takes the current value as theirs. A current `identityRegistry` of `address(0)` does not interpret itself, because a domain that never declared a registry and a domain that declared one and withdrew it are the same state read.

The history that separates the two is recoverable by backfilling `DomainIdentityRegistryUpdated` from the registry deployment, but logs are unavailable to an on-chain consumer and are not guaranteed retrievable once history expiry applies. So a state read is a valid fallback for `programKey` and not for `identityRegistry`, which is what the line should say. Raised by @babyblueviper1 on the discussions thread. No normative text changes and no keywords introduced outside Specification.


babyblueviper1:
Nice, thanks for tightening this @zexoverz — the fallback distinction (programKey self-interprets, identityRegistry's zero-address doesn't) is exactly the gap. Good catch on the log-availability/history-expiry point too.

PR #1974, opened 2026-08-27
Update ERC-7730: define wallet behavior for non-zero native value

### LINKED ISSUE #2900: Missing defined wallet behavior for non-zero msg.value not referenced by descriptors
## Problem

Function selectors don't encode payability, so the same clear-signing format applies whether or not the transaction carries native value. If that format doesn't reference `@.value`, nothing in the current standard says the wallet should show it.

As far as I can tell, the ERC doesn't address this at all. v2 seems to let an author assert:

```json
{"path": "@.value", "visible": {"mustMatch": ["0"]}}
```

so the wallet refuses to sign when value is attached (draft [#1923](https://github.com/ethereum/ercs/pull/1923) moves that into boundaries). If, however, the binding is simply omitted, it is currently up to each wallet to decide on behaviour.

This could result in transactions carrying an arbitrary native value that some wallets never show, which could easily attract malicious attention. While this may overlap with the general descriptor-credibility discussion (and related attestation registries), this seems easily preventable by specifying the expected wallet behaviour in the standard.

## Proposed change

I would propose adding the following to **Security Considerations**:

> A descriptor is bound to a contract deployment, and a format specification within it is selected by function selector. A selector does not encode payability, so the same format applies to transactions carrying any native value. Wallets MUST show `@.value` to the user when it is non-zero, even when no field references it. For each format describing a payable function, authors SHOULD either bind `@.value` as a displayed field, or assert that `@.value` is zero where the flow is not expected to carry native value.

## Open question

One point worth discussing: protocols that charge a mandatory fee in native value would surface dust on every legitimate call, which is presumably why some implementations suppress it.

The draft boundaries object looks like a good fit here (an author could declare the expected range for `@.value`), though it's opt-in, so I think the standard still needs to say what a wallet should do when nothing is declared.

soyrubio:
Resolves https://github.com/ethereum/clear-signing-erc7730-registry/issues/2900.

## Motivation

Function selectors encode parameter types only, not payability, so the same format
specification applies whether or not a transaction carries native currency value. When a
format does not reference `@.value`, the standard currently says nothing about what the
wallet should display, which leaves the decision to each implementation. A transaction can
therefore carry arbitrary native value that some wallets never surface to the user.

## Change

Adds a `Native currency value (@.value)` subsection to Security Considerations, placed after
`Binding context`:

- Wallets MUST display `@.value` when it is non-zero, even when no field in the format
  specification references it.
- For formats describing a payable function, authors SHOULD either include `@.value` as a
  displayed field, or assert that it is zero when the described operation is not expected to
  carry native currency value.

The author-side recommendation is intentionally mechanism-agnostic. The schema already
provides ways to constrain a field's value, and ongoing proposals would change how such
constraints are expressed, so naming a specific key here would date the text.

Protocols that charge a mandatory fee in native value would surface dust on every legitimate
call, which is presumably why some implementations suppress it. Future spec or schema changes
may introduce better ways to encode an expected range for a value, which would partially
address this. Any such mechanism is opt-in, though, so the default wallet behavior still needs
to be defined for the case where nothing is declared.

### DIFF
--- a/ERCS/erc-7730.md
+++ b/ERCS/erc-7730.md
@@ -1682,6 +1682,14 @@ ERC-7730 creates a risk surface where malicious actors could abuse formatting to
 
 The binding `context` mitigates misuse by specifying exactly which structured data a given ERC-7730 file may format. app developers MUST set restrictive constraints; wallets MUST verify all constraints and ensure formatting is cryptographically bound to the reviewed data and not tamperable in transit (including on resource-constrained hardware wallets).
 
+### Native currency value (`@.value`)
+
+Wallets select a format specification by function selector. Since a selector does not encode payability, the same format applies whether or not the transaction carries native currency value. A format that omits `@.value` could therefore hide an unexpected native currency transfer.
+
+Wallets MUST display `@.value` to the user when it is non-zero, even when no field in the format specification references it.
+
+For each format describing a payable function, descriptor authors SHOULD either include `@.value` as a displayed field, or assert that it is zero when the described operation is not expected to carry native currency value.
+
 ### Registry poisoning
 
 Curation and registry of ERC-7730 files are out of scope for this specification, but any external registry or wallet ecosystem MUST assume it will be targeted. A secure registry should (a) require cryptographically verifiable provenance and attestations for each ERC-7730 file and its maintainer, (b) keep a public, tamper-evident history of submissions, approvals, and revocations, (c) implement a mechanism to credibly link contract ownership or authority to the submitted file, and (d) adopt a clear governance model with multi-party sign-off and automated monitoring to detect anomalies or mass poisoning attempts. These measures mitigate attacks on ERC-7730 files in transit and make registry compromise significantly harder without constraining this ERC itself.


shukudaidayo:
I support this change, but I worry that wallets might present `@.value` as the total or net native-currency value transfer for a tx, which it is not. Execution can still transfer ETH from contract balances, including balances controlled by the user via a smart account, delegated EOA, or multisig, even when the top-level `@.value` is zero.

I propose adding something to the effect of:
> Displaying `@.value` only discloses the native currency attached to the transaction. It does not disclose native-currency transfers performed during execution, including transfers from smart accounts, delegated EOAs, or multisig accounts. Wallets MUST NOT present `@.value` alone as the transaction's total or net native-currency value transfer.

forshtat (commented):
@PrisionMike @soyrubio agreed to re-review the PR during the call

kuzdogan:
> Execution can still transfer ETH from contract balances, including balances controlled by the user via a smart account, delegated EOA, or multisig, even when the top-level @.value is zero.

Shouldn't these then be part of the descriptor of the function being called? My understanding is this proposal is only about the `@.value`

shukudaidayo:
Yes, when an internal transfer can be calculated from calldata, the descriptor should present it. However, this isn't always possible under ERC-7730 as it stands today.

For example, the registry’s [Feral File English Auction descriptor](https://github.com/ethereum/clear-signing-erc7730-registry/blob/master/registry/feral-file/calldata-feralfile-english-auction-0.json) formats `settleAuctionFund(uint256,address)`. The function is nonpayable, but it can transfer `highestBids[auctionID].amount` of ETH held by the auction contract to the supplied vault. Depending on stored bid state, it can also transfer nothing. There's not enough information in the calldata for the descriptor to determine whether a transfer will occur, or to calculate how much ETH will be transferred.

Transaction simulation can generally provide a better view of transaction effects, but even then, the estimate can become stale if the relevant state changes before the tx is included.

I agree that internal transfers are out of scope for this PR, but still feel it's worth clarifying that `@.value` represents only the native currency attached to the transaction, not all native-currency transfers that can occur during execution. 

See the attached wallet mockups for an example of a "bad" and "good" presentation of `@.value`:

feral-file-value-comparison

PrisionMike:
Another counter-example where we would invalidate the descriptor, but it kinda does the job, is: 1inch AggregationRouter V6. Where there are no paths directing to `@.value` but the swap is payable.

They encode the maximum ETH value in the calldata itself, then use the `nativecurrencyaddress` and the sentinel address to denote the same amount. The contract logic ensures that sufficient ETH exists in `msg.value` .

If we force the show, the wallets would be showing the same amount twice, which would make for a slightly confusing UI.

As much as I wanted this before, I would want us to not enforce this validation anymore :)

PR #1979, opened 2026-08-29
Add ERC: Account Authority Lifecycle

AnkushinDaniil:
## Add ERC: Frame Account Authority Lifecycle

Defines a common lifecycle for the authorities that authorize an EIP-8141 frame-transaction account: how an authority (a key, or an authorization predicate such as a lock or a limit) is added, rotated, and revoked. An authority is a leaf in a per-account committed tree whose root is published as an EIP-8272 recent root. A transaction carries a membership proof and names the root it proves against; the account verifies it in a frame that runs the account's own code and withholds approval on failure.

The standard adds no transaction type, opcode, or consensus rule. It is a convention over EIP-8141 and EIP-8272 primitives, so its category is ERC.

Two conformance tiers make the revocation-latency trade-off explicit: a root-bound tier with no mutable reads (mempool-cheap, revocation bounded by the EIP-8272 recency window), and a floor-bound tier with one mutable read (next-block revocation). Both keep invalidation account-bounded.

Preflight: `eipw --config config/eipw.toml` passes, `codespell` passes.

Open item: the `discussions-to` Ethereum Magicians thread is not yet created; the field is a placeholder and will be updated. Number to be set to the PR number.


ERCS/erc-1979.md
```diff
@@ -0,0 +1,110 @@
+---
+eip: 1979
```
abcoathup:
```suggestion
eip: 8403
```

ERCS/erc-8403.md
```diff
@@ -0,0 +1,110 @@
+---
+eip: 1979
+title: Frame Account Authority Lifecycle
+description: A committed-root add, rotate, and revoke model for frame-transaction account authorities, checked in the account's own frame
+author: Daniil Ankushin (@AnkushinDaniil)
+discussions-to: https://ethereum-magicians.org/t/erc-frame-account-authority-lifecycle/24001
```
abcoathup: Please create a discussions topic

AnkushinDaniil:
A visual companion to the spec, since the prose is dense. First how the core is built on the existing primitives (EIP-8141 and EIP-8130), then worked examples showing that features which look unrelated are the same frame shape — `VERIFY: prove membership against a named root → APPROVE → act` — where only the leaf/predicate changes. Illustrative, not normative.

## How it's built on the primitives

**1. EIP-8141 binding — data flow.** The authorization step is a `VERIFY` frame running the account's own code; the proof comes in via `FRAMEDATALOAD`, the named root via `RECENTROOTREFLOAD` (checked against pre-state before any code runs, so the frame reads no mutable external storage — Tier 1).

```mermaid
flowchart TD
  d["frame data: authority_id, predicate, merkle_path"] -->|"FRAMEDATALOAD (0xB1)"| vf["VERIFY frame (mode 1): runs account code"]
  rr["recent-root reference: root_id under account source_id"] -->|"RECENTROOTREFLOAD (0xB6), checked vs pre-state"| vf
  vf --> chk["account code: leaf = keccak256(authority_id, predicate), verify path to root"]
  chk -->|valid| ap["APPROVE (0xAA)"]
  chk -->|invalid| halt["exceptional halt: fail closed"]
  ap --> sf["SENDER frame (mode 2): transfer / execute"]
```

**2. EIP-8130 binding — data flow.** The datum and its lifecycle already exist (`actorId`, `actor_config`, authorize/revoke batches). The one piece marked `[PROPOSED]` is the published current-state root — the family does not ship it yet (EIP-8387 names a compact "proof of chain" as unspecified), which is why this binding is informative. Until then, 8130 validates by reading `actor_config` live.

```mermaid
flowchart TD
  b["SignedAccountChanges batch: AuthorizeActor / RevokeActor (channel, sequence)"] -->|applySignedAccountChanges| cfg["actor_config slot: authenticator(20) / expiry(6) / scope(2) / reserved(4)"]
  cfg -->|"derive leaf = keccak256(actorId, actor_config), sorted by actorId  [PROPOSED]"| root["current-state authority root"]
  root -->|"publish as EIP-8272 recent root under keystore source_id  [PROPOSED]"| pub["referenceable root"]
  p["frame tx carries (actorId, actor_config, merkle_path, root_id)"] --> auth["authenticator + scope check: verify path vs published root"]
  pub --> auth
  auth -->|"valid and block.timestamp <= expiry"| ok["authorized"]
  auth -->|else| no["fail closed"]
```

**3. Primitive mapping.** Same core concept, two realizations:

| core concept | EIP-8141 | EIP-8130 |
|---|---|---|
| `authority_id` | frame data field | `actorId` (32B) |
| `predicate` | frame data (pubkey / authenticator / limit) | `actor_config`: authenticator ‖ expiry ‖ scope |
| add / rotate / revoke | publish a new root (account is sole writer) | `AuthorizeActor` / `RevokeActor` batch |
| root publication | EIP-8272 recent root under account `source_id` | current-state root over sorted actor leaves *(proposed)* |
| proof carriage | `FRAMEDATALOAD` (0xB1) | tx-carried `(actorId, actor_config, path)` |
| root read | `RECENTROOTREFLOAD` (0xB6), pre-state checked | same 8272 reference once published |
| authorization step | `VERIFY` frame (mode 1), self-`APPROVE` (0xAA) | authenticator + scope check |
| fail-closed | withhold `APPROVE` / halt | reject |

## Worked examples

The frame shape above is reused verbatim below; only the leaf/predicate box changes.

**A. Post-quantum key migration.** `rotate` swaps the predicate in the same authority slot; nothing else in the account changes. A PQ scheme is just another predicate.

```mermaid
flowchart TD
  subgraph before["authority root (before)"]
    L1["authority_id = owner / predicate = ECDSA pubkey"]
  end
  subgraph after["authority root (after rotate)"]
    L2["authority_id = owner / predicate = ML-DSA pubkey"]
  end
  before -->|"rotate: same slot, new predicate"| after
  L2 --> tx["later tx: VERIFY proof of ML-DSA leaf vs root, then APPROVE"]
```

Signature *size* (PQ aggregation) is a separate SIGNATURE-frame line (EIP-8288), not this standard.

**B. Gas sponsorship.** Two authorities, two roots, one tx — and the sponsor is proven, not read live.

```mermaid
flowchart TD
  tx["frame tx"]
  tx --> sp["proof: sender leaf vs sender's root"]
  tx --> pp["proof: sponsor leaf vs sponsor's OWN root, carried in tx"]
  sp --> sv["VERIFY sender, then APPROVE_EXECUTION"]
  pp --> pv["VERIFY payer, then APPROVE_PAYMENT (no live sponsor SLOAD)"]
```

The ERC-20-pays-gas mechanics are the 8141 paymaster frame; this standard covers only the authorization half.

**C. Private-pool withdrawal.** The authority is a ZK membership proof; the leaf is hidden; revocation is a nullifier. Same frame shape as a key auth — only the predicate differs.

```mermaid
sequenceDiagram
  participant W as Withdrawer tx
  participant V as VERIFY frame (pool code)
  participant R as pool recent root (unspent notes)
  W->>V: zk_proof(note in root, unspent) + nullifier
  V->>R: check membership vs named root
  V->>V: check nullifier unused (spend once)
  V-->>W: APPROVE, then SENDER withdraws and pays its own fee
```

Deposit adds a note leaf → new pool root. Withdrawal proves membership + nullifier. The frame is identical to a signature auth; only the predicate changed.

**D. Recovery after a lost wallet.** The authority set is on-chain state, not a wallet's private record, so a wipe is a re-read.

```mermaid
flowchart TD
  lost["wallet cache lost: uninstall / vendor gone / device wiped"]
  lost --> read["read current authority leaves from ANY node (cold path)"]
  read --> rebuild["rebuild canonical tree, sorted by authority_id"]
  rebuild --> check{"recomputed root == on-chain root?"}
  check -->|yes| done["proofs reconstructed: no backup, no event replay"]
  hot["hot path: tx carries proof, reads NO tree"] -. stateless .-> done
```

**All examples reduce to one shape.** Only the leaf/predicate box differs.

```mermaid
flowchart LR
  a["ECDSA or PQ key leaf"] --> core
  b["sponsor leaf (own root)"] --> core
  c["ZK note + nullifier"] --> core
  core["VERIFY: prove membership vs named root, then APPROVE"] --> act["SENDER: transfer / execute / withdraw"]
```

Provenance: opcodes and frame modes are EIP-8141; `RECENTROOTREFLOAD` / recent roots / `source_id` are EIP-8272; `actorId` / `actor_config` / authorize-revoke are EIP-8130; PQ aggregation is EIP-8288; the current-state root derivation is the binding this standard proposes.

AnkushinDaniil:
Updated the informative EIP-8130 binding and its security note.

EIP-8130 has a proposed optional checkpointed mode that commits the current configuration as an image hash, with actors supplied as authenticated witnesses proven against it (see https://github.com/ethereum/EIPs/pull/12293). That is a referenceable committed root in this standard's sense, so the binding now records it as a Tier 2 realization (read from settled keystore state) and marks a no-mutable-read Tier 1 admission for keystore accounts as still open, instead of saying the keystore model publishes no current-state root. Also replaced informal wording with direct model names.

Discussion: https://ethereum-magicians.org/t/erc-8403-account-authority-lifecycle/29570

AnkushinDaniil:
Added a **Multichain authority** subsection to the Specification, plus two Security Considerations that replace the old "cross-chain rotation is not atomic / not resolved here" note.

The point is that multichain is not a new mechanism: the authorization check is already a membership proof against a named committed root, and it does not require that root to originate on the executing chain. An account that wants one authority set across chains publishes its root on a single home chain; each executing chain reads it through a cross-domain root-delivery primitive it already imports (for example the beacon block root of EIP-4788, an L1-storage-read precompile, or a bridged state root). A transaction then carries two proofs, leaf-membership under the authority root and that this root is the committed value at the account's home slot, both checked against a root the validator already holds. So there is no live cross-domain read per transaction, the two tiers / fail-closed / isolation carry over verbatim, and cross-chain revocation latency is just the imported root's freshness window under the same root-bound vs floor-bound knob.

Kept honest: this standard does not define or claim the delivery primitive, and requires a finalized home root so an unfinalized reorg cannot revert a rotation. On a chain with no such primitive an account falls back to a same-chain committed root or a replayed change chain.

AnkushinDaniil:
https://excalidraw.com/#json=qWinGdaf528ErnqnyjIdZ,VJOnH8q7nKcG4NOMx-_hQA

ERCS/erc-8403.md
```diff
@@ -0,0 +1,158 @@
...
+
+- **Tier 2 (floor-bound, instant revocation).** In addition to Tier 1, the check reads one account-owned monotonic value (a rotation-floor slot) with a single storage read and rejects any reference whose slot is below the floor. Raising the floor on rotate or revoke makes the change effective at the next block, at the cost of the mempool-cheap admission tier. A single-authority account MAY instead read its current authorized root directly, which is the degenerate case of the same rule.
+
+The account MUST fail closed: if no accepted authorization is proven, the authorization step MUST NOT authorize execution, and the transaction is invalid.
+
+### Isolation
+
+Authorization MUST be keyed by the account: an authority proven for one account MUST NOT authorize another. Because each account proves against its own subtree under its own root source, revoking an authority affects only that account and does not invalidate authorities held by other accounts.
```
chugarchugarr, line 79: This keeps the core model-neutral while defining the optional property wallets actually need.

```suggestion
Authorization MUST be keyed by the account: an authority proven for one account MUST NOT authorize another. Because each account proves against its own subtree under its own root source, revoking an authority affects only that account and does not invalidate authorities held by other accounts.

### Portable recovery

Portable cross-wallet recovery is an optional property of an account's authority tree. An account that claims portable recovery MUST make both the current authority records and the reconstruction scheme discoverable from account-controlled on-chain data without relying on wallet-local state. The recovery scheme MUST define how a wallet discovers those records and MUST deterministically specify the canonical byte encoding of each `(authority_id, predicate)`, authority ordering, tree arity and shape, padding and empty-tree rules, internal-node hashing and domain separation, and a canonical membership-proof encoding.

Given the same current authority records and the same recovery scheme, two conforming implementations MUST derive the same committed root. If they do not, the account does not satisfy portable recovery. The meaning of the recovery scheme for a published root MUST remain stable for as long as that root is referenceable.

A recovery scheme MAY be account-specific. Standard recovery schemes MAY be defined for cross-wallet interoperability; using an account-specific scheme preserves conformance to the core but makes recovery portable only to wallets that implement that scheme.
```

ERCS/erc-8403.md
```diff
@@ -0,0 +1,158 @@
...
+
+An account SHOULD keep its authority tree in enumerable on-chain state, with the published root a deterministic function of that state. Recovery is then a read of current state from any node, followed by rebuilding the canonical tree and checking its root against the published root; a wallet's local copy is a cache, and losing it is a re-read rather than a loss. This inherits the availability of state, which every full node retains, and needs no separate backup, event replay, or checkpoint. Reconstruction is unambiguous because the leaf encoding and tree shape are deterministic.
+
+This availability is separate from the authorization check's cost tier. The check stays stateless on the hot path, a carried proof against a named root with no state read, while recovery is an infrequent, account-initiated read off that path. An account MAY instead publish only the root and keep the leaves off-chain, for privacy or to minimize state, in which case it is responsible for their availability and accepts that a lost witness cannot be rebuilt from chain data; where an on-chain record is still wanted, leaf preimages can be emitted as logs so the set can be reconstructed from history.
```
chugarchugarr, line 190: Enumerable state guarantees availability, but it does not by itself guarantee that a foreign wallet can reconstruct an account-defined tree. This makes the rationale match the core and the recovery diagram precisely.

```suggestion
### Authority-set availability and recovery

Because an authority is proven by a membership proof, a wallet must be able to obtain the current leaf set to build one. A committed root is a hiding commitment: it verifies a proof but cannot by itself reconstruct the leaves, so the leaf data must be available from a source that outlives any single wallet installation.

An account SHOULD keep its authority tree in enumerable on-chain state, with the published root a deterministic function of that state. This makes the authority records recoverable from any node, but enumerable records alone do not determine an account-defined tree. An account that implements portable recovery additionally exposes the recovery scheme defined above; a recovering wallet reads the current authority records and scheme, deterministically reconstructs the tree, and accepts the reconstruction only if the recomputed root equals the published root. The wallet's local copy is then a cache, and losing it is a re-read rather than a loss.

This availability and reconstruction path is separate from the authorization check's cost tier. The check stays stateless on the hot path, a carried proof against a named root with no tree-state read, while recovery is an infrequent, account-initiated read off that path. An account MAY instead publish only the root and keep the leaves off-chain, for privacy or to minimize state, in which case it is responsible for their availability and accepts that a lost witness cannot be rebuilt from chain data; where an on-chain record is still wanted, leaf preimages can be emitted as logs so the set can be reconstructed from history. An account that does not expose a portable recovery scheme may still keep enumerable authority records, but a foreign wallet cannot assume those records are sufficient to reproduce the committed root.
```

chugarchugarr (commented):
The recovery question can be resolved without choosing one global tree shape. The missing invariant is deterministic reconstruction: enumerable leaves solve availability, while portable recovery additionally requires a discoverable reconstruction scheme whose semantics are stable for the root being interpreted. These suggestions preserve the account-defined-tree core and make the lost-wallet claim mechanically testable.

ERCS/erc-8403.md
```diff
@@ -0,0 +1,158 @@
...
+
+- **Authority-set availability.** A committed root cannot reconstruct its own leaves. An account that publishes only the root and keeps leaves off-chain MUST ensure their availability, or a lost witness locks the account out of proving membership even though the root stays valid; keeping the tree in enumerable on-chain state removes this dependency. Reading on-chain leaf data reveals the account's authority structure but confers no authorization: membership is public, while authorizing still requires a valid signature by the acting authority, so this is a confidentiality trade-off, not an authorization weakness.
```
chugarchugarr, line 212: The security consideration should distinguish data availability from deterministic reconstruction.

```suggestion
- **Authority-set availability and reconstruction.** A committed root cannot reconstruct its own leaves. An account that publishes only the root and keeps leaves off-chain MUST ensure their availability, or a lost witness locks the account out of proving membership even though the root stays valid. Keeping the authority records in enumerable on-chain state removes that availability dependency, but it does not by itself make an account-defined tree reconstructible by another wallet; portable recovery additionally requires the discoverable deterministic recovery scheme defined above. Reading on-chain authority data reveals the account's authority structure but confers no authorization: membership is public, while authorizing still requires satisfying the acting authority's predicate, so this is a confidentiality trade-off, not an authorization weakness.
```

AnkushinDaniil:
Visual for the two things the latest push adds (both informative in the spec).

**One transaction, one check.** The account is a committed root. The transaction carries a membership proof (public, binds the authority to the account) and a witness (secret, binds the action to the authority). The verifier is resolved from the proven leaf, never declared by the sender, so it cannot be swapped for an always-accept one.

```mermaid
flowchart TD
  R["account = committed root R"]
  L["proven leaf: authority_id, verifier, predicate"]
  TX["tx carries: authority_id, root_ref, membership_proof, witness"]
  M["membership proof: leaf is under R (keyless, public)"]
  W["witness: satisfies the predicate over the tx digest (secret)"]
  V["verifier resolved from the proven leaf, not from the tx"]
  C{"authorization step: proof AND witness pass"}
  OK["approve"]
  NO["withhold, fail-closed"]
  TX --> M --> C
  TX --> W --> C
  R --> L --> V --> C
  C -->|yes| OK
  C -->|no| NO
```

**What a node prices on is the verifier kind.** Cost tracks how much mutable state the check reads, not how arbitrary the code is: a succinct verifier is arbitrary code but a cheap class, an arbitrary contract that reads state is the costly one. The four kinds line up with the EIP-8130 authenticator classes; the field shape is a sketch for convergence, not an agreed wire format.

```mermaid
flowchart LR
  VF["leaf verifier: kind then ref"]
  VF --> K1["CANONICAL: ECDSA / P256 / keystore"]
  VF --> K2["CIRCUIT: succinct, reads 0 mutable state"]
  VF --> K3["BOUNDED: one fixed slot pattern"]
  VF --> K4["CONTRACT: arbitrary code, arbitrary state"]
  K1 --> CHEAP["priced from table, mempool Tier 1"]
  K2 --> CHEAP
  K3 --> T2["Tier 2: one monotone read"]
  K4 --> SIM["simulate under the verify-gas cap, else private inclusion"]
```

The same resolve step unifies the three ways to run one account model on another's transaction (header delegation, proxy selector, frame reinterpretation): each picks which verifier the root selects, and forcing a single verifier is the degenerate case a general resolve contains.

Full lifecycle visual, editable: https://excalidraw.com/#json=3xX1urbDXKw1m_wMpYO8q,Fgs-LwvVbWRhrU3C45ZdZA

AnkushinDaniil:
The last push makes explicit what the prose implied: **where the committed root is referenced is a per-tier choice, not a fixed external publication.** The root, the membership proof, and the witness are the same in both tiers; only the root's source differs, and `root_ref` is what picks it.

```mermaid
flowchart TD
  R["account commits its authority tree to one root R"]
  R --> T1["Tier 1 root-bound: name R as an EIP-8272 recent root, read with RECENTROOTREFLOAD, checked vs pre-state"]
  R --> T2["Tier 2 floor-bound: hold R in an account-owned slot, read it directly in the VERIFY frame"]
  T1 --> C1["mempool-cheap; revocation bounded by the recency window; depends on a live root source"]
  T2 --> C2["one mutable read; next-block revocation; no recency window and no external publisher"]
  T1 --> PUB["publish R to an external mechanism only when a reader that cannot see account state must verify: another chain, a keystore peer, a validity-proof verifier"]
  T2 --> PUB
```

The transaction stays `(authority_id, root_ref, membership_proof, witness)` in both tiers. A Tier-2 account that only ever validates at home reads its root from its own slot and publishes nothing.

**The enumerable tree on-chain is a recovery choice, not part of the check.** Neither tier reads the leaves at authorization time; a wallet reads them only to rebuild proofs after losing its local cache.

```mermaid
flowchart LR
  HOT["hot path, either tier: carry the proof, verify against R, read no leaves"]
  LEAVES["enumerable leaves in on-chain state"]
  REC["cold recovery path: read leaves from any node, rebuild the canonical tree, check its root equals R"]
  LEAVES -. only consumed by .-> REC
  HOT -. never reads .-> LEAVES
```

PR #1989, opened 2026-09-03
Update ERC-8354: bind expiry as a public input in the reference circuit

Echo-Merlini:
## What

Binds `expiry` as a public input in the ERC-8354 reference circuit (`assets/erc-8354/circuits`).

The reference circuit was still the pre-fix shape: `expiry` — a `Verdict` field the on-chain Guard checks — was **not** a public input, so a proof was not bound to it and the *same proof verified under any `expiry`* the submitter chose. The ERC's own Rationale requires every `Verdict` field to be a public input.

This mirrors the merged deployment-side fix in the reference implementation — `zexoverz/confidential-agent-policy-verdicts#4` ("bind expiry as a public input on all three programs"). Flagged by @zexoverz in the working group; this brings the ERC reference asset in line with it.

## Changes

- **`src/main.nr`** — add `expiry: pub Field`, appended last so no existing public-input index moves, plus `let _ = expiry;`. It is deliberately *not* constrained in-circuit: the Guard performs the time check and the circuit has no notion of block time. Making it public means the **proof commits to it** (bound by the verifier), which is the whole requirement — proving at one expiry and verifying at another fails in the reduction step. `test_allow_verdict` updated to pass it.
- **`Prover.example.toml`** — add `expiry`, and add the previously-missing `policy_kind` (a required public input the example omitted, so it did not execute as published).

## Verification (nargo 1.0.0-beta.26)

- Baseline (unmodified): `nargo check` clean, `test_allow_verdict` passes.
- Modified: `nargo check` clean, all 4 tests pass.
- Completed example: `nargo execute` → *"Circuit witness successfully solved."*

The binding is by construction — public inputs are bound by the verifier; the deployment repo's #4 additionally proves it at the verifier level (`ExpiryBinding.t.sol`).

### DIFF
--- a/assets/erc-8354/circuits/Prover.example.toml
+++ b/assets/erc-8354/circuits/Prover.example.toml
@@ -6,7 +6,9 @@ policy_root = "0x053d4542d140ad2350a0ee79fae4a522821274e428bd881e7e803ecd816635a
 action_commitment = [124, 203, 122, 78, 157, 81, 18, 139, 149, 28, 190, 221, 239, 174, 193, 20, 1, 128, 163, 209, 63, 110, 174, 111, 6, 89, 109, 196, 50, 5, 124, 250]
 nullifier = "0x041271fcaf479f6ab927df3a03f74d3809e9f49d880cd7a9595c8dc0a58a5e03"
 decision = "1"
+policy_kind = "0"   # 0 = ALLOWED (was missing from this example — required public input)
 executor = "0xE0"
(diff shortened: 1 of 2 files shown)


zexoverz (approved):
Approving. `expiry` goes on the end so no public-input index moves, same shape as the CAPV fix, and the Guard already passes it inside `abi.encode(v)` so nothing on the Solidity side changes.

PR #1991, opened 2026-09-03
Update ERC-7730: allow threshold formatting for amount formatters

PrisionMike:


kuzdogan:
Doesn't this require a change in the spec as well?

PrisionMike:
@kuzdogan it does. Do I need to do something different for it? It is backwards compatible so existing formatters don't have to be updated and even CI tests don't need to change, I think.

forshtat (commented):
@PrisionMike please also update the schema file
This opens the v2.1.0 in my opinion

forshtat (commented):
LGTM
SO, @kuzdogan once merged this becomes the v2.1.0 version immediately? This is the process?

kuzdogan:
@forshtat according to the spec yes. It's one feature - one version. The positive I can think is that it's easier to add tests for each feature and keep track. What woudl be the downside? Any thoughts?

JIOjosBG:
will this also handle cases where the amount is also relevant to other parts of the intent or interpolated intent, like when an approval's value is 0, it makes more sense for the action to be `Revoke approval` instead of `Approve` and then a 0 value?

PrisionMike:
@JIOjosBG good suggestion. Unfortunately that would be tricky for this version of the spec. We don't have cross-field references yet. The `label` of a field being conditional on a value of the field or others.
I would say that could be complicated for wallets to implement in the current state of the parser, with `intent` or `label` becoming "non-static" fields.

But I'm open to suggestions on how we could implement something like that. It would certainly be nice. :)

kuzdogan:
@JIOjosBG I think it's akin to this draft, right? https://github.com/ethereum/ERCs/pull/2027

Currently it's written just for the date formatter but maybe we can generalize this conditioning

JIOjosBG:
I am not deeply familiar with the eip, but those are the thing that make sense to me as a wallet dev
 
@PrisionMike 
> But I'm open to suggestions on how we could implement something like that. It would certainly be nice. :)

```ts
{
 "intent": [text:'Revoke approval', requirements:[ {type:'eq',val:"0",ref:'#.amount' } ]]
  ...
}
```
Broken down, each decision is:

```ts
{intent: string | T[] }
T = { text:string, requirements: M[] }
```
where the used text will be picked based on a first match basis 
```ts
usedIntent = intent.find( d=>d.requirements.every(isSatisfied))?.text
```
I picked T[] instead of Record<string,T> for explicit ordering
This allows for the same text to be used in different cases, not sure if there are major cases for needing this
Implicitly this structure will allow you to have a fallback text when requirements is `[]`

```ts
M = {type:'eq'|'neq'|'lte'|'gte'|'gt'|'lt'|'in','not_in', value: Reference|string|number|array, ref:ref_to_val}
```
Allowing the type comparison to be `eq` solves the Approval. 
Having requirements be M[] allows for more complex rules like a `transferFrom` function can be `Transfer USDC to 0xA`, `Move USDC from 0xA to 0xB`, `Take USDC from 0xA`, `Burn USDC`, `Burn USDC from 0xA`
where 0xA and 0xB can be owner of tokens or reciever, while the current user is just approved to manage them. For this to be possible value has to be reference in some cases to compare the executor and `from` or `to`

The above proposal could be be verbatim used to interpolated intent or label

tbh this is complex and will add debt, BUT most wallets will either use one of few major implementations or vibe-code a internal solution as the whole spec is designed in a way to be easily testable a verifiable

JIOjosBG:
>@JIOjosBG I think it's akin to this draft, right? https://github.com/ethereum/ERCs/pull/2027

yes, both this and your proposal seem overlapping.
I did not understand initially, this has nothing to do with what i outlined above

PrisionMike:
@JIOjosBG agreed with the relevance of it. Would you like to open it as an issue here: https://github.com/ethereum/clear-signing-erc7730-registry/tree/master so we can discuss it more? I am thinking if we would be accepting of such robust validation, we could be using a lot of them in the fields. Right now all we have is `mustMatch` and not many authors using it. It would be nice to grow it more and wallets would certainly appreciate a stronger validation logic.

PrisionMike:
@forshtat @kuzdogan this could actually be `v2.0.1` as we are not 'growing the set' here really. Wdyt?
Ref. : https://github.com/ethereum/ERCs/pull/2018

PR #1992, opened 2026-09-04
Add ERC-8410: Portable Execution Plan Artifact

moodysalem:
Adds ERC-8410, a JSON document carrying an ordered call sequence for one sender on one chain from a keyless preparer to the wallet that holds the key, plus a keccak256 plan digest over exactly the broadcast-determining fields, plus a reference envelope for handing the plan across an untrusted relay.

The digest is the substance of the proposal. Because it covers what will be broadcast and the labels shown beside it — and excludes gas, extension bags, and producer advice — a human approval, a policy decision, and a later audit record can all name the same bytes, and a producer cannot change what executes after approval without changing the digest.

**Relationship to EIP-5792.** That interface carries a batch of calls too, and this borrows its capability negotiation. The difference is lifetime: an interface call lives between two live endpoints, while an artifact is a value that can be stored, forwarded, queued, reviewed on another device, and compared against what was broadcast later. A normative mapping to and from `wallet_sendCalls` parameters is included so the two compose rather than compete.

**Delivery by reference.** Plans are increasingly relayed by the least trustworthy link in the chain — often an AI agent acting on untrusted text. The proposal defines an envelope naming a stored plan by URL and keccak256 digest, and specifies the fetch admission rules (public unicast only, no redirects, no credentials, bounded bodies, no response echo) because a wallet that fetches caller-supplied URLs is otherwise a request forwarder inside the user's network.

**Included**
- Normative canonicalization for the digest, with an explicit warning that the member order is *not* lexicographic and RFC 8785 would produce different bytes.
- `assets/erc-8410/execution-plan.schema.json` and `assets/erc-8410/artifact-reference.schema.json`, validated with ajv.
- `assets/erc-8410/digest-test-vectors.json` — four cases, one cross-checked between independent Rust and JavaScript implementations, one changing only excluded fields (same digest), one changing a single calldata byte (different digest).

**Reference implementation.** Ekubo Wallet (`github.com/EkuboProtocol/wallet`), where this format is the only route from a plan producer to a broadcast transaction. Named in prose rather than linked, per the EIP-1 external-link rule.

**Notes for editors**
- Three companion drafts were opened alongside this and have since been withdrawn (#1993, #1994, #1995); their useful content is either folded in here or belongs in project documentation rather than a standard.

CI checks run locally before opening: eipw (via the pinned `eipw-lint-js`, against the same merged EIP+ERC corpus the workflow builds), markdownlint with `config/.markdownlint.yaml`, codespell, and markdown-link-check.


ERCS/erc-8409.md
```diff
@@ -0,0 +1,665 @@
+---
+eip: 8409
```
abcoathup:
```suggestion
eip: 8410
```
Numbers are assigned by editors & associates.  Please don't attempt to self assign.

ERCS/erc-8409.md
```diff
@@ -0,0 +1,665 @@
...
+title: Portable Execution Plan Artifact
+description: A digest-bound JSON document carrying an ordered call sequence from an untrusted preparer to the wallet that holds the key
+author: Moody Salem (@moodysalem)
+discussions-to: https://ethereum-magicians.org/t/erc-8409-portable-execution-plan-artifact/0
```
abcoathup: Please create an actual discussions topic on Eth Magicians.

ERCS/erc-8409.md
moodysalem: Done — renumbered to 8410: preamble, filename, asset directory, schema $id values, and internal links all updated in be43d74…ad3817bf. Apologies for self-assigning.

ERCS/erc-8409.md
moodysalem: Done — https://ethereum-magicians.org/t/erc-8410-portable-execution-plan-artifact/29587

moodysalem:
Note on the red HTMLProofer check: it is a stale-cache artifact of the renumber, not a link in the proposal. The job restores `jekyll-Linux-eips-9207c601` from the earlier run that built `erc-8409`, and the incremental Jekyll build leaves the cached index pages (`all.html`, `erc.html`) still listing `/EIPS/eip-8409`, which no longer exists. The renamed `erc-8410` page itself passes. Because the cache is re-saved on every run under the same key, it will clear when the EIPs checkout advances or if someone with access evicts that cache entry.

eraandriyanerandriyan-blip:
> Adds ERC-8410, a JSON document carrying an ordered call sequence for one sender on one chain from a keyless preparer to the wallet that holds the key, plus a keccak256 plan digest over exactly the broadcast-determining fields, plus a reference envelope for handing the plan across an untrusted relay.
> 
> The digest is the substance of the proposal. Because it covers what will be broadcast and the labels shown beside it — and excludes gas, extension bags, and producer advice — a human approval, a policy decision, and a later audit record can all name the same bytes, and a producer cannot change what executes after approval without changing the digest.
> 
> **Relationship to EIP-5792.** That interface carries a batch of calls too, and this borrows its capability negotiation. The difference is lifetime: an interface call lives between two live endpoints, while an artifact is a value that can be stored, forwarded, queued, reviewed on another device, and compared against what was broadcast later. A normative mapping to and from `wallet_sendCalls` parameters is included so the two compose rather than compete.
> 
> **Delivery by reference.** Plans are increasingly relayed by the least trustworthy link in the chain — often an AI agent acting on untrusted text. The proposal defines an envelope naming a stored plan by URL and keccak256 digest, and specifies the fetch admission rules (public unicast only, no redirects, no credentials, bounded bodies, no response echo) because a wallet that fetches caller-supplied URLs is otherwise a request forwarder inside the user's network.
> 
> **Included**
> 
> * Normative canonicalization for the digest, with an explicit warning that the member order is _not_ lexicographic and RFC 8785 would produce different bytes.
> * `assets/erc-8410/execution-plan.schema.json` and `assets/erc-8410/artifact-reference.schema.json`, validated with ajv.
> * `assets/erc-8410/digest-test-vectors.json` — four cases, one cross-checked between independent Rust and JavaScript implementations, one changing only excluded fields (same digest), one changing a single calldata byte (different digest).
> 
> **Reference implementation.** Ekubo Wallet (`github.com/EkuboProtocol/wallet`), where this format is the only route from a plan producer to a broadcast transaction. Named in prose rather than linked, per the EIP-1 external-link rule.
> 
> **Notes for editors**
> 
> * `discussions-to` is a format-valid placeholder; the Ethereum Magicians thread is not yet created and I will update this before requesting review.
> * The number was taken as the next free one across both repositories at the time of writing; happy to renumber.
> * Three companion drafts were opened alongside this and have since been withdrawn ([Add ERC: Integrity Verified Artifact References #1993](https://github.com/ethereum/ERCs/pull/1993), [Add ERC: Declarative Wallet Signing Policy #1994](https://github.com/ethereum/ERCs/pull/1994), [Add ERC: Agent Wallet Execution Interface #1995](https://github.com/ethereum/ERCs/pull/1995)); their useful content is either folded in here or belongs in project documentation rather than a standard.
> 
> CI checks run locally before opening: eipw (via the pinned `eipw-lint-js`, against the same merged EIP+ERC corpus the workflow builds), markdownlint with `config/.markdownlint.yaml`, codespell, and markdown-link-check.

eraandriyanerandriyan@gmail.com

moodysalem:
@abcoathup Can you please review again? AFAICT the CI issue is a bug in the HTMLProofer check (using a stale cache that used EIP-8409)

moodysalem:
Note on the red HTMLProofer check: it is a stale-cache artifact of the renumber, not the proposal content. The job restores the incremental Jekyll `_site` cache built when this branch was still numbered 8409, so the regenerated index pages (`all.html`, `erc.html`) still link to `/EIPS/eip-8409`, which no longer exists. All other checks are green.

Mitigation needs a maintainer: evict the `jekyll-Linux-*` Actions caches on this repo (`gh cache delete --repo ethereum/ERCs` on the stale keys). The next run will do a clean build and pass. Nothing in the PR branch itself can clear it.

lightclient (approved):
LGTM

PR #1993, opened 2026-09-04
Add ERC: Integrity Verified Artifact References

moodysalem:
Adds ERC-8410, an envelope that names a wallet-bound payload by location and keccak256 digest instead of carrying it, so that a relaying party holds authority over *which* document is used and none at all over *what it says*.

The motivating case is a payload assembled by one party, relayed by a second, and executed by a third, where the middle link is the least trustworthy — a browser extension, an automation runner, or an AI agent acting on text from an untrusted source. Today that link is handed the full bytes and asked to forward them faithfully. It can rewrite them, it can corrupt them by re-serializing rather than copying (a failure mode that is disproportionately common when the relay is a language model), and it destroys provenance.

**Why not Subresource Integrity.** SRI is SHA-2 where the rest of this ecosystem is keccak256, has no type discriminator, has no inline form, and is defined as an HTML attribute rather than a relayable value. Each of those gaps matters here.

**The retrieval rules are half the proposal.** A wallet that fetches a caller-supplied URL is a request forwarder positioned inside the user's network. The spec fixes admission — public unicast addresses only, no redirects, no credentials, no non-default ports, no fragments, bounded bodies enforced during read, bounded concurrency, and errors that never echo the response body — because specifying the envelope alone would standardize the easy half.

**Included**
- `assets/erc-8410/artifact-reference.schema.json`, validated with ajv.
- A deliberate asymmetry, explained in the Rationale: the envelope tolerates unknown members so producers can enrich it, while the `integrity` block rejects them, because an integrity object containing something unverified is not an integrity object.

**Reference implementation.** Ekubo Wallet (`github.com/EkuboProtocol/wallet`), `crates/ekubo-wallet-core/src/plan_fetch.rs`. Named in prose rather than linked, per the EIP-1 external-link rule.

**Notes for editors**
- Companion proposals are open alongside this one and are referred to in prose rather than by number, since `markdown-refs` cannot resolve an unmerged sibling. Numbers and `requires` entries will follow once assigned.

### DIFF
--- a/ERCS/erc-8410.md
+++ b/ERCS/erc-8410.md
@@ -1,0 +1,343 @@
+---
+eip: 8410
+title: Integrity Verified Artifact References
+description: An envelope naming a wallet payload by URL and keccak256 digest, so a relaying party can pass it along without being trusted with its bytes
+author: Moody Salem (@moodysalem)
+discussions-to: https://ethereum-magicians.org/t/erc-8410-integrity-verified-artifact-references/0
+status: Draft
+type: Standards Track
+category: ERC
+created: 2026-09-04
+---
+
+## Abstract
+
+This proposal defines the *artifact reference*: a small JSON envelope that names
+a wallet-bound payload by its location and its keccak256 digest, rather than
+carrying the payload itself.
+
+A producer stores a document — an execution plan, a bundle of read calls, a
+curated token list — and returns an envelope holding an `https` URL or a bounded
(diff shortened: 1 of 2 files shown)


moodysalem:
Withdrawing as a separate proposal. The envelope, retrieval rules, and provenance semantics now live in #1992 as the execution plan's transport section, where they are needed; a reference format for one payload type did not warrant its own number.

PR #1994, opened 2026-09-04
Add ERC: Declarative Wallet Signing Policy

moodysalem:
Adds ERC-8411, a document that states in advance which transactions a wallet may sign without asking a human: an ordered list of `allow` / `review` / `deny` rules over each planned call, the prepared transaction envelope, and the channel that delivered the request.

A wallet that signs only when a human clicks is safe and does not scale. A wallet that signs whatever it is asked scales and is not safe. Agents now submit transactions on a schedule, in response to text read from untrusted sources, at hours when nobody is available — and prompt injection makes the requester's *intent* untrustworthy even when its code is honest. The only durable control in that setting is one the user wrote down beforehand.

**Relationship to existing work.** ERC-7715 and ERC-7710 grant scoped authority to a *third party*, typically enforced onchain by a smart account. This constrains the signer itself: a local, offchain limit on what the key will produce, applicable to an EOA, evaluated before a signature exists. A delegation says who may act; a policy says what this wallet will put its name to.

**Three things I would particularly like editor eyes on**

1. **The authority ordering `deny < review < allow`.** It makes "this edit only removes permission" mechanically checkable, which is what lets an untrusted party propose a policy at all: a provable tightening installs at low friction, and anything else — including anything undecidable — requires explicit human authorization. This is also why the matcher vocabulary is deliberately flat rather than a general expression language; a richer grammar would make the tightening proof undecidable and collapse every edit into full authorization.

2. **`review` as a third effect.** Most users do not want to automate a category of transaction; they want to stop being asked about everything else so the prompts they do see mean something. `review` is what makes a policy adoptable incrementally.

3. **The proved/claimed distinction on `source`.** A dapp's `domain` and an agent's `client` are self-asserted and attested by nothing; the channel, an automation id, and the TLS-vetted `plan_host` are proved. The spec requires implementations to surface which is which, because a vocabulary that does not say this invites users to build security boundaries out of claims. I would rather this be normative than left to each wallet.

**Included.** `assets/erc-8411/policy.schema.json`, validated with ajv against a conforming policy.

**Reference implementation.** Ekubo Wallet (`github.com/EkuboProtocol/wallet`), `crates/ekubo-wallet-core/src/core/policy.rs`, which implements the parts that are easy to omit: the tightening proof, rejection of provably shadowed rules, envelope matchers evaluated only after the wallet's own transaction preparation, and agents that can propose but never install. Named in prose rather than linked, per the EIP-1 external-link rule.

**Notes for editors**
- Companion proposals are open alongside this one and are referred to in prose rather than by number, since `markdown-refs` cannot resolve an unmerged sibling.

### DIFF
--- a/ERCS/erc-8411.md
+++ b/ERCS/erc-8411.md
@@ -47,8 +47,8 @@ requesting software is honest. In that setting the only durable control is one
 the user wrote down beforehand and the wallet enforces regardless of what it is
 asked.
 
-Existing work does not cover it. [ERC-7715](./erc-7715.md) lets a dapp request
-permissions and [ERC-7710](./erc-7710.md) lets contracts redeem delegated
+Existing work does not cover it. [ERC-7715](./eip-7715.md) lets a dapp request
+permissions and [ERC-7710](./eip-7710.md) lets contracts redeem delegated
 authority; both grant scoped power to a *third party*, typically enforced
 onchain by a smart account. This proposal is about the signer itself: a local,
 offchain constraint on what the key will produce, applicable to an externally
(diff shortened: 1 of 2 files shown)


moodysalem:
Withdrawing this draft. On reflection the material here is wallet behaviour that an agent reads from the wallet's own self-describing interface, not a wire format another party has to parse, so it belongs in project documentation rather than an ERC. The companion execution-plan proposal (#1992) continues.

PR #1995, opened 2026-09-04
Add ERC: Agent Wallet Execution Interface

moodysalem:
Adds ERC-8412, the wallet methods an autonomous client — an AI agent, a scheduled job, an automation runner — uses to submit and track a transaction it is structurally unable to approve.

Existing wallet interfaces assume a human is present: a request goes out, a modal appears, an answer comes back in seconds, and if nobody answers the promise rejects. Every part of that breaks when the requester runs unattended. The gap is not that agents need more power; it is that the honest arrangement — the agent asks, a human decides, possibly hours later — has no vocabulary.

**The centerpiece is that waiting for a human is a state, not an error.** A client submits, learns approval is required, polls, and is told the outcome. A rejected promise cannot distinguish "the user said no" from "the user has not looked yet", so a client receiving one does the wrong thing. Polling is worse in latency and better in everything else: it survives client restarts, it is inspectable, and it answers the same whether asked once or a thousand times.

**Two parts I would like editor input on**

1. **The capability boundary is normative rather than left to each wallet.** The spec lists exactly what this interface must never expose: deciding a pending request, exporting key material, installing or editing policy, accepting legal terms, mutating owner-only settings. Given a client may be relaying prompt-injected instructions, the set of things it *cannot* do is the whole security story. Leaving it to each implementation produces a market where the most permissive wallet sets the norm and a user cannot know what connecting an agent means.

2. **Signature methods are review-only, with no automatic path at all.** A policy can constrain targets, values, and decoded calldata. A typed-data signature has a domain and a payload whose consequences live in a contract the wallet never called; an order signed today executes against the account tomorrow. Rather than invent a second authorization vocabulary of unclear strength, the proposal declines to offer automation here.

**Framing.** This is written as transport-neutral `wallet_` methods plus a lifecycle state machine, not as a tool surface for any particular agent protocol. One binding (MCP over a local bridge) is described in a paragraph, and it adds no authority — every restriction is enforced by the wallet, not the transport.

One unusual affordance worth flagging: the spec asks wallets to return the correct next action in-band with each result. The clients here are frequently language models, which follow the most recent explicit instruction and are prone to plausible wrong next steps — calling the wrong wait method, resubmitting a request whose nonce is gone, or handing back a request id and stopping. This converts a class of client bugs into a wallet responsibility.

**Reference implementation.** Ekubo Wallet (`github.com/EkuboProtocol/wallet`), `src/mcp.rs`, where the session receives no key store and no arbitrary signing operation — only two guarded execution paths. Named in prose rather than linked, per the EIP-1 external-link rule.

**Notes for editors**
- This proposal carries transactions as execution plans and refers to a signing policy, both specified in companion proposals open alongside it. They are referred to in prose rather than by number, since `markdown-refs` cannot resolve an unmerged sibling; I will add the numbers and `requires` entries once assigned.

### DIFF
--- a/ERCS/erc-8412.md
+++ b/ERCS/erc-8412.md
@@ -150,11 +150,16 @@ It takes two OPTIONAL parameters:
   approve anything, cannot widen the policy, and cannot make a denied plan
   sendable. A wallet MUST reject `mustReview` combined with `requestId`, whose
   bytes are already reviewed and signed.
-- `onSimulationFailure` — whether a failed simulation should be reported to the
-  client or queued for a human. Its default SHOULD be to queue.
+- `onSimulationFailure` — `"queue"` to put a plan whose simulation failed in
+  front of a human, or `"fail"` to report the failure to the client instead. The
+  default MUST be `"queue"`. Neither value affects a `rejected` outcome, which
+  fails regardless.
 
-The method MUST return one of: a submitted transaction; a `requestId` in the
-`approvalRequired` state; or a failure. It MUST NOT return a `requestId` for a
+The method MUST return either a failure or a `requestId` together with the
+request's current state, which on success is `submitted` when the policy allowed
+the plan and `approvalRequired` otherwise. Every accepted submission has a
+`requestId`, including one that signed automatically, so that the same status
+and wait methods apply to both. The method MUST NOT return a `requestId` for a
 `rejected` outcome — a denying rule produces a failure, never a queued request.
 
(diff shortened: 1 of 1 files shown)


moodysalem:
Withdrawing this draft. On reflection the material here is wallet behaviour that an agent reads from the wallet's own self-describing interface, not a wire format another party has to parse, so it belongs in project documentation rather than an ERC. The companion execution-plan proposal (#1992) continues.

PR #2002, opened 2026-09-07
Add ERC: Preregistered Acceptance Criteria

richard7463:
## Summary

This PR adds a draft ERC for **Preregistered Acceptance Criteria**.

Existing agent verification standards reach a verdict by recomputation: derive
the artifact again and compare digests. That only works where a verdict is a
function of the artifact. Most of what agents pay for is judged rather than
computed — a generated deliverable, a human review, an on-site inspection.

This ERC freezes acceptance criteria and typed evidence obligations on-chain
*before* evidence exists, and requires the final attestation to be itemized
against that frozen list (MET / WAIVED / UNMET / NOT_APPLICABLE). A verdict that
declares an outcome satisfied while a required obligation is recorded UNMET is
rejected at the contract level rather than merely flagged.

Scope is deliberately narrow: registry interface, two off-chain document
schemas, a packed itemized-verdict encoding, and contract-enforced consistency.
Evidence formats, payment, and identity are out of scope, so it composes with
existing standards (ERC-8126 verdicts, ERC-8183/8195 settlement, ERC-8353
verifier selection, ERC-8004 identity) instead of competing with them.

## Discussion

- Ethereum Magicians: https://ethereum-magicians.org/t/pre-erc-discussion-preregistered-acceptance-criteria/29609

## Validation

- Preamble follows EIP-1 (title/description length, `discussions-to`,
  `type`, `category`, `created`, `requires`).
- Draft content reviewed against the recomputation boundary of ERC-8404 and the
  outcome-switching literature (ICMJE registration, COMPare) cited in the
  motivation.

### DIFF
--- a/ERCS/erc-8411.md
+++ b/ERCS/erc-8411.md
@@ -43,11 +43,11 @@ value and comparing it (status as of 2026-09-04):
 
 | Standard | Verification primitive |
 | --- | --- |
-| ERC-8281 (Observation Commitment Protocol, Draft) | `recompute → compare → confirm inclusion` |
-| ERC-8274 (AI Inference Proof Verification, Draft) | `verify(inputHash, outputHash, metadata, proof)` |
-| ERC-8299 (WYRIWE, Draft) | `raw_input_hash → sanitization_pipeline_hash → input_hash` |
-| ERC-8263 (Onchain Proof Layer, Draft) | `anchor(agentIdScheme, agentId, proofHash)` |
-| ERC-8404 (Recomputable Verification Receipts, Draft) | `REPRODUCED / DIVERGED / CANNOT_RECOMPUTE` |
+| [ERC-8281](https://ethereum-magicians.org/t/erc-8281-observation-commitment-protocol-ocp/28399) (Observation Commitment Protocol, Draft) | `recompute → compare → confirm inclusion` |
+| [ERC-8274](https://ethereum-magicians.org/t/erc-8274-ai-inference-proof-verification/28083) (AI Inference Proof Verification, Draft) | `verify(inputHash, outputHash, metadata, proof)` |
+| [ERC-8299](https://github.com/ethereum/ERCs/pull/1810) (WYRIWE, Draft) | `raw_input_hash → sanitization_pipeline_hash → input_hash` |
+| [ERC-8263](https://ethereum-magicians.org/t/erc-8263-onchain-proof-layer-for-ai-agents/28577) (Onchain Proof Layer, Draft) | `anchor(agentIdScheme, agentId, proofHash)` |
+| [ERC-8404](https://ethereum-magicians.org/t/erc-8404-recomputable-verification-receipts/29521) (Recomputable Verification Receipts, Draft) | `REPRODUCED / DIVERGED / CANNOT_RECOMPUTE` |
 
 The primitive is sound wherever a verdict is a function of the artifact. The
 boundary it cannot cross is not digital versus physical — it is **computed versus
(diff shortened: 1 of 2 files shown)


ERCS/erc-8411.md
```diff
@@ -0,0 +1,600 @@
+---
+eip: 8411
```
abcoathup:
```suggestion
eip: 8412
```
Numbers are assigned by editors & associates.  Please don't attempt to self assign.

ERCS/erc-8412.md
```diff
@@ -0,0 +1,600 @@
+---
+eip: 8411
+title: Preregistered Acceptance Criteria
+description: Acceptance criteria and typed evidence obligations frozen on-chain before evidence exists, with verdicts itemized against them
+author: Richard (@richard7463)
+discussions-to: https://ethereum-magicians.org/t/pre-erc-discussion-preregistered-acceptance-criteria/29609
```
abcoathup, line 6:
```suggestion
discussions-to: https://ethereum-magicians.org/t/erc-8412-preregistered-acceptance-criteria/29609
```

renezander030:
The Reference Implementation section says the registry contract and conformance vectors are the next deliverable before Review is requested. I would like to contribute the vectors, if that is useful and not already in progress.

Proposed shape: a JSON fixture set covering E1-E9 and O1-O4, each case naming the invariant it exercises, plus a minimal conforming registry to run them against so the fixtures are executable rather than descriptive. Either landing in `assets/erc-8412/` here, or in a standalone repo you vendor, whichever you prefer.

One thing worth settling before the vectors are written, because a fixture can only pin behaviour the text has already decided. The expiry boundary is currently specified with two different comparisons:

- E7: `attestOutcome` MUST revert once `block.timestamp > expiry`, so an attestation at exactly `expiry` is permitted.
- E8: `resolveExpired` MUST revert before `expiry`, so a resolution at exactly `expiry` is permitted too.

In the block where `block.timestamp == expiry` both calls are available. E8 guards one ordering, reverting if an attestation exists, but there is no mirror: E3 bounds attestations to one per preregistration rather than one attestation after a resolution, and E9 only rejects `None` as an input verdict. So `resolveExpired` can write `ExpiredUnresolved` and a later `attestOutcome` in the same block can overwrite it with `Satisfied` without violating anything in section 7. Section 9 hands `ExpiredUnresolved` straight to settlement layers as non-satisfaction, so this is reachable by any contract acting on `ExpiredResolved`.

Two lines close it: make E7 and E8 test the same comparison so the windows are disjoint by construction, and give `attestOutcome` the guard E8 already has in the other direction, reverting when a terminal verdict is recorded. Same shape one level down, section 9 says `resolveExpired` applies when neither an attestation nor a superseding preregistration exists, but E8 only names the attestation.

(Also raised on the discussions thread, post #5.)

Happy to write the vectors either way. I just do not want to encode a boundary the text has not chosen yet.

babyblueviper1:
Independently reread E3/E7/E8/E9 against @renezander030's point above — the gap is real, not just a boundary-comparison nitpick.

At `block.timestamp == expiry` with no prior attestation: E7 doesn't revert (`>` not `>=`), E8 doesn't revert (`expiry` isn't strictly *before* `expiry`, and no attestation exists yet to trigger its other guard). Both calls are live in the same block. `resolveExpired` firing first writes `ExpiredUnresolved`; nothing then stops `attestOutcome` firing after it in the same block and writing `Satisfied` over it — E3's "one attestation per preregistration" and E9's "reject `verdict == None`" both still hold, since neither speaks to whether a *resolution* already exists.

This isn't just a same-block tie — tightening the comparison operators (`>=`/`<=`) only removes the exact-timestamp overlap, it doesn't touch the underlying gap: nothing in section 7 makes `ExpiredUnresolved` itself terminal. E8 already has the right shape for this ("MUST revert...if an attestation exists") — it's just one-directional. E7 needs the mirror: `attestOutcome` MUST also revert if the preregistration already has a non-`None` verdict (from a prior `resolveExpired`, not just a prior attestation). That closes it independent of exact timestamp comparisons, and matches the doc's own stated goal in section 9 (`ExpiredUnresolved` as attributable, settlement-relevant state) — right now that state is reachable but not actually final.

Worth pinning as its own invariant (something like E10: `attestOutcome` MUST revert if a resolution already exists for this `preregistrationId`) before @renezander030's conformance vectors get written — a fixture can only pin behavior the text has already decided, same point already made above.

PR #2003, opened 2026-09-08
Update ERC-7730: Bind proxy descriptors to implementation addresses

manuelwedler:
This PR corrects the proxy guidance in ERC-7730. This change was discussed in an earlier Clear Signing Working group call but has been forgotten to implement.

The spec previously recommended binding ERC-7730 files to the *proxy* address for upgradeable and Diamond proxies, and to the *implementation* address only for multi-instantiation. A file bound to a proxy address keeps matching after an upgrade and displays intents reviewed for the old implementation while the new one executes. This is exactly the attack clear signing is meant to prevent.

Changes:
- Proxy support: ERC-7730 files SHOULD be bound to the implementation address for all proxy patterns, with an explanation why.
- Upgradeable proxies: bind to the implementation address. A new implementation gets a new file.
- Composable contracts: one file per facet, bound to the facet address.
- `contract.deployments` and `eip712.deployments`: a proxy matches when its resolved implementation address matches a deployment option.

Related: https://github.com/ethereum/clear-signing-erc7730-registry/issues/2960

### DIFF
--- a/ERCS/erc-7730.md
+++ b/ERCS/erc-7730.md
@@ -499,7 +499,7 @@ The `contract` sub-key is used to introduce an EVM smart contract binding contex
 
 An array of deployments options. Wallets MUST verify that the target chain and contract address of the containing transaction either:
   * Match one of these deployment options
-  * Is a *proxy* pointing to one of the deployment option (see [Proxy support](#proxy-support))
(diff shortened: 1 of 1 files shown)


manuelwedler:
decided to accept the current risk/limitation until we have https://github.com/ethereum/ERCs/pull/1738

PR #2012, opened 2026-09-19
Add ERC: Know-Your-Agent (KYA) Framework

garyyang-finchip:
New ERC draft: a scheme-agnostic framework for recording, resolving and presenting trust assertions (KYA) about AI agents, with a zero-knowledge profile and a normative ERC-8004 binding.

Reference implementation, schemas, vectors and an end-to-end test suite: https://github.com/garyyang-finchip/kya-standard

Discussion thread: [ERC-8419: Know-Your-Agent (KYA) Framework](https://ethereum-magicians.org/t/erc-8419-know-your-agent-kya-framework/29735).


babyblueviper1:
Nice, concrete framework — L1-L4 with a real ERC-8004 bridge and an end-to-end test suite (not just a spec doc) is the right shape for this to actually get adopted rather than argued about.

One question on scope, from the angle of a project running an adjacent but different layer (WYRIWE, `ethereum/ERCs#1810`, pending review — an ERC for per-decision judgment attestation: a signed verdict on whether *one specific agent action/output* was correct, published with a recomputable proof, not a durable claim about the agent in general).

KYA's model reads as durable and agent-scoped: a scheme + level + validity window that says something like "this agent, as of this assertion, satisfies scheme X at level Y" — closer to a credential than a per-event judgment. A WYRIWE-style verdict is the opposite shape: it's scoped to one action (this specific trade, this specific tool call), has no natural "level," and its validity is really "was this correct," not "is this agent trustworthy."

Is that distinction intentional — i.e. is KYA deliberately scoped to durable/general trust assertions and per-action judgment is explicitly a different registry's job (with the `kya:` bridge tag namespace maybe someday hosting both) — or would a scheme descriptor whose "level" is actually a per-action verdict outcome (bound to a specific `actionId`/timestamp rather than a validity window) fit inside the existing `IKYAVerifier`/attestWithProof() shape without forcing a new registry? Asking because if it fits, WYRIWE's own recompute-based proof format could plug in as a real `attestWithProof()` mode rather than staying a parallel, unconnected standard.

ERCS/erc-9999.md
```diff
@@ -0,0 +1,543 @@
+---
+eip: 9999
```
```suggestion
eip: 8419
```

ERCS/erc-9999.md
```diff
@@ -0,0 +1,543 @@
...
+title: Know-Your-Agent (KYA) Framework
+description: Scheme-agnostic registries and handshake for trust assertions about AI agents, with a zero-knowledge profile and agent-registry binding
+author: Gary Yang (@garyyang-finchip)
+discussions-to: https://ethereum-magicians.org/t/draft-erc-know-your-agent-kya-framework-trust-assertions-for-agents-zk-kya-profile-erc-8004-binding/29735
```
```suggestion
discussions-to: https://ethereum-magicians.org/t/erc-8419-know-your-agent-kya-framework/29735
```
Updated Eth Magicians title with assigned number and ERC title

PR #2015, opened 2026-09-20
Update ERC-8048: Fix Base Chain Identifier example

LiamVisionary:
Fixes a typo in the ERC-721T profile's "Chain-keyed addresses" example.

The ERC-7930 Chain Identifier given for Base has a stray `0x02` byte:

| | value |
| :-- | :-- |
| current | `0x000100000202210500` |
| corrected | `0x0001000002210500` |

Derivation, following ERC-7930 and the CAIP-350 `eip155` profile (chain id as a big-endian unsigned integer of the minimum necessary bytes, so `8453` is `0x2105`):

```
0x0001   version
  0000   chain type (eip155)
    02   chain reference length
  2105   chain reference (8453)
    00   address length (zero for a Chain Identifier)
```

The mainnet example in the same sentence, `0x00010000010100`, already follows this layout (`0001 0000 01 01 00`). The current Base value does not parse under it: after a reference length of `02` the reference would be `0221`, leaving an address length of `05` and a single trailing byte.

This matters in practice because the identifier is used verbatim inside the `address[<chain-id>]` and `account[<chain-id>][<index>]` keys, and ERC-8048 compares keys as exact bytes. A client that derives the key from ERC-7930 and a contract that copied the example text will not find each other.

Found while implementing the profile for a collection on Base. The author (@nxt3d) has confirmed it is a typo. One line changed, no normative change.


nxt3d:
thx, looks good.

nxt3d (approved):
LGTM

PR #2016, opened 2026-09-20
Update ERC-8048: metadata_contract is a plain EVM address

LiamVisionary:
Changes the value of `metadata_contract` in the Onchain Metadata Contract Reference extension from an ERC-7930 Interoperable Address to a plain EVM address, as agreed with the author (@nxt3d).

Why:

- The referenced contract MUST implement `metadata(uint256,string)`, so it is always an EVM contract. The chain-agnostic part of ERC-7930 (Solana, Bitcoin) cannot apply to it.
- A plain address is what an implementer reaches for first. The first deployment of this extension (HivePunkz, on Base) published a bare address before anyone noticed the spec asked for ERC-7930 bytes, and nothing reads the field yet, so now is the cheap moment to change it.
- The metadata contract is on the same chain as the token registry, so no chain needs encoding.

What changes:

- the value is a `0x`-prefixed 20-byte address; clients accept any letter case
- the example uses `0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045`
- client behavior loses the "decode ERC-7930 bytes" and "parse the target chain" steps
- the cross-chain paragraph is replaced by a sentence stating the contract is on the token registry's chain

ERC-7930 is untouched where it earns its keep: the chain-keyed `address[<chain-id>]` and `account[<chain-id>][<index>]` keys of the ERC-721T profile. `requires` is unchanged.

Separate from #2015 (the Base Chain Identifier typo). The two touch different lines.


nxt3d (approved):
LGTM

PR #2017, opened 2026-09-21
Update ERC-7730: fix the tokenAmount example and state the threshold comparison

kuzdogan:
There is no parameter named `thresholdLabel` in either schema; the parameter is `message`. The ERC-20 approve example under the interface-binding section is the only place that uses the wrong name, and a descriptor written from that example fails validation against `erc7730-v2.schema.json` (`additionalProperties: false` on `tokenAmountParameters`).

The `tokenAmount` reference table describes `threshold` as the value "above which" the message is displayed, without saying whether the threshold value itself counts. Implementations read this differently, so this PR states the comparison explicitly: the message is displayed when the field value is greater than or equal to the threshold.

That is the reading the specification's own examples already show. The example rows under the table give a field value of `0xFFFFFFFF` with a threshold of `0xFFFFFFFF` and expect `Unlimited DAI` and `Max DAI`. The registry depends on it as well: the Uniswap Permit2 descriptor marks an unlimited allowance on a `uint160` field with a threshold equal to the `uint160` maximum. No `uint160` value is strictly above its own maximum, so under a strict reading the message could never appear, and for such a field an author could not express "unlimited" at all. The other thresholds in the registry (2^255, and `0xff..fe` for `uint256` fields) give the same result under both readings for every value a transaction sends.

The same "above" wording sits in the `threshold` and `message` descriptions of the schema files. This PR updates them in `erc7730-v2.schema.json` and in the `erc7730-v3.0.0-next.schema.json` draft. These are description strings only; the set of valid descriptors does not change, so no schema version changes. The deprecated `erc7730-v1.schema.json` is left as it is.

Registry discussion: ethereum/clear-signing-erc7730-registry#2897 (ERC-7730 cannot ship a fix to its active schema).


PrisionMike (approved):

PrisionMike (commented):
The schema description might also need updation.
https://github.com/ethereum/ERCs/blob/12827ab9e332226fd9911fef0b5a79e78863c300/assets/erc-7730/erc7730-v2.schema.json#L909

kuzdogan:
@PrisionMike great catch! thanks

PR #2018, opened 2026-09-21
Update ERC-7730: let the active schema line receive patch and minor releases

kuzdogan:
The versioning section merged in PR 1838 (introduce schema semantic versioning) makes every published schema file immutable and allows one mutable file, the draft of the next major version. Together with the fixed file name that descriptors point at, this means the active line cannot receive a fix at all: a correction to `2.0.0` can only ship inside `3.0.0`, or by editing every descriptor that points at `erc7730-v2.schema.json` (389 files in the registry today).

Two open pull requests show the effect. PR 1991 (allow threshold formatting for the plain `amount` format) adds a parameter block, which is a MINOR change. It edits only the text, because the only file it may edit is the 3.0.0 draft. A descriptor that uses the feature fails validation against `2.0.0` (`must NOT have unevaluated properties: params`). PR 1782 (introduce an explicit extensibility process) edits `erc7730-v2.schema.json` in place, which the section forbids.

This PR keeps the rule that a published version never changes, and adds a file that descriptors may point at which follows the newest release of its major line. Three changes:

1. **Exact files and one alias per major line.** `erc7730-v2.0.0.schema.json` is added as a byte-for-byte copy of the current `erc7730-v2.schema.json`. Exact files are immutable. `erc7730-v2.schema.json` becomes the alias of the 2.x line and holds a copy of the newest 2.x release. Only a release changes an alias. Descriptors keep pointing at the alias, so a patch reaches them without an edit. The same is done for the 1.x line (`erc7730-v1.0.0.schema.json`) for consistency.
2. **PATCH, MINOR and MAJOR are defined by the set of valid descriptors.** PATCH: the set is unchanged. MINOR: the set grows. MAJOR: the set shrinks. This is what makes a moving alias safe, and it is checkable by machine: the registry validates every descriptor against a changed schema file (ethereum/clear-signing-erc7730-registry#3036).
3. **The released-versions table.** Only the Status column may change. The other columns are immutable.

The rule that the text describes the draft only is unchanged. How a wallet handles a key it does not know is also unchanged and out of scope here.

Registry discussion: ethereum/clear-signing-erc7730-registry#2897 (ERC-7730 cannot ship a fix to its active schema).


ERCS/erc-7730.md
```diff
@@ -128,24 +128,32 @@ In this example, the `to` parameter and the `value` parameter SHOULD both be dis
...
 
 Schema versions follow Semantic Versioning with an optional pre-release suffix: `-next` for the in-development draft, `-rc.N` for release candidates.
 
-A MAJOR bump indicates a breaking schema change, a MINOR bump indicates a backward-compatible addition, and a PATCH bump indicates a non-breaking fix or editorial change.
+The version parts are defined by the set of descriptors that the schema accepts:
 
-Schema changes merged between releases accumulate in a single mutable draft file, named `erc7730-v<VERSION>-next.schema.json` after the targeted version.
+* A PATCH bump does not change the set of valid descriptors. Only titles, descriptions and other non-normative text in the schema change.
+* A MINOR bump grows the set of valid descriptors. Every descriptor that was valid stays valid.
+* A MAJOR bump shrinks the set of valid descriptors. Some descriptor that was valid stops being valid.
```
forshtat, line 147:
```suggestion
* A MAJOR bump shrinks the set of valid descriptors or meaningfully changes their presentation to users. Some descriptor that was valid stops being valid.
```

